PR #24, opened 2022-10-18
Remove wee_alloc dependency

danheuck:
wee_alloc has a security advisory ([RUSTSEC-2022-0054](https://rustsec.org/advisories/RUSTSEC-2022-0054)). It looks like you recently removed it from the examples, this just finishes the job by removing it from the leptos_dom cargo.toml since it's not actually in use.


gbj:
Great catch, thanks!

PR #30, opened 2022-10-20
Add console error panic hook to all existing examples

### LINKED ISSUE #29: Add more helpful error messages in `leptos_reactive`
Much of the reactive system is built on `RefCell`, which for some reason gives really unhelpful error messages in the browser, even in debug mode, because it doesn't give line numbers as it does in the console. (At least for me. Chrome tends to be better than Firefox with Wasm errors, but still not great.)

I usually don't see panic messages in Wasm error output either, so I have a `debug_warn!` macro in `leptos_reactive` that uses `println!` syntax to log a warning to the console but is stripped out in release builds for bundle size.

It would be good to replace all the `borrow_mut()`a in `leptos_reactive` with `try_borrow_mut()`s that catch and log an appropriate error with `debug_warn!`

Some of these don't even need to panic: for example if you're trying to run an effect or update a signal they can log the warning and then do nothing. Reading a signal that has been disposed does need to panic (unless someone has a better solution), because otherwise there's no value to return.

This helps with debugging issues like #25

oceantume:
Fixes #29

### DIFF
--- a/examples/counter/Cargo.toml
+++ b/examples/counter/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 leptos = { path = "../../leptos" }
 console_log = "0.2"
 log = "0.4"
+console_error_panic_hook = "0.1.7"
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.0"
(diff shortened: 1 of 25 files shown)


gbj:
Thanks so much — I push the same changes a couple minutes after you and hadn't seen your PR yet but there were no real conflicts so I'm just merging this.

Issue #34, opened 2022-10-22
Forcing all resources to be serializable seems overly restrictive


danheuck:
It seems like resources are currently forced to be serializable to support server side rendering (and potentially resumability in the future). This makes sense, but it excludes some use cases that would otherwise make sense as resources. For example, I was experimenting with accessing a webcam in leptos (using nokhwa). This is an async operation and, reading the docs, it seemed like a resource was a perfect fit. However, a webcam handle isn't serializable so I ended up with this slightly more awkward implementation:

``` rust
use leptos::*;
use nokhwa::{js_camera::JSCameraConstraintsBuilder, JSCamera};
use std::{cell::RefCell, rc::Rc};


async fn get_camera(set_camera: WriteSignal<Option<Rc<RefCell<JSCamera>>>>) {
    let constraints = JSCameraConstraintsBuilder::new().build();
    let camera = JSCamera::new(constraints).await;

    match camera {
        Ok(camera) => {
            log::info!("Got camera");
            set_camera(Some(Rc::new(RefCell::new(camera))));
        }
        Err(e) => {
            log::error!("Failed to get camera: {}", e);
        }
    }
}

#[component]
pub fn UserWebcam(cx: scope) -> Element {
    let (camera, set_camera) = create_signal::<Option<Rc<RefCell<JSCamera>>>>(cx, None);
    spawn_local(get_camera(set_camera));

    create_effect(cx, move |_| {
        if let Some(camera) = camera.get() {
            if let Err(e) = (*camera).borrow_mut().attach("camera_out", false) {
                log::error!("Failed to attach camera: {}", e);
            } else {
                log::info!("Attached camera");
            }
        }
    });

    view! {cx,
        <video id="camera_out" class="user_webcam"/>
    }
}
```
It works, but it's not ideal and it can't integrate with functionality like suspense contexts (unless I'm missing something).

It feels like the most natural solution would be to split resources into serializable and non-serializable types. Serializable resources would keep the current behavior but non-serializable ones would be handled fully on the client side regardless of rendering configuration. Happy to take a stab at implementing this if you're open to a PR, but figured I'd open an issue first to at least make sure this functionality doesn't already exist somewhere.

gbj:
1. This is really cool.
[...]

PR #35, opened 2022-10-23
Unserializable resources

danheuck:
This PR allows for the creation of unserializable "client" resources as discussed in #34. The motivating example from that issue can now be rewritten using a resource:
```rust
use leptos::*;
use nokhwa::{js_camera::JSCameraConstraintsBuilder, JSCamera};
use std::{cell::RefCell, fmt::Debug, rc::Rc};

#[derive(Clone)]
struct Camera(Rc<RefCell<JSCamera>>);

impl Debug for Camera {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_tuple("Camera")
            .field(&self.0.borrow().id())
            .finish()
    }
}

async fn get_camera(_: ()) -> Option<Camera> {
    let constraints = JSCameraConstraintsBuilder::new().build();
    let camera = JSCamera::new(constraints).await;
    camera.ok().map(|c| Camera(Rc::new(RefCell::new(c))))
}

#[component]
pub fn UserWebcam(cx: scope) -> Element {
    let camera = create_client_resource(cx, || (), get_camera);

    create_effect(cx, move |_| {
        if let Some(Some(camera)) = camera() {
            if let Err(e) = (*camera.0).borrow_mut().attach("camera_out", false) {
                log::error!("Failed to attach camera: {}", e);
            } else {
                log::info!("Attached camera");
            }
        }
    });

    view! {cx,
        <video id="camera_out" class="user_webcam"/>
    }
}
```
Sadly, I still couldn't get it to work properly with Suspense. There's probably a way to do it, but the camera has to use an effect to attach the media stream to the video HTML element and, if I wrap the video element in suspense, getting the timing to work properly such that the element was created before the effect ran didn't seem to work out for me.

I tested this with the existing examples that use resources (fetch and hackernews) which is why this PR also has an unrelated fix for an issue in hackernews-app's cargo.toml. The added code, particularly the new resource creation functions (`create_client_resource()` and `create_client_resource_with_initial_value()`), still needs documentation but I figured I'd throw the implementation up here for review.

### DIFF
--- a/examples/hackernews/hackernews-app/Cargo.toml
+++ b/examples/hackernews/hackernews-app/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 anyhow = "1"
 console_log = "0.2"
-leptos = { path = "../../../leptos", default-features = false }
+leptos = { path = "../../../leptos", default-features = false, features = ["serde"] }
 leptos_meta = { path = "../../../meta", default-features = false }
 leptos_router = { path = "../../../router", default-features = false }
 log = "0.4"
--- a/leptos_reactive/src/resource.rs
+++ b/leptos_reactive/src/resource.rs
@@ -16,12 +16,18 @@ use crate::{
 };
 
 /// Creates [Resource](crate::Resource), which is a signal that reflects the
-/// current state of an asynchronous task, allowing you to integrate
-/// `async` [Future]s into the synchronous reactive system.
+/// current state of an asynchronous task, allowing you to integrate `async`
+/// [Future]s into the synchronous reactive system.
 ///
 /// Takes a `fetcher` function that generates a [Future] when called and a
-/// `source` signal that provides the argument for the `fetcher`. Whenever
-/// the value of the `source` changes, a new [Future] will be created and run.
+/// `source` signal that provides the argument for the `fetcher`. Whenever the
+/// value of the `source` changes, a new [Future] will be created and run.
+///
+/// When server-side rendering is used, the server will handle running the
+/// [Future] and will stream the result to the client. This process requires the
+/// output type of the Future to be [Serializable]. If your output cannot be
+/// serialized, or you just want to make sure the [Future] runs locally, use
+/// [create_local_resource()].
 ///
 /// ```
 /// # use leptos_reactive::*;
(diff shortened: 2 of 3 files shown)


gbj:
Yeah this looks really good — Existing automated tests still pass, the HackerNews demo works fine, and more than that I think it's a nice clean implementation. 

If you want to really go above and beyond feel free to add some doctests. I still need to write docs for Resources in general so don't worry too much about it. Otherwise, I'm happy to merge this any time.

Here's one super-simple doctest just to make sure it compiles and runs with a type that isn't `Serialize, Deserialize`. I tried this and it seems fine. Feel free to use this as a template or expand.

/// ```
/// # use leptos_reactive::*;
/// # create_scope(|cx| {
/// #[derive(Debug, Clone)] // doesn't implement Serialize, Deserialize
/// struct ComplicatedUnserializableStruct {
///   // something here that can't be serialized
/// }
/// // any old async function; maybe this is calling a REST API or something
/// async fn setup_complicated_struct() -> ComplicatedUnserializableStruct {
///   // do some work
///   ComplicatedUnserializableStruct { }
/// }
///
/// // create the resource that will
/// let result = create_client_resource(cx, move || (), |_| setup_complicated_struct());
/// # }).dispose();
/// ```

gbj:
Oh my one other thought was on naming: Do you think it should actually be `create_local_resource` or something? It could just as well be a server-only resource as a client-only one! Open to your thoughts.

danheuck:
Updated documentation and changed "client" resources to "local" resources. Do you think it makes sense to replace all references to "unserializable" resources with "local" as well? I think it makes sense since you could have a resource that is technically serializable but you still want to run locally. My only issue is that I'm not sure what to call the "serializable" resources then. "Remote" would be the obvious counterpart, but in client-side rendering they'll still be run locally, so that name feels wrong. Maybe just "resource" and  "local resource"?

gbj:
Nah I think it's fine to use "resource" and "local resource," and not worry about alternatives for serializable/unserializable... Those ones are pretty much descriptive, and not really part of the public API anyway.

Thanks so much for this contribution. I'm going to merge the PR. Great work!

Issue #36, opened 2022-10-24
Clone bounds for signals/resources/memos are mostly not necessary


danheuck:
Discussed briefly in #34, but I'm opening a new issue since it's mostly unrelated.

I took a look at the usage of Clone in the signals API and found it could be removed fairly easily. Resources and Memos were slightly less simple:
- Resource sources still need to be Clone, though the Future's output does not need to be
- A `Resource::with()` was added, with similar behavior to `Signal::with()` to allow access without cloning
- Memo outputs do not need to be Clone, but it now takes `impl FnMut(Option<&T>) -> T` instead of `impl FnMut(Option<T>) -> T`
 
I have a branch for these changes, but it's built on top of my branch for #35, so I'll open a PR for it once that gets merged.

gbj:
So, since I alluded to it in the previous discussion: The only thing this would really affect is transitions.

The `Clone` requirement was originally because I had a `use_transition()` like the one in Solid or React, implemented in a way that required you to be able to clone the existing value. Basically, transitions "fork" reality so that the reactive system keeps moving along and updating "under transition" but not actually updating the UI until all the asynchronous work being done is complete. This does require `T: Clone` because you need to clone `value` to get `transition_value`. 

This isn't necessary for something like our `<Suspense/>`, and in fact I've removed it because my implementation was buggy anyway. It enables some nicer UI for things like routing, but I'm sure that can be added manually.

I'm very open to removing the `Clone` requirement since it's not _currently_ necessary and maybe reimplementing transitions in the future in a different way (or restricting which signals can have transitional values).

danheuck:
Interesting note on the transitions. Based only on your original comment, I had the impression that the Clones were mainly just legacy code that hadn't been removed yet but, if there's a reason they'll be needed in the future, it seems like you might as well keep them. I'd have to take a look at that old transitions implementation to make sure I actually understand what they do/how they do it to give any real feedback I'm afraid.

gbj:
Re: transitions — If you're reading this (Dan or anyone else) and have no idea what I'm talking about, here's the classic Solid demo: [without transitions](https://www.solidjs.com/tutorial/async_transitions) and [with transitions](https://www.solidjs.com/tutorial/async_transitions?solved). Transitions are basically a way of "holding" an original state, while asynchronously transitioning to a new state, rather than flashing back to the `<Suspense>` fallback.

I've rewritten the reactive system since then, but here's how transitions would basically work in the current reactive system:
[...]

PR #37, opened 2022-10-24
Remove clone bounds

danheuck:
As mentioned in #36, this removes every `Clone` bound that didn't seem necessary. The main notes are:
- Types held in a Signal no longer need to be Clone
- The output type of a Resource's future no longer needs to be Clone, though the bounds for the source type are unchanged
- A `Resource::with()` method was added, with similar behavior to `Signal::with()` to allow access without cloning
- Memo output types do not need to be Clone
   - To accommodate this, Memo now takes `impl FnMut(Option<&T>) -> T` instead of `impl FnMut(Option<T>) -> T` which is a breaking change, albeit a minor one

This still needs some documentation updates which I'll hopefully get to tomorrow. I'll probably also drop the webcam component that spawned this PR and #35 in as an example unless you object. Tested with all of the examples, but let me know if you see any issues.

### DIFF
--- a/examples/counters-stable/src/main.rs
+++ b/examples/counters-stable/src/main.rs
@@ -33,7 +33,7 @@ pub fn Counters(cx: Scope) -> web_sys::Element {
             let signal = create_signal(cx, 0);
             new_counters.push((next_id, signal));
         }
-        set_counters.set(new_counters.clone());
+        set_counters.update(|counters| counters.extend(new_counters.iter()));
     };
 
     let clear_counters = move |_| {
--- a/examples/counters/src/lib.rs
+++ b/examples/counters/src/lib.rs
@@ -27,7 +27,7 @@ pub fn Counters(cx: Scope) -> web_sys::Element {
             let signal = create_signal(cx, 0);
             new_counters.push((next_id, signal));
         }
-        set_counters(new_counters.clone());
+        set_counters.update(move |counters| counters.extend(new_counters.iter()));
     };
 
     let clear_counters = move |_| {
--- a/leptos_core/src/for_component.rs
+++ b/leptos_core/src/for_component.rs
@@ -15,7 +15,7 @@ where
     G: Fn(Scope, &T) -> Element,
     I: Fn(&T) -> K,
     K: Eq + Hash,
-    T: Eq + Clone + 'static,
+    T: Eq + 'static,
 {
     pub each: E,
     pub key: I,
@@ -34,7 +34,7 @@ where
     G: Fn(Scope, &T) -> Element + 'static,
     I: Fn(&T) -> K + 'static,
     K: Eq + Hash,
-    T: Eq + Clone + Debug + 'static,
+    T: Eq + Debug + 'static,
 {
     let map_fn = (props.children)().swap_remove(0);
     map_keyed(cx, props.each, map_fn, props.key)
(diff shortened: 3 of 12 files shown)


danheuck:
Also includes an unrelated fix for a bug in the two "counters" examples which I noticed while testing.

gbj:
Yes, I would *love* it if you'd add the webcam example. This is the kind of thing that's great to have in the examples folder in addition to the very typical counters and todos.

gbj:
I'm working on a few changes to this that will allow Memo to continue taking an Option<T> for its previous value without cloning T, which preserves the current API, relaxes the Clone requirement, and should end up making memos even more efficient than they were before by saving a clone. That should be especially helpful for the <For> component. 

Basically the idea is to use RefCell::take() to get the previous value out of the signal inside the memo effect. That signal will temporarily be empty, but it's replaced synchronously by the update so I don't think it's actually possible for anything to read it while it's empty. 

Stay tuned. I'm terrible at GitHub so not quite sure, but hopefully I can push some changes to the PR later today.

danheuck:
The fundamental issue I ran into when trying to remove the clone bound from Memo without changing the API is that we need to pass the memo function the previous state and then compare that previous state against the value returned by the function. This means we need to have ownership (or a reference to) the previous state after we call the memo function. As far as I can tell, this can be achieved by:

1. Cloning the state and only giving the memo function a copy
   - This is what we started with. It works fine, but requires the state to implement Clone and means a Memo will _always_ clone the state no matter what
2. Giving the memo function ownership and requiring it to give the old state back in addition to the new state
   - This is both an inconvenient API and raises the possibility that the memo function will modify the previous state, invalidating our comparison
3. Giving the memo function an immutable reference and keeping ownership ourselves
   - This is where I ended up, obviously. This allows the state to not implement Clone but, if it does, the memo function can clone it manually to achieve the same behavior as the first option.

This could also be circumvented by introducing a Hash bound on the state. That way we could hash the state every time the memo function returns and just compare that hash against the previous hash to determine if it changed. I feel like hash is perhaps more annoying to implement than clone, but you're less likely to have unhashable data than uncloneable data. This does have the advantage though that we never have to clone and the memo function can mutate the state in place rather than allocating new memory if it wants, so this might be more performant?

gbj:
You're completely correct, of course.  I'd forgotten that we'd need still need to have the old value around, to checked whether it had changed!

I think taking `Option<&T>` for `prev` in a memo is the correct API, in fact, since it lets the user decide whether to clone or not in the memo function. Clearly preferable to cloning by default.

You mentioned docs changes but I think you've left everything in a consistent state, as far as I can tell. All right with you if I go ahead and merge it?

danheuck:
Yeah, hopefully not too many documentation changes, but I wanted to add something for `Resource::with()` and `Resource::read()`. Everything looks good to me, not much of a performance difference in those benchmarks (which is probably an indication of how little Cloning was actually happening despite all the things that needed to be `Clone`) but it's nice to see that things haven't gotten slower at least.

gbj:
Yeah the difference was way more impressive on the first run but even if the actual changes are only 0.01 - 0.02 on the total, that's closing 10-20% of the gap between Leptos and Solid.

At this point the benchmarking is a little silly — I'm genuinely focusing on other things like the RPC design — but it's good to feel like we're making things slightly faster, and certainly not slower!

I'm going to merge this. Really good work, and thanks for the contributions.

PR #44, opened 2022-11-01
Make base libs

akesson:
Excluded examples and created a Makefile. 

Install cargo-make with:
> cargo install --force cargo-make

Build, from workspace root with:
> cargo make build

Test:
> cargo make test

Note that the Makefile should be ok, but the build of leptos_reactive with `hydrate` currently fails.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,28 +11,9 @@ members = [
   # libraries
   "meta",
   "router",
+]
 
-  # examples
-  "examples/counter",
-  "examples/counter-isomorphic/client",
-  "examples/counter-isomorphic/server",
-  "examples/counter-isomorphic/counter",
-  "examples/counters",
-  "examples/counters-stable",
-  "examples/fetch",
-  "examples/gtk",
-  "examples/hackernews/hackernews-app",
-  "examples/hackernews/hackernews-client",
-  "examples/hackernews/hackernews-server",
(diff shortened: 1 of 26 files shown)


gbj:
I've fiddled to fix some build errors but come up with the following error on `cargo make build`:

> Building crate=leptos_router features=[regex]
> Package `leptos_router v0.0.1 (/Users/gjohnston/Documents/GitHub/leptos/router)` does not have feature `regex`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.

I'm not sure why it's trying to build with this feature. I did have a feature `regex` in a much earlier version of the router but have removed it; it looks like it's trying to run through the optional dependencies as features? Any idea?

gbj:
I'm running into an issue: it seems like `cargo make test` is running with stable, not nightly, which means almost all the doctests are broken. I see you're specifying `+nightly` in the args so I'm not sure why this is the case. Any idea?

akesson:
Fixed stable issue. It was running on nightly toolchain, but with `stable` feature on. But still fails on `Testing crate=leptos_macro features=[]`. Seems to be the macros (as you pointed out previously) that don't handle having none of the `ssr`/`csr`/`hydrate` features set.

You can test it directly in _leptos_macro_ with `cargo test --no-default-features`

PR #46, opened 2022-11-03
Remove mutually exclusive features

gbj:
@akesson I started from scratch with a new PR because I had to reorganize so much stuff, but all of the features (?) should now no longer be mutually exclusive. This means that plain `cargo make build` and `cargo make test` should work without all the excluded sets, etc. Honestly I'm still waiting for `cargo make build` to finish running on my local because this takes forever, as it includes all the examples.


gbj:
Unless I'm mistaken this should now pass `cargo make build` and `cargo make test`, including all the examples. 

I want to give the examples a manual test (to make sure the changes haven't just compiled but broken the UI) before I merge it. Hopefully I'll do that tonight. But I think we should be in good shape.

gbj:
Yep this all looks good.

PR #63, opened 2022-11-08
Readme vocabulary

mendeljacks:
As per chris biscardi's recommendation
The word universal is more trendy to indicate multiple execution environments.

https://www.youtube.com/watch?v=9BolA2WTH7U&ab_channel=chrisbiscardi


gbj:
Thanks; I mean isomorphic in a slightly different sense, which I should probably clarify in the Readme, but which is more specific than "universal." I'll edit my version slightly. Thanks for bringing it up!

PR #67, opened 2022-11-09
Amend docs on feature flags

dglsparsons:
It looked a bit confusing that there were 3 `serde` feature flags and all of them were defaults.


gbj:
Thanks—copy and paste typo from me, I'm sure!

Issue #72, opened 2022-11-11
Migrate examples to single-folder structure


gbj:
Some of the examples (that use SSR + hydration) are divided into three directories: one for the server, one for the client, and one for the app. This used to be necessary to allow you to compile the hydration code to Wasm and the SSR code for the server without causing feature conflicts.

This problem no longer exists because the feature flags are no longer contradictory. So the `hackernews` and `counter_isomorphic` examples can be rewritten to use a single-create/single-directory structure instead.

See the model [here](https://github.com/gbj/leptos-sfa).

benwis:
This seems like a perfect starter task for me, I'll do this while I create an Axum example.

gbj:
Awesome. Thanks so much!

PR #76, opened 2022-11-14
Updated hacker_news and counter-isomorphic to SFA format, fixed Router example, and added some READMEs

benwis:
Working on #72, and have moved the Hacker News and counter_isomorphic example to the SFA format. Fixed the router example so that it runs. Added some READMEs. Take a look and let me know if I missed anything obvious. I'll move the isomorphic counter, add an Axum example, and a bunch of READMEs if it looks good


examples/hackernews/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+
+            view! { cx, <App/> }
+        };
+
+        let head = r#"<!DOCTYPE html>
+            <html lang="en">
+                <head>
+                    <meta charset="utf-8"/>
+                    <meta name="viewport" content="width=device-width, initial-scale=1"/>
+                    <script type="module">import init, { main } from '/pkg/hackernews_client.js'; init().then(main);</script>"#;
```
gbj, line 42: `/pkg/hackernews_client.js` here should probably be `/pkg/leptos_hackernews.js`

benwis: Yep, missed that one. Corrected!

gbj:
This is all passing on my local so I'm going to merge it. Thanks for your work here.

Issue #77, opened 2022-11-14
Nested route rendering bugs in `router` example


gbj:
1. `<Suspense/>` rendering broken: "Loading contacts..." never goes away
2. Navigating to other than the route root concatenates a node to the page, rather than replacing the root

gbj:
Okay I've fixed 1 but it's clear there's an issue with the rendering of `<Outlet/>` when you have nested routes.

PR #78, opened 2022-11-15
Fix rendering issues, including in router example

gbj:
Per issue #77 

There are a few issues that have cropped up w.r.t. the ordering of children within elements recently. This is trying to systematically fix some of those issues.


gbj:
I'm going to take this in two parts, fixing the more urgent rendering issue mentioned in #79 now and dealing with half of #77, and then dealing with the `<Outlet/>` bug that's the rest of #77 in a separate PR.

PR #86, opened 2022-11-17
Working Axum Example!

benwis:
I wanted to get the Hacker News Axum example up for review. It seems to be working completely, except for the small performance issue @gbj mentioned with the streaming of suspense objects


gbj:
Tested and this passes everything with flying colors. 

Later I'll try to refactor so we can hide some of the ugliness of spawning that other process. I'm still curious about the performance implications there but I honestly am weak on that side of things (i.e., life outside the browser...)

PR #89, opened 2022-11-18
Add ability to get/set signals untracked

### LINKED ISSUE #85: Consider adding the ability to get/set a signal's value without being tracked
It can be somewhat easy to end up in an infinite reactivity loop when accessing/setting signals. Often times, it is desirable to access/set a value silently, without triggering that signal's dependents. This should be possible on all signal types, such as `ReadSignal`, `Memo`, etc. If [`ReadSignal::derive` mentioned here](#84) were to be implemented, it would make this even more useful, as I could imagine it not being possible to silently get a value if using just a plain `impl Fn() -> _`.

jquesada2016:
Closes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ Leptos is a full-stack, isomorphic Rust web framework leveraging fine-grained re
 Here are some resources for learning more about Leptos:
 
 - [Examples](https://github.com/gbj/leptos/tree/main/examples)
-- [API Documentation](https://docs.rs/leptos/latest/leptos/) (in progress)
+- [API Documentation](https://docs.rs/leptos/latest/leptos/)
+- [Common Bugs](https://github.com/gbj/leptos/tree/main/docs/COMMON_BUGS.md) (and how to fix them!)
 - Leptos Guide (in progress)
 
 ## `nightly` Note
@@ -107,6 +108,7 @@ The gold standard for testing raw rendering performance for front-end web framew
 ### Can I use this for native GUI?
 
 Sure! Obviously the `view` macro is for generating DOM nodes but you can use the reactive system to drive native any GUI toolkit that uses the same kind of object-oriented, event-callback-based framework as the DOM pretty easily. The principles are the same:
+
 - Use signals, derived signals, and memos to create your reactive system
 - Create GUI widgets
 - Use event listeners to update signals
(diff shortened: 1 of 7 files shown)


jquesada2016:
`get` is implemented, just need to do `set`

gbj:
Thanks! I would love some tests... I need to go to bed but if you want to add some you can model them on the ones [here](https://github.com/gbj/leptos/blob/main/leptos_reactive/tests/effect.rs) and use the untracked setters/untracked getters to make sure effects aren't rerunning.

Otherwise I can add tests in the morning and merge if it all looks good.

jquesada2016:
@gbj I was going to try and impl `UntrackedGettableSignal` for `Signal` and `MaybeSignal`, but I don't think this is possible. I wonder if there might be another way to go about doing this? Such as perhaps adding some kind of property to the scope that says "Hey, we are going into an `untracked` state, any subscriptions that come in, ignore them", which may work universally for any signal kind?

gbj:
I agree that it's not possible on `Signal` and `MaybeSignal` because there's no way to do an untracked get on a derived signal (as opposed to doing untracked gets of the signals within that derived signal, if you see what I mean) 

The more general solution already exists with [`Scope::untrack()`](https://docs.rs/leptos/0.0.16/leptos/struct.Scope.html#method.untrack), which can run an arbitrary closure without tracking anything.

Having separate untracked get/set on signals makes sense, because
a) it's more ergonomic to write `a.get_untracked()` than to write `cx.untrack(move || a.get())`
b) it's faster to use `get_untracked` as you implemented it than to go through the work for `Scope::untrack`

jquesada2016:
I never saw the `Scope::untrack()` method, but we can use this meanwhile which should work for `Signal` and `MaybeSignal`?

jquesada2016:
For example, something like this:

```rust
fn get_untracked(&self) -> T
    where
        T: Clone,
    {
        match &self.0 {
            SignalTypes::ReadSignal(s) => s.get_untracked(),
            SignalTypes::Memo(m) => m.get_untracked(),
            SignalTypes::DerivedSignal(f) => {
                let f = f.to_owned();
                let val = Rc::new(RefCell::new(None));

                let val_clone = val.clone();

                let disposer = create_scope(move |cx| {
                    *val_clone.borrow_mut() = Some(cx.untrack(|| f()));
                });

                disposer.dispose();

                let mut val_borrow = val.borrow_mut();

                val_borrow.take().unwrap()
            }
        }
    }
```

Would this actually work? I would hope yes, but not sure if the `Scope::untracked` would actually not track the get since it belongs to a parent scope.

gbj:
I haven't tried it but I'd be very surprised if that worked; `create_scope` there is creating an entirely new reactive system, so I don't think you'll actually be able to access the values that are within the derived signal. It would probably panic as you're trying to access signal IDs that don't exist.

The best way to do this would actually be to change `Signal::derive` so that it's `fn derive<T>(cx: Scope, derived_signal: impl Fn() -> T) -> Signal<T>` and change `Signal` so that it has a variant `DerivedSignal((Scope, Rc<dyn Fn() -> T>)`.

Then you could just do

```rust
fn get_untracked(&self) -> T
    where
        T: Clone,
    {
        match &self.0 {
            SignalTypes::ReadSignal(s) => s.get_untracked(),
            SignalTypes::Memo(m) => m.get_untracked(),
            SignalTypes::DerivedSignal((cx, f)) => cx.untrack(f)
        }
    }
```

I just checked and the docs for `create_scope` are completely wrong; I must have mixed them up with the `child_scope` docs at some point and not noticed.

gbj:
> @gbj we should be good to go, although I haven't written any tests yet

Thanks, I'm fixing the doctests and adding some tests, and will push and then merge if CI passes.

gbj:
All set. Thanks so much for the PR!

Issue #104, opened 2022-11-21
Add Binary Encoding as an Option for Server Functions


benwis:
So I've been digging through the leptos code, and I think I understand how server functions are processed in Leptos.
I'd like to add MessagePack as an option to encode arguments, so that binary data payloads can be more efficiently sent from client to server. I see application for this in image processing, encryption, and other things. This would also prevent us from having to use an action or handling the client side function separately.

As I understand it, the server macro in leptos_macro is responsible for annotating the function with the ServerFn trait, which creates the fn calls from preexisting boilerplate for the client and the server.

So the way I see it, I could do this like so.
1. Add a toggle to the server macro to be encoded with MessagePack
2. Edit the call_server_fn to use the different serializer on the client
3. Edit the register() function to recognize the different serializer and process it differently
4. Create an example of handle_server_fns and an example using that structure.

Does that sound reasonable? I'm debating trying to make it generalizable, but I think MessagePack is sufficient

gbj:
Yep very reasonable. I think it makes a *ton* of sense to do it in the macro, rather than as a feature, because then you can mix and match MessagePack for one type and a URL-encoded form for another. (The form thing is good in the default case because you can use it to create an actual `<form>` that just works.)

So in your steps:

1. Add a toggle to the server macro to be encoded with MessagePack. *Perfect.*
2. Edit the call_server_fn to use the different serializer on the client. *Exactly.*
3. Edit the register() function to recognize the different serializer and process it differently. *Even better: edit `leptos::call_server_fn` in `leptos_server` to take a `Content-Type` header. Use the server macro to set this to `application/messagepack` or whatever the right mime-type would be in the line where it invokes `::leptos::call_server_fn`, depending on whether you're in form or MessagePack mode. Then the server function handler on the server can just listen for a `Content-Type` header and decode using form encoding or MessagePack, as appropriate.*
4. Create an example of handle_server_fns and an example using that structure. *See 3.*

This is a great idea. Do you want to take a crack at it with a PR?

(I'm not touching any of `leptos_server` or `leptos_macro` with this other thread-local/memory-leak thing, so there shouldn't be any conflicts)

benwis:
I'll take a stab at it tomorrow. The macro is the weakest part for me, but I should figure out something

benwis:
Also looks like I'll have to rework Serializable to include both

gbj:
[...]

PR #107, opened 2022-11-22
Binary encoding as an option for server functions

benwis:
I wanted to get this up, so that it can be reviewed. I think it mostly meets the structure defined in #104 but I'm super new to macros. Here are the issues as I see them.

1. Encoding adds a lot of imports to leptos_server, not sure we want to do that there or move it into leptos_macro
2. cargo run produces an error about running a wasm_bindgen function outside wasm. Not sure why.

Other than that, this looks solid.


gbj:
Cool! I like this. I've been taking a look tonight and will do some more work in the morning.

gbj:
So I was running into a little trouble getting a working example going. It sounds like you have an example you're testing on... Would you mind adding it to the repo so I'm working on the same test example as you are?

benwis:
Sure, let me polish it a bit

benwis:
> So I was running into a little trouble getting a working example going. It sounds like you have an example you're testing on... Would you mind adding it to the repo so I'm working on the same test example as you are?

I added an example, todo-app-cbor, and decided to switch from MessagePack to CBOR. I think the remaining problem is in the output of the `run-sever_fn` var in `register()` that returns a `Pin<Box<dyn Future<Output = Result<String, ServerFnError>>>>`

It would appear I need to generalize that Box to contain an Enum instead of a Srtring, but I am unsure if that will have unforeseen consequences

gbj:
The GitHub CI ran out of disk space but it passed for me locally, so I'm merging. Great work and thanks again!

PR #108, opened 2022-11-23
Assign correct ids in `add_many_counters`

jasonrhansen:
The counters and counters-stable examples didn't assign the correct ids when clicking the "Add 1000 Counters" button if there were already counters added, which meant clicking the "x" to remove them would remove the wrong counter.


gbj:
Thanks!

Issue #115, opened 2022-11-23
Router sometimes exponentially renders routes


jquesada2016:
I've noticed that sometimes, the router seems to load the same route multiple times, especially when visiting the same route. In general, the more times you navigate to the same page, the more the number of re-renders. At minimum, it seems linear, but I can go up exponentially sometimes. Here's a repro:

```rust
use leptos::*;
use leptos_router::*;
use wasm_bindgen::prelude::*;

#[wasm_bindgen(start)]
pub fn start() {
  console_error_panic_hook::set_once();

  mount_to_body(view_fn);
}

fn view_fn(cx: Scope) -> impl Mountable {
  provide_context(cx, RouterIntegrationContext::new(BrowserIntegration {}));

  let rw_loaded = create_rw_signal(cx, false);

  provide_context(cx, rw_loaded);

  wasm_bindgen_futures::spawn_local(async move {
    rw_loaded.set(true);
  });

  view! { cx,
    <span>
      <Router>
        <A href="/">"Come Home"</A>
        <main>
          <Routes>
            <Route path="/" element=home />
            <Route path="/settings" element=settings_guard>
              <Route path="/" element=settings_home />
              <Route path="/me" element=me />
              <Route path="/group" element=group />
            </Route>
          </Routes>
        </main>
      </Router>
    </span>
  }
}

fn home(cx: Scope) -> Element {
  gloo::console::debug!("rendering home page");

  let nav = use_navigate(cx);

  view! { cx,
    <button on:click=move |_| nav("/settings", Default::default()).unwrap()>"Settings"</button>
  }
}

fn settings_guard(cx: Scope) -> impl IntoChild {
  gloo::console::debug!("rendering settings guard page");

  /* Guard this page */

  Outlet(cx, OutletProps {})
}

fn settings_home(cx: Scope) -> impl IntoChild {
  gloo::console::debug!("rendering settings page");

  let loaded = use_context::<RwSignal<bool>>(cx).unwrap();

  let nav = move |href| {
    let nav = use_navigate(cx);

    nav(href, Default::default()).unwrap()
  };

  move || {
    loaded().then(||view! { cx,
    <>
      <h1>"All Settings"</h1>
      <div><button on:click=move |_| nav("/settings/me")>"Me"</button></div>
      <div><button on:click=move |_| nav("/settings/group")>"Group"</button></div>
    </>
  })
  }
}

fn me(cx: Scope) -> impl IntoChild {
  gloo::console::debug!("rendering me page");

[...]

PR #117, opened 2022-11-24
Fix issue #115

gbj:
For the moment, this seems to be work to limit the number of `<Outlet/>` rerenders to 2, rather than the linearly or exponentially increasing number of rerenders.

I'll work a little more on getting that down to 1 where it should be.


gbj:
As you can see from the commits I had to ping-pong this one around a LOT to get unnecessary rerenders to 0, but it turns out it was basically a logic error in `<Outlet/>` in addition to the disposers. I don't *think* this breakings routing in general but if it does, please open another issue...

PR #126, opened 2022-11-28
Mostly working version of axum with server functions

benwis:
Created an Axum version of todo-sqlite with Axum to demonstrate server functions. It seems to work, but there are some weird router and render issues. What does `use_router()` even mean?

I also commented out adding the request to provide_contex() because it's not Clone and Axum won't let you get away with using it twice. We could edit that just to be the headers.


gbj:
Thanks so much! I ran out of steam to check this tonight but looks like you were still fixing things. Will take a look in the morning. Really appreciate your submitting it.

gbj:
There's an issue in the server function macro implementation under weird circumstances (when all features are disabled) that I need to figure out but which is unrelated to this. Merging and I'll release a new `leptos_axum`.

PR #127, opened 2022-11-28
Fix path deps' going one level too high

akesson:
The path dependencies go one step too high, meaning that they include the name of the repo. This won't compile since I named the repo *gbj_leptos*. This PR fixes it.


gbj:
Thanks!

PR #128, opened 2022-11-28
Workspace features

akesson:
Using workspace-level defined version and [dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacedependencies-table) in order to make sure that the versions match and for reducing the work needed for creating a release.

The `main` branch doesn't compile, so I have not been able to test it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ members = [
 ]
 exclude = ["benchmarks", "examples"]
 
+[workspace.package]
+version = "0.1.0-beta"
+
 [profile.release]
 codegen-units = 1
 lto = true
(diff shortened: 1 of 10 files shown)


gbj:
Thanks! This seems great. I will fix the merge conflicts and give it a test in the next couple days.

gbj:
I'm really bad at resolving merge conflicts correctly and I think you know more about Cargo configuration than I do... when you have some time could you try resolving these and then I'll check it out and run the CI on it on my machine? Not a rush.

akesson:
Ok, so I finally rebased it on main.

There is now only one version defined for the libs: the one specified in the root workspace Cargo.toml file. If this is not desired, then it's easy to go back to using package-specific versions.

Equally, all local dependencies are defined at the workspace level for ease of versioning.

One important exception is all the examples since they are excluded from the workspace.

Issue #129, opened 2022-11-28
Server resource called 3 times on ssr render


saolsen:
If you run the todo-app-sqlite example there's a print statement in the `get_todos` function that shows this happening.

On a server side render of the page you will see 3 logs of `req.path = '/'` but if you add or a delete a todo, which re-fetches the resource, it'll only log `req.path = "/api/examples-todo-app-sqlite-src-todo.rs-get_todos"
` once.

benwis:
The same thing happens on the Axum versions

gbj:
Thanks!

For anyone curious, this was actually applying to *any* `Resource` being run on the server.
1) an optimistic attempt to check whether the `Future` was synchronously ready and could be rendered, which I forgot would also drive the actual async work
2) running the `Future` normally
3) running the `Future` a separate time to serialize it to JSON and send it down to the client

I removed #1, fixed #3 so it was working with the reactive value being stored in the resource by #2, and left #2 so that now every resource just runs once.

Good eye, and thanks for the report.

PR #131, opened 2022-11-29
Fix issue in which server-side resource are called 3x

gbj:
See issue #129


gbj:
This passes locally, although CI continues to have issues on GitHub.

Issue #130, opened 2022-11-28
Router panic


akesson:
```rust
#[component]
pub fn App(cx: Scope) -> Element {
    view! { cx,
        <div id="app">
          <Router>  
            <div id="WITH THIS DIV HERE THERE'S A PANIC">
               <ContentLayer />
            </div>
          </Router>
        </div>
    }
}

#[component]
pub fn ContentLayer(_cx: Scope) -> Element {
    view! {
        _cx,
        <div class="layer content">
           <Routes>
              <Route path="/doc/about" element=move |cx| view! { cx,  <div>"about"</div> } />
           </Routes>
        </div>
    }
}
```

From what I can get from the stack trace it's happening here:
> $leptos_router::components::routes::Routes::{{closure}}::h6cc2a448d4d924f1

akesson:
Full stack: [localhost-1669653084164.log](https://github.com/gbj/leptos/files/10105795/localhost-1669653084164.log)

gbj:
Okay so my impression is that this is strange behavior that has to do with the order of evaluation of element children in hydration mode, and therefore executes `<ContentLayer/>` before executing `<Router/>`, and therefore not having access to the context `<Router/>` is supposed to provide.

If you need a very shady temporary fix to unblock you, or if anyone else comes across this issue in the meantime, it does not panic if `<Router/>` has another child that precedes `<ContentLayer/>`, like this

```rust
view! { cx,
        <div id="app">
          <Router>  
            <div></div>
            <div><ContentLayer /></div>
          </Router>
        </div>
    }
```

Obviously I'll work on actually fixing it.

akesson:
Thanks for the tip!

PR #133, opened 2022-11-29
Fix issue #130

gbj:


gbj:
This is the wrong brach...

PR #134, opened 2022-11-29
Fix out-of-order hydration issue

gbj:
Fixes issue #130


gbj:
Passed local testing. Still having trouble with Github CI.

PR #142, opened 2022-11-30
Update example lib.rs

FDiskas:


gbj:
Nice catch, and thanks for the PR!

PR #149, opened 2022-12-02
Allow styling `<A/>` tags with `class` property

gbj:


gbj:
CI passed but ran out of space when caching. Merged!

PR #152, opened 2022-12-04
Add config crate and generate file for cargo-leptos to watch

benwis:
In an attempt to standardize config with cargo-leptos and use our integrations, I've modified leptos to have a central config for SSR, and to generate a KDL file with config params for cargo-leptos to read.

I also updated the todo-sqlite and todo-sqlite-axum to show what that would look like.

I'd love to get input on this structure. I haven't fixed the tests or most of the examples yet, so this is WIP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ blob.rs
 Cargo.lock
 **/*.rs.bk
 .DS_Store
+.leptos.kdl
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "leptos",
   "leptos_dom",
   "leptos_core",
+  "leptos_config",
   "leptos_macro",
   "leptos_reactive",
   "leptos_server",
--- a/examples/hackernews-axum/src/main.rs
+++ b/examples/hackernews-axum/src/main.rs
@@ -41,7 +41,7 @@ if #[cfg(feature = "ssr")] {
         // `GET /` goes to `root`
         .nest_service("/pkg", pkg_service)
         .nest_service("/static", static_service)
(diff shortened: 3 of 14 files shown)


benwis:
All right, I think this is ready for merging. I've updated the rest of the examples and made the integration tests pass. I'd like to wait until @akesson has a chance to review it as working for cargo-leptos

akesson (commented):
Looking great. Only a few comments:
- Could the socket reload JS be factored out into a single file? Maybe inside of *leptos_config*
- To me, the *.leptos.kdl* file is a compiler artefact and should thus be inside of the target folder.
- Wouldn't the *asset* dir be something that would be nice to have in RenderOptions? It would be called *static*, in the web world, but that conflicts with the Rust keyword.

benwis:
I thought about having the JS code in a separate file, but I don't want to do a file read on every request. I could put in a lazy_static though. I'm not sure how configurable things need to be, ideally we'd extend MetaContext for most things in head.
The trouble with the target folder is that folder changes location based on whether it's in a workspace or not. It's either in the crate root or the workspace root, and that's a bit tricky

akesson:
Regarding the JS: it's just to avoid having the same JS in two different places. 

Regarding the target folder: When doing the parallel build I'll have to set the target folder explicitly when running `cargo build` so the issues that arise in a workspace setup are something I will have to face anyway. But ok, I see that there's no reasonably easy way to know where the target folder is without pulling in things like `cargo-metadata`.

benwis:
@gbj After talking with @akesson I think this is ready to merge.
The crate root is a predictable place to have the `.leptos.kdl` file.
I'm OK with two copies of the head html/JS for now. There's a bunch of duplicated code in processing for the head(not just the html) that might warrant it's own spot, but it's not too bad yet.

PR #157, opened 2022-12-08
Integration html updates

akesson:
For the hard-coded html in the integrations:
- Added a missing _wasm path_ param to init.
- Added preloading of wasm and js.

@benwis look ok to you?


benwis:
Yeah, looks good to me

PR #165, opened 2022-12-13
update_returning handlers

Gentle:
added update_returning variants to all Signal types, update_returning runs the closure and then returns its return value. Since the closure isn't guaranteed to be run, update_returning has to return Option<U>


gbj:
Oops I forgot to merge this after the tests passed. Merging now, thanks so much for your contribution!

PR #168, opened 2022-12-14
CF worker example

gbj:


gbj:
There's a better example [here](https://github.com/perry-birch/leptos-cloudflare).

PR #175, opened 2022-12-21
fix warning, set initial value of counter

ultrasaurus:
trying out `leptos` for the first time, I noticed a warning in this example, which pointed to what appears to be a bug:

```
warning: unused variable: `initial_value`
  --> examples/counter/src/lib.rs:11:33
   |
11 | pub fn SimpleCounter(cx: Scope, initial_value: i32, step: i32) -> web_sys::Element {
   |                                 ^^^^^^^^^^^^^ help: try ignoring the field: `initial_value: _`
   |
   = note: `#[warn(unused_variables)]` on by default
```

I think this resolves the problem


gbj:
Thanks, nice one! Missed this when I updated it to add info about documenting components.

PR #176, opened 2022-12-21
info about how to set up nightly -> README

ultrasaurus:
I haven't used nightly much and at first couldn't get examples to run.  
I found helpful notes in the tailwind example README, and thought they
would be good to include in the main README for other newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ To set up your rustup toolchain using nightly and
 add the ability to compile Rust to WebAssembly:
 
 ```
-rustup toolchain install nightly --allow-downgrade
+rustup toolchain install nightly
 rustup default nightly
 rustup target add wasm32-unknown-unknown
 ```
--- a/README.md
+++ b/README.md
@@ -76,7 +76,18 @@ Here are some resources for learning more about Leptos:
 
 ## `nightly` Note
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,7 +76,18 @@ Here are some resources for learning more about Leptos:
 
 ## `nightly` Note
 
-Most of the examples assume you’re using `nightly` Rust. If you’re on stable, note the following:
+Most of the examples assume you’re using `nightly` Rust.
+To set up your rustup toolchain using nightly and
+add the ability to compile Rust to WebAssembly:
+
+```
+rustup toolchain install nightly --allow-downgrade
```
afiqzx, line 84: Is `--allow-downgrade` needed here?

gbj: Yeah I think I agree; @ultrasaurus I'm assuming `--allow-downgrade` snuck in because you copy and pasted it from somewhere (`tailwind` example?) and it's there because I copy and pasted it from somewhere. But I agree with the feedback that it's probably not necessary. Unless there's a good reason for it to be there, could you just remove it and then I'd be happy to merge this!

ultrasaurus: you are correct -- copied from tailwind example.  I'll remove

gbj:
Awesome -- thanks very much for your contribution!

PR #177, opened 2022-12-21
Changes to leptos, leptos_meta, leptos_actix/leptos_axum, and leptos_config to support cargo-leptos-beta

benwis:
Hi @gbj and @akesson, I have made several changes to leptos to support the cargo-leptos beta.

I hope both of you can take some time to review this, especially @akesson since it needs to work with cargo-leptos. I think I have it working with my stack, but would love your input.

Changes:
- Add id prop to `<Stylesheet/>` component to allow cargo-leptos' new css feature to work
- Rename RenderOptions to LeptosOptions and delete KDL file writing
- Read LeptosOptions from Cargo.toml in the [package.metadata.leptos] section, sharing keys with cargo-leptos
- Conditional WASM filename based on PACKAGE_NAME env var. If deploying a bundle built with cargo-leptos, PACKAGE_NAME env var will need to manually set
- Allowed Server Functions to set status of form redirects

This should allow for one source of truth for people using cargo-leptos or wasm-pack, and streamline some things.

It appears that cargo-leptos beta also does not support workspaces, as a bunch of paths are relative to the cargo root, which changes to the workspace root when in a workspace. This will prevent modifying the examples in leptos(and having them run with cargo-leptos watch), at least with the present example format. It shouldn't deter operation with wasm-pack


examples/counter-isomorphic/Cargo.toml
```diff
@@ -42,3 +42,28 @@ ssr = [
...
+# The IP and port (ex: 127.0.0.1:3000) where the server serves the content. Use it in your server setup.
+site_address = "127.0.0.1:3000"	
+# The port to use for automatic reload monitoring
+reload_port = 3001	
+# [Optional] Command to use when running end2end tests. It will run in the end2end dir.
+end2end_cmd = "npx playwright test"
+#  The browserlist query used for optimizing the CSS.
+browserquery = "defaults" 	
+# Set by cargo-leptos watch when building with tha tool. Controls whether autoreload JS will be included in the head
+watch = false
```
akesson, line 67: The watch parameter is managed by `cargo-leptos` and should not be user-visible.

akesson (commented):
I really like what you did with the Config and TypedDerive for the LeptosOptions.

Why do you need to read the `Cargo.toml` file? It's fine now, but with workspace support, you'll have a difficult (read impossible) time finding the right one. I'm implementing `cargo-leptos` command-line arguments and workspace config options for choosing which project to run.

The live-reload JS should be:

```rust
r#"
<script crossorigin="">(function () {{
    var ws = new WebSocket('ws://{addr}/live_reload');
    ws.onmessage = (ev) => {{
        let msg = JSON.parse(event.data);
        if (msg.all) window.location.reload();
        if (msg.css) {{
            const link = document.querySelector("link#leptos");
            if (link) {{
                let href = link.getAttribute('href').split('?')[0];
                let newHref = href + '?version=' + new Date().getMilliseconds();
                link.setAttribute('href', newHref);
            }} else {{
                console.warn("Could not find link#leptos");
            }}
        }};
    }};
    ws.onclose = () => console.warn('Live-reload stopped. Manual reload necessary.');
}})()
</script>
"#
```

examples/counter-isomorphic/Cargo.toml
benwis: Yeah, I don't think this needs to be a param

benwis:
> Why do you need to read the `Cargo.toml` file? It's fine now, but with workspace support, you'll have a difficult (read impossible) time finding the right one. I'm implementing `cargo-leptos` command-line arguments and workspace config options for choosing which project to run.

Mostly because if cargo-leptos is not used, then I need it to work with wasm-pack without any env vars set. I'm not sure about workspaces, but the Cargo.toml file still exists in the crate in a Workspace, and I reference it relative to the crate.

The examples seem to work, but it'll need more testing

benwis:
> > I'm not sure about workspaces, but the Cargo.toml file still exists in the crate in a Workspace, and I reference it relative to the crate.
> 
> I'm saying that this is not necessarily a valid assumption when running in a Workspace. The workspace support I'm working on adds the possibility to split the server project and the front-end project into two different Cargo packages. Besides, when running from the workspace root, I'm not sure if I will/can change the current working directory to the sub-project.

Fair point. I've added an option to specify the path to Cargo.toml in get_configuration(). But it's use isn't required, a user could still build a `LeptosOptions` with `Default::default()` and the Builder,  and skip the conf file reading entirely. This default approach is just to make it easier for the two build methods to coexist.

benwis:
@akesson @gbj I've updates the examples and config once more to support the cargo-leptos beta. With this, it should be ready to merge again. Happy to have @akesson add features to it.

PR #185, opened 2022-12-24
Fix build errors in `counter-isomorphic` when using `stable`

snapbug:
I'm investigating using this framework, but prefer to use stable Rust, so just a couple of fixes to get the `counter-isomorphic` example building when enabling the `stable` feature.

Tested by enabling stable in the [counter isomorphic deps](https://gist.github.com/snapbug/49e0ad2512511e2ef86c85d9ea761a37), and building. Checked functionality remains correct for all the methods shown in both Safari and Chrome.

Note: I don't know that editing the example is necessarily a completely desireable fix, given the separation of `counter` and `counters-stable`, and I also noted in router/src/components/link.rs that there is a cfg_if separation of csr or hydrate and not, whereas components/form.rs also has a stable/not cfg_if.


gbj:
Thanks very much!

PR #186, opened 2022-12-25
Update COMMON_BUGS.md to fix typo

luckynumberke7in:
```rust
let (b, set_a) = create_signal(cx, false); // should be set_b
```


gbj:
Thanks!

Issue #163, opened 2022-12-10
leptos - tailwind - axum - cargo-leptos compatibility issue


ebadta81:
Hi,

There are some incompatibilities between ```leptos_axum::render_app_to_stream``` and cargo-leptos.

render_app_to_stream use the pkg_path variable to refer to the generated wasm, js files, while cargo-leptos use constants, like:

```
const DEST: &str = "target/site/pkg/app.css";
...
let wasm_path = "target/site/pkg/app.wasm";
...
fs::write("target/site/pkg/app.js", js).await.dot()?;
```

This way you always need to use "/pkg/app" for ```RenderOptions::builder().pkg_path```.

It would be nice, if you could improve axum, actix integration to use generated.rs, or  integrate cargo-leptos to the leptos project.

gbj:
You're correct. The development processes for `cargo-leptos` and for the Axum/Actix integrations overlapped and are slightly incompatible in this way. I understand this is being worked on in `cargo-leptos` and will be resolved in one of its upcoming updates but feel free to open an issue in the [`cargo-leptos` repo](https://github.com/akesson/cargo-leptos) or ask on Discord (link in the Readme in this repo)

benwis:
I have pushed some changes as part of #167 that will restore functionality for non cargo-leptos users. cargo-leptos is under development, and I will adapt it again once it's ready for the work

ankarhem:
@benwis

Still doesn't work though since cargo-leptos wants `app.wasm` and the changes changed it to `app_bg.wasm`?

benwis:
Yes, the changes were reverted to make manual compilation with trunk and cargo run work as expected. cargo-leptos is undergoing some changes with the goal of streamlining things and allowing both to work as expected

------- Original Message -------

Issue #122, opened 2022-11-26
Update examples so they're compatible with `cargo-leptos`


gbj:
We now have a great build tool released that supports things like synchronized client-server live-reloading. It would be great to update (some or all) of the examples, in particular the ones with both client and server code (`hackernews`, `counter-isomorphic`, and `todo-app-sqlite` and `todo-app-cbor` in particular) so you can use `cargo-leptos` to run them.

GreatGreatGreg:
And you can add tailwindCSS, the most weekly downloaded CSS now.  

Leptos is becoming best rust web framwork with great developer experience! 
1. Fast: Almost fastest rust web framework according to js-framework-benchmark.
2. HotReload: Instant UI developing feedback. It's already the fastest one.
3. TailwindCSS Intellisense: Some rust web framework do not work because of micro and DSL. Leptos works perfectly.

Thank you

PR #209, opened 2023-01-01
Update all examples to use cargo-leptos and patch integrations for latest cargo-leptos support

benwis:
This is the big one that should address #163 and #122. I've updated the actix and axum integrations to support the latest cargo-leptos release, and to still support wasm-pack.

All of the SSR examples should build with cargo leptos watch without any changes, and the READMEs have been updated to mention the new install method.

Huge thanks to @akesson for developing cargo-leptos!


gbj:
Yay! This is awesome.

I notice that in the `todo_app_sqlite`, `todo_app_sqlite_axum`, `hackernews`, and `hackernews_axum` examples there are CSS files included in the source (usually just `style.css` in the root of the example) but `style-file` isn't included in `[package.metadata.leptos]` Should those be added?

Otherwise looks good — I tested all the examples using `cargo-leptos` and they work great.

I'm pretty sure I can add `cargo make build` CI support back in using `cargo make` task dependencies, which I can do in a separate PR. 

So, yeah, if you want to add the CSS files (or tell me why not, either way) I'm happy to merge this. And great work!

benwis:
Yeah, I hadn't changed them over to use cargo-leptos. The last commit should have done so and fixed a few bugs. Should be ready to go!

PR #220, opened 2023-01-02
added on:eventname:undelegated support

jquesada2016:


jquesada2016:
@gbj What'd I miss here? I see CI is failing to pull this?

gbj:
Sorry this is the CI, not you... sometimes it fails because it already has `cargo-all-features` installed which I haven't quite been able to understand. I'll check out your copy and run the CI.

My only concern has to do with `parse_event` — I think the requirement to be `on:event` or `on:event:undelegated` is too strict. For example, `greglibrary:my-event` is a valid ([and sometimes even recommended](https://gomakethings.com/custom-event-naming-conventions-in-vanilla-js/)) custom event name, which this would panic on.

How about using [String::strip_suffix](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_suffix) instead, such that *any* attribute name starting with `on:` and ending with `:undelegated` would be `force_undelegated`, and any other attribute name starting `on:` is an event that's not `force_undelegated`?

jquesada2016:
Sorry, I didn't see the rest of this message, only the CI part!

Yeah, totally fine by me! I didn't know that `:` could be used in event names!

jquesada2016:
Alrighty, we should be good to go now.

The only concern I have now is that if the user adds a typo, they won't get the event they want...ideally, we would emit a warning if the name is close, but not exactly sure on how to go about doing this, and the tolerances, on top of not knowing how to support silencing the warning

We can get around this if we add modifiers before `on:`, so `undelegated:on:blah-lah-lah`

gbj:
Great: thanks! Yeah, we can think about how to guard against this kind of typo in general, I guess, but it's a little hard...

jquesada2016:
It is hard if we do it the way it was committed here, but not if we mave the modifiers _before_ the `on` keyword.

```rust
view! { cx,
  <button undelegated:on:click=|_| {} />

  // This won't compile, because undeligated is not a valid modifier
  <button undeligated:on:click=|_| {} />
}
```

gbj:
Well it's a little awkward because `undeligated:on:click` is a valid HTML attribute name, right? So we'd have to make the choice that any attribute name that contains `:on:` can only be an event listener with modifiers.

jquesada2016:
That is a very good point...we don't have many parsing options for this, as `syn-rsx` is currently limitted here :( 

we could potentially create our own parsing logic with the following syntax, because `syn-rsx` supports arbitrary expressions within `{}` on attribute positions, but it looks ugly, imho

```rust
view! { cx,
  <button {on:click|undelegated} />
}
```

or whatever other syntax we would want

jquesada2016:
Actually, now that I come to think of it, `on:eventname` is also a valid attribute...I say it may be fine to assume `:on:` in an attribute means it's an event listener. The user can always add it the attribute in manually if this is not the case.

PR #228, opened 2023-01-03
Add a redirect() function and some helpful utility functions for ResponseParts and ResponseOptions

benwis:
Inspired by your video and Remix's `redirect()`, I've added some helper functions to ResponseParts/ResponseOptions, and a handy `redirect()` function for easy redirection in server functions.

So this code
```rust
 let response =
        use_context::<ResponseOptions>(cx).expect("to have leptos_actix::ResponseOptions provided");
    let mut response_parts = ResponseParts::default();
    let mut headers = HeaderMap::new();
    headers.insert(
        SET_COOKIE,
        HeaderValue::from_str(&format!("darkmode={prefers_dark}; Path=/"))
            .expect("to create header value"),
    );
    response_parts.headers = headers;

    std::thread::sleep(std::time::Duration::from_millis(250));

    response.overwrite(response_parts).await;
```
could be this now:
```rust
let mut response = use_context::<ResponseOptions>(cx).expect("to have leptos_actix::ResponseOptions provided");
response.insert_header( SET_COOKIE, HeaderValue::from_str(&format!("darkmode={prefers_dark}; Path=/")).expect("to create header value"),)
std::thread::sleep(std::time::Duration::from_millis(250));
```

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -21,8 +21,8 @@ impl ResponseParts{
     pub fn insert_header(&mut self, key: header::HeaderName, value: header::HeaderValue){
         self.headers.insert(key, value);
     }
-     /// Insert a header, overwriting any previous value with the same key
-     pub fn append_header(&mut self, key: header::HeaderName, value: header::HeaderValue){
+    /// Append a header, leaving any header with the same key intact
+    pub fn append_header(&mut self, key: header::HeaderName, value: header::HeaderValue){
         self.headers.append(key, value);
     }
 }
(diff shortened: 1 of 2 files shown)


gbj:
This looks really awesome! Three comments:
1. The docs for `insert_header` and `append_header` are identical, which I imagine is probably not intentional: maybe one overwrites existing header (`insert_header`?) and one doesn't? (`append_header`?)
2. Maybe `.status()` should be `.set_status()`? I know it takes `&self` because of the interior mutability, but `&self.status()` feels like it should return the current status, not update it. 
3. Oh and `.status()` also has that same identical doc comment.

benwis:
Curse VsCode, I did change those I swear.

gbj:
Haha beautiful, thanks.

Issue #226, opened 2023-01-03
Custom events seem to be broken


jquesada2016:
The following view:

```rust
view! { cx,
  <button on:blah-lah:dah=|_| {} />
}
```

expands to:

```rust
leptos::leptos_dom::button(cx).on(leptos::ev::Custom, |_| {})
```

which is incorrect, because `Custom` needs to be constructed.

PR #232, opened 2023-01-04
`leptos_macro` improvements to `class:`, `prop:`, `on:`, `:undelegate…

gbj:
…d`, and events


gbj:
Addresses #226, but I also went a little crazy on DX improvements by spanning to the correct helpers...

PR #241, opened 2023-01-05
Serve from site_root

akesson:
This PR changes so that the site is served from the `site_root`, without including it in the http links.

It is currently only applied to the hackernews example, as I would like to get an OK on this way of functioning before modifying the other examples.

So the idea is that: 

1. the `site_root` is basically the path to the folder that should be served. I.e. the root of the website. 
2. the contents of the `assets_dir` is synchronised to the `site_root`. Ex: a favicon that is placed in the root of the `assets_dir` ends up in the root of the `site_root` 
3. the server serves the app html only when html is requested (Accept header checked in the HtmlGuard)
4. the last service is the file service that serves whatever is in the `site_root`. That includes the `/pkg/*` stuff

@benwis I might have broken something so please review.


gbj:
Yeah I mean, this definitely seems better to me from the user perspective! (not having to have `/target/site/...` for CSS etc)`

benwis:
With #269, I added all your changes and a new proposed routing, except for setting the site_root default to ".", which gbj recommended we wait on until the check is added to cargo-leptos that prevents it deleting things.  If this checks out, we can close this PR.

gbj:
I'm going to close this per Ben's updates in #269, which made the same changes for the other examples while moving things over to the new routing system.

PR #242, opened 2023-01-05
impl `Default` for `SignalSetter<T>`

jquesada2016:


jquesada2016:
@gbj, I'm ready now for a review. To sum up, here was the original problem example:

```rust
#[component]
fn CheckboxA(
  cx: Scope,
  #[prop(optional)] checked: Option<Signal<bool>>,
) -> impl IntoView {
  view! { cx,
    <CheckboxB checked=checked.unwrap_or_default() />
  }
}

#[component]
fn CheckboxB(
  cx: Scope,
  #[prop(optional)] checked: Option<Signal<bool>>,
) -> impl IntoView {
  view! { cx,
    <input prop:checked=move || checked.map(|checked| checked()) />
  }
}
```

But now, I added `Signal::default(cx)`, so we can do instead in `CheckboxA`

```rust
view! { cx,
  <CheckboxB checked=checked.unwrap_or_else(|| Signal::default(cx)) />
}
```

I had to make a custom default method because I didn't see a way to impl this without adding a `Default` bound on `T`, which wouldn't work.

I was able to impl `Default` for `SignalSetter<_>`, which is good!

gbj:
Yep, this is a very smart way of doing this. Nice work! I'm going to merge.

PR #251, opened 2023-01-06
Bump actions/checkout to version@3

martinfrances107:
I was looking into the test failures.

I see out of memory errors.

I have seem similar errors in a unrelated project, and the solution was to bump the major version of actions/checkout to 3

This may or may not be the problem here ....


gbj:
Oh, thanks! The CI stuff is definitely my weak point.

PR #260, opened 2023-01-07
added `IntoSignal` and `IntoSignalSetter` helper traits

jquesada2016:
I'm not sure what name to give the `IntoSignalSetter` method...I thought that `.mapped_signal_setter()` might be appropriate because we already mapped a signal in the closure, and now just need to convert it to the right type. Lmk if you don't like it!


jquesada2016:
...and hopefully this time, CI passes!

gbj:
LGTM.

PR #264, opened 2023-01-07
Removing clone call where possible(round2).

martinfrances107:
Sorry for the noise, 

It seems with all these minor changes we have changed the way chippy perceives the code-base
(Earlier today I found only 2 places where .clone() was called without reason.)

Now under iteration I see a  few more places, so this is round two.


martinfrances107:
Had to force push, after rebase, keeping up with main 

First PR was generated mid way through multiple commits to the main branch yesterday.

PR #269, opened 2023-01-07
Generate Routes and pass them to Actix/Axum

benwis:
This PR provides a function to walk the Leptos router's app tree and produce a list of routes. Each integration has conversions to parse them into a format for either Actix or Axum, and then a trait to provide a `.leptos_routes()` extension for their respective routers.

This lets us avoid any wildcard routes for routing leptos! And allows you to define routes in only one place! We can do some server side error handling, and be a better neighbor for other services.

I also put in the changes from #241 which are simplified by this change. Take a look and let me know if you spot any issues. The Actix version especially needs some prodding, it's got some fairly involved regexs, as its pathing is quite different from leptos_router.

The todo_app_sqlite and todo_app_sqlite_axum extensions have been updated to the new format. I'll update the rest if it looks good

### DIFF
--- a/examples/todo_app_sqlite/src/main.rs
+++ b/examples/todo_app_sqlite/src/main.rs
@@ -31,7 +31,7 @@ cfg_if! {
             let addr = conf.leptos_options.site_address.clone();
             println!("BEFFOORE");
             let routes = generate_route_list(|cx| view! { cx, <TodoApp/> });
-            println!("HIIIIIIIIIIII");
+
             println!("Routes: {:#?}", routes);
 
             HttpServer::new(move || {
--- a/examples/todo_app_sqlite/src/main.rs
+++ b/examples/todo_app_sqlite/src/main.rs
@@ -1,5 +1,6 @@
 use cfg_if::cfg_if;
 use leptos::*;
+use leptos_router::generate_route_list;
 mod todo;
 
 // boilerplate to run in different modes
@@ -28,6 +29,10 @@ cfg_if! {
 
             let conf = get_configuration(Some("Cargo.toml")).await.unwrap();
             let addr = conf.leptos_options.site_address.clone();
+            println!("BEFFOORE");
+            let routes = generate_route_list(|cx| view! { cx, <TodoApp/> });
+            println!("HIIIIIIIIIIII");
+            println!("Routes: {:#?}", routes);
 
             HttpServer::new(move || {
                 let leptos_options = &conf.leptos_options;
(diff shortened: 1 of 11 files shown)


gbj:
CI is failing on this at the moment
```
error[E0433]: failed to resolve: use of undeclared crate or module `tokio`
 --> integrations/actix/src/lib.rs:8:5
  |
8 | use tokio::sync::RwLock;
  |     ^^^^^ use of undeclared crate or module `tokio`
```

akesson:
Some Actix feedback:
1.  Really neat with the `App::new().leptos_routes(...`. Great DX.
2. Makes you dream about having an `App::new().leptos_server_functions(...` as well 😄 
3. The app is now created twice: `let routes = generate_route_list(|cx| view! { cx, <TodoApp/> });` and `App::new().leptos_routes(leptos_options.to_owned(), routes.to_owned(), |cx| view! { cx, <TodoApp/> })` which is problematic for me because I pass in some rather heavy owned data at app creation i.e. `view! { cx, <MyApp mydata /> }`. Would it be possible to do it in one go? Probably not, but asking just in case...
4. ResponseOptions accessible from cx is really nice

benwis:
> Some Actix feedback:
> 
> 1. Really neat with the `App::new().leptos_routes(...`. Great DX.
> 2. Makes you dream about having an `App::new().leptos_server_functions(...` as well 😄
> 3. The app is now created twice: `let routes = generate_route_list(|cx| view! { cx, <TodoApp/> });` and `App::new().leptos_routes(leptos_options.to_owned(), routes.to_owned(), |cx| view! { cx, <TodoApp/> })` which is problematic for me because I pass in some rather heavy owned data at app creation i.e. `view! { cx, <MyApp mydata /> }`. Would it be possible to do it in one go? Probably not, but asking just in case...
> 4. ResponseOptions accessible from cx is really nice

Glad you like it! I was thinking about server functions, I think routes could be extracted from the `register()` function. Although I'm not super concerned, as it only requires giving up `/app` as a route.

Double rendering the app is a bit of a drawback. I could store the results of the render and pass that to all the functions, but I think that's probably unnecessary because we only pay the cost on startup once and rendering is fast. 

Sounds like I'm good to go on updating the other examples and docs, I'll get that up soon.

Not sure why CI is failing, leptos_actix is defined in the Cargo.toml for that example, and it compiles fine locally

gbj:
It's failing on `todo_app_sqlite` when `features = []`. Should `cargo-all-features` try to build `features = []` in our case? Probably not, but I haven't figured out how to stop it.

I'd guess `leptos_actix` is `optional = true` and gated on the `ssr` feature, but is imported somewhere that isn't gated. 

*Actually checks* Okay yeah move line 4 in `main.rs` down into the section that's behind the `cfg_if` gate for `ssr`.

benwis:
Yeah, caught that after I said it. I've updated all the examples. So far so good, I think it's just about ready to merge

PR #272, opened 2023-01-08
fix component macro lifetime parsing

DPM97:
image
image


gbj:
Oh thanks so much; I'm glad this was such a quick fix, appreciate your taking it on.

Issue #274, opened 2023-01-08
Policy change: Add a github/workflow for 'Cargo fmt'?


martinfrances107:
This a controversial,  everyone has their IDE setup slightly differently. This leads to  PR's  having slightly different tabs and spacing 

RUST being a new-ish language has 

```cargo fmt```

"love a rule" or "hate a rule"... is does not matter 
all discussions end with "is that standard or not?"  

I too have loves and hates BUT I like it only because it settles needless discussion

When I look at the code-base today, running "cargo fmt" creates a moderately sized PR/patch.

I propose a one time PR that fixes all the non-standard issues. 
A one time patch accompanied with a pre-commit hook that enforce the standard 

Nothing fancy just something of the form

[sample pre commit  hook](https://gist.github.com/hryniuk/e3f4ae34e37947d0ff55a3f63add90f8)

Anyway I thought I had better make this an issue, so others can express an opinion. I am new to this code-base .. and I don't to make waves..

As a discussion point against my proposal ... a PR of this form is bad for anyone using "git Blame". it means that they will have to iterate a little....that is why I want to make this a one time thing.

Separately, the discussion  in sample pre-hook gist, talks about including cargo clippy, my own view on this is Not to do it. Clippy has saved me several times but that is a step too far.

I will create a speculative PR so people can see what is what.

martinfrances107:
As a change to what I said above 

pure git hooks are problematic on github.

We can get the same effect with "github/workflows".

Gentle:
I would opt for running `cargo fmt --check` in CI so Pull Requests will show that formatting is off, but as you said commit hooks are too restrictive during active development in branches

martinfrances107:
This is stage 2 #294

gbj:
Thanks, this is done.

PR #275, opened 2023-01-08
Policy change: Workflow now enforce "cargo fmt".

martinfrances107:
#274 - I suggest enforcing "cargo fmt" before every commit.

This is a speculative patch ... if we get the go ahead for this rule change.


martinfrances107:
Hmm, I think the best way to proceed is to split this into 2 PRs

1) A PR after running "cargo fmt"
2) A second PR which enforces the policy change

```
       name: Run Rustfmt
        run: cargo fmt -- --check
```

This is still a speculative PR, subject to approval of the policy change... but I am just going to convert this to the first PR, just so other can see the typical formatting changes.

martinfrances107:
force push, as I chose to rebase ... next time I will merge I makes reviewing cleaner.

martinfrances107:
> Does cargo fmt do more or less or something else than cargo clippy?

No there are very different animals 

Cargo fmt ... is just tabs, whitespaces, brackets and indnetations.

Cargo clippy .. is more of a code analyser.. a linter  ... it has a comprehensive set of rules based on common misconceptions and code patterns that will lead to a sub-optimal binary.

It will for example spot when .clone() is being called when it does not need to be 

https://rust-lang.github.io/rust-clippy/master/#redundant_clone

This one can reduce the number of memory allocations

https://rust-lang.github.io/rust-clippy/master/#needless_pass_by_value

When I am using lifetimes when I don't need to  this rules kicks in 
https://rust-lang.github.io/rust-clippy/master/#needless_lifetimes

Here is the full list.

https://rust-lang.github.io/rust-clippy/master/

Hope that helps

PR #277, opened 2023-01-08
examples/hackernews - Cargo clippy fixes.

martinfrances107:
Previously I  have been sweeping up errors by running cargo clippy from the root directory.

Today I ran  "cargo clippy" from examples/hackernews.

The feature flags in this example change the build environment

running  "cargo clippy" here picks up a few more issues.

I WAS trying to fix issues class-by-class, but in this PR there are previously visited rules : -

Redundant clone issues.
clippy::uninlined_format_args

I hope that explains why some of these rules are re-appearing

Here is the clippy warning for a new ambiguous namespace issue found in the view macro

```
warning: `leptos_router` (lib) generated 3 warnings (run `cargo clippy --fix --lib -p leptos_router` to apply 3 suggestions)
    Checking hackernews v0.1.0 (/home/martin/build/leptos/examples/hackernews)
warning: The view macro is assuming this is an HTML element, but it is ambiguous; if it is an SVG or MathML element, prefix with svg:: or math::
  --> src/routes/stories.rs:51:30
   |
51 | ...                   <a class="page-link"
```

### DIFF
--- a/examples/hackernews/src/main.rs
+++ b/examples/hackernews/src/main.rs
@@ -1,5 +1,4 @@
 use cfg_if::cfg_if;
-use leptos::*;
 
 // boilerplate to run in different modes
 cfg_if! {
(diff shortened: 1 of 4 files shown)


gbj:
Thanks! All good.

PR #282, opened 2023-01-09
Use envs for workspace config

akesson:
When using workspaces there can be many different `[[workspace.metadata.leptos]]` sections. 

There is no way for the LeptosConfig to know which one is used and it cannot be set by the server binary package either as one of them can be used in many `[[workspace.metadata.leptos]]` sections.

This solution uses creates the LeptosConfig from the environment variables set by `cargo-lepos`. When running the server binary outside of `cargo-leptos` (in prod) the env vars `LEPTOS_SITE_ADDR`, `LEPTOS_SITE_ROOT` and `LEPTOS_SITE_PKG_DIR` can be set using alternative ways (bash script, .env etc).

@benwis I know you are heavily opposed to this solution, but I frankly don't see any other.


benwis:
Actually one of the reasons I set it up this way was that all the LeptosOptions are already overridden by env vars. You can find that here:
https://github.com/gbj/leptos/blob/main/leptos_config/src/lib.rs#L157

I also agree with your comment regarding workspaces, I put that in an earlier version. I added reading of that in an earlier version, and now it is dead code.

The only thing not overridden is output_name, but that can be set with a one-liner in the same place.

akesson:
I'm sorry, I misunderstood your intentions due to:

1. The current version throws a `LeptosConfigError::ConfigNotFound` if no _Cargo.toml_ file is specified.
2. It fails when there are more than one `[[workspace.metadata.leptos]]`
3. The `Config::builder()` fails if no `[leptos_options]` is found
4. The field `output_name` has no default (which makes sense).

In order to minimize the changes to your code, I chose to add an `fn try_from_env()` but I guess I can fake a toml section instead:
```rust
format!("[leptos_options]\n output_name = {}", env_output_name)
```

Would you mind if I make reading the Cargo.toml file optional? It feels odd to need to have a copy of it along with the binary in order to run the server.

akesson:
Actually, faking it doesn't seem to work well. Seems like the defaults in LeptosOptions don't work. I get: 
> ConfigError("missing field `site_root`")

I think the defaults only work for the TypedBuilder, not for the Config.builder()

benwis:
> I'm sorry, I misunderstood your intentions due to:
> 
> 1. The current version throws a `LeptosConfigError::ConfigNotFound` if no _Cargo.toml_ file is specified.
> 2. It fails when there are more than one `[[workspace.metadata.leptos]]`
> 3. The `Config::builder()` fails if no `[leptos_options]` is found
> 4. The field `output_name` has no default (which makes sense).
> 
> In order to minimize the changes to your code, I chose to add an `fn try_from_env()` but I guess I can fake a toml section instead:
> 
> ```rust
> format!("[leptos_options]\n output_name = {}", env_output_name)
> ```
> 
> Would you mind if I make reading the Cargo.toml file optional? It feels odd to need to have a copy of it along with the binary in order to run the server.

Not at all, I was considering a similar thing. Loosening the restrictions on the current function should make things easier for cargo-leptos

benwis:
> Actually, faking it doesn't seem to work well. Seems like the defaults in LeptosOptions don't work. I get:
> 
> > ConfigError("missing field `site_root`")
> 
> I think the defaults only work for the TypedBuilder, not for the Config.builder()

Ah yeah, the config create makes you specify the defaults all manually. It might be simpler to avoid it or go with your approach

benwis:
Latest version looks good to me, assuming it works for @akesson. Should give us maximum flexibility

akesson:
> Can you please change the env var name in the interrogations from OUTPUT_NAME to LEPTOS_OUTPUT_NAME?
Done

@gbj or @benwis please review my last commit "Default to serve root" https://github.com/leptos-rs/leptos/pull/282/commits/1f556cefb07aa8bad6594b4e5bab006c5726427f as I don't really know if this is correct. When I updated the `cargo-leptos` examples no routes were created. This is probably because the examples don't use any Routes and just display a message. 

Other than that, I'm good to go. The cargo-leptos examples are updated, but I need this PR to be merged before I can push those changes.

PR #284, opened 2023-01-09
Reducing build workload, wasm-bindgen crates moved to root workspace.

martinfrances107:
This is similar to #283 

I am slowly reducing the duplication of crates downloaded.

The following crates are used multiple times, and so have been moved them to root workspace 

```
+wasm-bindgen = "0.2"
+wasm-bindgen-futures = "0.4"
+wasm-bindgen-test = "0.3.0"
```


martinfrances107:
This won't work I was assuming each example was in its own workspace. sorry for the noise.

PR #296, opened 2023-01-10
create_slice in leptos_reactive

Gentle:
create_slice can take a complex signal and only react to parts of it. It has a green test but requires a better Docstring explaining what it is meant for


martinfrances107:
@Gentle, sorry I feel a little responsible for the build failure here...

The build was rejected because of a indentation issue.. its is a new thing

If, you run 

``` 
cargo fmt 
``` 

It will automatically fix the white-space/indentation issue.

gbj:
In this particular case rebasing should fix it, as this is a formatting issue from outside your code that I've subsequently fixed on main.

gbj:
Yeah this looks good. @Gentle, at your convenience could you add a slightly more descriptive doc comment as you mentioned, or do you want me to do it?

Gentle:
I changed the generic callback signatures to impl signatures since RA handles those more nicely (and everything else in leptos_reactive does it that way)

The example is now domain-specific and with more documentation about what a slice does and should be used for

gbj:
This looks awesome! Thanks so much.

PR #300, opened 2023-01-11
Pre resolved data

akesson:
So basically I provide a way to resolve data before calling the app. This is how it looks when using it:

```rust
        App::new().leptos_preloaded_data_routes(
                leptos_options.clone(),
                routes.to_owned(),
                data_populate,
                |cx, (data, theme)| view! { cx, <App theme data /> },
            )
    })
    
    async fn data_populate(req: HttpRequest) -> Result<(JsonResponse, Theme), actix_web::Error> {
       ....
    }
```

In the end, I made it as a new LeptosRoutes function, because like this it doesn't break backward compatibility. 

I will also need to refactor the integrations to support custom Actix Guards (my root path has a wildcard which means it needs to use a guard for Accept=text/html in order to not be used for things like favicon requests. 

Once the second refactoring is done, then we could discuss implementing any breaking changes.

Ping @benwis

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -284,7 +284,7 @@ pub fn render_data_app_to_stream<Data, Fut, IV>(
 ) -> Route
 where
     Data: 'static,
-    Fut: Future<Output = Data>,
+    Fut: Future<Output = Result<Data, actix_web::Error>>,
     IV: IntoView + 'static,
 {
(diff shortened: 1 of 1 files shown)


benwis:
As far as the second refactoring goes, I was wondering if an exclude param might be sufficient. Then you could exclude your wildcard root route and define it manually, and then still define the others.

Or maybe just not use `leptos_routes()`? The integrations work fine doing it the manual way.

PR #318, opened 2023-01-15
Switch get_configuration calls to None and add a note in the docs for get_configuration()

benwis:
Since we're assuming cargo-leptos is the default, we want to do `get_configuration(None)` so as to avoid issues with dashes/underscores and questions.


gbj:
Looks good to me. Thanks for updating the Readmes too.

Issue #322, opened 2023-01-16
Impossible to Provide fallback to Leptos Router


benwis:
It appears the type signature of the fallback method makes it impossible to provide one to leptos_router.
From Jose:
If you see the docs for RouterProps, fallback takes fn() -> View as a type, which means you cannot pass in an environment-capturing closure...in other words, fallback is broken as is because there's no way to return a View without using Scope to create it...and since there is no Scope, no dice.

gbj:
Oh shoot this used to be possible in 0.0. I get the problem...

benwis:
Well, the PR fixes this issue, but it looks like the fallback is never actually rendered regardless, more details in the Discord.

PR #323, opened 2023-01-16
fixed `fallback` type signature

jquesada2016:
Since `fallback` doesn't provide a `Scope` to the view function, there's no way for the view function to return any view that is not `View::Unit`, or `View::Text`, as the `fallback` must be a function pointer, and not a closure. Therefore `Scope` can't be grandfathered in from it's environment.


jquesada2016:
I forgot to mention, but this closes #322 .

PR #324, opened 2023-01-17
Fix `<Router fallback>` (signature and functionality)

gbj:


gbj:
See #322. Overlaps with #323 but without formatting issues.

PR #325, opened 2023-01-17
Sets status code 404 when router can't match a route and returns fallback

gbj:


benwis:
I'm still kinda torn about this PR. I think what's bugging me is that it assumes a specific use of fallback without an easy exit. So say, for example, the route is not found, and we render a fallback. Does that always mean that it's a 404, or could there be a scenario where we want to render the fallback without a 404? Changing the status in the integrations means we're forcing people to choose 404 or rewrite it.

I'm wondering if the correct scenario here is to have Router add a specific error to Errors, and then handle it in an ErrorBoundary
```
<ErrorBoundary fallback=...>
<Router>
</Router>
</ErrorBoundary>
```
For SSR apps I want to catch a route not found error during routing, and if you check out the todo_app_sqlite_axum example, you can see an example of that being done

PR #330, opened 2023-01-17
typed-builder inconsistent version.

martinfrances107:
This is really minor.

typed-builder is used in 4 sub crates, it looks like there is a small oversight.

3 crates use 

```
typed-builder = "0.11"
```

1 crate 
```
typed-builder = "0.10"
```

"One of these kids is doing his own thing."


gbj:
Thanks, they updated this fall.

Issue #311, opened 2023-01-13
Idea for <ErrorBoundary/>


gbj:
I just realized that I think we could implement `IntoView` for `Result<impl IntoView, Error>` by creating an `<ErrorBoundary/>` component, which provides a context with a `Signal<Box<dyn Error>>`... any time you try to render `Ok`, it renders, and any time you try to render `Err`, it renders `()` and triggers the error boundary.

benwis:
I like this idea! I think this is a great way to handle Errors returned from server functions and components. Coupled with error handling in Actix/Axum, that's the entire error handling story!

gbj:
Done in #338. Thanks @benwis!

PR #338, opened 2023-01-18
ErrorBoundary Component

benwis:
Spent some time working on #311 because I need it for various things. I think I've got a fairly straightforward implementation based on Solid's version.

I also have a prototype of a better error handler fallback. Basically I wrapped Axum's ServeDir to return either the file or a prerendered page in Leptos.  That should cover not found errors before anything is downloaded.

Then #311 gives an ErrorBoundary, which adds the ability to return `Result<T,E>` from a Component, and on Error, render an <ErrorBoundary /> component at the nearest parent <ErrorBoundary />. That should allow us to capture Errors from server fns and resources run inside Views


benwis:
Alright, this should be ready to go. Works in CSR and SSR, although I didn't edit the examples. We can do that later

Issue #336, opened 2023-01-18
thread 'actix-rt|system:0|arbiter:1' has overflowed its stack


tqwewe:
I created a new leptos project, and added some HTML, and stragely I get this error:

```
thread 'actix-rt|system:0|arbiter:1' has overflowed its stack
fatal runtime error: stack overflow
```

It seems like its when I have too much HTML for some reason?

The `<li></li>` here causes the error, and when you comment it out, the error goes away.

```html
<ul
  class="pf-c-dropdown__menu pf-m-align-right"
  aria-labelledby="dashboard-demo-masthead-settings-button"
  hidden
>
    // vvvvvv Try commenting this li out, and the error goes away
    <li></li>
</ul>
```

<details>
    <summary>View full <code>app.rs</code> file</summary>
    
```rust
use leptos::*;
use leptos_meta::*;
use leptos_router::*;

#[component]
pub fn App(cx: Scope) -> impl IntoView {
    // Provides context that manages stylesheets, titles, meta tags, etc.
    provide_meta_context(cx);

    view! {
        cx,

        // injects a stylesheet into the document <head>
        // id=leptos means cargo-leptos will hot-reload this stylesheet
        <Stylesheet id="leptos" href="/pkg/leptos_start.css"/>

        // sets the document title
        <Title text="Welcome to Leptos"/>

        // content for this welcome page
        <Router>
            <Routes>
                <Route path="" view=|cx| view! { cx, <HomePage/> }/>
            </Routes>
        </Router>
    }
}

/// Renders the home page of your application.
#[component]
fn HomePage(cx: Scope) -> impl IntoView {
    view! { cx,
      <div class="pf-c-page" id="dashboard-demo">
      <a
        class="pf-c-skip-to-content pf-c-button pf-m-primary"
        href="#main-content-dashboard-demo"
      >
        "Skip to content"
      </a>
      <header class="pf-c-masthead" id="dashboard-demo-masthead">
        <span class="pf-c-masthead__toggle">
          <button
            class="pf-c-button pf-m-plain"
            type="button"
            aria-label="Global navigation"
          >
            <i class="fas fa-bars" aria-hidden="true"></i>
          </button>
        </span>
        <div class="pf-c-masthead__main">
          <a class="pf-c-masthead__brand" href="#">
            <picture
              class="pf-c-brand pf-m-picture"
              style="
                  --pf-c-brand--Width: 180px;
                  --pf-c-brand--Width-on-md: 180px;
[...]

PR #344, opened 2023-01-20
Reenable optimizations for SSR using the `view!` macro

gbj:
This reimplements the fast path for server-side rendering HTML elements by compiling static parts into an HTML string. I've tested it against the examples in the repo but it could definitely use more testing.

It's probably not as optimized as it could be. In particular, I think the need to insert so many hydration IDs on what are actually static elements slows it down significantly. This is something we should be able to address in the future.


gbj:
Forgot to mention that this should also fix issues like #336 by significantly reducing stack usage for apps with large numbers of static HTML elements.

PR #347, opened 2023-01-20
doc: add link for leptos watch

imalexlab:
Why:
- didn't know when reading the documentation if I had to serve the app myself or if it was listening on a port

What was done:
- Explain to open browser on localhost:3000 after leptos watch


gbj:
Thanks!

PR #353, opened 2023-01-22
Clippy fixes: redundant clone and .to_string() issues.

martinfrances107:
There have been lots of improvements recently

Clippy is very good at spotting redundant .clone() calls, which I think are always good to removed, without spending the time to see if they are on a hot code path or not.

Might as well fix some reported .to_string() while I see them.

### DIFF
--- a/leptos_dom/src/components/errors.rs
+++ b/leptos_dom/src/components/errors.rs
@@ -20,7 +20,7 @@ where
           Some(errors) => {
             let id = HydrationCtx::id();
             errors.update({
-              let id = id.clone();
+              let id = id;
               move |errors: &mut Errors| errors.insert(id, error)
             });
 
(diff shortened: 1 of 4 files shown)


gbj:
Thanks — There are a couple lines I'll mark that now seem redundant without the unnecessary clones. Want to remove those while you're at it?

leptos_dom/src/components/errors.rs
```diff
@@ -20,7 +20,7 @@ where
           Some(errors) => {
             let id = HydrationCtx::id();
             errors.update({
-              let id = id.clone();
+              let id = id;
```
gbj, line 23: This is now redundant, which makes the added `{ ... }` within `.update()` redundant.

router/src/components/router.rs
```diff
@@ -216,7 +216,7 @@ impl RouterContextInner {
             match resolved_to {
                 None => Err(NavigationError::NotRoutable(to.to_string())),
                 Some(resolved_to) => {
-                    let resolved_to = resolved_to.to_string();
+                    let resolved_to = resolved_to;
```
gbj, line 219: This is now redundant, too -- same as the other one, I think.

PR #371, opened 2023-01-25
Make Errors Sync

benwis:
Basically I want to pass Errors as a prop into render_app_to_stream(), and I can't do that unless it's thread safe


benwis:
Closing in favor of #372

PR #373, opened 2023-01-26
Add a counter example that does not use macros

flosse:


gbj:
Thanks, this is great to have for the next time someone asks about the builder syntax!

PR #379, opened 2023-01-26
added `hgroup` element

### LINKED ISSUE #378: `hgroup` isn't supported
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup

I get this error trying to use it in the `view!` macro:
> cannot find function `hgroup` in crate `leptos::leptos_dom`

ModProg:
fixes #378


gbj:
Thanks!

PR #382, opened 2023-01-26
leptos_axum::handle_server_fns was also duplicated

Gentle:
same as the last one but different function


Gentle:
I made an oops and will fix the branch first

PR #392, opened 2023-01-27
doc/book updated leptos version.

martinfrances107:
Recently a large number of doc issue were fixed...

When I run

```cargo outdated```

I see some outdated packages were being used in the doc area. So this is a minor tidy


martinfrances107:
The last fix, removed a lingering issue reported by "CLIPPY". which was not what I was going for but is nice.

gbj:
Thanks! FYI I'm rewriting all the book stuff so feel free to ignore any of this, it will disappear soon.

PR #401, opened 2023-01-28
Update ErrorBoundary to use miette::Diagnostic instead of Error, and various other tweaks

benwis:
So my goal here has always been to have one standard, pretty page for errors, and a general way to catch errors both at the routing level and inside the app.

This PR should allow one to use one template for both kinds of Errors, and provide additional metadata. It also creates a neat(IMO) way to map between unique Rust Errors and HTTP statuscodes.

You can see it working in the todo_app_sqlite_axum example. 

What's still left would be making error_template actually return the error code from the Rust error, but I wanted to get input on this app structure. 

It also adds miette as a dependency, which should have a small impact on WASM size, but we should test it.


benwis:
So this should be ready to merge, assuming the test passes. The todo_sqlite_axum example has been updated to show a more complete error handling scheme, and there are still some convenience changes to unify Error handling and clear up some API quirks

gbj:
Yeah looks good. But I think you can remove the `miette` dependency from `leptos_dom`, right?

I'm also wondering if we should abstract away from making the `HashMap` that's internal to `Errors` public: i.e., give ways to insert and remove and get an iterator, but without promising that it's backed by a `HashMap`? (For example given the likely small `n` here it's almost always more efficient to store them in a `Vec<_>` and do a linear search.)

This PR is already a breaking change from `0.1.3` (because the callback is `RwSignal<Errors>` instead of `Option<RwSignal<Errors>>` so if there are any other API changes we should make them now.

benwis:
Forgot to save that file.

I mean we certainly could limit visibility to the HashMap, I'm just not sure I want to reimplement all or some of hashmap's methods.

I thought you wanted HydrationKey so you can do cleanup?

PR #410, opened 2023-01-30
Escape <HTML> and <BODY> tokens in documentation markup.

martinfrances107:
Clears warnings found by running

```cargo doc```

Markup not escaping HTML or BODY tokens could lead to "code injection" security problems.


gbj:
Thanks!

PR #423, opened 2023-02-01
fixes cx not found on components marked with #[component(transparent)]

jquesada2016:
Currently, if you use `#[component(transparent)]` on a component, and the `cx` has any other name besides `cx`, we get an error. This fixes this.

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -167,7 +167,7 @@ impl ToTokens for Model {
 
         let component = if *is_transparent {
             quote! {
-                #body_name(cx, #prop_names)
+                #body_name(#cx, #prop_names)
             }
         } else {
             quote! {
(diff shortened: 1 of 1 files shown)


gbj:
```
error[E0425]: cannot find value `cx` in this scope
   --> leptos_macro/src/component.rs:170:29
    |
170 |                 #body_name(#cx, #prop_names)
    |                             ^^ not found in this scope
    ```

jquesada2016:
> ```
> error[E0425]: cannot find value `cx` in this scope
>    --> leptos_macro/src/component.rs:170:29
>     |
> 170 |                 #body_name(#cx, #prop_names)
>     |                             ^^ not found in this scope
>     ```
> ```

Hmmm...well this is true, but the odd thing is that when I ran it, I got no compile error. I'll fix it tomorrow, thanks!

jquesada2016:
@gbj, okay, _now_ we should be good to go....

PR #427, opened 2023-02-01
Several Minor Updates on Examples

Indrazar:
* Added a button to errors_axum example to generate a 500 error with a \<Form\>
* Correct README.md on errors_axum example
* Ran clippy on errors_axum and todo_app_sqlite_axum examples
```
cargo clippy --no-deps --features="csr"
cargo clippy --no-deps --features="ssr"
cargo clippy --no-deps --features="hydrate"
```
* Changed a few cfg_if blocks to help rust_analyzer code completion
* Added "http://" to log so the page can be directly opened from the terminal when server starts

### DIFF
--- a/examples/errors_axum/README.md
+++ b/examples/errors_axum/README.md
@@ -1,9 +1,8 @@
-# Leptos Todo App Sqlite with Axum
-
-This example creates a basic todo app with an Axum backend that uses Leptos' server functions to call sqlx from the client and seamlessly run it on the server.
+# Leptos Errors Demonstration with Axum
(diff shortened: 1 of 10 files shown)


gbj:
The CI runs `cargo fmt --check` before running tests, and it failed — could you `cargo fmt` and push the changes?

gbj:
@requaos This has happened occasionally in the past; I'm surprised it's started occurring again. I generally finish running the CI on my computer, which is not ideal but is fine. I had changed a GitHub setting immediately before this happened and have reverted it... It may have been related.

PR #440, opened 2023-02-02
Minor: Stripped warnings from test results.

martinfrances107:
This is minor 

I have been crawling over the code a lot recently, might as well clear 
lint warnings while I read.


martinfrances107:
bad idea.

PR #443, opened 2023-02-02
use latest tokio in leptos_axum

Gentle:
there was a version number error in Cargo.toml which caused leptos_axum to depend on a separate older version of tokio


gbj:
Thanks!

PR #448, opened 2023-02-03
feature: allow `on:` event listeners on `<Component/>` nodes

gbj:


ModProg:
One thing that I noticed is, that if you register an event listener on a component that will override the internal event listener if they happen to apply to the same element:

Component = `<input on:input= |_| do_some_thing()/>`
Callsite: `<Component on:input=|_| do_some_thing_else()/>`

`do_some_thing` will never be called, but:


Component = `<div><input on:input= |_| do_some_thing()/></div>`
Callsite: `<Component on:input=|_| do_some_thing_else()/>`

works as expected, and both `do_some_thing` and `do_some_thing_else` are called.

gbj:
Yeah that makes sense to me, as I don't think we currently support adding multiple of the same event listener:
`<input on:input= |_| do_some_thing() on:input=|_| do_some_other_thing()/>`

Maybe we should -- if this would be helpful for your use case you could open a feature request issue to be able to add multiple event listeners of the same type to an element.

Components don't "exist" in any sense, so there's nothing to attach a listener to other than the elements.

PR #449, opened 2023-02-03
fix: update leptos dependencies paths to the workspace

TurboTobias:
Thank you for Leptos!

Just a tiny update of the dependency path to Leptos from the `examples/todo-app-sqlite-axum` so that all of the examples build successfully


gbj:
Ahh thanks!

PR #462, opened 2023-02-04
Convert site_address to site_addr

### LINKED ISSUE #446: `site-addr`, `site_addr`, `site_address`
`leptos_config` and `cargo-leptos` have used different names in the past and there's some inconsistency left. (See https://github.com/leptos-rs/start/pull/13)

We just need to make sure we're consistent and using the correct name across examples and in `leptos_config`, possibly maintaining backwards-compatibility with the incorrect versions but trying not to replicate them.

benwis:
Should fix #446, by changing leptos internal options to site_addr

### DIFF
--- a/examples/counter_isomorphic/src/main.rs
+++ b/examples/counter_isomorphic/src/main.rs
@@ -37,7 +37,7 @@ cfg_if! {
             // when not using cargo-leptos None must be replaced with Some("Cargo.toml")
             let conf = get_configuration(None).await.unwrap();
 
-            let addr = conf.leptos_options.site_address.clone();
+            let addr = conf.leptos_options.site_addr.clone();
             let routes = generate_route_list(|cx| view! { cx, <Counters/> });
 
             HttpServer::new(move || {
--- a/examples/errors_axum/src/main.rs
+++ b/examples/errors_axum/src/main.rs
@@ -44,7 +44,7 @@ async fn main() {
     // Setting this to None means we'll be using cargo-leptos and its env vars
     let conf = get_configuration(None).await.unwrap();
     let leptos_options = conf.leptos_options;
-    let addr = leptos_options.site_address;
+    let addr = leptos_options.site_addr;
     let routes = generate_route_list(|cx| view! { cx, <App/> }).await;
 
     // build our application with a route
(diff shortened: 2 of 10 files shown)


gbj:
```
  --> src/main.rs:47:31
   |
47 |     let addr = leptos_options.site_address;
   |                               ^^^^^^^^^^^^ help: a field with a similar name exists: `site_addr`
   ```

So this fails because this example uses `leptos_options.site_address`. We could get it to pass by changing the failing example but I imagine the starter templates and every app someone's created also uses `site_address`, so everyone's apps would suddenly stop compiling.

I think a `serde` alias that maps either `site_addr` or `site_address` to the `site_address` field might be better, to maintain that backwards compatibility.

benwis:
If we were using serde for this, that'd be a great idea.

So if they're doing get_configuration(None), which should be all cargo-leptos users, cargo-leptos reads all the vars and leptos uses env vars. Since it's site-addr not site-address, the host/port is never set in leptos, and it uses the default. So it's a change on cargo-leptos' side.

If they're using get_configuration(Some) or the builder, it would be working currently. I could try to special case that into working, but I suspect there are so few deployments of it we can get away with it.

gbj:
Isn't `get_configuration(Some("Cargo.toml"))` using `serde`, though? I see we derive `Deserialize`
https://github.com/benwis/leptos/blob/b14411029f8b1a968f8da3b0284606dd7f2214dc/leptos_config/src/lib.rs#L24

and then deserialize from the TOML
https://github.com/benwis/leptos/blob/b14411029f8b1a968f8da3b0284606dd7f2214dc/leptos_config/src/lib.rs#L160-L162

I'm probably missing something.

benwis:
Technically we are, in that the `config` crate implements its traits and uses it to convert from the `Config` type to the `ConfFile` type, but the crate appears to validate and read the values itself before it does so. I tried various incantations of adding rename to the site_addr line, but none caused anything to change

gbj:
Yeah, #2 sounds good; our next release will be `0.2.0` because of some other breaking changes so this makes sense to me. The error Rust gives you when you try to access `leptos_options.site_address` instead of `.site_addr` is pretty useful anyway because the names are similar. And #2 is basically what your PR is at present, correct? So this should be ready to merge.

benwis:
Correct, #2 is this PR and should be ready to go

PR #469, opened 2023-02-04
error on non meta input for prop attribute

### LINKED ISSUE #465: Support non-literal defaults with `#[prop(default=...)]`
After some investigation, it turned out that this is in fact a missing feature of the attribute parsing, we currently do not support non-literal defaults due to only supporting meta attributes.

> ```
> struct A;
> 
> #[component]
> fn CustomButton(
>   cx: Scope,
>   #[prop(default=A)] non_default: A,
>   #[prop(default=true)] default: bool) -> impl IntoView {}
> ```
> 
> `non_default` will not be optional even though a default is provided, while `default` will be.
> 
> It T implements Default should not be relevant here.

ModProg:
Doesn't fix #465 but improves it in showing an error on invalid usage.

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -4,6 +4,7 @@ use convert_case::{
 };
 use itertools::Itertools;
 use proc_macro2::{Ident, TokenStream};
+use proc_macro_error::ResultExt;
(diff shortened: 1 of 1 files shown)


ModProg:
We should add UI tests for the proc macros e.g. with https://crates.io/crates/trybuild/1.0.77

PR #472, opened 2023-02-04
impl From<&str> for MaybeSignal<String>

g-re-g:
Once again I'm not sure if there are any unforeseen consequences of doing this, just something I ran into while writing some code.
Enables not having to call `.to_string()` on a `&str` passed as a `MaybeSignal` like this:
```rust
<A class="navbar-brand" href="/">"link text"</A>
```
instead of this:
```rust
<A class="navbar-brand".to_string() href="/">"link text"</A>
```

### DIFF
--- a/leptos_reactive/src/signal_wrappers_read.rs
+++ b/leptos_reactive/src/signal_wrappers_read.rs
@@ -586,6 +586,12 @@ impl<T> From<Signal<T>> for MaybeSignal<T> {
     }
 }
 
+impl From<&str> for MaybeSignal<String> {
+    fn from(value: &str) -> Self {
+        Self::Static(value.to_string())
+    }
+}
+
 #[cfg(not(feature = "stable"))]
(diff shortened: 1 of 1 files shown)


gbj:
Hm... I'd like to have a more generic version of this, for sure. Unfortunately Rust won't allow
```rust
impl<T, U> From<U> for MaybeSignal<T>
where
    U: Into<T>,
{
    fn from(value: U) -> Self {
        Self::Static(value.into())
    }
}
```
Because it conflicts with the blanket `From<T> for T`, which is very annoying.

It won't even allow
```rust
impl<T> From<T> for MaybeSignal<String>
where
    T: std::fmt::Display,
{
    fn from(value: U) -> Self {
        Self::Static(value.to_string())
    }
}
```
Because, of course, `String` implements `Display`.

It feels weird to have a special case like this but it's so common in HTML that it probably makes sense.

PR #483, opened 2023-02-07
workspace `rustfmt`

jquesada2016:
I am commiting my personal preferred settings, but these should in no means be what the project actually uses, just a potential starting point. I also have not applied the rules, as this would be a full project-wide recommit.


jquesada2016:
Whoaps, I see this brought in the changes from another branch. This is because I accidentally branched not off main. Let me force push to fix this

gbj:
This is fine by me, except I'd prefer either the default `tab_spaces = 4` or, possibly better, `hard_tabs = true` so you can set tab width in your editor to 2 and others can set to 4 (but this would involve changing every line in the project once)

jquesada2016:
If you check out the config file, it is actually using the default tab size of 4. I wouldn't worry about the spaces vs tabs thing because I saw that VS code is able to figure it out on it's own, so I think it should be fine.

The thing of not would be to comb through the bevy of rustfmt settings and see if there's anything we like. For example, I didn't know format_docs existed, and I can't go back after it! There may be more settings like this hiding around that may benefit the code aesthetics of the project. I also think it'd be a good idea to change these settings as little as possible for the same reason you mentioned, each change will potentially change ALL files in the repo.

Another slightly related thing...we're using `cfg_if` quite a bit....and it makes code look really out of place because it doesn't format anything inside :(

gbj:
> If you check out the config file, it is actually using the default tab size of 4. I wouldn't worry about the spaces vs tabs thing because I saw that VS code is able to figure it out on it's own, so I think it should be fine.

```
tab_spaces = 2
```
?

I may be misunderstanding.

jquesada2016:
Hmm...on my commit, that line isn't there?

jquesada2016:
We were on the same branch, but not the same commit. I forced it again...idk what happened that caused this upstream PR to be diferent from my local one.

PR #490, opened 2023-02-07
Signal traits

jquesada2016:
This is intended to be a largish PR. The theme of this game is to make all core shared methods on signals into traits. I have added `try_` variants to all of them. The `Untracked*` traits also need to be updated, but perhaps in a future PR, as this one is large enough as is.


gbj:
Thanks. I just looked over this and it looks like reverting the changes to `StoredValue` is the only thing missing that I can see.

I am also thinking that `signals` could be a better module name than `signal_prelude`, so that someone can `use leptos::signals::*;`

I'm going to do some work on reorganizing the other re-exports now too.

PR #516, opened 2023-02-13
leptos_dom erros.rs remove<E>() does not need to be generic.

martinfrances107:
Clippy spotted this.


gbj:
Thanks! It's the semicolon. I need to take a final look at this API anyway so I'll take this clippy fix for now.

PR #517, opened 2023-02-13
URLs with a scheme in A component rendered correctly

g-re-g:
Currently if you supply a URL to an `A` component that has a scheme or something that looks like a scheme it will silently fail. Notice this is the capital `A` component not the lowercase `a` component.

This code:
```rust
<A href="http://google.com">"Url"</A>
<A href="google.com">"No Http"</A>
<A href="localhost:3000">"No Http with port"</A>
<A href="mailto:greg@greg.greg">"Mailto"</A>
<A href="tel:+11111111111">"Tel"</A>
```

Currently produces:
Screenshot 2023-02-13 at 2 48 01 PM

In which all but the second one are not what a normal `a` tag would do.

After this PR it produces:
Screenshot 2023-02-13 at 2 47 21 PM

Which are all what a normal `a` would produce.

The other option here is to not compile and recommend that the user use an `a` tag rather than an `A` tag which I also think is a reasonable solution but I picked this one simply so there's less friction for a user. Happy to make it do whatever folks think is best!

### DIFF
--- a/router/src/hooks.rs
+++ b/router/src/hooks.rs
@@ -70,7 +70,7 @@ pub fn use_resolved_path(
         if path.starts_with('/') {
             Some(path)
         } else {
-            route.resolve_path(&path).map(String::from)
+            route.resolve_path(&path).or(Some(path)).map(String::from)
         }
     })
 }
--- a/router/src/matching/resolve_path.rs
+++ b/router/src/matching/resolve_path.rs
@@ -74,7 +74,7 @@ pub fn join_paths<'a>(from: &'a str, to: &'a str) -> String {
 
 const TRIM_PATH: &str = r#"^/+|/+$"#;
 const BEGINS_WITH_QUERY_OR_HASH: &str = r#"^[?#]"#;
-const HAS_SCHEME: &str = r#"^(?:[a-z0-9]+:)?//"#;
+const HAS_SCHEME: &str = r#"^[a-z0-9+.-]+:"#;
 const QUERY: &str = r#"/*(\*.*)?$"#;
 
 #[cfg(not(feature = "ssr"))]


gbj:
Yeah, this is interesting. This usage of the `<A/>` component instead of the `<a>` element is basically an antipattern. You should *never* use `<A/>` for any of those examples, but I guess hypothetically someone could be passing in a signal that, if it's on the same domain, should resolve a relative route, and otherwise should handle it as a normal link. So I guess we should handle it correctly.

gbj:
> Thinking about this a little more, if folks wanted to link to `login/a:b` with this PR they can't unless something like `./a:b` is supported. So perhaps this is all intended behavior and users just need to make sure they are using the right component. I personally got here by using the `A` for an absolute URL link like `https://google.com` and it didn't work. Maybe that's just user error.

So I had the chance to play around with this a little and I think this is actually the core issue here... Colons are definitely an allowed character in URL paths (although somewhat unusual), so treating every `a:b` as a scheme and opting out of relative routing actually makes it impossible to resolve a URL like this as a relative route. On the other hand, every use case where this is "broken" can be solved by simply using an `<a>` element, since the `<A/>` component's only functionality has to do with your own routes. 

Can you think of a way to distinguish between the two cases 1) "resolve this to the relative route `a:b`" and 2) "use the absolute URL `a:b`" (i.e., with scheme `a`)? My only idea is to whitelist certain schemes (`mailto`, `localhost`, `tel`, etc.)

If not, I'll probably close this PR and maybe add some documentation to the `<A/>` component indicating that it shouldn't be used for external links.

gbj:
> Maybe the best way forward is to not treat any input as special? Currently if you start the `href` in an `A` tag with something that looks like a `scheme://`, the `A` component will swallow that input and output an empty `href`. Similarly if you start the `href` with a `/` it will output the `href` as also starting with a `/` and not relative to the current URL. The combination of those two is what led me to believe this was a bug and I'm guessing will lead others to feel the same or it might be hard for them to debug what's wrong with their code. Relatedly `my-site.com/login/https://google.com` is a valid url, and though it's not a smart one to use, it's not possible to use it with the `A` tag.
> 
> Anyway, I'm wondering if the least ambiguous and most predictable thing to do is to always outpt the `href` as an extention of the current url without any special logic for detecting intention. I think that'd help folks debug their links and keep them from thinking somethings up. Might simplify the docs as well.

This makes sense — the exception is that if the URL starts `/`, you do still want it to be resolved as an absolute path on the current domain. This means that if I'm on `/users/gbj`, then
```
<A href="about"/> // to /users/gbj/about
<A href="/about/> // to /about
```

gbj:
Sounds good! I'll close this then.

PR #523, opened 2023-02-14
Add example: Login with API token (CSR only)

flosse:
Here is an example that demonstrates how you can use leptos for SPAs that use some kind of a API token (e.g. JWT).

**Motivation**:
I am considering porting an existing web application whose server/API cannot be changed and SSR is not needed initially. 
To get a sense of what an example looks like beyond a counter example, I chose this scenario.
Since I'm new to FRP, I'm interested in any feedback that could make the code cleaner.

### DIFF
--- a/examples/login_with_token_csr_only/Cargo.toml
+++ b/examples/login_with_token_csr_only/Cargo.toml
@@ -1,0 +1,7 @@
+[workspace]
+members = ["client", "api-boundary", "server"]
+
+[patch.crates-io]
+leptos = { path = "../../leptos" }
+leptos_router = { path = "../../router" }
+api-boundary = { path = "api-boundary" }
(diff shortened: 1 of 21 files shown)


gbj:
Hi! Could you maybe provide a little more commentary on how close you think this is to ready? It's quite a large example... I'm happy to sit down and really try to understand it but I also see various FIXME comments and things. Are you still working on it or do you want a review at this point?

flosse:
Actually, it is already working but I'd not call it finished because of the three `FIXME` points for which I have not yet found a solution. Two of them could be phrased as questions:

1. How to read the initial value of an input field?
2. How to call `use_navigate` outside of a `Router`?

The third is not really a problem but only something that feels a bit fragile.

So my suggestion is that I mark this PR as draft and I will get back to you as soon as I have solved the three issues so that I don't waste your time unnecessarily.

flosse:
@gbj I think it's ready for review now. I expect some parts are not written in a leptops/FRP ideomatic way, but that is due to my lack of knowledge about it. So please be picky and point out anything to me that isn't pretty.

gbj:
Honestly this looks pretty good to me. Thanks so much for the contribution!

PR #524, opened 2023-02-14
Identify CSS to reload from the href

akesson:
Identify the CSS to reload from the href, which is already sent to the browser in the WebSocket message, instead of using the CSS `id`. 

This opens the possibility to have more than one CSS file that is hot-reloaded by `cargo-leptos`. I plan on using that for the Tailwind implementation.

This is backwards compatible with previous `cargo-leptos` versions.

@benwis Ok for you? I have to admit that I didn't test it with an Axum setup as I don't have one easily available. Would appreciate it if you could test it!

@gbj I would be really good to include this in release 0.2.0.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -589,14 +589,15 @@ fn html_parts(
                     let msg = JSON.parse(ev.data);
                     if (msg.all) window.location.reload();
                     if (msg.css) {{
-                        const link = document.querySelector("link#leptos");
-                        if (link) {{
-                            let href = link.getAttribute('href').split('?')[0];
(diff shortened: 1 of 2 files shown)


benwis:
@akesson Is the actix one supposed to be matching "fail"?

akesson:
🤦  Thanks @benwis for spotting that. It was for testing the error message.

benwis:
Other than that, It looks fine to me. I see no reason Axum would work any differently than Actix here. I tried it and it seems sane

gbj:
Yep looks good to me. I'm doing a little refactoring into a `leptos_integration_utils` package as part of the in-order streaming/async stuff, by the way, so this code will live in one place in the future. Will merge right now.

PR #536, opened 2023-02-18
fix(examples): hackernews_axum styles href

ApplY3D:
```bash
2023-02-18T15:49:42.957124Z ERROR ❌ error
error from HTML pipeline

Caused by:
    0: error getting canonical path for "/leptos/examples/hackernews_axum/./static/style.css"
    1: No such file or directory (os error 2)
```


ApplY3D:
https://github.com/leptos-rs/leptos/issues/537

PR #550, opened 2023-02-20
Fix issue with redirects in server fns creating multiple Location headers

benwis:
Because headers from `ResponseOptions` were set before the default referral on form submission, you could use `redirect()` in your server fn, and the default referral would add another Location header, if it was a form submission. Then everything would blow up.


fundon:
Or use [is_success](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.is_success) and [is_redirection](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.is_redirection).

benwis:
> Or use [is_success](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.is_success) and [is_redirection](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.is_redirection).

I did not know those existed! But I don't think we a number of 3xx error codes are valid responses for our Form component

benwis:
@gbj You'll probably want to take a look at this. I believe I've fixed the behavior in a quite kludgy way.
Because gloo_net follows redirects, we'll never get a 3xx response. It'll always be either a 200 or a 4xx with a redirected=true value on it.

I've changed it to check for that, and redirect based on the url() field in the 200 response. But to do that I had to import the url crate(which looked like it was disabled for hydration), to convert the absolute path into a relative one for navigate()

Also, navigating this way produces a warning and a panic in the leptos reactive code in the firefox console, something about updating non-existent signals.

But hey, it redirects, so......

gbj:
> @gbj You'll probably want to take a look at this. I believe I've fixed the behavior in a quite kludgy way. Because gloo_net follows redirects, we'll never get a 3xx response. It'll always be either a 200 or a 4xx with a redirected=true value on it. I tried disabling that, but then it returns status code 0 and no headers, which is even less helpful.

Ah okay, great! Good work figuring this out.

By the way you seem to have committed some Nix-related files inadvertently (`.direnv`, `.envrc`, `flake.nix`)

> I've changed it to check for that, and redirect based on the url() field in the 200 response. But to do that I had to import the url crate(which looked like it was disabled for hydration), to convert the absolute path into a relative one for navigate()

There's already a `Url` implementation in the router that uses the browser's native URL parsing on the client or the `url` crate on the server, so I've changed this just to use that and avoid the import in Wasm.

> Also, navigating this way produces a warning and a panic in the leptos reactive code in the firefox console, something about updating non-existent signals.
> 
> But hey, it redirects, so......
Believe it or not this is because it's using a server action and deserializing the JSON from the response is asynchronous, so it was actually trying to update the values in the action *after* navigating away from the page. I've fixed the panics; the warnings will still show but I don't think that's really an issue.

PR #554, opened 2023-02-20
feature: add class option to Html component

polarmutex:
allows you to set class on the Html component.

do not like line 95, but is was done to fix a compile error, not sure how to resolve it

### DIFF
--- a/meta/src/html.rs
+++ b/meta/src/html.rs
@@ -14,12 +14,28 @@ pub struct HtmlContext {
 impl HtmlContext {
     /// Converts the `<html>` metadata into an HTML string.
     pub fn as_string(&self) -> Option<String> {
-        let lang = self.lang.borrow().as_ref().map(|val| format!("lang=\"{}\"", val.get()));
-        let dir = self.dir.borrow().as_ref().map(|val| format!("dir=\"{}\"", val.get()));
-        let class = self.class.borrow().as_ref().map(|val| format!("class=\"{}\"", val.get()));
-        let mut val = [lang, dir, class].into_iter().flatten().collect::<Vec<_>>().join(" ");
+        let lang = self
+            .lang
+            .borrow()
+            .as_ref()
+            .map(|val| format!("lang=\"{}\"", val.get()));
+        let dir = self
+            .dir
+            .borrow()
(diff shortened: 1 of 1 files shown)


meta/src/html.rs
```diff
@@ -8,18 +8,29 @@ use std::{cell::RefCell, rc::Rc};
...
 
 impl HtmlContext {
     /// Converts the `<html>` metadata into an HTML string.
     pub fn as_string(&self) -> Option<String> {
-        match (self.lang.borrow().as_ref(), self.dir.borrow().as_ref()) {
-            (None, None) => None,
-            (Some(lang), None) => Some(format!(" lang=\"{}\"", lang.get())),
-            (None, Some(dir)) => Some(format!(" dir=\"{}\"", dir.get())),
-            (Some(lang), Some(dir)) => {
+        match (self.lang.borrow().as_ref(), self.dir.borrow().as_ref(), self.class.borrow().as_ref()) {
```
gbj, line 17: This is going to grow exponentially if we add any more attributes... With 4 conditions it was fine but this is a lot of duplication. Can you try a refactor?

meta/src/html.rs
```diff
@@ -76,10 +90,19 @@ pub fn Html(
                     _ = el.set_attribute("dir", &value);
                 });
             }
+
+            if let Some(class) = class {
+                let el = document().document_element().expect("there to be a <html> element");
```
gbj, line 95: Instead of this, you should add a `let el = el.clone();` at line 88 before it moves `el` into the effect for `dir`, just as the `lang` attribute clones at line 80 before it moves into its effect.

gbj (commented):
Thanks! I added a couple comments, should be pretty straightforward.

gbj:
Oh could you also please run `cargo fmt` before you push again? The CI runs `cargo fmt --check` first thing.

gbj:
Thanks — the APIs for the HTML bit are slightly finicky. Your version would render something like `<html lang="Some(TextProp)" dir="None" class="None">` I took the liberty of fixing it up! I think this should be good to go once it passes tests. Thanks for the contribution.

Issue #227, opened 2023-01-03
`leptos_macro` hygiene


jquesada2016:
We need to optimize hygiene as much as possible for both the view! and #[component] macros until macros 2.0 make life easier. Specifically, we should:

- Convert all imports to absolute paths. For example, instead of leptos::ev::click, we should use ::leptos::ev::click. This ensures that the import is unambiguously absolute to the crate.
- Explicitly reference all trait methods, such as Clone, IntoView, etc., as they cannot be assumed to be in scope. For example, instead of value.clone(), we should use ::core::clone::Clone::clone(&value).
- Minimize the use of imports from std to facilitate support for #![no_std] environments, such as embedded microcontrollers in the future.
- Remove all macro panics from view! and use proc_macro_error macros instead to improve DX.

martinfrances107:
In terms of review 

Points 1-2 make sense and appear frequently in guides on "how to write macros". 
Points 3-4 make sense and I think should be regard as good rules.


+ 1 from me

PR #568, opened 2023-02-22
Macro `panic` hygiene

remkop22:
I noticed that the `view` macro panic's on incorrect input. This can be a problem for language servers and semantically it's not really the correct thing to do.

I tried to minimize codepath's that are able to panic and replace panic's with `abort` where applicable.

Related: #227


gbj:
Thanks. This looks really great — much appreciated.

Issue #572, opened 2023-02-24
Link doesn't output correct html with as_ attribute


SleeplessOne1917:
When you use the `Link` meta tag like so:

```rust
<Link
      rel="preload"
      href="/fonts/MedievalSharp-Regular.ttf"
      as_="font"
      type_="font/ttf"
      crossorigin="anonymous"
/>
```
The link tag that gets output is this:

```html
<link id="leptos-link-4" as_="font" crossorigin="anonymous" href="/fonts/MedievalSharp-Regular.ttf" rel="preload" type="font/ttf">
```

Notice that the underscore is still there for the output html. I think I found where in the code causes this and will open a PR.

PR #573, opened 2023-02-24
Make meta link as output correct attribute html

SleeplessOne1917:
Fixes [this issue](https://github.com/leptos-rs/leptos/issues/572).


gbj:
Thanks!

PR #575, opened 2023-02-25
Ssr modes axum

tanguy-lf:
Create Axum Ssr modes example, and add In-order routes.
Fix Async Ssr mode in axum integration.


gbj:
Thanks!

PR #577, opened 2023-02-25
feat: Support `rkyv` encoding

### LINKED ISSUE #528: Support `rkyv` encoding
Could we add a feature flag to support [`rkyv`](https://github.com/rkyv/rkyv) encoding in `Serializable`?

Motivation: we use rkyv encoding for our API (built on Actix) and would like to load data from it using SSR. [More](https://david.kolo.ski/blog/rkyv-is-faster-than/) details [why](https://rkyv.org/motivation.html).

617a7a:
I have managed to get past the lifetime issues I was experiencing (due to rkyv's zero-copy deserialisation). Therefore, this pull request adds a feature flag, `rkyv`, to support [`rkyv`](https://rkyv.org/) encoding for SSR'd pages. 

If the flag is enabled, rkyv will replace JSON as the serialisation format used in Leptos. We encode the bytes using base64, so the trait is identical for JSON/rkyv.

Closes #528


gbj:
Okay, this looks good. Thanks for your work!

PR #585, opened 2023-02-27
feat: (fouc) reduce flicker on unstyled content

perry-birch:
Injects css to hide the html element and then restores visibility once wasm is loaded.

Examples of hiding html to reduce flickering:
https://stackoverflow.com/questions/3221561/eliminate-flash-of-unstyled-content/43823506

Inject css to restore visibility via js:
https://stackoverflow.com/questions/524696/how-to-create-a-style-tag-with-javascript


gbj:
This is a really clever solution! I like this.

Re: naming, I wonder if something really explicit like `<BlockingStylesheet/>` is a good idea? Just to make it really clear what this is doing?

perry-birch:
Updated the name as noted

gbj:
Looks like you're still making some changes — just let me know when you want me to review.

perry-birch:
I found an issue with the timing to detect load using style tags so that lay push was a quick fix.

Should be stabilized now and the tailwind example is updated to use it for a quick test if you like.

I put the network speed throttling on the slowest 3g option to make it obvious the difference.

gbj:
Thanks so much for your work on this. Now that I sit down and ran through the Tailwind example on 3G, I see exactly the problem you're seeing.

However I'm sorry to say this is actually an XY kind of problem — The problem is different from what I thought it was initially and what it was intended to solve, and I think there's an even better solution here! On 3G, I think you'll notice, the load time of the CSS is not the actual issue: for me, it appears basically instantly with the initial HTML. The real issue is that when the WASM loads for initial hydration, it is rerunning the app and resetting all of the attributes on the page. I hadn't realized the implications of this, but when one re-sets, for example, the `href` on a `link` or the `src` on an `img`, it actually reloads that resource from the server, even if it is the same exact URL.

So you'll notice that the FOUC comes immediately once the WASM finishes loading and resets that `href` on the stylesheet... Then there's no style until the CSS has loaded again.

I tried implementing a quick guard against this, and it turns out that if we guard against re-setting the attributes during hydration, we actually prevent the FOUC and I saw better better performance doing this (no flash at all) than with the `<BlockingStylesheet/>` solution which causes the page to blip out of existence for 20ms (or 40 or 60 depending on speed, of course).

So
1) I'm going to close this PR as not needed
2) I'm going to make changes in the hydration code so that we don't reset all those attributes and other static values
3) This actually saves a huge amount of work during hydration, of course. So thank you *so* much for surfacing a performance issue that was slowing down every single Leptos app.

perry-birch:
Awesome!  I'll take a solve over a patch any day ;)

PR #588, opened 2023-02-27
make counter test compile

eiswind:
a had to make some small tweaks to let the counter test example compile and pass.

I am not sure if web-sys is really needed, but it is used with websys::HtmlElement in the test.

please have a look at it.

### DIFF
--- a/examples/counter/tests/mod.rs
+++ b/examples/counter/tests/mod.rs
@@ -1,9 +1,9 @@
+use counter::*;
+use leptos::*;
+use wasm_bindgen::JsCast;
 use wasm_bindgen_test::*;
 
 wasm_bindgen_test_configure!(run_in_browser);
-use counter::*;
-use leptos::*;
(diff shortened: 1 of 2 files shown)


examples/counter/tests/mod.rs
```diff
@@ -84,22 +84,22 @@ fn inc() {
     let clear = div
         .first_child()
         .unwrap()
-        .dyn_into::<HtmlElement>()
+        .dyn_into::<web_sys::HtmlElement>()
```
perry-birch, line 87: I'm not certain these added turbofish are needed, the following command seems to run with just the other changes you proposed:
wasm-pack test --node

gbj:
Thanks! I think you could probably consolidate some of the `web_sys::HtmlElement` into a single `use web_sys::HtmlElement` import at the top, but this looks fine to me and I'll merge it if it passes the CI tests which it should.

PR #571, opened 2023-02-24
change: swap from `stable` feature to `nightly` feature

gbj:
See discussion in #567.

PR #591, opened 2023-02-27
swap `stable` feature with `nightly` and empty feature set testing

remkop22:
@gbj I based this branch of your previous work here #571.

I found out by some trial and error that specifying an empty set `[]` in `skip_feature_sets` actually skips the empty set 🤯.

I also added a separate test that tests all non-doc-tests with an empty feature set on a stable compiler. This should at least verify that Leptos works as it should on stable without additional feature flags.

Also added feature flags to the integration libraries (@gbj  If you have reasons to dislike this due to breakage I'll happily remove it).


gbj:
Thanks — This is great to know, and seems really helpful. One thing that would also be good would be to remove `"csr"` from the default features for `leptos`, so that people need to opt in to CSR and not opt out. This would mean people don't automatically do `default-features = false` for SSR and then have to opt back in to nightly.

I'm happy to schedule this for some upcoming `0.3` when we're reading to make breaking changes.

gbj:
I started this over in #1227, and just migrtaed all the tests over to work with stable.

PR #596, opened 2023-02-28
Make server functions framework agnostic

Demonthos:
This PR pulls out the server function implementation into a separate crate that is framework agnostic and makes leptos_server wrap that crate

### DIFF
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -1,7 +1,7 @@
 #![forbid(unsafe_code)]
 #![deny(missing_docs)]
 
-//! # Leptos Server Functions
+//! # Server Functions
 //!
 //! This package is based on a simple idea: sometimes it’s useful to write functions
 //! that will only run on the server, and call them from the client.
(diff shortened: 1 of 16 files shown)


gbj:
Oh absolutely fantastic. I'll take a look at this later.

server_fn/src/lib.rs
```diff
@@ -0,0 +1,450 @@
+#![forbid(unsafe_code)]
+#![deny(missing_docs)]
+
+//! # Leptos Server Functions
```
gbj, line 4: This title can be more framework-agnostic if you'd like. Doesn't matter to me either way.

server_fn_macro/src/lib.rs
```diff
@@ -148,36 +173,47 @@ pub fn server_macro_impl(
...
+        impl #server_fn_path::ServerFn<#server_ctx_path> for #struct_name {
             type Output = #output_ty;
 
             fn prefix() -> &'static str {
                 #prefix
             }
 
             fn url() -> &'static str {
-                #url
+                #server_fn_path::const_format::concatcp!(#fn_name_as_str, #server_fn_path::xxhash_rust::const_xxh64::xxh64(concat!(env!("CARGO_MANIFEST_DIR"), ":", file!(), ":", line!(), ":", column!()).as_bytes(), 0))
```
gbj, line 201: We had some concerns from people about leaking the structure of their app in server function names, but it looks like this is a good way of obscuring that while maintaining unique IDs per source file and avoiding collisions — Thanks! I'm assuming this is not (easily?) reversible.

gbj (approved):
This looks really great. Thanks for your work in doing this, and I'm hopeful it will be a good way to integrate these into Dioxus (and maybe Sycamore too), and to pool work. Feel free of course to make PRs for other changes in the future.

This looks good to me, so let me know if you're ready and I'll happily merge it and release the package.

server_fn_macro/src/lib.rs
Demonthos: It is not a cryptographic hash, but it should be lossy so you cannot recover the path from the hash. My main reason for doing this approach was it works on stable, but not exposing the file path is a nice bonus.

PR #604, opened 2023-03-01
In `Action::dispatch()` set new value before resetting input signal

Qwox0:
This PR changes when the value signal is set during `Action::dispatch()` and `MultiAction::dispatch()`. This aligns the behavior of `Action::dispatch()` with `<ActionForm/>`, which already sets the value signal before resetting the input signal ([see form.rs](https://github.com/leptos-rs/leptos/blob/ceb7bd398df73ba3bb0059467a06a42299484283/router/src/components/form.rs#L238)).

Before there was a brief moment, after the future resolves, where the state of input and value were the same as before the dispatch call.

### DIFF
--- a/leptos_server/src/action.rs
+++ b/leptos_server/src/action.rs
@@ -191,9 +191,9 @@ where
         pending.set(true);
         spawn_local(async move {
             let new_value = fut.await;
+            value.set(Some(new_value));
             input.set(None);
             pending.set(false);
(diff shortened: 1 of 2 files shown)


gbj:
Okay, this makes sense to me. Future changes to the reactive system may well make this irrelevant, but for now it seems good to have the consistency in order.

PR #611, opened 2023-03-02
feat: support expr for #[prop(default=...)]

ModProg:
- Improved error messages
- Added testing for component macro

fixes: #465

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -1,3 +1,4 @@
+use attribute_derive::Attribute as AttributeDerive;
 use convert_case::{
     Case::{Pascal, Snake},
     Casing,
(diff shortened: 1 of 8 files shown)


leptos_macro/src/component.rs
```diff
@@ -407,106 +365,43 @@ impl Docs {
...
-        }
-    }
+#[derive(Clone, Debug, AttributeDerive)]
+#[attribute(ident = prop)]
+struct PropOpt {
+    #[attribute(conflicts = [optional_no_strip, strip_option])]
+    optional: bool,
+    #[attribute(conflicts = [optional, strip_option])]
+    optional_no_strip: bool,
+    #[attribute(conflicts = [optional, optional_no_strip])]
```
ModProg, line 375: these are not needed as already implied by the other two fields, just added them for completenes.

ModProg:
Not sure why this fails, and how to fix it

gbj:
Ah bummer yes, looks like we've overwhelmed poor GitHub Actions. I'll run the CI locally later and review.

ModProg:
> Ah bummer yes, looks like we've overwhelmed poor GitHub Actions

Maybe we could split the one test up into multiple actions?

gbj:
Looks great — And thanks for the added tests! (This may well have been what caused the CI to kick the bucket but we'll see.)

PR #613, opened 2023-03-02
Remove the register function from server_fn

Demonthos:
This removes the register function from the ServerFn trait and instead uses the [inventory](https://github.com/dtolnay/inventory) crate to collect the functions.
This requires each framework to create it's own wrapper around server functions that has a const function called from_generic_sever_fn and implement a small amount of boilerplate in there own crate.

I have tried to update any docs around the register function I could find, but I'm not familiar with Leptos' docs so I could have missed some areas

**This is a breaking change**
- [x] Fix docs building

### DIFF
--- a/examples/counter_isomorphic/src/counters.rs
+++ b/examples/counter_isomorphic/src/counters.rs
@@ -9,13 +9,6 @@ use std::sync::atomic::{AtomicI32, Ordering};
 use broadcaster::BroadcastChannel;
 
 #[cfg(feature = "ssr")]
-pub fn register_server_functions() {
-    _ = GetServerCount::register();
-    _ = AdjustServerCount::register();
-    _ = ClearServerCount::register();
-}
-
-#[cfg(feature = "ssr")]
 static COUNT: AtomicI32 = AtomicI32::new(0);
 
 #[cfg(feature = "ssr")]
--- a/examples/counter_isomorphic/src/main.rs
+++ b/examples/counter_isomorphic/src/main.rs
@@ -30,9 +30,6 @@ cfg_if! {
 
         #[actix_web::main]
         async fn main() -> std::io::Result<()> {
-
-            crate::counters::register_server_functions();
-
             // Setting this to None means we'll be using cargo-leptos and its env vars.
             // when not using cargo-leptos None must be replaced with Some("Cargo.toml")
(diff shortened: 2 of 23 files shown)


gbj:
I'm fine with this being a breaking change for `server_fn` but I'd rather mark the `register` function deprecated and make it a no-op for now in Leptos, just so I don't need to make a breaking change there. And likewise we can leave it in the examples for now, so it's not confusing for people who are on `0.2` rather than git main and do still need to call `register()`.

I haven't released a version that depends on `server_fn` yet so it's okay to bump the version there without Leptos needing a version bump.

Otherwise this is awesome and I'm glad you realized this was possible. It should make these much easier to use.

gbj:
Reading up on `inventory`:
> Platform support includes Linux, macOS, iOS, FreeBSD, Android, and Windows. Other platforms will simply find that no plugins have been registered. Support for other targets would need to be added in the ctor crate.
We've been pretty careful about trying to maintain support for `wasm32-wasi` as a target for server-side rendering, on platforms like Cloudflare Workers, Deno Deploy, etc. I'm curious about whether this is going to fail silently on that platform.

Demonthos:
> Reading up on `inventory`:
> 
> > Platform support includes Linux, macOS, iOS, FreeBSD, Android, and Windows. Other platforms will simply find that no plugins have been registered. Support for other targets would need to be added in the ctor crate.
> 
> We've been pretty careful about trying to maintain support for `wasm32-wasi` as a target for server-side rendering, on platforms like Cloudflare Workers, Deno Deploy, etc. I'm curious about whether this is going to fail silently on that platform.

It appears it will https://github.com/dtolnay/inventory/issues/13, and a similar crate linkme https://github.com/dtolnay/linkme/issues/6 both do not support WASM
It appears WASM support is blocked by some work on the compiler supports. We will need to find a new approach or wait until the compiler supports ctor

gbj:
I'd actually like to reopen this topic, either via this PR or a new one given the time that's passed/merge conflicts. (I'll take a look at the conflicts and see how easy it would be to merge.)

I think I was being much too conservative. If we simply preserve the possibility of explicitly registering server functions, I think that's fine and will be a huge DX win for the very large number of people using one of the supported platforms.

PR #615, opened 2023-03-03
Minor: Clippy - Simplified conditional logic in transition.rs.

martinfrances107:
Just a couple of clippy fixes, only one is worth highlighting

all I have to say is when I draw out the equivalent logic elements this makes sense to me.

```
warning: this boolean expression can be simplified
   --> leptos/src/transition.rs:122:24
    |
122 |                       if (has_local_only && child_runs.get() > 0)
    |  ________________________^
123 | |                         || !has_local_only
    | |__________________________________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#nonminimal_bool
    = note: `#[warn(clippy::nonminimal_bool)]` on by default
```

### DIFF
--- a/leptos/src/transition.rs
+++ b/leptos/src/transition.rs
@@ -119,9 +119,7 @@ where
                 if is_first_run(&first_run, &suspense_context) {
                     let has_local_only = suspense_context.has_local_only();
                     *prev_children.borrow_mut() = Some(frag.nodes.clone());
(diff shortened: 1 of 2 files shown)


martinfrances107:
This build failed 

"No space left on device"

gbj:
Thanks! We've added a couple more examples and tests so it's straining CI now occasionally, will fix independently of this.

PR #625, opened 2023-03-03
fix: added missing attributes of events that don't bubble

WafflePersonThing:
references used:
- https://developer.mozilla.org/en-US/docs/Web/API/
- web archives of the above before jul 11th 2022, relevant: https://github.com/mdn/content/issues/19590


gbj:
Thanks!

PR #626, opened 2023-03-04
Example of app with broken out pages and components

brettearle:
Hello, 
I just added an example with modules representing pages and components within the app module. 

Module Tree of this example.

crate leptos_start
- mod app: pub
    - - mod components: pub
                   - - - mod content_paragraph: pub
    - - mod pages: pub
                            - - - mod homepage: pub

Is this something of interest for leptos? 

Also happy to clean up and format in a way that fits with leptos if interested


gbj:
Thanks for offering! I'm not sure this adds much as an example, beyond how to use Rust modules to import components. We have a pretty extensive set of examples at this point, so I'm happy to take additional contributions to the examples that demonstrate particular features or model real apps, but I think I'll close this for now.

brettearle:
Sweet as, after digging deeper found it was pretty useless if you know rust properly.

PR #628, opened 2023-03-04
Unit tests for `leptos-config` and new functions for loading configuration

vaniusrb:
There were missing unit tests for `leptos-config` so I created it. I tried to follow the same pattern of Leptos' crates, where the tests are located in the `tests` folder. But for this approach, it can't test the private functions, I created a second test file called `tests.rs` inside `src`.

Today `get_configuration` expects a `Cargo.toml` filename or `Option::None` (to load from env variables). In some cases, it would be useful to receive the String content of the file instead of its location, for example when we want to generate a single binary and embed a config file content using `include_str!`.
For a better design, new functions `get_config_from_file` (file name), `get_config_from_str` (content), and `get_config_from_env` were introduced. The `get_configuration` function still exists. These changes here do not break any backward compatibility.

Let me know what you think. I hope I'm helping.


gbj:
Thanks! This all looks good to me. I'd like @benwis to take a look since he's done more work in this area, so I've added him as a reviewer.

benwis:
This PR looks good to me, I've been considering making the configuration functions a bit more explicit. The tricky part here, and maybe @vaniusrb has an idea, is that I'd love to patch the `.replace(`-`, `_`)` line we have in read_to_string(). The TOML keys need it, but overwriting it in values has caused some issues with filepaths.

Appreciate the work to write tests and generalize this

gbj:
Awesome. Thanks to you both. I'll merge this and if anyone has an idea we can do a separate PR for the replacement function.

PR #631, opened 2023-03-05
docs: fix instruction typos

Banyc:
Previously, the console warning message tells user to add `feature = ["csr"]`. However, it's incorrect. This change rectifies the instruction to be adding `features = ["csr"]`.


gbj:
Thanks! Good catch.

PR #632, opened 2023-03-05
Put the Request on the Context for Axum

benwis:
Inspired by @gbj I realized that I can con provide the Request on the Context by wrapping it in an Arc, in addition to RequestParts, so that people could run extractors on it in their components or in their server fns.

I am tempted to delete RequestParts and just say people should just use Request, but that's a breaking change, and I'm curious what you think


gbj:
Thanks! It's worth saying that this particular implementation is also a breaking change, as it changes the signature of `generate_request_parts`, which is a `pub fn`. We can easily do it in a non-breaking way that would involve creating another function and deprecating the earlier one with `#[deprecate]`.

It would be good to work on some examples to make sure we have the API right, here. For example, I'm pretty sure the `Arc<Request<Body>>` in its present form can't actually be used. The reason is that the `FromRequest` trait consumes the request, but we can't get an owned `Request<_>` from an `Arc<Request<_>>` because the context system will still own a handle to it. 

I'd propose actually wrapping it in a `leptos_axum::Response<B>` struct that's something like
```rust
pu struct Response<B>(Arc<RwLock<Option<Request<B>>>>)

impl<B> Response<B> {
  pub fn take(&self) -> Option<Response<B>> {
    self.0.write().take()
  }
}
```
This means that 
a) we could share the response safely via context via the `Arc<_>`
b) but could still consume it somewhere during the request by taking the value from the `Option`

For `RequestParts`, we should probably `impl From<RequestParts>` for [`Parts`](https://docs.rs/http/0.2.9/http/request/struct.Parts.html) so that you can just do
```rust
let foo = use_context::<RequestParts>(cx).unwrap().into().from_request_parts();
```

benwis:
I think you meant Request here, as I haven't changed the Response types. 
Do we really need an RwLock and an Option here? I can't really think of a reason why someone would want to mutate the Request object

gbj:
> I think you meant Request here, as I haven't changed the Response types.

Yep sorry, I meant request.

> Do we really need an RwLock and an Option here? I can't really think of a reason why someone would want to mutate the Request object

Part of the idea here, in my understanding, is to allow people to use extractors without needing a special handler. Extractors are either `FromRequest`, which takes an owned `Request`, or `FromRequestParts`, which takes `&mut Parts`. So if we want people to be able to use a `FromRequest` extractor, we need to let them take control of an owned `Response`. The only way to do this, as far as I can tell, as it have something like an `Arc<RwLock<Option<Response<_>>>` so that we have
1) cloneability (for context)
2) interior mutability around an `Option<_>`, so that we can not just mutate but actually take ownership of the `Request`

Does that make sense?

benwis:
Yeah, thinking about it a bit more, it seems like that might be the only way to do it. 

I don't think we're going to be able to write a From for Parts, it has a private field on it to prevent construction. We can definitely do it from the Request though.

In the end, that suggests to me we might want to deprecate RequestParts in favor of the Request type with helpers.

gbj:
> I don't think we're going to be able to write a From for Parts, it has a private field on it to prevent construction. We can definitely do it from the Request though.

I think this is actually okay because if you have access to `&mut Request` you can use `extract_parts` to apply a parts extractor to it using [`extract_parts`](https://docs.rs/axum/0.6.10/axum/trait.RequestExt.html#tymethod.extract_parts).

So as long as we provide the request object it should be fine.

benwis:
Yeah, meant to add that, but that was my understand as well. I'm editing the PR to create the new LeptosRequest type, and setting this up. Are we OK with deprecating RequestParts?

gbj:
Deprecate away!

benwis:
OK, so I have a new version up to play with. Unfortunately I still have to change generate_request_parts(), because it consumes the request, and I need it not to so that generate_leptos_request() can consume it. 

Since it's so clunky to get access to the Request types, I'm thinking about leaving RequestParts undeprecated, to make that easier for people

integrations/axum/src/lib.rs
```diff
@@ -110,20 +110,54 @@ pub fn redirect(cx: leptos::Scope, path: &str) {
...
+#[derive(Debug, Default)]
+pub struct LeptosRequest<B>(Arc<RwLock<Option<Request<B>>>>);
+
+impl<B> Clone for LeptosRequest<B> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+impl<B> LeptosRequest<B> {
+    /// A less boilerplatey way to overwrite the contents of `ResponseOptions` with a new `ResponseParts`
```
gbj, line 144: This doc comment should be updated I think.

integrations/axum/src/lib.rs
```diff
@@ -110,20 +110,54 @@ pub fn redirect(cx: leptos::Scope, path: &str) {
...
+/// Requred by Request not being Clone. See this issue for eventual resolution: https://github.com/hyperium/http/pull/574
+#[derive(Debug, Default)]
+pub struct LeptosRequest<B>(Arc<RwLock<Option<Request<B>>>>);
+
+impl<B> Clone for LeptosRequest<B> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+impl<B> LeptosRequest<B> {
```
gbj, line 143: `LeptosRequest` probably also needs some kind of `.take_request()` method or something, right? I think because it's a private inner field there's not any way to get the request out at present.

integrations/axum/src/lib.rs
```diff
@@ -110,20 +110,54 @@ pub fn redirect(cx: leptos::Scope, path: &str) {
 }
 
 /// Decomposes an HTTP request into its parts, allowing you to read its headers
-/// and other data without consuming the body.
-pub async fn generate_request_parts(req: Request<Body>) -> RequestParts {
+/// and other data without consuming the body. Creates a new Request from the
+/// original parts for further processsing
+pub async fn generate_request_parts(
```
gbj, line 115: I'd suggest renaming this to `generate_request_and_parts`, leaving the original `generate_request_parts` alongside it (deprecated), and using the new `generate_request_and_parts`. This would allow us to merge this without it being a breaking change for semver purposes, even if nobody is actually using `generate_request_parts` outside the integration.

integrations/axum/src/lib.rs
benwis: Yeeah, accidentally didn't commit that, fixed.

integrations/axum/src/lib.rs
benwis: Yep

benwis:
OK, I've added a with(), update(), fixed my typos, and restored take_request()

PR #638, opened 2023-03-07
Bump tower-http upto 0.4.

martinfrances107:
This is as result of looking at cargo outdated 

As the code base currently  stands there is a  inconsistent policy in the stating of of version number of tower-http

in many places the full MAJOR.MINOR.PATCH is used.

in one place it stated only as MAJOR.MINOR .. 

The MAJOR.MINOR option  relies on the goodness of the crate below to not break anything with a PATCH update.

As I updated the version number I changed/unified the policy to MAJOR.MINOR only, If any reviewer objects I will unify  the patch to the  MAJOR.MINOR.PATCH pattern

There are lots of tabulation changes, but I think that is as result of my IDE picking up on the recently introduced 

rustfmt.toml file.

### DIFF
--- a/examples/errors_axum/Cargo.toml
+++ b/examples/errors_axum/Cargo.toml
@@ -13,7 +13,7 @@ console_error_panic_hook = "0.1.7"
 futures = "0.3.25"
 cfg-if = "1.0.0"
 leptos = { path = "../../../leptos/leptos", default-features = false, features = [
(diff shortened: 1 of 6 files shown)


gbj:
Thanks, I prefer MAJOR.MINOR for Rust packages in general so this is fine by me, especially as this is all in example code.

PR #643, opened 2023-03-07
Make server functions work outside of WASM

Demonthos:
Uses [reqwest](https://docs.rs/reqwest/0.11.5/reqwest) instead of gloo_net to make server function work outside of web contexts. This makes it possible to use the server_fn library in native renderers. It exposes a new set_server_url function that should be called once per app before any server functions are run. This defaults to the current origin for WASM targets.

# Alternatives
1) Use gloo_net for WASM and reqwest for non-WASM targets
2) Expose a fetcher trait that allows users to choose between any request frameworks


gbj:
I think this is a great idea. I'm concerned about the effects on binary size in the browser context, though. I built our `todo_app_sqlite` example against the `gloo_net` version (`main`) and the `reqwest` version (PR) (release mode, opt-level = Z, etc.) It's 552kb with `gloo_net` and 862kb with `reqwest`. This is way too big a difference for me.

So I like your alternative plan #1 to use `gloo_net` for WASM and `reqwest` for non-WASM, where binary size doesn't really matter.

Demonthos:
> I think this is a great idea. I'm concerned about the effects on binary size in the browser context, though. I built our `todo_app_sqlite` example against the `gloo_net` version (`main`) and the `reqwest` version (PR) (release mode, opt-level = Z, etc.) It's 552kb with `gloo_net` and 862kb with `reqwest`. This is way too big a difference for me.
> 
> So I like your alternative plan #1 to use `gloo_net` for WASM and `reqwest` for non-WASM, where binary size doesn't really matter.

Sounds good, it now uses gloo_net for WASM. Other parts of Leptos use gloo_net as well, so it could be including both in the final binary. Either way, fine defaulting to gloo_net especially if the binary size is effected

gbj:
Yeah this looks good. I pushed one commit adding an import for `MessageEvent` over in `leptos_dom` that was breaking the tests here; I assume this is probably something that `gloo_net` was importing from `web_sys` but that we needed to add now that `gloo_net` is disabled on non-wasm32 targets.

Will wait for the new CI run to finish and then merge if it's green. Thanks so much!

PR #654, opened 2023-03-10
🔧 Fix(leptos_dom/logging): Fix showing warning instead of error

pikhosh:


gbj:
Thanks, much appreciated!

PR #656, opened 2023-03-10
cargo machete: leptos_macro - Removed unused crates.

martinfrances107:
"Say hello, wave goodbye" to leptos_server.


gbj:
Cheers! I wonder if we can set up some checks for this during the CI so you don't have to keep cleaning up after us all the time 😄

PR #660, opened 2023-03-11
Copy & Clone for MaybeSignal<T>

CharlesTaylor7:
Hello, first PR here!
I implemented Copy & Clone for MaybesSignal<T> so that the api has better parity with Signal<T>. 

Are there any additional steps I should make here, e.g. updating a changelog?


gbj:
Thanks very much, looks good! No further action needed.

PR #666, opened 2023-03-12
Added example using Tailwind, CSR (only) and Trunk

vascokk:
The title is self-explanatory.
I wasn't able to run the original Tailwind example with Trunk, so here is a simple one that should work out of the box.


gbj:
Nice, thanks! I added this to the root `Makefile.toml` so it will be checked during CI, looks good otherwise. Much appreciated.

PR #691, opened 2023-03-17
feat: add a `debounce` helper for event listeners

gbj:


leptos_dom/src/helpers.rs
```diff
@@ -134,10 +148,95 @@ pub fn set_timeout(cb: impl FnOnce() + 'static, duration: Duration) {
...
-    );
+    window()
+        .set_timeout_with_callback_and_timeout_and_arguments_0(
+            cb.as_ref().unchecked_ref(),
+            duration.as_millis().try_into().unwrap_throw(),
+        )
+        .map(TimeoutHandle)
+}
+
+/// Handle that is generated by [set_interval] and can be used to clear the interval.
```
elliotwaite, line 159: This comment looks like it should be updated. Maybe to something like:
```
/// Handle that is generated by [set_timeout_with_handle] and can be used to clear the timeout.
```

leptos_dom/src/helpers.rs
```diff
@@ -134,10 +148,95 @@ pub fn set_timeout(cb: impl FnOnce() + 'static, duration: Duration) {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct TimeoutHandle(i32);
```
elliotwaite, line 161: This struct and its impl could potentially be moved up in this file to be above the `set_timeout` function for consistency with how `set_iterval` has its handle above its function call, which might make it easier for people to find things if these functions and their handles all follow the same pattern.

leptos_dom/src/helpers.rs
gbj: Ah yes of course, thanks!

leptos_dom/src/helpers.rs
gbj: Agreed.

PR #692, opened 2023-03-17
examples: remove duplicate `console_error_panic_hook::set_once()` calls

elliotwaite:
I noticed that in a lot of the examples, `console_error_panic_hook::set_once()` was being called twice:
```
console_error_panic_hook::set_once();
_ = console_log::init_with_level(log::Level::Debug);
console_error_panic_hook::set_once();
```
so I removed the first one since most of the other examples were using this pattern:
```
_ = console_log::init_with_level(log::Level::Debug);
console_error_panic_hook::set_once();
```
Lef me know if you would prefer that I instead delete the second one, so that the order would be like this:
```
console_error_panic_hook::set_once();
_ = console_log::init_with_level(log::Level::Debug);
```
Also, I noticed that the [examples/tailwind/src/lib.rs](https://github.com/leptos-rs/leptos/compare/main...elliotwaite:leptos:fix-duplicate-panic-hook-calls?expand=1#diff-e699beee9abb30108fa822048736e0e46473b21529e0c09e8d12b5bdeea6c644) example was using that other order (calling `console_error_panic_hook::set_once()` then `console_log::init_with_level(log::Level::Debug)`), so I swapped those calls so that all the examples would be consistent. I also fixed some indent issues I noticed in that file.

### DIFF
--- a/examples/counter_isomorphic/src/lib.rs
+++ b/examples/counter_isomorphic/src/lib.rs
@@ -10,7 +10,6 @@ cfg_if! {
 
         #[wasm_bindgen]
         pub fn hydrate() {
-            console_error_panic_hook::set_once();
             _ = console_log::init_with_level(log::Level::Debug);
             console_error_panic_hook::set_once();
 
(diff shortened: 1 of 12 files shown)


gbj:
Thanks!

Issue #627, opened 2023-03-04
feat: allow routes to accept other HTTP verbs


gbj:
It's entirely possible for routes to opt into accepting `POST` or `DELETE` requests or whatever, even though they're currently only set up for `GET` requests. This would allow the traditional multi-page-app/PHP-style process in which, for example, a form page can post to itself and render a response.

We would want to be somewhat careful here in thinking about the client-side-routing story. It's fine if we want to just use a `<form>` and do a page refresh. What does it look like to upgrade this to a `<Form/>` and post to the route without a full-page refresh?

This could work the same way that SSR modes work now: an optional `methods` prop (or `verbs`?) on the `<Route/>` components. If any method is present on any nested route in a full path, then the path can accept that verb, i.e., if I have

```html
<Routes>
   <Route
        path="about"
        view=move |cx| view! { cx,  <About/> }
    >
    <Route
        path="contacts"
        view=move |cx| view! { cx,  <ContactList/> }
        methods=[Method::Post]
    >
        <Route
            path=":id"
            view=move |cx| view! { cx,  <Contact/> }
            methods=[Method::Delete]
        />
        <Route
            path="/"
            view=move |_| view! { cx,  <p>"Select a contact."</p> }
        />
    </Route>
 </Routes>
```
Then `"/about"` only accepts `GET`, `"/contacts"` accepts `GET` or `POST`, and `"/contacts/42"` would accept `GET`, `POST`, or `DELETE`, even if they don't necessarily handle all of those verbs.

Tagging in @underscorefunk whose idea it was and @benwis who was also in this discussion. No obligation for either of you to reply or take any action, just wanted you to know it's on my radar.

underscorefunk:
I think this is a pretty elegant way to progressively enhance the happy path. 

I'm inclined to suggest that methods defaults to [Method::Get]... but if you provide methods=[...] then it's explicit. In the case of /contacts, it would only handle Post. It's maybe a bit more verbose in that you need to specify both, but I think it makes it more clear. I can see people being confused about a route handling a GET when the method only says "Method::Delete". 

In cases like that we could have Method::Any?

My gut tells me that this might be a very ergonomic way to setup a simple JSON api with a set of routes and their path arguments. Almost like defining your interface and then the views or handlers are the implementations elsewhere. 

Totally off topic but it does make me want to have something like this:
```
<Api>
    <ActionRoute
        path="contacts/:id"
        action=get_contacts
        />
</Api>
```
[...]

PR #695, opened 2023-03-17
feat: allow multiple HTTP request methods/verbs

gbj:
See discussion in #627.

To do
- [x] underlying stuff
- [x] Axum integration
- [x] Viz integration
- [x] Actix integration


benwis:
Looking good! Is the expectation that people will get the Method from the Request to do conditional stuff in the component?

gbj:
> Looking good! Is the expectation that people will get the Method from the Request to do conditional stuff in the component?

Yeah I've been messing with an example so I've got to include that so it makes more sense. It's interesting to see the patterns emerge. 

I found that the Axum FromRequest extractor is async, so I actually ended up pulling the form data parsing out of the component into a server function, which... kind of makes sense anyway! Saving me some cfg flags inside the component. And with async rendering it just held the hold page so the page waits for the resource to load anyway. And then it can be reused as a server action etc if it needs to be.

benwis:
Nice to see I'm not the only one haunted by rustftm.

I really like that idea, we colocate a server fn with the route it handles, and that fn handles everything!

PR #696, opened 2023-03-18
work around issue in wasm-pack

kulak:
Issue is described in:

- <https://github.com/rustwasm/wasm-pack/issues/650>
- <https://github.com/rustwasm/wasm-pack/issues/1234>


gbj:
Thanks! Could you provide a little more context about the issue? Do you know what the wasm-pack default features are and why disabling them fixes your issue?

If I search the repo I find `cargo install wasm-pack` in 11 places, I think all example readmes. This PR only changes one of them, so could you include the others?

Issue #693, opened 2023-03-17
feat: make `request_animation_frame`, `request_idle_callback`, and `set_timeout` cancellable


elliotwaite:
Currently, in [leptos_dom/src/helpers.rs](https://github.com/leptos-rs/leptos/blob/main/leptos_dom/src/helpers.rs), there are:
* `request_animation_frame` - not cancellable
* `request_idle_callback` - not cancellable
* `set_timeout` - not cancellable
* `set_interval` - returns a handle, `handle.clear()` → `window().clear_interval_with_handle(self.0)`

It would be nice if they were all cancellable:
* `request_animation_frame` - returns a handle, `handle.cancel()` → `window().cancel_animation_frame(self.0)`
* `request_idle_callback` - returns a handle, `handle.cancel()` → `window().cancel_idle_callback(self.0)`
* `set_timeout` - returns a handle, `handle.clear()` → `window().clear_timeout_with_handle(self.0)`
* `set_interval` - returns a handle, `handle.clear()` → `window().clear_interval_with_handle(self.0)`

I can work on a PR for this, but I wanted to check first to see if this change was actually wanted.

gbj:
Yep, makes perfect sense. I just had to do this [while implementing a simple debounce helper](https://github.com/leptos-rs/leptos/pull/691). Changing the current functions is a breaking change, so for the sake of semantic versioning, you could implement new functions that return their handles, then implement the old functions in term of them, just ignoring the handle.

elliotwaite:
Sounds good. I created a PR (https://github.com/leptos-rs/leptos/pull/698) to also add `request_animation_frame_with_handle(...)` and `request_idle_callback_with_handle(...)`.

However, now `set_interval` seems like the odd one out since it does return a handle. To me, it would seem more consistent if it didn't return a handle, and then `set_interval_with_handle` was also added. But it's somewhat justifiable that `set_interval` does return a handle since it's probably quite rare that someone would want to set an interval callback that they never wanted to cancel.

Do you have a preference between making `set_interval` consistent with the others versus leaving it as is?

Cause if it was preferred to make it consistent, we could add `set_interval_with_handle`, which would be an alias of `set_interval`, and then mark `set_interval` as deprecated until it is updated to not return a handle with one of the future breaking changes updates.

gbj:
Yes adding `set_interval_with_handle` and marking `set_interval` is deprecated is a good approach. I'll open a PR to do that right now.

[...]

PR #698, opened 2023-03-18
feat: add cancellable versions of `request_animation_frame` and `request_idle_callback`

elliotwaite:
Issue: https://github.com/leptos-rs/leptos/issues/693

This PR adds:
* `AnimationFrameRequestHandle`
* `request_animation_frame_with_handle(...)`
* `IdleCallbackHandle`
* `request_idle_callback_with_handle(...)`

Please review this PR thoroughly as I am very new to Rust. All nits are welcome. I just tried to follow the same structure of the other handle structs and their corresponding `*_with_handle` functions.

For naming the handle structs, the reason I went with `AnimationFrameRequestHandle` and `IdleCallbackHandle` was because that seemed to align most with the language used in the MDN Web Docs, but the alternatives I considered were:
* `AnimationFrameHandle`
* `IdleCallbackRequestHandle`

However, in the [window.cancelAnimationFrame()](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame) docs, they say it cancels an "animation frame request", not just an "animation frame":
> The `window.cancelAnimationFrame()` method cancels an **animation frame request** previously scheduled through a call to `window.requestAnimationFrame()`.

And in the [window.cancelIdleCallback()](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelIdleCallback) docs, they just say it cancels a "callback", not a "callback request":
> The `window.cancelIdleCallback()` method cancels a **callback** previously scheduled with `window.requestIdleCallback()`.

And in the [window.requestIdleCallback()](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) docs, they refer to it as an "idle callback":
> You can call `requestIdleCallback()` within an **idle callback** function to schedule another callback to take place no sooner than the next pass through the event loop.

However, if you think either of those alternative handle struct names would be better, let me know and I can change them.


gbj (approved):
This all looks very good to me. Thanks for the idea and the contribution!

elliotwaite:
Thanks for the review. This is the first actual Rust code I've submitted to any repo. I've been wanting to learn Rust, and Leptos made me feel like I finally had a reason to. So thanks for all the work you've been putting into this project.

PR #704, opened 2023-03-18
`view!` macro not compiling with a non-default scope name

alexisfontaine:
The following basic component is failing to compile on the current version of Leptos `v0.2.3`:

```
error[E0425]: cannot find value `cx` in this scope
```

```rust
use leptos::*;

#[component]
pub fn Button (scope: Scope) -> impl IntoView {
	view!(scope, <button>"Click"</button>)
}
```


gbj:
Good catch, and thanks for the PR. I'll merge this after the CI tests run, but it looks good to me.

PR #707, opened 2023-03-19
Clippy: less .clone() calls, simpler pointer passing.

martinfrances107:
As usual less .clone() calls

Just to signal things to come, there are two warnings which I am keeping behind, because I am going to suggest a change is approach which needs a different kind of review.

Here is the warning which I will cover in a upcomming PR

```
warning: unused `std::result::Result` that must be used
   --> leptos_dom/src/ssr_in_order.rs:145:5
    |
145 |     tx.unbounded_send(std::mem::take(&mut buffer));
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
```

### DIFF
--- a/leptos_dom/src/lib.rs
+++ b/leptos_dom/src/lib.rs
@@ -322,7 +322,7 @@ impl Element {
                 cx,
                 element,
                 attrs,
-                children: children.clone(),
+                children,
                 #[cfg(debug_assertions)]
                 view_marker,
             }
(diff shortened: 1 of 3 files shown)


gbj:
Oh yeah thanks, I've noticed the `unbounded_send` warning since I made this change and have been meaning to get it to. Looking forward to hearing your idea.

gbj (approved):
All looks good, thanks.

PR #714, opened 2023-03-20
Add `is_mounted` and `dyn_classes`

jquesada2016:
I added two methods to `HtmlElement`.

- `is_mounted` which returns true if the element is currently in the DOM
- - `dyn_classes` which takes any signal that yields iterable elements of class names which will dynamically add/remove the classes on the element as the signal changes

### DIFF
--- a/leptos_dom/src/html.rs
+++ b/leptos_dom/src/html.rs
@@ -573,6 +573,23 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
         self
     }
 
+    /// Checks to see if this element is mounted to the DOM as a child
+    /// of `body`.
+    ///
+    /// This method will always return [`None`] on non-wasm CSR targets.
+    pub fn is_mounted(&self) -> bool {
+        #[cfg(all(target_arch = "wasm32", feature = "web"))]
+        {
+            crate::document()
+                .body()
+                .unwrap()
+                .contains(Some(self.element.as_ref()))
+        }
+
+        #[cfg(not(all(target_arch = "wasm32", feature = "web")))]
+        false
+    }
+
     /// Adds an attribute to this element.
     #[track_caller]
     pub fn attr(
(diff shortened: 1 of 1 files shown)


leptos_dom/src/html.rs
```diff
@@ -679,6 +696,90 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
...
+            leptos_reactive::create_effect(
+                self.cx,
+                move |prev_classes: Option<
+                    SmallVec<[Cow<'static, str>; 4]>,
+                >| {
+                    let classes = classes_signal()
+                        .into_iter()
+                        .map(Into::into)
+                        .collect::<SmallVec<[Cow<'static, str>; 4]>>(
+                    );
```
gbj, line 724: I think the `.flat_map()` splitting into individual strings should be inserted before collecting into `classes`, shouldn't it? As it is, `classes` is returned with the unsplit values, which means `prev_classes` has the unsplit values, which means that the matching logic below will be inaccurate.

leptos_dom/src/html.rs
```diff
@@ -679,6 +696,90 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
...
+
+                    let mut new_classes = classes
+                        .iter()
+                        .flat_map(|classes| classes.split_whitespace());
+
+                    if let Some(prev_classes) = prev_classes {
+                        let mut old_classes = prev_classes
```
gbj, line 731: Ah okay never mind, I see we're splitting them again down here. But why not just split them once up top, instead of repeating it?

leptos_dom/src/html.rs
```diff
@@ -679,6 +696,90 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
...
+                            .iter()
+                            .flat_map(|classes| classes.split_whitespace());
+
+                        // Remove old classes
+                        for prev_class in old_classes.clone() {
+                            if !new_classes.any(|c| c == prev_class) {
+                                el.class_list()
```
gbj, line 738: Not a huge deal, but you could 
```rust
let class_list = el.class_list();
```
once before the two sets of iterations and save calling out to JS quite so many times.

leptos_dom/src/html.rs
```diff
@@ -679,6 +696,90 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
...
+                                },
+                            );
+                        }
+                    }
+
+                    classes
+                },
+            );
+        }
+
```
gbj, line 779: I notice this method just does nothing in SSR mode. I think it should, instead, read the initial value of the signal and add it to the classes in the element. Otherwise, if there are classes here, they'll pop in once WASM is loaded and there be unstyled content beforehand.

gbj (commented):
Thanks! `is_mounted` looks good. I left a few comments on `dyn_classes`.

jquesada2016:
> Ah okay never mind, I see we're splitting them again down here. But why not just split them once up top, instead of repeating it?

The reason is lifetimes. If I did them all at top, I'd need to collect into an owned `String`, vs the it's now, I can collect `&str` which respects the `Cow<'static, str>`

> once before the two sets of iterations and save calling out to JS quite so many times.

Sorry, I didn't follow what you meant here?

> I notice this method just does nothing in SSR mode. I think it should, instead, read the initial value of the signal and add it to the classes in the element. Otherwise, if there are classes here, they'll pop in once WASM is loaded and there be unstyled content beforehand.

100% agree! I completely forgot about SSR. :S

Also, there's something funny going on. There seems to be a bug somewhere, butif I do:

```rust
let input_ref = create_node_ref(cx);

input_ref.on_load(cx, |input| {
  input.dyn_class(Some("my-class"))
});

view! { cx,
  <input ref=input_ref class="class-a class-b" />
}
```

The resulting classlist on `input` is `class-a class-b` intead of `my-calss class-a class-b`. I checked to make sure, and the class is being set, it's just being replaced by `class-a class-b`.

If I change the above code to be:

```rust
let input_ref = create_node_ref(cx);

input_ref.on_load(cx, |input| {
  wasm_bindgen_futures::spawn_local()async move {
    input.dyn_class(Some("my-class"))
  });
});

view! { cx,
  <input ref=input_ref class="class-a class-b" />
}
```

then we get the correct classes

PR #718, opened 2023-03-21
Minor: Simplifiy Box::Pin() call.

martinfrances107:
-        Box::pin(async move { fut.await }) as Pin<Box<dyn Future<Output = O>>>
+        Box::pin(fut) as Pin<Box<dyn Future<Output = O>>>

My reading of this is the simplification: - 
The type coercion into the final concrete type only relies "Can I make what I have  into a 'Pin<Box<dyn Future<Output = O>>>'


martinfrances107:
PS I should say this removes  "Clippy" warning

gbj:
Looks good.

PR #726, opened 2023-03-22
Add arbitrary attributes to Html meta tag

step4:
Adds a additional `attributes` property to the Html component so one can add several arbitrary attributes to the document's `<html>` element.

### DIFF
--- a/meta/src/html.rs
+++ b/meta/src/html.rs
@@ -1,4 +1,4 @@
-use crate::TextProp;
+use crate::{additional_attributes::AdditionalAttributes, TextProp};
 use cfg_if::cfg_if;
 use leptos::*;
 use std::{cell::RefCell, rc::Rc};
@@ -9,7 +9,7 @@ pub struct HtmlContext {
     lang: Rc<RefCell<Option<TextProp>>>,
     dir: Rc<RefCell<Option<TextProp>>>,
     class: Rc<RefCell<Option<TextProp>>>,
-    attributes: Rc<RefCell<Option<Vec<(String, TextProp)>>>>,
+    attributes: Rc<RefCell<AdditionalAttributes>>,
 }
 
 impl HtmlContext {
@@ -30,7 +30,7 @@ impl HtmlContext {
             .borrow()
             .as_ref()
             .map(|val| format!("class=\"{}\"", val.get()));
-        let attributes = self.attributes.borrow().as_ref().map(|val| {
+        let attributes = self.attributes.borrow().0.as_ref().map(|val| {
             val.iter()
                 .map(|(n, v)| format!("{}=\"{}\"", n, v.get()))
                 .collect::<Vec<_>>()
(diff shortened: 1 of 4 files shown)


meta/src/html.rs
```diff
@@ -60,9 +67,11 @@ impl std::fmt::Debug for HtmlContext {
 /// fn MyApp(cx: Scope) -> impl IntoView {
 ///     provide_meta_context(cx);
 ///
+///     let attrs = vec![(String::from("data-theme"), TextProp::from("dark"))];
+///
 ///     view! { cx,
 ///       <main>
-///         <Html lang="he" dir="rtl"/>
+///         <Html lang="he" dir="rtl" attributes=attrs/>
 ///       </main>
```
step4, line 75: I think there is a way to implement the proper `From` trait for the `attrs` vector such one can use:
`let attrs = vec![("data-theme", "dark")];`
But I did not manage to solve that.

gbj:
I started working on implementing `From<_>` for the vector and... got a little carried away. (You can't implement a "foreign" trait for a foreign type, so I thought I'd add an `AdditionalAttributes` wrapping type, and then added the same thing for `<Body/>`)

Thanks so much for getting things rolling on this. Take a look and see if the changes I made make sense to you, and if so, I'll merge!

step4:
Thanks for the help.
I would never come up with that `From<_>` implementation 😃 But I think I learned a bit more about rust.

Nevertheless I tried to set the attributes now in a reactive way like this:

```rust
let (attrs, set_attrs) = create_signal(cx, vec![("dark-theme", "dark")]);

<Html attributes=attrs/>
```

But I get the compilation error:
```
error[E0277]: `ReadSignal<Vec<(&str, &str)>>` is not an iterator
  --> src/counters.rs:54:34
   |
54 |         <Html attributes=attrs/>
   |                       ---------- ^^^^^ `ReadSignal<Vec<(&str, &str)>>` is not an iterator
   |                       |
   |                       required by a bound introduced by this call
   |
   = help: the trait `Iterator` is not implemented for `ReadSignal<Vec<(&str, &str)>>`
   = note: required for `ReadSignal<Vec<(&str, &str)>>` to implement `IntoIterator`
   = note: required for `leptos_meta::additional_attributes::AdditionalAttributes` to implement `From<ReadSignal<Vec<(&str, &str)>>>`
   = note: required for `ReadSignal<Vec<(&str, &str)>>` to implement `Into<leptos_meta::additional_attributes::AdditionalAttributes>`
```
Does that mean `ReadSignal` needs a very generic implementation of `IntoIterator`?


So settings these statically works, if this is the only use case for this branch I guess we can merge it.

gbj:
Ah yeah to be reactive we will need to change the signature so that attributes takes something like `Option<MaybeSignal<AdditionalAttributes>>`

gbj:
Okay, so I've added the reactive piece here. Now, you do need to manually specify `AdditionalAttributes::from` but it can be reactive or not reactive.

Either 
```rust
<Html attributes=AdditionalAttributes::from([("data-theme", "garden")])/>
```
or something like this (I did this in the todo example)
```rust
let todo_attrs = create_rw_signal(cx, AdditionalAttributes::from(vec![("data-theme", "garden")]));

view! { cx,
    <Html attributes=todo_attrs/>
}
```

step4:
Hey
sorry couldn't have a look into it the last days.

Thanks for your Implementation and will definitely try it out

PR #727, opened 2023-03-22
fix: include query string in new navigation if an `<ActionForm/>` response redirects

gbj:


gbj:
#728 is better, closing this.

Issue #694, opened 2023-03-17
Allow adding scripts above default head scripts


chenasraf:
### Description

I'm trying to implement a wasm bridge crate ([firebase-wasm](https://github.com/jquesada2016/firebase-wasm-rs)), which loads modules via their npm name. I want to use ESM modules, so I would like to use an [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) to specify a remote CDN instead of NPM module.

This causes an error:

```
An import map is added after module script load was triggered.
```

### How to reproduce

1. Make a minimal app using [leptos-start](https://github.com/leptos-rs/start) (or manually)
2. Add the following code to the `app.rs` right above or below the `<Stylesheet/>` call:
    ```rust
        <script type="importmap">
          {"{
            \"imports\": {
              \"firebase/auth\": \"https://www.gstatic.com/firebasejs/9.18.0/firebase-auth.js\",
              \"firebase/app\": \"https://www.gstatic.com/firebasejs/9.18.0/firebase-app.js\",
            }
          }"}
        </script>
    ```
3. See console - the error appears there. The importmap doesn't work and breaks the rest of the import process.

### Source of issue

This is caused by the file [integrations/utils/src/lib.rs](https://github.com/leptos-rs/leptos/blob/a9f778459a568c27472e37db41a3af85fbf80305/integrations/utils/src/lib.rs#L67)

```rust
    let head = format!(
        r#"<!DOCTYPE html>
            <html{html_metadata}>
                <head>
                    <meta charset="utf-8"/>
                    <meta name="viewport" content="width=device-width, initial-scale=1"/>
                    <link rel="modulepreload" href="/{pkg_path}/{output_name}.js">
                    <link rel="preload" href="/{pkg_path}/{wasm_output_name}.wasm" as="fetch" type="application/wasm" crossorigin="">
                    <script type="module">import init, {{ hydrate }} from '/{pkg_path}/{output_name}.js'; init('/{pkg_path}/{wasm_output_name}.wasm').then(hydrate);</script>
                    {leptos_autoreload}
                    "#
    );
```

As you can see from the code above, there is a `<script type="module">` being loaded before any user-specified script is provided in the source.

### Expected behavior

I expect some way to hook into the actual HTML content generated at custom points, more specifically hopefully at the top of the `<head>` element, or at the very least above the `<link rel="modulepreload" href="/{pkg_path}/{output_name}.js">` line.

### Possible solutions

[...]

PR #731, opened 2023-03-23
change: insert `<head>` metadata tags at the beginning of the `<head>`

gbj:
This allows, specifically, for JS import maps, which must occur before any module loading begins (which we do right away so we can start loading WASM).

This addresses #694.


toastxc:
Hi, I know this issue is already closed but I'm having the same problem.
Are there any decent examples of using `importmap`s?

gbj:
@toastxc please open a discussion or an issue with a minimal reproducible example so we can discuss it there.

PR #737, opened 2023-03-24
leptos-reactive: Bumped serde-lite from 0.3 to 0.4.

martinfrances107:
Just a minor change. 

PS by way of discussion when I look at bumping 

syn   1.0.109  ->   2.0.8 

it not possible at the moment, there is a messy sub dependency issue 

syn-rsx is moving more slowly that syn. -- in needs to internal starting using syn 2.x before we can increment to syn2.x


gbj:
Thanks. Yes, agreed... `syn` going to `2.0` is going to cause dependency issues for a bunch of libraries, and I think it's honestly going to be a mess to coordinate it across the ecosystem. We'll have to wait for all our dependencies to update to `2.0` before we can update our own, I think.

PR #738, opened 2023-03-24
fix: always run individual classes after the `class` attribute (closes #735)

### LINKED ISSUE #735: [Bug] Hard-coded CSS gets overwritten by conditional CSS in certain cases
As discussed on discord, the following input does not work as expected:

```html
<div class:dark=color_scheme_signal class="dummy">
```

It ends up outputting only the `dark` class.

On the other hand, changing the order of the hard-coded class to be first, before the conditional class, seems to work well:

```html
<div class="dummy" class:dark=color_scheme_signal >
```

In this case, both `dummy` as well as `dark` gets set to the element in the DOM.

CC @benwis

gbj:


gbj:
This passes CI locally but it's the GitHub SSH key issue, and I'm too checked out on a Friday afternoon to fix that, so I'm going to merge it...

PR #741, opened 2023-03-24
feat: Added ProtectedRoute component to route file

Kaszanas:
As discussed on Discord with @gbj it might be a good idea to introduce some sort of a ```ProtectedRoute``` that evaluates a condition and either redirects the user, or exposes a Route.

I am unsure if the condition is called correctly in its current form.

The overarching idea was taken from:  https://www.robinwieruch.de/react-router-private-routes/

### DIFF
--- a/router/src/components/route.rs
+++ b/router/src/components/route.rs
@@ -257,7 +257,7 @@ where
     P: std::fmt::Display + 'static,
     C: Fn(Scope) -> bool + 'static,
 {
-    if condition {
+    if condition(cx) {
         return view! {cx, <Route path=expose_path view=view />}.into_view(cx);
     } else {
         return view! {cx, <Redirect path=redirect_path /> }.into_view(cx);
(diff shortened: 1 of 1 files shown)


gbj:
I hadn't fully thought it through when we were discussing, but your version isn't quite compatible with how the router actually works. If you try it in your own app you'll find that it won't detect this route at all. 

I've made some adjustments so that `ProtectedRoute` is more similar to how the plain `Route` component works. I've done some local testing and it works as I'd expect it to.

This can then be used like this:
```xml
<Router>
    <main>
        <Routes>
            <Route path="" view=|cx| view! { cx, <p>"home"</p> }/>
            <Route path="blocked" view=|cx| view! { cx, <p>"blocked!"</p> }/>
            <ProtectedRoute
                path="protected"
                redirect_path="/blocked"
                condition=|_| true
                view=|cx| view! { cx, <p>"protected route content"</p> }
            />
        </Routes>
    </main>
</Router>
```

Kaszanas:
Yeah I suspected that there would be a need of a deeper dive into how a Route is constructed. Didn't have enough knowledge about the internals to try and apply it.

I will be testing this approach with a CSR, JWT login example and the ```ProtectedRoute``` will be receiving a signal containing a boolean specifying if the user is logged in.

My plan is to let you know if this works as intended.

PR #745, opened 2023-03-25
Document cargo workspace feature resolver footgun

jmintb:
Due to no rust edition being present in a workspac's Cargo.toml, non
WASM compatible code can end up being built for a WASM target.

This commit documents this error and how to resolve it.


gbj:
Yes, this looks great to me — I actually understand this much better after reading your short description here than after following along with all the conversation! Really helpful, I appreciate it.

Issue #747, opened 2023-03-26
Component macro does not allow usage of `::leptos::Scope`


lpotthast:
When writing a comopnent, the `component` macro requires the scope to be defined as `Scope` only.
That does not allow doing something like the following, which one might want to do in code generation:
```
#[::leptos::component]
pub fn MyComp(cx: ::leptos::Scope) -> impl ::leptos::IntoView {
   ::leptos::view! { cx,
        "ok"
    }
}
```
It would be nice if an absolute usage of `Scope` like shown above would not trigger the "this method requires a `Scope` parameter" compile-error.

PR #748, opened 2023-03-26
Allow component decl without `use leptos::*` in scope

lpotthast:
This  improves the behaviour / hygiene of the `component` macro, allowing for the following type of leptos component declarations, not requiring `use leptos::*` to be in scope.
```
#[::leptos::component]
pub fn MyComponent(cx: ::leptos::Scope) -> impl ::leptos::IntoView {
    ::leptos::view! { cx,
        "ok"
    }
}
```

Closes: https://github.com/leptos-rs/leptos/issues/747

### DIFF
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -16,6 +16,7 @@ attribute-derive = { version = "0.5", features = ["syn-full"] }
 cfg-if = "1"
 html-escape = "0.2"
 itertools = "0.10"
+once_cell = "1"
 prettyplease = "0.1"
 proc-macro-error = "1"
 proc-macro2 = "1"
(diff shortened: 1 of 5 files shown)


gbj:
Is there a more general solution here? My impression is that, for example, your PR now allows for `cx: Scope` and `cx: ::leptos::Scope` but not `cx: leptos::Scope`, and for `-> impl IntoView` and `-> impl ::leptos::IntoView` but not `-> impl leptos::IntoView`.

lpotthast:
I couldn't come up with a reason why `leptos::Scope` (for exmample) should not be supported as well. Added that option to the checks. Also created two helper functions, where each performs its syntax check in a single place while being backed by a vec of allowed cases.

I'm seeing now, that these helpers are always called negated. Can switch theri naming and remove the negation if you prefer that. I just generally start out with function names not containing any negation.

gbj:
Looks good -- thanks!

PR #785, opened 2023-04-01
Fix server functions with non-copy server contexts

Demonthos:
An unused bit of code in the server functions macro causes it to fail when non-copy types are set as the server context. This works for Leptos because scope is clone, but for other libraries like Dioxus this can cause issues.


gbj:
Yeah looks good, I think this some legacy logic to prevent warnings about an unused argument to the server function but I don't think it's even needed for that any more.

gbj:
Okay I added a commit to suppress the warning in the correct place without the assignment, so this should be good assuming it's still green. 

Thanks!

PR #787, opened 2023-04-01
Remove unused fs dependency from leptos_config

valeth:
This removes an unused outdated dependency from `leptos_config`.

It was tripping up my `cargo deny` as it was pulling in a very old `bytes` and `futures 0.1` version that I blacklisted.

I found one function that I think it was intended to be used in, as said function is async despite not using any await.


gbj:
Thanks, good catch!

Issue #788, opened 2023-04-02
feat: Allow Server Functions to be submitted via GET requests


benwis:
Allow server fns to be submitted via GET requests, and their responses cached by CDNs.

PR #789, opened 2023-04-02
Add the ability for server fns to be submitted via GET requests

benwis:
Per popular request, I added the ability for server fns to be submitted via GET requests, with their args included in a query string! The encoding "GetJSON" and "GetCBOR" will do that, and return the response either via JSON or CBOR.

Let me know if I broke something, because I probably did. I modified the axum integration to make this work, but not the Actix one yet, and there's plenty of docs that still need to be updated.

Closes issue #788

### DIFF
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -85,8 +85,7 @@ use proc_macro2::TokenStream;
 use quote::TokenStreamExt;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 pub use server_fn_macro_default::server;
-use std::sync::Arc;
-use std::{future::Future, pin::Pin, str::FromStr};
+use std::{future::Future, pin::Pin, str::FromStr, sync::Arc};
 use syn::parse_quote;
 use thiserror::Error;
 // used by the macro
@@ -299,7 +298,7 @@ where
     ) -> Pin<Box<dyn Future<Output = Result<Self::Output, ServerFnError>>>>;
 
     /// Registers the server function, allowing the server to query it by URL.
-    #[cfg(any(feature = "ssr", doc, ))]
+    #[cfg(any(feature = "ssr", doc,))]
     fn register_in<R: ServerFunctionRegistry<T>>() -> Result<(), ServerFnError>
     {
         // create the handler for this server function
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -85,7 +85,6 @@ use proc_macro2::TokenStream;
 use quote::TokenStreamExt;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 pub use server_fn_macro_default::server;
-#[cfg(any(feature = "ssr", doc))]
 use std::sync::Arc;
 use std::{future::Future, pin::Pin, str::FromStr};
 use syn::parse_quote;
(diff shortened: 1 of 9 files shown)


gbj:
@Demonthos We're expanding server functions to allow GET requests, with either JSON or CBOR responses. In other words, rather than POSTing a URL-encoded set of arguments, you could GET with them in the URL query string. This is useful because it enables people to set `Cache-Control` and other headers, and if they're behind a CDN they can cache server function requests across users that way.

Wanted to get your input as this would affect your implementation as well. Basically this just adds a couple new encoding options to be handled.

Demonthos:
The new encoding options look great! It might be worth considering changing the get function to return a complete server function object that contains both the encoding and the server function so that we don't need to maintain a separate hashmap for the encoding.

benwis:
> The new encoding options look great! It might be worth considering changing the get function to return a complete server function object that contains both the encoding and the server function so that we don't need to maintain a separate hashmap for the encoding.

I thought about that, the issue being that  REGISTERED_SERVER_FUNCTIONS returns a HashMap of ServerFnTraitObj. It would be fairly trivial to change that to return a struct containing both things though, if you and @gbj  think that's preferable

server_fn/src/lib.rs
```diff
@@ -379,54 +427,94 @@ where
...
-            .header("Content-Type", content_type_header)
-            .header("Accept", accept_header)
-            .body(s)
-            .send()
-            .await
-            .map_err(|e| ServerFnError::Request(e.to_string()))?,
+                .map_err(|e| ServerFnError::Request(e.to_string()))?,
+        },
+        Encoding::GetCBOR | Encoding::GetJSON => match args_encoded {
+            Payload::Binary(_) => panic!(
```
gbj, line 462: Checking out different combinations I ran into this and was a little surprised. I would have expected that `GetCBOR` would serialize the arguments as a URL-encoded string and the response as CBOR. As it stands, is there any scenario in which `GetCBOR` doesn't panic?

server_fn_macro/src/lib.rs
```diff
@@ -274,6 +274,10 @@ impl Parse for ServerFnName {
             .map(|encoding| match encoding.to_string().as_str() {
```
gbj, line 274: For this block, it's probably worth matching on `encoding.to_string().to_lowercase()` so you don't have to handle different capitalization combinations.

server_fn/src/lib.rs
benwis: As far as I understand it, the Payload defined here is the Request Payload from the client, which cannot be binary because it's going into the query string.

server_fn_macro/src/lib.rs
benwis: Yeah, that would simplify things a bit in terms of code length.

gbj:
I was pointing to the wrong line with my comment about `GetCBOR`.

One of the two commits I just pushed should fix it by encoding `GetCBOR` requests as `GET` requests with URL-encoded arguments, but with `CBOR` responses. Otherwise trying to run a `GetCBOR` server function just panicked.

The other one tweaks the Actix handler so it can accept any HTTP method (although obviously mostly GET and POST). Note that `get()` handlers have to added manually in `main.rs` for Axum/Viz, which is fine.

...The other two commits are `cargo fmt` and `clippy`.

benwis:
Looks good to me, can't believe I missed the panic

gbj:
All good. Thanks for your great work as always, Ben! Merging now.

PR #794, opened 2023-04-03
fix: unused warning on cx in server functions

markcatley:
When running cargo clippy on server functions that use `cx: Scope` it has an unused variable error.

It appears that the logic for adding an `#[allow(unused)]` notation is inverted.


gbj:
Thanks!

PR #808, opened 2023-04-05
Optimize Runtime::mark_dirty

novacrazy:
Looking for places to remove memory allocations. Found this intermediate buffer and removed it.

I also marked `Runtime::mark` as `#[inline(always)]`. Due to opt-level s and z aggressively _avoiding_ inlining, such a hot function as `mark` would likely incur significant overhead to call in the loops.

Might even be worth looking into rerunning the js benchmarks after this.


novacrazy:
This needs some extra work on it to avoid duplicate traversals and accumulation onto `pending_effects`. Doesn't actually hurt anything, since `update_if_necessary` is a no-op if the node is already clean, but I'd like to find an efficient way to stop descension early.

PR #809, opened 2023-04-05
Optimize memory usage of update methods

novacrazy:
Removes a (potentially unused) clone from `Runtime::update` and shrinks what memory/operations are needed in `update_if_necessary`


gbj:
Thanks for both these contributions!

PR #815, opened 2023-04-06
Use local pools for axum handlers

### LINKED ISSUE #811: Make Axum integration less bad
The way I hacked together the `leptos_axum` code to allow for the fact that our reactive system is `!Send` is bad. Like, quite bad.
https://github.com/leptos-rs/leptos/blob/8b57ba7aa8498832cc5aade2764333dce2956dd9/integrations/axum/src/lib.rs#L296-L303
...and so on.

Spawning a new runtime per request and then blocking is not great.

Luckily, this has not stopped a bunch of people joining the community who are quite good. I'm opening this issue for accountability so the good people can help me make the bad code less bad.

Specifically, we've talked about the idea of creating a `thread_local` `LocalSet` and spawning these `!Send` futures into those instead. There may be other good ideas here, too.

akarras:
Should make axum handlers less bad should close #811

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -32,8 +32,13 @@ use leptos_integration_utils::{build_async_response, html_parts_separated};
 use leptos_meta::{generate_head_metadata_separated, MetaContext};
 use leptos_router::*;
 use parking_lot::RwLock;
-use std::{io, pin::Pin, sync::Arc};
-use tokio::task::{spawn_blocking, LocalSet};
+use std::{
+    io,
+    pin::Pin,
+    sync::{Arc, OnceLock},
+};
+use tokio::task::LocalSet;
+use tokio_util::task::LocalPoolHandle;
 
 /// A struct to hold the parts of the incoming Request. Since `http::Request` isn't cloneable, we're forced
 /// to construct this for Leptos to use in Axum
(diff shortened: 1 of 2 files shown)


integrations/axum/src/lib.rs
```diff
@@ -1214,3 +1141,13 @@ impl LeptosRoutes for axum::Router {
         router
     }
 }
+
+fn get_leptos_pool() -> LocalPoolHandle {
+    static LOCAL_POOL: OnceLock<LocalPoolHandle> = OnceLock::new();
+    LOCAL_POOL
+        .get_or_init(|| {
+            let pool = tokio_util::task::LocalPoolHandle::new(1);
```
akarras, line 1149: I will mention, at the moment there's only a single thread in the pool. We might want to consider making this configurable, or matching the existing tokio runtime settings.

benwis: Yes! How about a call to https://doc.rust-lang.org/std/thread/fn.available_parallelism.html so we create one for however many threads are available

gbj:
Anything else to be done on this? I think it's ready for review, right?

akarras:
Should be ready for review yep.

gbj:
Nice. Yeah, this all looks good to me! Thanks so much for taking it on, and glad you found this library, @FrankReh.

PR #817, opened 2023-04-06
Publish book ci

bram209:
After merging & running this workflow, you have to update a few settings in the repository:
image

After that, the docs should be accessible from: https://leptos-rs.github.io/leptos/


d4h0:
@gbj: Just in case you missed it, did you see @bram209's edit above? When I go to https://leptos-rs.github.io/leptos/ I still get a 404.

bram209:
> @gbj: Just in case you missed it, did you see @bram209's edit above? When I go to https://leptos-rs.github.io/leptos/ I still get a 404.

The reason is that the workflow has never ran yet. As soon as a change to the book is pushed, it will be published.
@d4h0 for now feel free to temporarily use https://bram209.github.io/leptos/ until that happens ;)

d4h0:
I see, thanks @bram209!

gbj:
Thanks @d4h0 @bram209, I've gotten my act together to update the book slightly and set it to publish, and the action is running now.

PR #831, opened 2023-04-07
Various optimizations, size reductions and stability improvements

novacrazy:
The goal of all these changes is to reduce WASM binary size with `opt-level = "z"` and reduce the number of allocations for tiny objects. 

While `#[inline(always)]` is usually discouraged, when using `opt-level = "z"` LLVM strongly avoids inlining to the point it can sometimes not inline enough, causing extra function calls/indirection and duplicate code. My internal guidelines for inlining here is to emulate what `opt-level=3` would do automatically for simple functions. "Simple" being barely more than a single function call within. Furthermore, functions that take `impl Fn()` almost always create a unique monomorphization of that function for the callback closure type, so we may as well force-inline it where that closure is, allowing the compiler to merge them and potentially optimize further. If necessary I could change them all to `#[cfg_attr(target_arch = "wasm32", inline(always))]`

Other changes:

* Quite a bit of non-generic code has been pulled out of generic functions. This is slightly worse for artificial benchmarks, but significantly reduces compiled binary size.
* `ScopeDisposer` no longer allocates. 
* Fixed nested `batch` calls. Previously, `batching` was set to `false` and effects were run on every call to `batch`, but now effects will only be run on the top-most call to batch, when nested. 
* `batch` and `untrack` are now somewhat resilient to panics, even if nothing else is yet.
* I changed the `fn bubbles(&self) -> bool` method on `EventDescriptor` to an associated constant `const BUBBLES: bool`, as this is not a dynamic property, nor is it ever used in a `dyn` context. Keeping it const potentially allows for dead-code elimination of certain branches. Hard to predict if this helps, but it doesn't hurt.
* `Attribute`, `IntoAttribute` and most standard meta components have been updated to use `Cow<'static, str>`. The downside of this is the removal of the `impl IntoAttribute for &'_ str` implementation. However, static attributes do not allocate now, which is a very common case. Unsure if it was even possible for attributes to be a non-static str in practice before.
* Added a "fast-path" in `mark_dirty` for linearly nested children. Avoids boxing the iterator and the single iteration. 
* Various other tiny cleanups I noticed along the way, e.g.: enum Defaults, some code organization. 

Here is an example of the binary size improvements seen for the Hackernews example:
```
lto = false (default):
  Hackernews 0.2.5: 498,180 bytes
  Hackernews main branch: 483,875 bytes
  Hackernews my changes: 448,513 bytes
lto = true:
  Hackernews 0.2.5: 410,247 bytes
  Hackernews main branch: 397,740 bytes (154KB gzipped)
  Hackernews my changes: 376,679 bytes (149KB gzipped)
```


gbj:
Having just read through every line, I want to say: Thank you, this is really fantastic. 

I think I agree with all of the decisions here, and wanted to call out especially the change to`ScopeDisposer` (very good clean up from an ancient version of this, avoiding the allocation) and preventing as-yet-unreported bugs with nested batching and untrack.

gbj:
Oh I should add: I've swapped over to `criterion` benchmarks inside `leptos_reactive`, and this PR gave a smallish but (supposedly) statistically significant ~4-5% improvement on "deep update" (chain of 1000 memos) and "fan out" (1 signal to 1000 memos), with no significant change to "narrow down" (1000 signals to 1 memo). The benchmarks for other frameworks do tend to bounce around a bit between runs, sometimes even reporting significant improvements within an actual version change, so I'd say this is all either within the margin of error or a slight improvement, in release mode SSR, so — very good.

PR #835, opened 2023-04-09
chore: deny warnings on examples in Github Actions

markcatley:


markcatley:
It doesn't seem to work in this case and I haven't figured it out. For some reason the environment variable doesn't get passed down the chain in cargo make. I'm not really sure why.

gbj:
This has been superseded by later work on verifying examples in CI so I'm going to close it.

PR #836, opened 2023-04-09
example: personal blog

LeVuMinhHuy:
a simple blog, with router and styles


LeVuMinhHuy:
will move it in here https://github.com/leptos-rs/awesome-leptos

PR #838, opened 2023-04-09
[WIP] Trigger prototype

novacrazy:
I would like to introduce the `Trigger` primitive to use in place of `RwSignal<()>`, simply to avoid unnecessary allocations and overhead in cases where many hundreds or thousands of them may be present.

`Rc<RefCell<dyn Any>>` will allocate 8-16 bytes (32-bit/64-bit) on the Rust heap even if the value stored is a unit type, while `Option<Rc<Refcell<dyn Any>>` will not allocate if `None`, and thanks to null-pointer optimization of `Option` will not increase the size of `ReactiveNode` at all. It also avoids the overhead of regular signal functionality. 

If not used by an application, this only adds about 200 bytes to the compiled WASM.

TODO:
- [x] Documentation
- [x] Testing

Which I'll likely get to tomorrow. Submitting this as a draft for early feedback. 

My later plan is to use triggers inside a reactive Store structure generated by proc-macros to emulate Proxies in Rust.

### DIFF
--- a/leptos_reactive/src/diagnostics.rs
+++ b/leptos_reactive/src/diagnostics.rs
@@ -1,5 +1,3 @@
-use cfg_if::cfg_if;
-
 // The point of these diagnostics is to give useful error messages when someone
 // tries to access a reactive variable outside the reactive scope. They track when
 // you create a signal/memo, and where you access it non-reactively.
@@ -23,7 +21,7 @@ pub(crate) struct AccessDiagnostics {}
 #[doc(hidden)]
 pub struct SpecialNonReactiveZone {}
 
-cfg_if! {
+cfg_if::cfg_if! {
     if #[cfg(debug_assertions)] {
         use std::cell::Cell;
 
@@ -66,7 +64,7 @@ impl SpecialNonReactiveZone {
 #[macro_export]
 macro_rules! diagnostics {
     ($this:ident) => {{
-        cfg_if! {
+        cfg_if::cfg_if! {
             if #[cfg(debug_assertions)] {
                 AccessDiagnostics {
                     defined_at: $this.defined_at,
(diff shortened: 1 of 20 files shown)


leptos_reactive/src/diagnostics.rs
gbj, line 1: Makes sense, thanks for the hygiene.

leptos_reactive/src/trigger.rs
```diff
@@ -0,0 +1,187 @@
...
+use crate::{
+    diagnostics,
+    diagnostics::*,
+    node::NodeId,
+    runtime::{with_runtime, RuntimeId},
+    Scope, ScopeProperty, SignalGet, SignalSet, SignalUpdate,
+};
+
+pub trait TriggerDirty {
+    fn dirty(&self);
```
gbj, line 13: I wonder whether there's a better name for `dirty` here. `.trigger()`? `.mark_dirty()`? It wasn't initially obvious to me whether this meant `.is_dirty()` or `.mark_dirty()`, since I don't often use "dirty" as a verb on its own.

leptos_reactive/src/signal.rs
```diff
@@ -1997,13 +1996,14 @@ impl NodeId {
                 None
             };
 
-            // mark descendants dirty
-            runtime.mark_dirty(*self);
-
             // notify subscribers
-            if updated.is_some() && !runtime.batching.get() {
```
gbj, line 2004: Yeah looks like this batching check is redundant with the one that's already in `run_effects`, so this makes sense.

gbj (commented):
Yeah this looks good to me overall!

novacrazy:
Apologies, I couldn't help but do some additional cleanup. Doc links (within `leptos_reactive`) to types and functions have been wrapped in backticks for markdown formatting. Also `mark_dirty` has been tweaked a bit more.

gbj (commented):
Okay, thanks! All looks good to me. Appreciate the formatting. 

I think `.notify()` is a good name. I'm ready to merge this if you are happy with it.

PR #847, opened 2023-04-11
feat: render resources and fragments together

akarras:
Might make things better might make things worse! I think things will probably be better though.

### DIFF
--- a/leptos_dom/src/ssr.rs
+++ b/leptos_dom/src/ssr.rs
@@ -7,7 +7,10 @@ use crate::{
     CoreComponent, HydrationCtx, IntoView, View,
 };
 use cfg_if::cfg_if;
-use futures::{stream::FuturesUnordered, Future, Stream, StreamExt};
+use futures::{
+    stream::{self, LocalBoxStream},
+    Future, FutureExt, Stream, StreamExt,
+};
 use itertools::Itertools;
 use leptos_reactive::*;
 use std::{borrow::Cow, pin::Pin};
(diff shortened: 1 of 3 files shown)


gbj:
I think the issue here is that this particular approach actually causes a race condition between `<Suspense/>`, resources, and WASM in certain conditions, which is why it's currently 1) all fragments, then 2) all resources.

Here's the race condition: Say you have one resource which is read under one `<Suspense/>`, and some WASM to hydrate.

In that case the resource will pretty much always be sent before the `<Suspense/>` that reads it, although I guess it depends on whether it takes longer to serialize the JSON or to render the HTML for the suspended fragment.

If things load in the order:
- Suspense Fragment - Resource - WASM or
- Suspense Fragment - WASM - Resource
then everything is fine. By the time the WASM is loaded and the resource has resolved, the resolved HTML will already be there and then fragment will hydrate correctly.

But if it loads in the order
- WASM - Resource - Suspense or
- Resource - WASM - Suspense
then when the page goes to hydrate it will *expect* to find the resolved Suspense fragment in the HTML, but it hasn't arrived yet. The renderer currently breaks on mismatches between the HTML present and what it expects during hydration.

So "fragments first, then resources" is generally a good idea. *Except* that if you have 5 resources read under 3 `<Suspense/>` components, there are cases in which it would actually be better to stream one of the fragments and its resources, because the other fragments don't depend on them, then the other fragments, etc. It involves some work to parse out the dependency tree here and only stream a given resource if it is not used in any fragments that are still pending.

akarras:
That's some great context thank you.
I think I'm going to go ahead and close this, but will continue thinking about ways we could improve this.

PR #854, opened 2023-04-12
fix: Strip `&` from the end of params queries

mondeja:
The method `ParamsMap.to_query_string()` currently appends a `&` character at the end of query strings when they have one param or more, which is very ugly, specially to share URLs copy- pasting them.


gbj:
Yep makes sense. Thanks!

PR #863, opened 2023-04-14
Fix the counter without macros test

agilarity:
These changes:

- Add a cargo make task to run `wasm` tests
- Address compile errors
- Rename test module to remove extraneous test output

Run `cargo make test-flow` to execute unit and `wasm` tests.


gbj:
Oh nice, thank you very much!

PR #864, opened 2023-04-14
feat: add `expect_context` function

markcatley:
Most of the time when using use_context it would be a bug if the context wasn't present and appropriate to panic. This is a convenience function that has that behavior.

### DIFF
--- a/leptos_reactive/src/context.rs
+++ b/leptos_reactive/src/context.rs
@@ -187,9 +187,14 @@ where
 ///     todo!()
 /// }
 /// ```
-pub fn use_unwrapped_context<T>(cx: Scope) -> T
+pub fn expect_context<T>(cx: Scope) -> T
 where
     T: Clone + 'static,
 {
-    use_context(cx).expect("context to be present")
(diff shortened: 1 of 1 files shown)


gbj:
This seems like a reasonable helper function to include, but I'm feeling uncertain about the name: no pun intended, but I expect an `expect` function to let me specify a panic message, and this is more like an `unwrap_context()` expect that it specifies an arbitrary panic message I can't control.

markcatley:
I'm easy on the name, maybe `use_unwrapped_context`?

gbj:
Sure I'm happy with `use_unwrapped_context`.

gbj:
CI only failed because the name wasn't updated in the doctest here, but it gave me an opportunity to have second thoughts. I think your initial `expect_context` is probably better, just because the goal here is to be more terse.

`use_unwrapped_context(cx)` is actually longer than 
`use_context(cx).unwrap()`, whereas
`expect_context(cx)` is significantly shorter and still follows the verb-noun pattern.

Sorry to waffle, I'll just switch it back to `expect_context` myself and then merge!

PR #876, opened 2023-04-16
docs: Add per-project toolchain override readme

kamilogorek:
This makes it easier to play around with Leptos without changing global settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,14 +65,22 @@ Here are some resources for learning more about Leptos:
 
 ## `nightly` Note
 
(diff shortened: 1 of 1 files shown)


d4h0:
Another option is to use the `+nightly` switch (e.g. `cargo +nightly run --example xy`). This is especially useful, when running example from the repo (which doesn't seem to specify that nightly is required).

gbj (approved):
Looks good to me, thanks!

Issue #857, opened 2023-04-13
`#server` url munging broken under Bazel


snapbug:
👋 Hi, the url munging for server functions breaks when using the Bazel build system.

Specifically [this line](https://github.com/leptos-rs/leptos/blob/main/server_fn_macro/src/lib.rs#L211), which uses `env!("CARGO_MANIFEST_DIR")` to munge the urls causes the following:

On the server side,
```rust
register_server_functions()
info!("{:#?}", leptos_server::server_fns_by_path());
info!(concat!(env!("CARGO_MANIFEST_DIR")));
```
results in the following output:
```
[
    "get_single_config4869621392562198152",
    "get_event_config2255429015822590071",
]
/private/var/tmp/_bazel_user/4b541d909237bb1ea4e282a886cf6762/sandbox/darwin-sandbox/5148/execroot/project/www
```

On the client side, sticking
```rust
log::info!(concat!(env!("CARGO_MANIFEST_DIR")));
```
results in the following output:
```
/private/var/tmp/_bazel_user/4b541d909237bb1ea4e282a886cf6762/sandbox/darwin-sandbox/5140/execroot/project/www
```
and using the chrome developer tools (and logging on the server side, a representative url being hit for serverfn is):
```
http://localhost:8000/api/get_event_config12333777864187817067
```
different from the registered `get_event_config2255429015822590071`, results in `http 400` errors.

Because the client wasm bundle and the server binary are different bazel targets they get executed in their own sandbox directory. This results from the minor difference in the `CARGO_MANIFEST_DIR` (`5148` vs `5140`).

I realize this is somewhat down to build tool choice, but I'm working in an environment where that's not changeable, but Bazel has plenty of other upsides even if it were a Greenfield project 🙂.

I _suspect_ that this was added in to help reduce potential of naming collisions? Can this clobbering be made an option in the macro? Happy to try and attempt once day job doesn't get in the way. This behavior wasn't exhibited in 0.2.1 which I'e been developing against for now, and I'd rather not be stuck on.

Thanks in advance.

gbj:
Makes sense to me. Yes, this is a way of preventing naming collisions that works on both stable and nightly. 

[...]

PR #878, opened 2023-04-16
Use override key, if available, for server function calling

snapbug:
Following up on #857 

I tried the suggested `option_env!(...).unwrap_or(env!(...))`, but that runs into an error within the `concatcp` macro ([note: erroneous constant used](https://gist.github.com/snapbug/6c24a18d7c8998f93e76bc0d8d023a3a)). I tried to get multiple alternatives within the generated code to work, but all ran into this, or very similar variants of it.

I thought that the build context of `server_fn_macro` would probably be the same as that of the library being built, meaning the env that exists there probably exists when the lib using `#[server...]` is being built.

So, take advantage of that to decide which env var to use when building the library.

Results from cargo expand on counter isomorphic file:

Override env not set:
`cargo expand --bin counter_isomorphic --features="stable ssr" > output.rs`
```rust
::leptos::server_fn::xxhash_rust::const_xxh64::xxh64(
  "/Users/mcrane/tmp/leptos/examples/counter_isomorphic:src/counters.rs:26:1",
  0,
)
```

`SERVER_FN_OVERRIDE_KEY=test cargo expand --bin counter_isomorphic --features="stable ssr" > output.rs`
and with it set:
```rust
::leptos::server_fn::xxhash_rust::const_xxh64::xxh64(
  "test:src/counters.rs:26:1".as_bytes(),
  0,
)
```

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -195,8 +195,8 @@ pub fn server_macro_impl(
         .unwrap_or_else(|| quote! { server_fn });
 
     let key_env_var = match option_env!("SERVER_FN_OVERRIDE_KEY") {
-         Some(_) => "SERVER_FN_OVERRIDE_KEY",
-         None => "CARGO_MANIFEST_DIR",
+        Some(_) => "SERVER_FN_OVERRIDE_KEY",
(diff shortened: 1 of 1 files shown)


gbj:
Yeah looks good to me! Thanks very much.

PR #884, opened 2023-04-17
Split counter without macros web test

agilarity:
This improves the test by extracting distinct scenarios. 

It also demonstrates how you can:

- Remount the counter
- Frame tests from the user's point of view
- Query elements by exact text

Run `cargo make test-flow` to verify.


agilarity:
This change covers #885.

gbj:
Looks good — thanks!

PR #891, opened 2023-04-18
Fixes #890 that was using OnceLock, which is nightly only, by adding the once_cell crate as a dependency.

### LINKED ISSUE #890: `leptos_axum` does not build on stable due to use of unstable `OnceLock`
Greets, `leptos_axum` does not build on stable due to use of unstable `OnceLock`, perhaps using the `once_cell` crate would be fine until that stabilizes?

Applying this minimal patch to use once_cell gets it working for me (I can submit a PR if wished, don't know if once_cell is wanted as a new dependency for just this though):

```diff
diff --git a/integrations/axum/Cargo.toml b/integrations/axum/Cargo.toml
index 58c7edfa..231038f5 100644
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -19,4 +19,5 @@ leptos_integration_utils = { workspace = true }
 serde_json = "1"
 tokio = { version = "1", features = ["full"] }
 parking_lot = "0.12.1"
-tokio-util = {version = "0.7.7", features = ["rt"] }
\ No newline at end of file
+tokio-util = {version = "0.7.7", features = ["rt"] }
+once_cell = "1.17"
diff --git a/integrations/axum/src/lib.rs b/integrations/axum/src/lib.rs
index 19cdb7c1..0972ecb3 100644
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -35,9 +35,10 @@ use parking_lot::RwLock;
 use std::{
     io,
     pin::Pin,
-    sync::{Arc, OnceLock},
+    sync::Arc,
     thread::available_parallelism,
 };
+use once_cell::sync::OnceCell;
 use tokio::task::LocalSet;
 use tokio_util::task::LocalPoolHandle;
 
@@ -1193,7 +1194,7 @@ impl LeptosRoutes for axum::Router {
 }
 
 fn get_leptos_pool() -> LocalPoolHandle {
-    static LOCAL_POOL: OnceLock<LocalPoolHandle> = OnceLock::new();
+    static LOCAL_POOL: OnceCell<LocalPoolHandle> = OnceCell::new();
     LOCAL_POOL
         .get_or_init(|| {
             tokio_util::task::LocalPoolHandle::new(

```

OvermindDL1:
This fixes #890 where leptos_axum is using the nightly only `OnceLock`, switching to use the `once_cell` dependency for now as per gbj.

### DIFF
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -20,3 +20,4 @@ serde_json = "1"
 tokio = { version = "1", features = ["full"] }
 parking_lot = "0.12.1"
 tokio-util = {version = "0.7.7", features = ["rt"] }
+once_cell = "1.17"
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -31,14 +31,9 @@ use leptos::{
 use leptos_integration_utils::{build_async_response, html_parts_separated};
 use leptos_meta::{generate_head_metadata_separated, MetaContext};
 use leptos_router::*;
-use parking_lot::RwLock;
-use std::{
-    io,
-    pin::Pin,
-    sync::Arc,
-    thread::available_parallelism,
-};
 use once_cell::sync::OnceCell;
+use parking_lot::RwLock;
+use std::{io, pin::Pin, sync::Arc, thread::available_parallelism};
 use tokio::task::LocalSet;
 use tokio_util::task::LocalPoolHandle;
 
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -35,9 +35,10 @@ use parking_lot::RwLock;
 use std::{
     io,
     pin::Pin,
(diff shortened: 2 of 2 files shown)


OvermindDL1:
Is there a todo list for things like saying to switch to stable when the minimum rust version is raised to an appropriate version (OnceLock is supposedly going to be stabilized in the next rust stable, or maybe the one after, but when will the minimum rust version raise for leptos as well)?

gbj:
Thanks for the contribution!

> Is there a todo list for things like saying to switch to stable when the minimum rust version is raised to an appropriate version (OnceLock is supposedly going to be stabilized in the next rust stable, or maybe the one after, but when will the minimum rust version raise for leptos as well)?

We've never had an official MSRV. We do use `once_cell` in a couple places across the project, so at some point will probably adopt the `std` version there, but maybe best once it's a few releases in the rear-view mirror. It may be wise to get more formal about this.

PR #899, opened 2023-04-19
Remove need for LEPTOS_OUTPUT_NAME env var after compilation

benwis:
Tweaked the get_configuration function and the checks to make LEPTOS_OUTPUT_NAME required for compilation, but not with the compiled binary.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -77,6 +77,7 @@ install_crate = "cargo-all-features"
 
 [env]
 RUSTFLAGS=""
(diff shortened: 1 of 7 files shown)


benwis:
Ha! This broke all the examples building(because we're not using cargo-leptos to build them) and some of the tests. Fantastic

benwis:
It's building past that at least, but the response types in the leptos_config tests have changed.

gbj:
Thanks! This is awesome.

PR #901, opened 2023-04-19
Check output name at build time

mbergkvist:
This removes the requirement to provide `LEPTOS_OUTPUT_NAME` at run time.


gbj:
I think this is addressing the same issue as #899. Maybe you and @benwis could both look at each other's solutions and see if there's anything missing, or any pros/cons you can think of? (Full disclosure: I haven't read either one very closely yet.)

mbergkvist:
The same issue _is_ addressed in #899. It also includes some documentation updates and a improved compile error message, so I'm closing this issue.

PR #902, opened 2023-04-19
Added get_untracked for node_ref.

dgsantana:
Avoids getting messages of unused reactivity.

When working with simple event handlers the use of NodeRef::get() triggers a diagnostic due to inner RwSignal that gets it `get` called. For this cases the simple get_untracked avoids this diagnostic message.


gbj:
Thanks!

Issue #210, opened 2023-01-01
CI: add Playwright or `wasm-bindgen-test` testing for examples and run them using `cargo-make`


gbj:

agilarity:
### Progress Report

#### GitHub Workflows

Example workflows only run when needed.

- `Check Examples` runs when a leptos source package is changed
- `Verfiy All Examples` runs daily and on version change
- `Verify Changed Examples` runs for examples that changed

#### Cargo Make

Custom `cargo make` tasks and files make it easy to configure and run `ci` consistently.

- `cargo make` tasks run at the workspace and project levels
- `cargo make` files provide a one-line strategy to add support for WASM Test, Playwright Test, Cargo Leptos Test, and Trunk- `test-info` task reports how example projects are configured for `ci` (#1411)

#### CI Testing

```console
CI test runners by example...

counter -C-W
counters -W
counters_stable -W-P
counter_without_macros -R-W
js-framework-benchmark -W
router -P
tailwind -L

Test Runners: C = Cargo Test, L = Cargo Leptos Test, P = Playwright Test, R = RS Test, W = WASM Test
```

- 7 examples (26.9%) have some `ci` testing
- Examples demonstrate testing with WASM and Playwright
- Examples demonstrate two ways to start the client for e2e testing (See `counters_stable` and `tailwind`)

#### What's Left?

We have a stable example testing platform with a nice DX.

Apart from adding more tests, two potential improvements come to mind.

- [ ] Pull up nightly `build` and `check` tasks into `cargo-make`
- [ ] Refactor `check-stable.yml` to use `run-cargo-make-task.yml`

agilarity:
### Progress Report

#### GitHub Workflows

Example workflows only run when needed.

- `Check Examples` runs when a leptos source package is changed
- `Verfiy All Examples` runs daily and on version change
- `Verify Changed Examples` runs for examples that changed

#### Cargo Make

Custom `cargo make` tasks and files make it easy to configure and run `ci` consistently.

- `cargo make` tasks run at the workspace and project levels
- `cargo make` files provide a one-line strategy to add support for WASM Test, Playwright Test, Cargo Leptos Test, and Trunk- `test-info` task reports how example projects are configured for `ci` (#1411)

#### CI Testing

```console
CI test runners by example...

counter -C-W
counters -W
counters_stable -W-P
counter_without_macros -R-W
js-framework-benchmark -W
router -P
tailwind -L

[...]

PR #904, opened 2023-04-20
Use cargo make to run tests for examples

agilarity:
This makes it easier to run checks and tests for all examples. 

It also covers the `cargo make` part of  #210.

The Solution:

- Makes the examples directory a workspace
- Adds task flows that support lint, unit, and web testing
- Makes web testing optional
- Demonstrates calling from the leptos workspace

Run `cargo make test-examples` in the root to verify.

The `counters` example is excluded from the members until the test compile errors are fixed.


gbj:
Thanks! This looks good to me.

Issue #769, opened 2023-03-30
RFC: Typed Component Children for `view!` Macro


jquesada2016:
# Summary

Allow specifying typed children of components at compile-time rather than at runtime.

# Motivation

Currently, components can only have a default `children` prop that accepts a fragment, which can contain any type of child. This can be restrictive for API possibilities, and is less compile-time safe, since components cannot explicitely opt in to requiring specific child components from being present.

Let's take as an example an `<If />` component. I will be using this, as it's a familiar construct to many, although this can apply to many other components, such as leptos' `<Router />`.

Let's say I would like to support the following API:

```rust
view! { cx,
  <If signal=a_is_true>
    <Then>"show content when a is true"</Then>
    <ElseIf signal=b_is_true>"b is true"</ElseIf>
    <ElseIf signal=c_is_true>"c is true"</ElseIf>
  <Else>"none of the aboe are true"</Else>
  </If>
}
```

As the API stands right now, the only way to support this, is to take a fragment which must be checked at runtime to make sure the children you require exist, and decide what to do with extrenuous children..

One possibility would be for us to do:

```rust
#[component]
pub fn If(
  cx: Scope,
  signal: MaybeSignal<bool>,
  then: Then,
  else_if: Option<Vec<ElseIf>>,
  else_: Option<Else>,
) -> impl IntoView { /* ... */ }
```

This would be far from ideal, because it promotes "flow breaks", which is what I call when you need to break out of the current `view!` in order to define your input outside of it, and then come pack to the flow of your component. In this example, this would be:

```rust
let then = view! { cx, /* ... */ };

view! { cx,
  <If signal then />
}
```

or, even worse:

```rust
view! { cx,
  <If signal then=view! { cx, /* ... */ } />
}
```

Thus, we need a better way to be able to declare inline where component children should go.

# Guide-level Explanation

Let's introduce the concept of `slots`. The idea is borrowed from Svelte. A slot is a pseudo attribute that can be added to children of a component to denote that they should not be included into a component's default children. For example:

```rust
view! { cx,
  <If signal>
    <Then slot:then>/* ... */</Then>
  </If>
}
```

Here, whe're telling the `view` macro that the `<Then />` child should not be included in `.children()`, but rather in `.then()`. This can be rewritten, and is equivalent to:

```rust
view! { cx,
  <If signal then=view! { cx, <Then>/* ... */</Then> } />
}
```

This has the advantage of components which relied on the following pattern:

```rust
#[component]
[...]

PR #909, opened 2023-04-20
feat: add slots

yuuma03:
This adds support for slots as described in #769. Prop passing (as described in the RFC) is still missing.
Recursively adding event listeners to the children of the prop is also missing.
Instead of using the same component macro+props, I made a new macro specific for slots.
I'm still quite new to Rust (and to contributing to open source projects in general), so any opinions on the code/what to change are appreciated.


jquesada2016:
@yuuma03 thank you very much for trying your hand at this RFC! Just a side note, I just updated the RFC's reference-level guide to include mentions of using the `Into` trait when taking props, and an example as to why we might want this.

@gbj and myself will take a look at this PR as soon as we get a chance.

leptos_macro/src/lib.rs
```diff
@@ -644,6 +645,128 @@ pub fn component(args: proc_macro::TokenStream, s: TokenStream) -> TokenStream {
...
+///           <SlotWithChildren slot:slot on:click=move |_| {}>
+///             <h1>"Hello, World!"</h1>
+///           </SlotWithChildren>
+///         </ComponentWithSlot>
+///     }
+/// }
+/// ```
+///
+/// ```
+/// // ✅ Until fixed, do this instead
```
gbj, line 728: FWIW I think this is probably the correct behavior and not something that needs to be fixed: If a "slot" is defined as a way of building a struct and passing it as one of the props to another component, I can't imagine what it would mean to add an event listener to it, except in the specific case of using a slot to pass through `children`, in which case — because `children` can return, for example, a fragment or nothing — the event listener either needs to be added on an element inside the slot, or needs to be added by the parent component that takes the slot.

leptos_macro/src/view.rs
```diff
@@ -1132,6 +1214,123 @@ pub(crate) fn parse_event_name(name: &str) -> (TokenStream, bool, bool) {
...
 
+pub(crate) fn slot_to_tokens(
+    cx: &Ident,
+    node: &NodeElement,
+    slot: &NodeAttribute,
+    parent_slots: Option<&mut HashMap<String, Vec<TokenStream>>>,
+    global_class: Option<&TokenTree>,
+) {
+    let parent_slots =
+        parent_slots.expect("slots can only be used inside components");
```
gbj, line 1225: Could this use `proc_macro_error::emit_error` instead of panicking?

gbj (approved):
Yeah, so this looks really good to me over all. I'm pretty much fine merging as is, but left a couple simple comments. @jquesada2016 what do you think?

leptos_macro/src/lib.rs
jquesada2016: I believe this is 100% correct. The parent component should be responsible for adding events to components.

leptos_macro/src/view.rs
```diff
@@ -715,30 +734,58 @@ fn set_class_attribute_ssr(
...
-        quote! {
-            #node.into_view(#cx)
+    let mut nodes = nodes
+        .iter()
+        .filter_map(|node| {
+            let node = node_to_tokens(
+                cx,
+                node,
+                parent_type,
+                if has_slots { Some(&mut slots) } else { None },
```
jquesada2016, line 758: Can be rewritten to `has_slots.then_sume(&mut slots)`

jquesada2016:
@gbj everything thus far looks good to me! I think we're missing just one edge case, which I did not think of from the RFC. I can add it if we want, or keep it here. But basically it's the same `.into()` thing, but instead of component props, for HTML elements, so:

```rust
fn Comp(cx: Scope, title: HtmlElement<H1>) -> impl IntoView { /* ... */ }

view! { cx,
  <Comp>
    <h1 slot:prop>"hello"</h1>
  </Comp>
}
```

Now, we need to handle the "mixed type" case

```rust
#[component]
fn Comp(cx: Scope, props: Vec<HtmlElement<AnyElement>>) -> impl IntoView {}

view! { cx,
  <Comp>
    <div slot:prop>"mixed"</div>
    <p slot:prop>"types"</p>
  </Comp>
}
```

The `Into` trait might save us here, to be honest, but we should add doctests for these cases just in case.

The last special case is multiple of the same type:

```rust
#[component]
fn Comp(cx: Scope, prop: Vec<HtmlElement<P>>) -> impl IntoView {}

view! { cx,
  <Comp>
    <p slot:prop>"a"</p>
    <p slot:prop>"b"</p>
  </Comp>
}
```

jquesada2016:
I think if we just assume that slot children that start with a lowercase letter are HTML elements, that would probably work very well for custom elements.

gbj:
@jquesada2016 Why don't you add that in a separate PR if separate work needs to be done, just so we can keep things neat. For now, I'll merge this and thank you @yuuma03 for your work!

PR #910, opened 2023-04-20
Add a window event listener with a callback that receives an event that has been cast to its non-generic type.

loganbnielsen:
# Overview

This PR introduces a new function `window_event_listener_with_precast` to remove a common boiler plate case of `window_event_listener`. This solution also enables better compile time checks by enumerating the types of events that can be listened to.

# Discussion

Presently we have a the following definition for `window_event_listener`

```
pub fn window_event_listener(
    event_name: &str,
    cb: impl Fn(web_sys::Event) + 'static,
) {
    cfg_if::cfg_if! {
      if #[cfg(debug_assertions)] {
        let span = ::tracing::Span::current();
        let cb = move |e| {
          leptos_reactive::SpecialNonReactiveZone::enter();
          let _guard = span.enter();
          cb(e);
          leptos_reactive::SpecialNonReactiveZone::exit();
        };
      }
    }

    if !is_server() {
        #[inline(never)]
        fn wel(cb: Box<dyn FnMut(web_sys::Event)>, event_name: &str) {
            let cb = Closure::wrap(cb).into_js_value();
            _ = window().add_event_listener_with_callback(
                event_name,
                cb.unchecked_ref(),
            );
        }

        wel(Box::new(cb), event_name);
    }
}
```

which results in extra boiler plate to process the event in the callback.

For example, consider the following listener:
```
window_event_listener(&"mousemove", my_mouse_callback);
```

Then then the callback will be of the form:
```
let my_mouse_callback = move |event: Event| {
    // function body
}
```
But usually we want access to the specific event object associated with the call. Which in this case is `MouseEvent`

```
let my_mouse_callback = move |event: Event| {
   let event = gen_event.dyn_into::<MouseEvent>().unwrap();
    // function body
}
```

This cast of the Event can be inferred from the `event_name` so there should be an option for it to be already done by the `window_event_listener`. 

This PR handles that case as well as having the benefit of checking that the `event_listener` string is correct by using an `EventDescriptor` object instead for better compile time checks.


gbj:
Looks good! I just made one small change (to replace `.dyn_into().unwrap()` with `.unchecked_into()`) and it should be good to go.

loganbnielsen:
On discord, dgsantana@ was suggesting that maybe this should be a change to window_event_listener itself. Since this would be a breaking change, we'd need to do it for 0.3

Is there a reason not to make this the new default behavior? I'm happy either way

Issue #802, opened 2023-04-04
Fixing scope handling on dynamic children


gbj:
This is basically a "yeah, I hadn't quite thought that all the way through..." issue.

## The Problem

Currently, you can render a dynamic child node like this:
```rust
{move || count()}
```
or this
```rust
{move || view! { cx, <Child/> }}
```

This is all well and good. However, consider the following example
```rust
#[component]
pub fn App(cx: Scope) -> impl IntoView {
    let (count, set_count) = create_signal(cx, 0);
    view! { cx,
        <button on:click=move |_| set_count.update(|n| *n += 1)>"+1"</button>
        {move || (count() % 2 == 1).then(|| view! { cx, <Child/> })}
    }
}

#[component]
pub fn Child(cx: Scope) -> impl IntoView {
    on_cleanup(cx, || leptos::log!("disappearing"));
    view! { cx,
        <p>"Child"</p>
    }
}
```

In this example, `disappearing` never logs, even though the `<Child/>` is added and removed.

## What's Happening?
So what's happening here?

`DynChild` (the internal representation for this kind of thing) does create a child scope, which it uses to call `.into_view()` when building the child content
https://github.com/leptos-rs/leptos/blob/64e056ffa97595fcedb65ff8c1e9afb49fa61bfc/leptos_dom/src/components/dyn_child.rs#L191-L192

And it correctly disposes of that scope
https://github.com/leptos-rs/leptos/blob/64e056ffa97595fcedb65ff8c1e9afb49fa61bfc/leptos_dom/src/components/dyn_child.rs#L200-L201

However, this `Scope` is actually *not* the `Scope` that the `<Child/>` is rendered with.

In particular, this line is a problem.
```rust
{move || (count() % 2 == 1).then(|| view! { cx, <Child/> })}
```
`cx` here is moved in from `<App/>`.
- Passing the `move ||` into the view creates a `DynChild`. When it renders its content, it creates and disposes child scopes descended from `<App/>`. 
- Creating the `<Child/>` component creates a child scope, and renders `<Child/>` in it. But note that in `view! { cx, <Child/>`, the `cx` is the `<App/>` scope.
This means that these two are *siblings*, not parent and child.
```
App
|_ DynChild
  |_ (created child scope)
|_ Child
```

## The Solution
The correct design for this is fairly simple: rendering a dynamic child should require a function that takes a `Scope` as its first argument, like this
```rust
{move |_| count()}
```
or this
```rust
[...]

PR #918, opened 2023-04-21
change: shift from `Scope`-based ownership to reactive ownership

gbj:
# To Do
- [x] Reactive system
- [x] Renderer
- [x] SSR
- [x] Meta
- [x] Router
- [x] Integrations
- [x] Router animations
- [x] Fix OOO streaming
- [x] Zero-prop components
- [x] Make sure server functions continue in the correct runtime if they `.await`
- [x] TODOs
- [ ] Clean up exposed runtime methods
- [x] Update tests
- [x] Test examples and edge cases

***

#802 raises the issue that `on_cleanup` is tied to the current `Scope`, not the current reactive effect. As a result, the cleanups in conditionally-rendered children don't always run:
```rust
#[component]
    let (count, set_count) = create_signal(cx, 0);
    view! { cx,
        <button on:click=move |_| set_count.update(|n| *n += 1)>"+1"</button>
        {move || (count() % 2 == 1).then(|| view! { cx, <Child/> })}
    }
}

#[component]
pub fn Child(cx: Scope) -> impl IntoView {
    on_cleanup(cx, || leptos::log!("disappearing"));
    view! { cx,
        <p>"Child"</p>
    }
}
```

This PR shifts `on_cleanup` such that cleanups are owned by the reactive "owner" node, i.e., whichever effect or memo `on_cleanup` is called within. This correctly solves the problem and allows us to remove manual scope manipulation from `<Show/>` and possibly elsewhere.

However, that raises the question: If I create a signal inside a conditional like this, will it be disposed? The answer is no; a child scope is created for the component, but is not disposed. This suggests that life would be better if signals, resources, and effects were also owned by the observer.

At that point, the place of the nested tree of `Scope`s begins to be taken over by the tree of reactive owners, and `Scope` is left managing only contexts. Which is fine, but why not simplify and merge those into effects as well?

In fact that's not a bad idea, and all of this happens to be how SolidJS organizes itself as well. It's taken me about nine months to understand why, but this is why.

~~Todos
- [x] Move signal/effect/memo/resource ownership over to owners instead of scopes
- [x] Move contexts over to owners instead of scopes
- [x] Fix `<ErrorBoundary/>`: This breaks `<ErrorBoundary/>` by cleaning up the current errors, even if a new error is introduced. Should investigate.
- [x] This will require some extensive testing.~~


gbj:
Update: working on this has demonstrated to me that once you start going, you might as well continue going. There are problems that this will solve that also allow getting rid of manual scope passing over-all. It will take a little more thought, and lots of testing, but once you start relying on the reactive graph itself for ownership all the manual scopes become unnecessary:
```rust
#[component]
pub fn Counter() -> impl IntoView {
    let (value, set_value) = create_signal(0);

    view! {
        <div>
            <button on:click=move |_| set_value.update(|value| *value += 1)>
                {value}
            </button>
        </div>
    }
}
```

duttaoindril:
Excited for this change!

doeixd (approved):

PR #919, opened 2023-04-21
fix: features related compile error

### LINKED ISSUE #916: counter_isomorphic compile error: `hydrate` and `ssr` enabled as features
Run `cargo make test -p counter_isomorphic` from `leptos/examples` to reproduce.

agilarity:
This fixes #916.

`cargo make test` sets the --all-features flag by default. This change clears it.

Run `cargo make test-examples` from the project root.


gbj:
Oh, great! I was worried at first this would turn it off for tests in the library itself, where we do want to tell all features, but I see it's just in the examples Makefile so should be no problem. Thanks for figuring this out.

PR #925, opened 2023-04-22
Added (potentially broken) timer example

bluenote10:
This is an attempt to add a timer example. Currently it doesn't quite work yet, probably related to #802, and it might also be an interesting use case in that context.

In general, wrapping `setInterval` can be an insightful example, because it touches upon the ["impedance mismatch"](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#the-impedance-mismatch) (as Dan Abramov calls it)  between reactive and non-reactive APIs. From a Rust perspective it also addresses a few challenges, like how to capture and move properly into the nested closures.

It is quite possible that I got it wrong, even if `on_cleanup` changes its semantics to effect-based.


gbj:
Here's a version of the `use_interval` hook that works as you're intending. It's quite a bit simpler than the way that doesn't work, I think. The thing to know here is that effects (and memos) take their previous return value as their argument, so if there's something you need to keep alive, returning it from the effect and consuming it on the next iteration works well. 

Conveniently, this also means that if we return the interval handle from the effect call, we always have a cancelable interval handle. Because the effect will only re-run when the interval changes, we always want to cancel a handle if we have one available. (Of course this is `None` on the first run.)

I also used the [`set_interval_with_handle`](https://docs.rs/leptos/0.2.5/leptos/fn.set_interval_with_handle.html) helper that's built into Leptos — we have these for all the timeout/interval/animation-frame/idle-callback functions. All it does is provide the web-sys wrapping you were doing manually, there's no reactive magic happening.

```rust
pub fn use_interval<T, F>(cx: Scope, interval_millis: T, f: F)
where
    F: Fn() + Clone + 'static,
    T: Into<MaybeSignal<u64>> + 'static,
{
    let interval_millis = interval_millis.into();
    create_effect(cx, move |prev_handle: Option<IntervalHandle>| {
        // effects get their previous return value as an argument 
        // each time the effect runs, it will return the interval handle 
        // so if we have a previous one, we cancel it
        if let Some(prev_handle) = prev_handle {
            prev_handle.clear();
        };

        // here, we return the handle
        set_interval_with_handle(
            f.clone(),
            // this is the only reactive access, so this effect will only 
            // re-run when the interval changes
            Duration::from_millis(interval_millis.get()),
        )
        .expect("could not create interval")
    });
}
```

gbj:
Just as a fun side-by-side comparison for anyone else reading —

The React version
```js
function useInterval(callback, delay) {
  const savedCallback = useRef();

  useEffect(() => {
    savedCallback.current = callback;
  });

  useEffect(() => {
    function tick() {
      savedCallback.current();
    }

    let id = setInterval(tick, delay);
    return () => clearInterval(id);
  }, [delay]);
}
```

The Leptos version
```rust
pub fn use_interval<T, F>(cx: Scope, interval_millis: T, f: F)
where
    F: Fn() + Clone + 'static,
    T: Into<MaybeSignal<u64>> + 'static,
{
    let interval_millis = interval_millis.into();
    create_effect(cx, move |prev_handle: Option<IntervalHandle>| {
        if let Some(prev_handle) = prev_handle {
            prev_handle.clear();
        };

        // here, we return the handle
        set_interval_with_handle(f.clone(), Duration::from_millis(interval_millis.get()))
          .expect("could not create interval")
    });
}
```

gbj:
Actually there's a way to do this that's even shorter and more fun! 😄 

`gloo_timers` implements Rust-y RAII wrappers for all the timers and intervals. They are sometimes annoying to use because they return an `Interval` or `Timer` or whatever that clears the timeout or interval on `Drop`. However, in this case it's hilariously good.

Behold—`use_interval`
```rust
use gloo_timers::callback::Interval;

pub fn use_interval<T, F>(cx: Scope, interval_millis: T, f: F)
where
    F: Fn() + Clone + 'static,
    T: Into<MaybeSignal<u32>> + 'static,
{
    let interval_millis = interval_millis.into();
    create_effect(cx, move |_| {
        Interval::new(interval_millis.get(), f.clone())
    });
}
```

bluenote10:
Thanks a lot for the suggestions & help!

Avoiding that "double cleanup" is obviously much nicer. I missed the possibility to make the (Rust) closure itself `Clone`, which is why I though I had to turn it into a JS closure on the outer capture level, because I knew `JsValue` can be cloned. But yeah, since we mostly rely on `'static` move closures anyway, it is natural to simply require `Clone` as well. Is such a double cleanup supposed to work though?

The suggested versions indeed work well! I've switched to the first one suggested, assuming that this is the most idiomatic / helpful to new users?

gbj:
> I missed the possibility to make the (Rust) closure itself `Clone`, which is why I though I had to turn it into a JS closure on the outer capture level, because I knew `JsValue` can be cloned. But yeah, since we mostly rely on `'static` move closures anyway, it is natural to simply require `Clone` as well.

Yeah the nice thing here is that even if the closure itself isn't `Clone` the user can pass `Rc::new(move || { /* some closure that captures a value that isn't Cloneable */})` I'm pretty sure. So it's relatively robust.

> Is such a double cleanup supposed to work though?

I *think* after the ownership for `on_cleanup` moves from the `Scope` to the owning effect (see #918) then the double cleanup version should work. But I'd always recommend the "pass ownership through an effect's arguments" version, when you have an explicit `create_effect` already.

The impedance mismatch thing is interesting. This is actually a case where there's *not* a mismatch between Leptos's execution model and the way things tend to work in Rust. For example, the reason the `gloo-timers` version works so well is that when you call `Timer::new()` it creates a new timer, and implements `Drop` such that it clears it when dropped. This is how most things are designed to work in Rust... And the keep-alive effect of a Leptos efffect just kind of works with that perfectly, where returning something from the effect keeps it alive inside the reactive system, and taking it in as an argument to the effect and then dropping it cleans it up naturally.

Anyway now I'm just pontificating. Merging this PR and thanks for the example!

PR #927, opened 2023-04-22
Test business logic for counter_without_macros

agilarity:
This demonstrates extracting and testing business logic.

Run `cargo make test-all` from the `examples/counter_without_macros` directory

You can also run `cargo make test-examples` from the `leptos/examples` directory.

### DIFF
--- a/examples/counter_without_macros/Cargo.toml
+++ b/examples/counter_without_macros/Cargo.toml
@@ -17,6 +17,7 @@ console_error_panic_hook = "0.1.7"
 wasm-bindgen = "0.2.84"
 wasm-bindgen-test = "0.3.34"
 pretty_assertions = "1.3.0"
+rstest = "0.17.0"
 
 [dev-dependencies.web-sys]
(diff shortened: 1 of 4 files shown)


gbj:
Thanks. I like what you're doing but have two questions
1. What's up with the casting back and forth between `i32` and `u32`?
2. It looks like you create the `Count` struct and its logic but don't actually use it in the component, unless there's a missing commit or something. You should be able to use `Count` to handle the actual logic in the component: create the signal with an initial `Count` value, call `.increase()` and `.decrease()` in the click handlers, etc. I think this would make more sense of the added testing.

agilarity:
Oops!

That's my mistake on both.

I will fix it.

Thank you for the considerate feedback.

agilarity:
I have pushed up the missing use of the `Count`.

Only one cast is to prevent a negative `step` value.

Thank you for catching the problems.

gbj:
Looks good, thanks!

PR #928, opened 2023-04-22
fix: allow nested slots

### LINKED ISSUE #926: Slots are not constructed recursively
Currently, slots are not constructed recursively, so the following won't compile.

```rust
#[slot]
struct AProps { b: BProps }

#[slot]
struct BProps {}

#[component]
fn Comp(cx: Scope, a: AProps) -> impl IntoView {}

fn example(cx: Scope) {
  view! { cx,
    <Comp>
      <A slot>
        <B slot />
      </A>
    </Comp>
  };
}
```

Currently, this will first create a compile error that says only components can have slots, but if we comment out that part of the code which bails out, then the nested slot isn't constructed recursively, but rather is treated like a normal component.

yuuma03:
Closes #926

### DIFF
--- a/leptos_macro/src/view.rs
+++ b/leptos_macro/src/view.rs
@@ -1239,7 +1239,7 @@ pub(crate) fn slot_to_tokens(
     let span = node.name.span();
 
     let Some(parent_slots) = parent_slots else {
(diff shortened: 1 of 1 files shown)


gbj:
Slots in slots in slots! Looks good, thanks.

PR #929, opened 2023-04-22
chore: specify dependency version for cached

rodrigc:
This eliminates cargo warning:

```
dependency (cached) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
```


gbj:
... oops that was me. Thanks.

PR #931, opened 2023-04-22
Add From for RequestParts into Parts for Axum and add an option to ge…

benwis:
Per our earlier discussion, I added a way to exclude routes from .leptos_routes() and implemented From Parts for RequestParts


benwis:
This does introduce a breaking change, requiring the addition of None as an arg to all the examples that use generate_route_list(). I think we're OK with that?

gbj:
Hm it might be worth avoiding the breaking change by adding `generate_route_list_excluding()` and keeping `generate_route_list`, which just calls that and passes `None`.

benwis:
@gbj Not sure I understand why this Test run failed

PR #932, opened 2023-04-22
Link to actual type instead of Into trait for component properties

ModProg:
The link-region could be done differently as well, but the main point is to not link to std's `Into` but the actual type in question.


gbj:
Thanks!

PR #934, opened 2023-04-22
feat: add allow missing_docs to `children` field

ModProg:
I use `#[warn(missing_docs)]` to spot missing docs. But children don't need
documentation in many cases.

An alternative would be to allow passing-through a `#[allow(missing_docs)]` put
on the parameter, which is currently only supported for doc attributes.


gbj:
Yep this makes sense to me.

PR #938, opened 2023-04-23
Adds the GetCbor and GetJson to server macro documentation.

dgsantana:
I was looking into how I could use the Get (for caching) and found out that it was missing documentation for the `server` macro for those encodings. This PR add that documentation. Of course the user must register the leptos_axum::handle_server_fns for `get`.


gbj:
Looks good, thanks!

PR #939, opened 2023-04-23
Fix ServerFnError::Serialization error string

mtsr:
Fix a tiny error, that still sent me in the wrong direction while debugging.


gbj:
Ack oh no! Sorry about that, thanks for fixing it.

Issue #933, opened 2023-04-22
Forcing the return type of `#[component]` to be `impl IntoView` is overly restrictive


ModProg:
As there are concrete types that implement `IntoView` there are situations where you might want to return those directly instead of obfuscating them trough `-> impl IntoView`.

I propose removing the following from the component attribute macro:
```rs
// Make sure return type is correct
if !is_valid_into_view_return_type(&item.sig.output) {
    abort!(
        item.sig,
        "return type is incorrect";
        help = "return signature must be `-> impl IntoView`"
    );
}
```

## Contra
- It makes it harder to spot the mistake when one forgets to add `-> impl IntoView`

## Pro
- makes `#[component]` more vercitile

## Alternative
add a parameter to `#[component(disable_arbitrary_checks_that_are_ensured_by_rustc_as_well)]` to disable the check when needed.

gbj:
The `#[component]` macro adds some debugging and tracing information in debug mode that is genuinely useful — for example, it shows the boundaries between different components in your browser's DOM inspector with named comments:
https://github.com/leptos-rs/leptos/blob/9a4f3ab08c93f735b57e9443ea4635bdbd835f2b/leptos_macro/src/component.rs#L181-L190

It does that by wrapping the component body in a lightweight `Component` type. Because `Component` implements `IntoView` and the type you return must also implement `IntoView`, this means that both the wrapped type and the unwrapped type meet the type definition, namely `-> impl IntoView`

I'm not sure I understand the argument for changing this. The title of the issue says that "forcing the return type to be `-> impl IntoView` is overly restrictive." The alternative would be to make the `#[component]` requirement *less* restrictive by allowing you to return types that are *more* restrictive, do I have that right? So you could say for example
```rust
#[component] 
pub fn MyButton(cx: Scope, children: Children) -> HtmlElement<Button> {
  view! { cx,
    <button class="fancy-button-class>{children(cx)}</button>
}
``` 
Is that the idea?

I can see the advantage of returning accurate types in a situation like this, where you might want to have access to the underlying HTML element for some reason. I'm not really sure what other advantages there would be. Did you have an example in mind where the status quo is causing you an issue?

ModProg:
Well I ran into the issue that I couldn't call `.into_any()` on  a `view!{}` containing only one of my components.
```rs
view! {
    cx,
    <Card kind=CardKind::Filled>
        <Actions>
            <ActionForm action=delete_timer>
[...]

PR #940, opened 2023-04-23
examples: better practice for view types in todos

gbj:
See discussion in #933.

### DIFF
--- a/examples/todo_app_sqlite/src/todo.rs
+++ b/examples/todo_app_sqlite/src/todo.rs
@@ -156,11 +156,11 @@ pub fn Todos(cx: Scope) -> impl IntoView {
                             todos.read(cx)
                                 .map(move |todos| match todos {
                                     Err(e) => {
-                                        vec![view! { cx, <pre class="error">"Server Error: " {e.to_string()}</pre>}.into_any()]
+                                        view! { cx, <pre class="error">"Server Error: " {e.to_string()}</pre>}.into_view(cx)
                                     }
                                     Ok(todos) => {
                                         if todos.is_empty() {
-                                            vec![view! { cx, <p>"No tasks were found."</p> }.into_any()]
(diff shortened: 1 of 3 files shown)


examples/todo_app_sqlite/src/todo.rs
```diff
@@ -175,9 +175,9 @@ pub fn Todos(cx: Scope) -> impl IntoView {
                                                             </ActionForm>
                                                         </li>
                                                     }
-                                                    .into_any()
                                                 })
                                                 .collect::<Vec<_>>()
+                                                .into_view(cx)
```
ModProg, line 180: Should it be possible to `.collect()` directly into the view here?

ModProg: my bad, `into_view` needs a Scope, and `.collect()` cannot provide that.

gbj: All good! Thanks for looking at it :-)

ModProg: Scratch that, while `Fragment::into_view` takes a `Scope` it is not used in the whole progress. So it should be possible to implement `FromIterator<View> for View`.
A naive implementation would just copy the code from `Fragment::into_view()` and `ComponentRepr::into_view()`:

```rs
impl FromIterator<View> for View {
    fn from_iter<T: IntoIterator<Item = View>>(iter: T) -> Self {
        let fragment = iter.into_iter().collect::<Fragment>();
        let mut frag = ComponentRepr::new_with_id("", fragment.id().clone());

        #[cfg(debug_assertions)]
        {
            frag.view_marker = fragment.view_marker;
        }

        frag.children = fragment.nodes;
        #[cfg(all(target_arch = "wasm32", feature = "web"))]
        if !HydrationCtx::is_hydrating() {
            for child in &self.children {
                mount_child(MountKind::Before(&self.closing.node), child);
            }
        }

        View::Component(frag)
    }
}

```

PR #954, opened 2023-04-25
Add wasm web test and common tasks

agilarity:
This configures the `counter`, `counter_without_macros`, and `counters` examples to use a set of reusable tasks to  support testing and verification.

**Verification:**

**From the root directory:**
- Run `cargo make test-examples`
- Run `cargo make verify-examples` (Currently fails checks for some examples but the task works)

**From the example directory:**
- Run `cargo make test-local`
- Run `cargo make web-test`
- Run `cargo make check-style` (`counter` and `counters` currently fail checks but the task works) 
- Run `cargo make verify-local` (`counter` and `counters` currently fail checks but the task works) 

*This PR will make it easier to work through the existing style issues.*


gbj:
Thanks!  Run the tests locally when I have time. Our CI issues have resolved with the most recent commit I made, I think, but just to double check.

agilarity:
That sounds good. Thanks for the update @gbj.

PR #956, opened 2023-04-26
feat: add collect_view(cx)

### LINKED ISSUE #953: Create an `Iteator<IntoView>::collect_view method to collect a view from an iterator
As we cannot implement `IntoView for impl IntoIterator<Item=impl IntoView>` due to trait constraints, I thought we could consider adding a `collect_view` method for everything that implements `IntoView`:
```rs
/// Collects an iterator or collection into a [`View`].
pub trait CollectView {
    /// Collects an iterator or collection into a [`View`].
    fn collect_view(self, cx: Scope) -> View;
}

impl<I: IntoIterator<Item = T>, T: IntoView> CollectView for I {
    #[cfg_attr(
        any(debug_assertions, feature = "ssr"),
        instrument(level = "info", name = "#text", skip_all)
    )]
    fn collect_view(self, cx: Scope) -> View {
        self.into_iter()
            .map(|v| v.into_view(cx))
            .collect::<Fragment>()
            .into_view(cx)
    }
}
```
this would allow calling collect_view(cx) on e.g. `Vec<impl IntoView>` but also a `.map(|_| view!{})` etc.

_Originally posted by @ModProg in https://github.com/leptos-rs/leptos/issues/941#issuecomment-1520428848_

ModProg:
I updated all applicable examples (maybe too aggressively).

fixes #953

### DIFF
--- a/docs/book/src/view/04_iteration.md
+++ b/docs/book/src/view/04_iteration.md
@@ -31,6 +31,22 @@ view! { cx,
 }
 ```
 
+Leptos also provides a `.collect_view(cx)` helper function that allows you to collect any iterator of `T: IntoView` into `Vec<View>`.
+
+```rust
+let values = vec![0, 1, 2];
+view! { cx,
+    // this will just render "012"
+    <p>{values.clone()}</p>
+    // or we can wrap them in <li>
+    <ul>
+        {values.into_iter()
+            .map(|n| view! { cx, <li>{n}</li>})
+            .collect_view(cx)}
+    </ul>
+}
+```
+
 The fact that the _list_ is static doesn’t mean the interface needs to be static.
 You can render dynamic items as part of a static list.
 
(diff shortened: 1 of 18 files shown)


docs/book/src/view/04_iteration.md
```diff
@@ -26,7 +26,7 @@ view! { cx,
     <ul>
         {values.into_iter()
             .map(|n| view! { cx, <li>{n}</li>})
-            .collect::<Vec<_>>()}
+            .collect_view(cx)}
```
gbj, line 29: In this specific case, since the context is talking about collect into a `Vec<_>`, it might be better to let it stand, and add a paragraph like

> Leptos also provides a `.collect_view(cx)` helper function that allows you to collect any iterator of views into `Vec<View>`.

And then include the modified example.

examples/counters/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 use leptos::{For, ForProps, *};
 
-const MANY_COUNTERS: usize = 1000;
+const MANY_COUNTERS: usize = 40000;
```
gbj, line 3: Is this related?

PR #958, opened 2023-04-26
fix: escapes in doc comments on component properties

ModProg:
The current implementation mangled doc comments containing escapes.


gbj:
Looks good to me, thanks!

PR #966, opened 2023-04-28
fix: use a HashMap on Branches::initialize

yuuma03:
Using a Vec<Branch> was breaking hosting multiple SPAs (on different paths) inside the same server.
I'm not sure if not allowing that is intended behaviour or not, but this patch seems to make it work.


gbj:
It makes sense that the new caching would cause that bug in those circumstances — thanks for fixing it.

Issue #834, opened 2023-04-08
Detached Nodes memory leak


novacrazy:
If you run an example with dynamically-present elements (such as from `<Show>` and `<For>`), like the Hackernews example, after switching pages a few times go into the Chrome dev-tools Memory tab and take a heap snapshot. You will see many thousands of detached nodes.
![detached2](https://user-images.githubusercontent.com/4708645/230742962-e829a7db-3135-4310-a031-1f18250f6a6c.PNG)

This is caused by the nodes being kept in the `heap` JS array variable used by wasm_bindgen to hold `JsValue`s, and not being freed. The WASM program refers to these heap values by index, and upon Drop of a `JsValue` in Rust code the value on the heap is replaced with an integer pointing to the previous free slot, freeing whatever other reference was there. Cloning a `JsValue` just makes a copy of the reference in `heap` at another index. It's pretty air-tight on the wasm_bindgen side despite JS references being weird.

Therefore, it seems somewhere in Leptos's code the `JsValue` is being retained. Unfortunately, tracking this down is difficult, as the devtools always point to the `heap` variable, not Rust code. For some of the retainer graphs of these detached nodes, I have seen them point to multiple indexes in `heap`, implying the `JsValue` has been cloned via `Clone` and all are being retained. 

Note that pretty much everything in `js_sys` and `web_sys` are built on `JsValue`, including `HtmlElement`. If it refers to something living in JS-land, it uses the `JsValue` system.

novacrazy:
It's definitely the `<For>` component, at least. `<Show>` by itself seems to not have any issues at the time of testing today, knock on wood.

Modifying the `counter` example to be:
```
<div>
    <div>
        <button on:click=move |_| set_value(0)>"Clear"</button>
        <button on:click=move |_| set_value.update(|value| *value -= step)>"-1"</button>
        <span>"Value: " {value} "!"</span>
        <button on:click=move |_| set_value.update(|value| *value += step)>"+1"</button>
    </div>
    <Show when={move || value() % 2 == 0} fallback=|_| ()>
        <For each={|| vec![1, 2, 3]} key=|key| *key view={move |cx, k| {
            view! {
                cx,
                <article>{k}</article>
            }
        }}/>
    </Show>
</div>
```
then clicking the increment button N times results in N*3/2 leaked `<article>` nodes, so a leak every time it becomes visible.

It also leaks N/2 number of `Comment` nodes.

gbj:
[...]

Issue #398, opened 2023-01-28
Suspense should cache children to avoid rerendering them if you fall back a second time


gbj:

gbj:
It turns out this was actually not a good idea, in part because it would prevent us from ever running cleanups for old `<Suspense/>` children! So, closing.

gbj:
I had closed this because I tried it several times and it seemed to be impossible to achieve. However it turns out that it was closely linked to the memory leak described [here](https://github.com/leptos-rs/leptos/issues/834#issuecomment-1524068660).

Implementing this requires implementing `DynChild` in such a way that a `DynChild` can be unmounted and remounted. It took me a few hours of work but I figured out that it could be implemented (PR coming soon.) 

So I'll reopen this for now and test if it addresses the rest of #834 as well or not.

gbj:
See commentary on #967 and #1006. I was eventually able to get this mostly working except for one specific case (the first refresh after hydration). The fact that DOM nodes move when shallow-cloned and inserted somewhere else in the DOM just makes it *really* hard to work with this sort of cloning correctly, and after working on it way too long for its benefit, I'm giving up again...

gbj:
Now that I've removed the nested fragments from `<Suspense/>`, this should be possible to do more easily with the renderer. It would also solve some unfortunate behavior in cases like this, in which a resource is created and read inside the children of the `<Suspense/>`, which means it loops continuously without ever resolving. (Because every time it resolves, it renders the children and then immediately falls back again.)

```rust
#[component]
fn Inner(cx: Scope, use_inner_suspense: bool) -> impl IntoView {
    let resource = create_local_resource(
        cx,
        || (),
        move |_| async move {
            gloo_timers::future::sleep(Duration::from_secs(2)).await;
            "Done"
        },
    );
    let text = move || resource.read(cx).unwrap_or("Resource loading");

    if use_inner_suspense {
        (view! { cx,
            <Suspense fallback=|| "Suspense Inner...">
                <p>{text}</p>
            </Suspense>
        })
        .into_view(cx)
    } else {
        (view! { cx,
            <p>{text}</p>
        })
        .into_view(cx)
    }
}

#[component]
fn Outer(cx: Scope, use_inner_suspense: bool) -> impl IntoView {
    view! { cx,
        <Suspense fallback=|| "Suspense Outer...">
            <Inner use_inner_suspense=use_inner_suspense />
        </Suspense>
    }
}
```

PR #967, opened 2023-04-28
fix: `<Suspense/>`-related leaks

gbj:
This finally solves issues that were blocking #398, namely making sure that `DynChild` moves its children back into its document fragment when it is being unmounted (from itself!)

This seems to solve the [third](https://github.com/leptos-rs/leptos/issues/834#issuecomment-1524068660) described in #834. I am also no longer able to reproduce the [second](https://github.com/leptos-rs/leptos/issues/834#issuecomment-1524068660) memory leak, for what I think is the same reason. I do still see the [initial](https://github.com/leptos-rs/leptos/issues/834#issue-1659654336) memory leak reported in HackerNews, so more investigation is needed.

I've also broken hydration in this version, so will need to go back in and fix that. 

To do
- [ ] fix hydration
- [ ] investigate HackerNews leak and determine whether it's `<For/>` causing a similar issue

### DIFF
--- a/examples/counter/Cargo.toml
+++ b/examples/counter/Cargo.toml
@@ -12,6 +12,7 @@ leptos = { path = "../../leptos" }
 console_log = "1"
 log = "0.4"
 console_error_panic_hook = "0.1.7"
+gloo-timers = { version = "0.2.6", features = ["futures"] }
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


gbj:
Closing this in favor of #1006, which solved the same problems I'd been working on for two weeks in about ten minutes.

PR #971, opened 2023-04-28
Fix the lint issues in the counter example

agilarity:
This cleans up the lint issues in the `counter` example.

Verification:

- Run `cargo make check-style` from the `examples/counter` directory
- Run `cargo make verify-examples` from the `leptos` root directory (`counter` passes but other examples fail)


gbj:
Thanks!

PR #973, opened 2023-04-29
Replace `serde_urlencoded` with `serde_html_form` to support cross-input vec serverfn

snapbug:
`serde_html_form` is a fork of `serde_urlencoded` that supports additional features that won't be supported by `urlencoded`.

The motivating example I have is an `ActionForm` with `create_server_action` where I want multiple fields to be combined into a single `Vec` argument to the `serverfn`, allowing for varying numbers of arguments. (Think: a todo list that you can select multiple items to mark done/delete in one action).

Typically this would be done with non-standard (from what I can tell) additions/encoding in www-form-urlencoded, `"field[]=1&field[]=2"` or `"field=1&field=2"`. The non-standard-ness is why it's not going to be added to urlencoded (https://github.com/nox/serde_urlencoded/issues/6 / https://github.com/nox/serde_urlencoded/issues/75).

The fork has already been adopted by axum for the axum-extras package https://github.com/tokio-rs/axum/issues/434 .

---

I have a hacked up [version of the todo-app-axum](https://gist.github.com/snapbug/d694398098697035a8395519f25127e8) as illustration that it mostly works correctly (there's still the an issue when non are selected, but that's work-aroundable with active state toggling the submit button at worst), and that form-encoding is needed even with a `"Cbor"` serverfn (makes total sense for an action form triggered server fn.

### DIFF
--- a/leptos_macro/src/lib.rs
+++ b/leptos_macro/src/lib.rs
@@ -816,7 +816,7 @@ pub fn slot(args: proc_macro::TokenStream, s: TokenStream) -> TokenStream {
 /// - **Arguments must be implement [`Serialize`](https://docs.rs/serde/latest/serde/trait.Serialize.html)
 ///   and [`DeserializeOwned`](https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html).**
 ///   They are serialized as an `application/x-www-form-urlencoded`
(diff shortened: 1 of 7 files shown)


gbj:
I'm fine with this, and glad to see it's in `axum-extras` as hopefully that will ensure it's maintained long term. I think over time we're probably interested in extending the range of the URL encoding support anyway (to support things like nested structs) so, this at least is a good start (and a question I've gotten from people before!) Thanks.

PR #980, opened 2023-05-02
Fix lint issues in counter_isomorphic example

agilarity:
This cleans up the lint issues in the `counter_isomorphic` example.

Verification:

- Run cargo make check-style from the examples/counter_isomorphic directory
- Run cargo make verify-examples from the leptos root directory (counter_isomorphic passes but other examples fail)
- Run `cargo leptos watch` from the leptos root directory and test manually


agilarity:
Excellent!

Thank you.

PR #987, opened 2023-05-03
feat: allow structs in server function arguments

gbj:
`serde_urlencoded` doesn't accept nested structs. (Makes sense: URL encoding is flat.) But server function arguments are already built into a struct, so they can be serialized. (Makes sense: otherwise how would you send named arguments?) But this means you can't send a struct as a server function argument: Sad panda.

~~I don't know if this will actually work so simply but there is a `#[serde(flatten)]` attribute that should allow this to be possible.~~

`#[serde(flatten)]` doesn't help without additional proc-macro work (because it can only be applied to structs, not primitives) but using `serde_qs` will solve the problem.


gbj:
@snapbug FYI This will supersede `serde_html_form` — It should cover both the `Vec<_>` args case that `serde_html_form` covered and the issue of passing structs as arguments (which caused nested-struct issues in `serde_urlencoded` and `serde_html_form`), so I think it's a net gain.

PR #1009, opened 2023-05-05
feat: add `style:` to view

### LINKED ISSUE #942: View macro should support style: syntax
Might be missing an actual feature as I've only just picked up leptos but the intuitive approach doesn't work. I didn't see anything in the book, examples or other issues that mention this.

What I thought might work:
```rust
view! { cx, <div style:--columns=5></div> }
```
```css
div { 
  display: grid;
  grid-template-columns: repeat(var(--columns), 120px);
}
```
Above is just an example of using [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) in a way I might expect to be able to in leptos.

A workaround is to assign the whole style like `style="--columns:5"` but it would be nice if the view macro supported this.

gbj:
Closes #942.

To do
- [x] docs in `view` macro
- [x] docs in guide
- [x] test edge cases

### DIFF
--- a/leptos_dom/src/html.rs
+++ b/leptos_dom/src/html.rs
@@ -865,9 +865,9 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
                     match value {
                         Style::Value(value) => Some(value),
                         Style::Option(value) => value,
-                        _ => unreachable!()
+                        _ => unreachable!(),
                     }
-                },
+                }
             };
 
             if let Some(style_value) = include {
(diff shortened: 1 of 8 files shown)


leptos_dom/src/html.rs
```diff
@@ -824,6 +824,69 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
         }
     }
 
+    /// Sets a style on an element.
+    ///
+    /// **Note**: In the builder syntax, this will be overwritten by the `style`
+    /// attribute if you use `.attr("class", /* */)`. In the `view` macro, they
```
FrankReh, line 830: Looks like a copy-paste mistake. "class" should be "style" maybe?

leptos_dom/src/macro_helpers/into_style.rs
```diff
@@ -0,0 +1,209 @@
...
+style_type!(u128);
+style_type!(isize);
+style_type!(i8);
+style_type!(i16);
+style_type!(i32);
+style_type!(i64);
+style_type!(i128);
+style_type!(f32);
+style_type!(f64);
+style_type!(char);
```
FrankReh, line 209: I don't know anything yet. It's not appropriate to handle a bool style_type also?

PR #1011, opened 2023-05-06
docs: adjust Dynamic Attributes page

FrankReh:
Adjust the intro in the Dynamic Attributes page to include the recent `dynamic style` feature. Also reorder a little given the order of the page body that follows.


gbj:
Thanks!

PR #1013, opened 2023-05-06
book: close iframe tag on global state management page

Wadu436:
The MDBook UI doesn't work on the Global State Management page:
https://leptos-rs.github.io/leptos/15_global_state.html

This turned out to be because the iframe tag for the codesandbox embed wasn't closed.

This fixes it.


gbj:
Thanks!

PR #1015, opened 2023-05-06
docs: add examples for manual server integration

ChristopherPerry6060:
This could be a bit more verbose for people wanting to write their own integration with the router.

I copied the examples from how `ServerIntegration` is used [here](https://github.com/leptos-rs/leptos/blob/72414b7945dee81c37833b44b54285a36393dfc8/router/src/extract_routes.rs#LL61C51-L61C51).

probably should have opened an issue first before opening this, but alas. Likely an obvious thing for someone familiar with web, but more docs on finer details can be appreciated by newer people.

### DIFF
--- a/router/src/history/mod.rs
+++ b/router/src/history/mod.rs
@@ -135,14 +135,14 @@ impl History for BrowserIntegration {
 
 /// The wrapper type that the [Router](crate::Router) uses to interact with a [History].
 /// This is automatically provided in the browser. For the server, it should be provided
-/// as a context.
+/// as a context. Be sure that it can survive conversion to a URL in the browser.
 ///
 /// ```
 /// # use leptos_router::*;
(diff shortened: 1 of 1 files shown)


router/src/history/mod.rs
```diff
@@ -167,7 +167,24 @@ impl History for RouterIntegrationContext {
     }
 }
 
-/// A generic router integration for the server side. All its need is the current path.
+/// A generic router integration for the server side.
+///
+/// This should match what the browser history will show.
+///
+/// Generally, this will already be provided if your are using the leptos
```
FrankReh, line 174: 'if you'

gbj (commented):
Thanks, this looks great! The need for a full URL to ensure consistency between the browser and the url crate is definitely one of those small gotchas.

PR #1016, opened 2023-05-07
build tailwind CSS with trunk

agilarity:
**Proposed**

This demonstrates how the tailwind build process can be automated with a Trunk hook.

The hook will run whenever a build is triggered by`trunk build` or `trunk serve`.


gbj:
Nice!

PR #1026, opened 2023-05-09
feat: `SsrMode::PartiallyBlocked`

gbj:
Using `create_blocking_resource` with `SsrMode::OutOfOrder` waits for the blocked fragments to resolve before returning any response, which means that they can affect headers and the `<head>`. However, it still uses JavaScript to replace the fallbacks with the rendered children in the browser. 

It can be desirable in some circumstances — for example, if there's a good chance JavaScript is not available — to replace the fallback with the children on the server instead. This makes for a marginally slower initial response, as additional string-replacement work has to be done on the server at an O(n) cost (n = size of the first HTML chunk). However, it potentially makes for a better experience if JS is not available at all.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -16,10 +16,9 @@ use actix_web::{
 use futures::{Stream, StreamExt};
 use http::StatusCode;
 use leptos::{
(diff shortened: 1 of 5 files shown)


gbj:
Also closes #1029.

PR #1036, opened 2023-05-12
Update slice.rs

sjud:
Split create_slice into read and write, then combine them in create slice, changing the generic parameters of create_slice so that Signal and SignalSetter return different values.


gbj:
Thanks! The failing test is just related to formatting -- if you run `cargo fmt` on it it should be fine. Otherwise, this looks good to me.

gbj:
Looks good, thanks.

PR #1040, opened 2023-05-13
examples: fix trunk config to run tailwind at the right time

flosse:
The current example only works if there is already a created CSS file but on the first run it fails.
This patch should fix is.


gbj:
Thanks!

PR #1043, opened 2023-05-15
fix: typo in actix extract documentation

markcatley:
Got excited here for a minute

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1028,7 +1028,7 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


gbj:
Whoops! Sorry about that. Axum extractors are rather harder to work with in our model unfortunately... (Specifically because Axum requests are not `Clone`, but you need `&mut` to run an extractor and they return a `Future`, so even if you store the request in a lock you immediately run into the issue of holding a lock across an `.await`, which means if we provided something similar for Axum you'd immediately get race conditions between your different server functions if more than one tried to use an extractor.)

PR #1048, opened 2023-05-16
fix: multipart forms on server fns

yuuma03:
As a consequence, any extractor that needs to consume the Payload (like MultipartForm) should now work on server fns.

### DIFF
--- a/integrations/actix/Cargo.toml
+++ b/integrations/actix/Cargo.toml
@@ -8,6 +8,7 @@ repository = "https://github.com/leptos-rs/leptos"
 description = "Actix integrations for the Leptos web framework."
 
 [dependencies]
+actix-http = "3"
 actix-web = "4"
 futures = "0.3"
 leptos = { workspace = true, features = ["ssr"] }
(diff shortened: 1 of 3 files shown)


gbj:
I like this idea but am wondering whether it's possible to check whether there is in fact multipart form data before cloning the body.

I looked at the implementation of `actix_web::Bytes` and honestly it's mostly over my head, but I think I'm correct in saying that cloning `body` here will be a bytewise clone, i.e., it really does copy the data. This is unnecessary except in the case in which you're actually trying to consume the body in the server function.

I haven't done much with multipart forms — Do they set a header with the `enctype` or anything that we can check, to avoid cloning the data unnecessarily for regular POSTs?

(BTW the broken CI is something unrelated that is affecting the CI on the main branch but is not reproducible on my computer. One of these weird things, I'll try to figure it out.)

yuuma03:
For that case, Content-Type should be `multipart/<something>; boundary=<boundary>` (normally the <something> is going to be `form-data`).
I'm gonna give a try at implementing that, and I'll commit in a bit.

yuuma03:
It seems to not have broken anything on the website I'm making, and also it should be only only cloning the payload for multipart forms now.

gbj:
Great, this looks good to me!

Issue #833, opened 2023-04-08
Allow unquoted text in view macro


wxyer:
i think the feature is very convenient，is this nightly rust give or the framework give

gbj:
Here’s a tracking issue for unquoted text in [`syn-rsx`](https://github.com/stoically/syn-rsx/issues/2), which is the underlying library we use for parsing the `view` macro.

wxyer:
thank you ,i expect it very much

brahimABD98:
quoted text is can have benefits  writing stuff in  Arabic is easier no  extra work is needed , maybe if it's possible to keep both

gbj:
This was added in #1054. Quoted text is still supported as usual, but unquoted text is also supported. Extra whitespace is trimmed, so you may find that you still need quoted text in situations where you need to ensure whitespace is included, and you can still use quoted text in any case.

This is available on both nightly and stable.

Issue #766, opened 2023-03-29
feat: Allow for a more concise syntax when passing views as props


elliotwaite:
Instead of having to write this:
```
<Routes>
  <Route path="/" view=|cx| view! { cx, <Home/> }/>
  <Route path="/other" view=|cx| view! { cx, <Other>"children"</Other> }/>
  <Route path="/*any" view=|cx| view! { cx, "just text" }/>
</Routes>
```
Allow for something like this:
```
<Routes>
  <Route path="/" view=<Home/>/>
  <Route path="/other" view=<Other>"children"</Other>/>
  <Route path="/*any" view=<>"just text"</>/>
</Routes>
```
Would it be possible for the `view!` macro to support this type of syntax?

bram209:
Would be nice if this was possible (didn't test code)

```rust
#[component]
fn ResourceOk<R, RV, IV, Props>(
    cx: Scope,
    resource: ReadSignal<Result<R, String>>,
    resource_view: RV,
) -> impl IntoView
where
    R: 'static,
    RV: FnOnce(Scope, Props) -> IV + 'static,
    Props: From<R>,
    IV: IntoView,
{
    todo!()
}
```

So you would be able to call it by:

```rust

struct User;

#[component]
fn UserProfile(cx: Scope, user: User) -> impl IntoView { }

#[component]
fn SomeComponent(cx: Scope) -> impl IntoView {
    let (name, set_name) = create_signal(cx, Ok(User {}));

    view! { cx,
        <ResourceOk resource=name resource_view=UserProfile  />
    }
```

bram209:
I also wonder if the view! macro could return a closure that takes the scope as an argument. This way you don't always have to explicitly pass the scope as the first argument

bram209:
> Another area this might apply to is when using views in code that's already inside a !view macro. 

This is not really feasible, considering that the `view!` macro parses rust code with the [`syn`](https://github.com/dtolnay/syn) crate, so it will ask syn to parse a full expression. `syn` is not aware that it is being used within a leptos context. So unless you fork `syn` and add a new variant to [`syn::Expr`](https://docs.rs/syn/latest/syn/enum.Expr.html) or write a full rust source code parser, I don't see how this can easily be implemented.

elliotwaite:
@bram209, ah, I see. Thanks for the explanation.

[...]

PR #1054, opened 2023-05-17
Rsx parser with recovery after errors, and unquoted text

vldm:
Hi.
Original author of syn-rsx is absent for a while.
I tried to reach him through the e-mail but have no luck with that.

I have created a fork of https://github.com/rs-tml/rstml 
With aim to better IDE support.
Fork contain few refactors which breaks the API, but during refactor all types became to be self-describing (every type contain all tokens that was parsed), which, i hope, will reduce future breaking changes.

In this PR i move from syn-rsx to rstml fork.
Which also bump minimum syn version to 2.0 (it was already in dependencies before)

## In summary, this pr introduce or changes :
1. Add support of unquoted text support.
2. Introduce recovery of parser after invalid closing tags, unclosed tags or invalid expression ( can be checked by semantic syntax highlight)
3. Simplified Node::Text ( no expressions, or other literals except string)
4. Attribute refactoring - not a node anymore, seperated Block and key=value attributes.
5. As part of 2. `NodeBlock::Invalid` was introduced, which represent block with invalid `syn::Block` syntax.

## To discuss:
- Component macro refactor.
 Currently if inside fn body any invalid expression was found, then macro will fail to expand. 
 To fix this, we can omit parsing of fn body.
 
- `#[allow(unused_braces)]`, `vec![]` and other macro usage inside expansion.
I have removed this macro usage, because with them `rust-analyzer` failed to give completion for invalid expression inside braces `<div> {v.<COMPLETION>} </div>`. Probably i can return it for `NodeBlock::Valid`, and keep it removed only for `InvalidBlock`.

### DIFF
--- a/leptos_macro/src/template.rs
+++ b/leptos_macro/src/template.rs
@@ -1,9 +1,9 @@
 use crate::attribute_value;
 use leptos_hot_reload::parsing::is_component_node;
 use proc_macro2::{Ident, Span, TokenStream};
-use quote::{quote, quote_spanned};
-use syn::spanned::Spanned;
-use syn_rsx::{Node, NodeAttribute, NodeElement, NodeValueExpr};
+use quote::{quote, quote_spanned, ToTokens};
+use syn::{spanned::Spanned, ExprBlock};
+use syn_rsx::{Node, NodeAttribute, NodeElement, KeyedAttribute, NodeBlock};
 use uuid::Uuid;
 
 pub(crate) fn render_template(cx: &Ident, nodes: &[Node]) -> TokenStream {
@@ -53,7 +53,7 @@ fn root_element_to_tokens(
                 .unwrap();
         };
 
-        let span = node.name.span();
+        let span = node.name().span();
 
         let navigations = if navigations.is_empty() {
             quote! {}
@@ -67,7 +67,7 @@ fn root_element_to_tokens(
             quote! { #(#expressions;);* }
         };
 
-        let tag_name = node.name.to_string();
+        let tag_name = node.name().to_string();
 
         quote_spanned! {
             span => {
@@ -104,9 +104,9 @@ enum PrevSibChange {
     Skip,
 }
 
-fn attributes(node: &NodeElement) -> impl Iterator<Item = &NodeAttribute> {
-    node.attributes.iter().filter_map(|node| {
-        if let Node::Attribute(attribute) = node {
+fn attributes(node: &NodeElement) -> impl Iterator<Item = &KeyedAttribute> {
+    node.attributes().iter().filter_map(|node| {
+        if let NodeAttribute::Attribute(attribute) = node {
             Some(attribute)
         } else {
             None
@@ -129,11 +129,11 @@ fn element_to_tokens(
 ) -> Ident {
     // create this element
     *next_el_id += 1;
-    let this_el_ident = child_ident(*next_el_id, node.name.span());
+    let this_el_ident = child_ident(*next_el_id, node.name().span());
 
     // Open tag
-    let name_str = node.name.to_string();
-    let span = node.name.span();
+    let name_str = node.name().to_string();
+    let span = node.name().span();
 
     // CSR/hydrate, push to template
     template.push('<');
@@ -145,7 +145,7 @@ fn element_to_tokens(
     }
 
     // navigation for this el
-    let debug_name = node.name.to_string();
+    let debug_name = node.name().to_string();
     let this_nav = if is_root_el {
         quote_spanned! {
             span => let #this_el_ident = #debug_name;
@@ -247,14 +247,14 @@ fn next_sibling_node(
                 if is_component_node(sibling) {
                     next_sibling_node(children, idx + 1, next_el_id)
                 } else {
-                    Ok(Some(child_ident(*next_el_id + 1, sibling.name.span())))
+                    Ok(Some(child_ident(*next_el_id + 1, sibling.name().span())))
                 }
             }
             Node::Block(sibling) => {
-                Ok(Some(child_ident(*next_el_id + 1, sibling.value.span())))
+                Ok(Some(child_ident(*next_el_id + 1, sibling.span())))
             }
             Node::Text(sibling) => {
-                Ok(Some(child_ident(*next_el_id + 1, sibling.value.span())))
+                Ok(Some(child_ident(*next_el_id + 1, sibling.span())))
             }
             _ => Err("expected either an element or a block".to_string()),
         }
@@ -263,7 +263,7 @@ fn next_sibling_node(
 
 fn attr_to_tokens(
     cx: &Ident,
-    node: &NodeAttribute,
+    node: &KeyedAttribute,
     el_id: &Ident,
     template: &mut String,
     expressions: &mut Vec<TokenStream>,
@@ -272,8 +272,8 @@ fn attr_to_tokens(
     let name = name.strip_prefix('_').unwrap_or(&name);
     let name = name.strip_prefix("attr:").unwrap_or(name);
 
-    let value = match &node.value {
-        Some(expr) => match expr.as_ref() {
+    let value = match &node.value() {
+        Some(expr) => match expr {
             syn::Expr::Lit(expr_lit) => {
                 if let syn::Lit::Str(s) = &expr_lit.lit {
                     AttributeValue::Static(s.value())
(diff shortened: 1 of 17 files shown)


leptos_macro/src/template.rs
```diff
@@ -387,9 +387,13 @@ fn child_to_tokens(
                 ))
             }
         }
+        //TODO Should we ewerywhere use syn::expr?
```
gbj, line 390: Could you say a little more about what you mean here?

leptos_macro/src/view.rs
```diff
@@ -484,42 +489,35 @@ fn element_to_tokens_ssr(
...
-                                    chunks.push(SsrElementChunks::String {
-                                        template: std::mem::take(template),
-                                        holes: std::mem::take(holes),
-                                    })
-                                }
-                                chunks.push(SsrElementChunks::View(quote! {
-                                  {#value}.into_view(#cx)
-                                }));
+                            // TODO: Is it was string? Original syn-rsx parse block with braces,
+                            // but value_to_string expect it without braces.
```
gbj, line 506: If I understand you correctly: `{"Text"}` should be handled as `SsrElementChunks::String` here, I think this was just a mistake in my origin `value_to_string`, which you've corrected.

leptos_macro/src/view.rs
```diff
@@ -1102,30 +1095,29 @@ fn element_to_tokens(
...
-                        (
-                            quote! {
-                                #[allow(unused_braces)] #value
-                            },
-                            false,
-                        )
-                    }
+                    (quote! { #node }, true)
+                }
+                // TODO: Implement html  escaping?
```
gbj, line 1100: I don't believe HTML escaping needs to be added here or on `Node::Text`. These static string types are escaped by the SSR optimization code for SSR, and don't need to be escaped in the browser because they're created using `document.createTextNode()` not an HTML element. I may be wrong.

leptos_macro/src/view.rs
```diff
@@ -1102,30 +1095,29 @@ fn element_to_tokens(
...
+                    let text = node.to_string_best();
+                    let text = syn::LitStr::new(&text, node.span());
+                    (quote! { #text }, true)
                 }
                 Node::Block(node) => {
-                    if let Some(primitive) = value_to_string(&node.value) {
-                        (quote! { #primitive }, true)
-                    } else {
-                        let value = node.value.as_ref();
+                    // TODO: Is there any static string possible in block?
```
gbj, line 1107: Removing the commented sections should be fine.

leptos_macro/src/view.rs
```diff
@@ -1313,9 +1305,9 @@ fn attribute_to_tokens(
         };
 
         // all other attributes
-        let value = match node.value.as_ref() {
+        let value = match node.value() {
             Some(value) => {
-                let value = value.as_ref();
+                let value = value;
```
gbj, line 1310: Is this redundant now?

gbj (commented):
This is very exciting! I have tested some examples locally and the unquoted text appears to work well on both stable and nightly, although of course for edge cases whitespace-trimming means that quoted strings are sometimes useful. I really appreciate the thought you've put into this.

The CI issues are 
1) `cargo fmt`
2) some unrelated issues I've fixed on main that should be fixed if you rebase

If you want to respond to my comments on some of the todos, rebase, and run `cargo fmt`, I'll be very happy to merge this.

Thanks for all your work on it.

This would close #833. I think `rstml` also provides an opportunity to do #766.

leptos_macro/src/template.rs
vldm: I am just wasn't sure what block_to_tokens is doing. And what it expect for different `syn::Expr`.

There also was `str_value match` that looks like `value_to_string` to string body.

So my thought about `block_to_tokens` was:
1. should it reuse `value_to_string`?
2. if it tries to handle primitive literals differently maybe this handling should be done on higher lever (split this function into two (text_to_tokens/dyn_block_to_tokens), and use them in place where needed?

vldm:
I have fixed fmt, regenerate test stderr for macro ui.

Also for `value_to_string`, added conversion for primitive `Node::Block` - now it should explicitly convert
`{"string literal"}` to `"string literal"` and use it `view::SsrElementChunks::String` or in `block_to_tokens`.



I'm not sure if this pr can move forward on https://github.com/leptos-rs/leptos/issues/766 - because the syntax that was declared in the first post can significantly complicate attribute parsing logic. I also left comment in #766 , with imaginary syntax that could be a bit easier to implement (at least on parsing level), but also solve same problem.

I also could mention, that my fork is also closes https://github.com/stoically/syn-rsx/issues/51
But i am not sure how you bind close_tag with name (you should probably emit some dummy tokens?), so if you can share artifacts of your previous work on it, i can update it.

leptos_macro/src/template.rs
gbj: Ah okay. No, I think it was just legacy code. Yes `str_value` is same as `value_to_string`.

gbj:
Agreed, #766 should be entirely separate from this PR. I just meant that having you maintaining rstml more actively than syn-rsx is maintained would mean we might have some more syntactic flexibility! (I see your issue list for example includes a few things Leptos is shipping workarounds for right now)

Re: https://github.com/stoically/syn-rsx/issues/51, I essentially added this to `leptos_dom/src/html.rs` after `HtmlElement::with_view_marker`
```
#[cfg(debug_assertions)]
    /// Marks the closing of this HTML element.
    #[inline(always)]
    pub fn closing_tag(self) -> Self {
        self
    }
```
and then added a call to `closing_tag()`, spanned to the closing tag, in `element_to_tokens`, for elements that had closing tags. This caused the syntax highlighting for the open and closing tags to match, but didn't do anything more interesting like try to link them to one another somehow. I'm not sure if there's a way to indicate to rust-analyzer that two different spans refer to the same thing. Conceptually, for elements and components both the opening and closing tags are linked to the `name` variable in `element_to_tokens`, but at present only the opening tag name token is connected to the element/component function invocation from rust-analyzer's perspective.

vldm:
>I'm not sure if there's a way to indicate to rust-analyzer that two different spans

I didn't find a way either.

>and then added a call to closing_tag(), spanned to the closing tag, in element_to_tokens, for elements that had closing tags

Theoretically macro also can emit dummy token before dom builder, example that i use in html-to-string-macro:

```rust
enum x {}; // for each html, head, body (open and close tag)
let _ = crate::docs::element; // for rest tags
```
I done this as an example, how one can colorize its output, and add output.

I think some hack way to keep close tag colorized and linked to the valid documentation would be emiting tokens inside `if fasle`, so rust will not execute it, but it still be possible to bould span to valid component Type.

```rust
{
if false {
  let _ = leptos::leptos_dom::html::dom; // for close dom tag
  let _ = Option::<Component>::Some; // for close component
}

leptos::leptos_dom::html::section().child(..)

}
```

gbj:
> I think some hacky way to keep close tag colorized and linked to the valid documentation would be emiting tokens inside `if fasle`, so rust will not execute it, but it still be possible to bould span to valid component Type.
> 
> ```rust
> {
> if false {
>   let _ = leptos::leptos_dom::html::dom; // for close dom tag
>   let _ = Option::<Component>::Some; // for close component
> }
> 
> leptos::leptos_dom::html::section().child(..)
> 
> }
> ```

Oh I think this is very clever.

I think I'll merge this, for now, and then add the closing tag piece in a separate PR. Does that sound good to you? I think this is complete otherwise, right?

vldm:
> > I think some hacky way to keep close tag colorized and linked to the valid documentation would be emiting tokens inside `if fasle`, so rust will not execute it, but it still be possible to bould span to valid component Type.
> 
> > 
> 
> > ```rust
> 
> > {
> 
> > if false {
> 
> >   let _ = leptos::leptos_dom::html::dom; // for close dom tag
> 
> >   let _ = Option::<Component>::Some; // for close component
> 
> > }
> 
> > 
> 
> > leptos::leptos_dom::html::section().child(..)
> 
> > 
> 
> > }
> 
> > ```
> 
> 
> 
> Oh I think this is very clever.
> 
> 
> 
> I think I'll merge this, for now, and then add the closing tag piece in a separate PR. Does that sound good to you? I think this is complete otherwise, right?
> 
> 
> 
> 


Yes, thanks. I will open separate PR.
This pr was simple refactoring, but for closing tags I want to understand more about view and template macro.

Issue #994, opened 2023-05-03
Option to wait until resources are loaded when navigating 


Skyliegirl33:
I think it'd be great if we can have an option to wait until the resources of a page are resolved before rendering the page.
Example: if I'm viewing a blog post and want to navigate to the list of all posts, currently the navigation occurs before the resource is resolved, which, depending on how long that takes, can result in a noticeable delay or flicker. In such a case, it might be preferable to not navigate to the page at all until the resource(s) are resolved.

Assuming this can be implemented on the route level, something like this would work:
```rust
<Route path="posts" view=|cx| view!{ cx, <Blog/> } load=LoadingMode::WaitResolved />
<Route path="" view=|cx| view!{ cx, <Home/> } load=LoadingMode::Fallback />
```
or
```rust
<Route path="posts" view=|cx| view!{ cx, <Blog/> } render=RenderMode::Blocking />
<Route path="" view=|cx| view!{ cx, <Home/> } render=RenderMode::Fallback />
```

gbj:
Yep this makes perfect sense and is built into the router in e.g., SolidJS, with its more-advanced transition system. I played around a little with wrapping `<Outlet/>` in `<Transition/>` and so on and got the dread hydration mismatch warnings, but this should definitely be doable and possibly even the default behavior.

PR #1055, opened 2023-05-18
feat: add "async routing" feature

gbj:
#994 describes the behavior used by many (but not all) frameworks: if you navigate to a new page in the client (after hydration or with client-side rendering) and async resources must be loaded to render it, rather than navigating immediately and showing a "Loading..." fallback, hold the current page until the resources have loaded.

I'm open to suggestions on how to opt into this. For the moment, I've done it by adding an optional `set_is_routing` prop to the `<Router/>` component that just takes a `WriteSignal<bool>` or similar.

```rust
#[component]
    let (is_routing, set_is_routing) = create_signal(cx, false);

    view! { cx,
        <Router set_is_routing>
```

In my opinion this is kind of clever because it requires the user to provide some kind of indicator that will be set during the routing process (i.e., while the previous page is still being held), which is necessary for good UX. If you don't use this signal, it looks like the page is just not responding while loading data, as the browser's own page-load indicators don't fire (since it's a client-side navigation).

I think this is much more polished than the "fallback" default, so this may become default behavior in a future release.

https://github.com/leptos-rs/leptos/assets/286622/cecd9e8b-1305-4959-b5d9-3708571e8940

To-do
- [x] Support this in `<Outlet/>` for nested routes as well as `<Routes/>` for top-level routes
- ~~[ ] Test against animated routing~~ (this is hard, maybe I'll leave it for later...)
- [x] provide a default progress-bar kind of component as well, that makes it easy to use that pending notification?

### DIFF
--- a/examples/hackernews/src/lib.rs
+++ b/examples/hackernews/src/lib.rs
@@ -1,5 +1,5 @@
 use cfg_if::cfg_if;
-use leptos::{component, view, IntoView, Scope};
+use leptos::*;
 use leptos_meta::*;
 use leptos_router::*;
 mod api;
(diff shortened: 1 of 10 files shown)


gbj:
`<RoutingProgress/>` component 

https://github.com/leptos-rs/leptos/assets/286622/6ba39958-06c7-43b8-9a71-1a39132ffc07

PR #1057, opened 2023-05-18
fix: form data deserialization

yuuma03:
The other patch introduced one problem that I initially overlooked: form data isn't guaranteed to be valid UTF-8, as it might contain arbitrary binary data inside.
The default deserialization method (URL) will fail on invalid UTF-8, even if there's no args to deserialize (which would be the case if you're extract()ing using actix-multipart).
For now, I'm marking this as a draft, as I'm not quite sure if this is the correct way of fixing this.
Probably it would be better to implement a way of deserializing multipart/form-data (as we do have access to the raw u8 buffer), and introduce proper support for it on ActionForm/MultiActionForm (as that is broken on multipart, and at the moment I'm manually using a Form instead), but I'm still out of ideas in how to do that properly.


yuuma03:
I'm going to close this draft in favor of another solution I'm writing at the moment.

PR #1060, opened 2023-05-18
Added Debug, PartialEq and Eq derives to trigger.

dgsantana:
Simple commit to add PartialEq, Eq and Debug derive traits to Trigger.


gbj:
Cheers!

PR #1062, opened 2023-05-19
test: verify tailwind example with playwright tests

agilarity:
This runs the Playwright tests for the **tailwind** example as part of the `verify-examples` task flow, via the `web-test` task.

Verification:

From the **examples/tailwind** directory...

- Run `cargo make setup`
- Run `cargo make web-test`


gbj:
Nice, thanks!

PR #1063, opened 2023-05-19
Docs edit:   added a hint for a common error when using use_navigate

sjud:
added a hint for a common error when using use_navigate


gbj:
Thanks for this, sorry for the doctest pain!

PR #1069, opened 2023-05-19
Specify Server Fn Paths

### LINKED ISSUE #1061: feat: Specify path of Server Functions
Basically I'd like to define an API with Leptos for external consumption, and use Leptos server functions on it. I believe I can do that if I can specify the path the server function will accept paths on.

As an example, I could define /login and the server function
```rust
// I know there are more args to the macro, but we'd add on or replace the prefix section
#[server("/login")]
pub fn login() -> Result<(), ServerFnError>{
...
}
```

benwis:
Add an option to specify the full server fn paths, so that one could define an API for external consumption with Leptos server Fns. Should close #1061

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -75,9 +75,11 @@ pub fn server_macro_impl(
         struct_name,
         prefix,
         encoding,
+        fn_path,
         ..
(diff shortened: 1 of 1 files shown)


gbj:
Looks good to me! Thanks, Ben.

PR #1085, opened 2023-05-23
Remove LeptosProvider trait and use Axum SubStates

### LINKED ISSUE #1079: Is axum with_state broken?
First: thank you for this wonderful project! I look forward to fully absorbing all this framework can do! I am entirely new to this project, so might be missing something obvious. Therefore, please bear with me.

**TL;DR** 
I have an issue using `with_state` setting up my axum routers. The `leptos_routes` call seems to force the axum router to become a `Router<()>`-router (i.e. without any meaningful state?).

**Long version:**
I have setup my leptos project using the `https://github.com/leptos-rs/start-axum` template using `cargo-leptos`. Because I want to pass other state through to non-leptos routes I wanted to use `with_state` to manage my app state in axum, rather than `Extension`-method the starter template is using. I see [`with_state` support](https://github.com/leptos-rs/leptos/pull/1032) was recently added, and from what I gather this should also have made it into the 0.3.0 release which I am using.

The problem I am observing is that as soon as I call [`leptos_routes`](https://github.com/sebastian/leptos_with_state/blob/main/src/main.rs#L41) the type of the router becomes `Router<()>` which then causes a compilation error when I subsequently call `with_state`:

```
   --> src/main.rs:43:21
    |
43  |         .with_state(state);
    |          ---------- ^^^^^ expected `()`, found `State`
    |          |
    |          arguments to this method are incorrect
    |
note: method defined here
   --> /Users/sebastian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/axum-0.6.18/src/routing/mod.rs:294:12
    |
294 |     pub fn with_state<S2>(self, state: S) -> Router<S2, B> {
    |            ^^^^^^^^^^
```

I created a minimal reproducible (well, it reproduces nicely on my end...) version here: https://github.com/sebastian/leptos_with_state

Could the reason be that `leptos_routes` calls into:

```
/// The default implementation of `LeptosRoutes` which takes in a list of paths, and dispatches GET requests
/// to those paths to Leptos's renderer.
impl LeptosRoutes for axum::Router {
    #[tracing::instrument(level = "info", fields(error), skip_all)]
    fn leptos_routes<IV>(
        self,
        options: LeptosOptions,
        paths: Vec<RouteListing>,
        app_fn: impl Fn(leptos::Scope) -> IV + Clone + Send + 'static,
    ) -> Self
    where
        IV: IntoView + 'static,
```

where `axum::Router` isn't specified as `axum::Router<S, B>`?

benwis:
Basically the goal is to remove the LeptosProvider trait and use Axum's inbuilt SubStates with FromRef. Should close #1079

### DIFF
--- a/examples/session_auth_axum/src/main.rs
+++ b/examples/session_auth_axum/src/main.rs
@@ -6,37 +6,38 @@ if #[cfg(feature = "ssr")] {
     use axum::{
         response::{Response, IntoResponse},
         routing::get,
-        extract::{Path, State, Extension, RawQuery},
+        extract::{Path, State, RawQuery},
         http::{Request, header::HeaderMap},
         body::Body as AxumBody,
         Router,
     };
     use session_auth_axum::todo::*;
     use session_auth_axum::auth::*;
+    use session_auth_axum::state::AppState;
     use session_auth_axum::*;
     use session_auth_axum::fallback::file_and_error_handler;
     use leptos_axum::{generate_route_list, LeptosRoutes, handle_server_fns_with_context};
-    use leptos::{log, view, provide_context, LeptosOptions, get_configuration};
+    use leptos::{log, view, provide_context, get_configuration};
     use sqlx::{SqlitePool, sqlite::SqlitePoolOptions};
     use axum_database_sessions::{SessionConfig, SessionLayer, SessionStore};
     use axum_sessions_auth::{AuthSessionLayer, AuthConfig, SessionSqlitePool};
 
-    async fn server_fn_handler(Extension(pool): Extension<SqlitePool>, auth_session: AuthSession, path: Path<String>, headers: HeaderMap, raw_query: RawQuery,
+    async fn server_fn_handler(State(app_state): State<AppState>, auth_session: AuthSession, path: Path<String>, headers: HeaderMap, raw_query: RawQuery,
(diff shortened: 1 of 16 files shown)


examples/session_auth_axum/src/main.rs
```diff
@@ -33,8 +32,8 @@ if #[cfg(feature = "ssr")] {
         }, request).await
     }
 
-    async fn leptos_routes_handler(Extension(pool): Extension<SqlitePool>, auth_session: AuthSession, State(options): State<Arc<LeptosOptions>>, req: Request<AxumBody>) -> Response{
-            let handler = leptos_axum::render_app_to_stream_with_context((*options).clone(),
+    async fn leptos_routes_handler(Extension(pool): Extension<SqlitePool>, auth_session: AuthSession, State(options): State<LeptosOptions>, req: Request<AxumBody>) -> Response{
```
akarras, line 35: Might be nice to use this example as a place to give examples of substates

akarras (approved):
Looks good! Tested with my local repo and seems to work.

examples/session_auth_axum/src/main.rs
benwis: A good idea, I did that in my project, but not to the example

gbj:
Okay great, thank you!

PR #1086, opened 2023-05-23
fix: duplicate headers (like Set-Cookie) on the actix integration

yuuma03:
Ignoring any header with a `None` key means that we can't set multiple cookies (for example).
The correct behaviour seems to be reusing the last non-None key from what I saw in the docs.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -254,17 +254,11 @@ pub fn handle_server_fns_with_context(
                             }
 
                             // Use provided ResponseParts headers if they exist
-                            let mut last_key = None;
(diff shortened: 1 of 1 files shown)


gbj:
Hm maybe this should've just been using [`.into_iter()`](https://docs.rs/actix-web/latest/actix_web/http/header/struct.HeaderMap.html#impl-IntoIterator-for-HeaderMap) in the first place rather than `.drain`. 

`.into_iter()` is documented to clone the header name for each header but that's what we want anyway. Maybe cleaner than either the old or new drain versions?

gbj:
Looks good, thanks!

PR #1096, opened 2023-05-25
Improve fetch example

anacrolix:
I was keen to try out the fetch example, and was confused by what it was showing me initially. This version is a lot of fun, individual details are in the commit messages.


gbj:
Oh thanks! Yes, as you saw I messed this up at some point by accidentally committing a version that removed the `` tags, and believe it or not the Cats API is actually broken so it returns 10 cats for any number over 1 or whatever :facepalm: I like your iterator solution.

PR #1101, opened 2023-05-26
Fix: make template macro independent.

vldm:
In my work on "highlighting closing tags" i have found that `template!` macro cannot be used dirrectly in examples (i tested in counter).

Few imports was missing, and some extern crates was needed to use `template!` instead of `view!`.
I have added feature gate and hide all needed imports behind it. 
But if you want i can remove custom feature, reimport all crates in `csr` mode, and add fallback `use macro::view as template` in `ssr` or `hydration` mode. 

I also have copied code from [this benchmark](https://github.com/krausest/js-framework-benchmark) to the examples, to keep `template` testable in future.

If you plan to remove `template!` or deprecate it in future, feel free to close this PR. But if you want to keep `template!`, i can add support of semantic highlighting to it (currently it is completely broken).


gbj:
Thanks! This looks reasonable to me.

Yes, the semantic highlighting's completely messed up.

Re: future plans — I will definitely be keeping `template` around. It was actually the original `view` macro but replaced by the more flexible current version of the renderer. There's a small but meaningful performance benefit in some cases using `template`. Removing the SSR/hydration aspect of it made it much more manageable, but it could definitely use some love. One possibility for the future involves merging some of the `template` logic back into the `view`, but I don't have any immediate plans to do it.

PR #1102, opened 2023-05-26
docs: fix generic type

abhikjain360:
Fix the signature for generic params in the ch18 of the docs, from `Memo<Result<T>, _>` to `Memo<Result<T, _>>`.

Thanks!


gbj:
Thanks very much!

PR #1105, opened 2023-05-26
added scrollend event

maccesch:
Please see https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event


gbj:
Thanks! Looks like it's new to the standard? (Or at least only implemented in Firefox and Chrome Canary.) I'm happy to add it in that case.

maccesch:
Thanks for the fast merge! I didn't know it existed until I started porting `use_scroll` from solidjs-use. https://synphonyte.github.io/leptos-use/sensors/use_scroll.html

PR #1122, opened 2023-05-29
perf: reduce overhead of hydration keys

gbj:
The current hydration-key system mirrors the one in SolidJS, with keys growing over time so you end up with things like `0-1-1-1-1-1-1-2-1-3-1-3-10-3` as the component tree is nested. Originally this was for the correct handling of suspense. However, as our renderer has grown apart from Solid's the way we actually render things has changed.

This way of generating hydration keys was also creating massively long chains, that actually grew linearly as the number of components used grew, which meant larger pages were built slower and slower (because you were generating longer and longer keys per component.)

I've done a bit of an experiment and this turns out to be almost entirely unnecessary, I think. Because the renderer is now built to run completely top-down, we can just use a single-number key, which should really help with SSR performance.

This is a draft for now because I ran out of time, but wanted to share it so people can do some early testing.

To dos
- [ ] Clippy/cleanup
- [ ] Test against more examples and real apps

### DIFF
--- a/leptos_dom/src/components/errors.rs
+++ b/leptos_dom/src/components/errors.rs
@@ -80,7 +80,7 @@ where
     E: Error + Send + Sync + 'static,
 {
     fn into_view(self, cx: leptos_reactive::Scope) -> crate::View {
-        let id = ErrorKey(HydrationCtx::peek().previous.into());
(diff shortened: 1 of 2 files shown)


luckynumberke7in:
It's a bit shocking to see so few lines capable of making such a large impact to performance! I look forward to making it to the ssr side of leptos =P

PR #1129, opened 2023-05-31
work around for remove inline0.js

tempbottle:
this pr use js_sys Reflect to remove snippets/inline0.js file, which can simplify the deployment of leptos apps.

### DIFF
--- a/leptos_reactive/src/spawn_microtask.rs
+++ b/leptos_reactive/src/spawn_microtask.rs
@@ -1,8 +1,4 @@
-// `queue_microtask` needs to be in its own module, which is the only thing
-// in this entire framework that requires "unsafe" code (because Rust seems to
-// that a `wasm_bindgen` imported function like this is unsafe)
-// this is stupid, and one day hopefully web_sys will add queue_microtask itself
-
+#![forbid(unsafe_code)]
 use cfg_if::cfg_if;
 
 cfg_if! {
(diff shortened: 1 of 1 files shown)


leptos_reactive/src/spawn_microtask.rs
```diff
@@ -14,11 +14,13 @@ cfg_if! {
...
-            inline_js = "export function microtask(f) { queueMicrotask(f); }"
-        )]
-        extern "C" {
-            fn microtask(task: wasm_bindgen::JsValue);
+        fn microtask(task: wasm_bindgen::JsValue) {
+            use js_sys::{Reflect, Function};
+            use wasm_bindgen::prelude::*;
+            let window = web_sys::window().expect("window not available");
+            let queue_microtask = Reflect::get(&window, &JsValue::from_str("queueMicrotask")).expect("queueMicrotask not available");
+            let queue_microtask = queue_microtask.dyn_into::<Function>().expect("queueMicrotask not a function");
```
gbj, line 22: This should probably just used `.unchecked_into()` rather than `.dyn_into()` and unwrapping.

gbj (commented):
Thanks, I think this is a great idea. It would both remove the extra inline JS file, and remove the need for this single module to allow unsafe code (see comment at the top of this file, which can also be removed).

PR #1133, opened 2023-06-02
actix: Removed calls to .collect() and .into_ier()

martinfrances107:
I want to reduce the number of calls to .collect()

The code below shows the variable "routes" and the  first .iterator -> .map() ->  .collect() code pattern .

Later on in the code the "collect()'ed"  routes are then converted by a .into_iter() call 
and fed into next .map() call 

The whole thing can be simplified by daisy-chaining the two .map()  calls.

Another way of thinking about this is  'This change signals to the compiler, don't spool the intermediate values of "routes" to memory 
before proceeding'

```rust 
    routes = routes
        .into_iter()
        .map(|listing| {
            let path = listing.path();
            if path.is_empty() {
                return RouteListing::new(
                    "/".to_string(),
                    listing.mode(),
                    listing.methods(),
                );
            }
            RouteListing::new(listing.path(), listing.mode(), listing.methods())
        })
        .collect();
```

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -891,6 +891,12 @@ where
 {
     let mut routes = leptos_router::generate_route_list_inner(app_fn);
 
+    // Actix's Router doesn't follow Leptos's
+    // Match `*` or `*someword` to replace with replace it with "/{tail.*}
+    let wildcard_re = Regex::new(r"\*.*").unwrap();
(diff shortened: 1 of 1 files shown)


martinfrances107:
It look like the tests runner timed out rebuilding the package cache ... I am force pushing after rebase..


time has passed and maybe things are more stable now.

gbj:
Looks good, thanks!

PR #1134, opened 2023-06-02
integrations: use lazy static to define Regular expressions.

martinfrances107:
This is a follow up to https://github.com/leptos-rs/leptos/pull/1133

The code change is in the same function, but I think separating activities in to mico PRs  is better as when reverting code ... allow for more fine-grained control

RegEx::new() is surprisingly expensive, so I think the "Compute Once" pattern should always be used.


gbj:
`generate_route_list` is typically just called once during server startup to create the list of Leptos-handleable routes that's passed into the Actix router. 

I think `lazy_static` is a small enough overhead that this is reasonable to add; and it certainly guards against the possibility of accidentally using `generate_route_list` in more scenarios in the future and recompiling the regex.

martinfrances107:
This PR is a few hours old and already out of date!

Yesterday rust 1.7 was released, and the third part crate "lazy_static" was stabilised in core.

My preferred solution **NOW** is to insist on 1.7 as the min version of rust, thereby cutting out a dependency.

I appreciate this will lead to a policy change and/or a major path version bump.

I appreciate Insisting on 1.7 will be disruptive in the short term  and not for everyone  and so I am about to file a new issue, so the wider community can comment on the disruption.

lazy_static and the 1.7 release is discussed here.

https://www.youtube.com/watch?v=_hNOTQRip1Q

martinfrances107:
The next action is to rewrite when we don't need to pull in lazy_static 

Its a minor point, and the PR will be disruptive,

PR #1136, opened 2023-06-02
Ran cargo clippy --fix.

martinfrances107:
Lots of minor changes 

Only one thing worth talking about is the removal of a .clone() call.

As with the recent release of  rust 1.7 : cargo clippy has added a few new rules.


gbj:
Looks good thanks! (The `.clone()` removal is fine, the global suspense moved from implementing only Clone to also Copy and it was left over)

PR #1146, opened 2023-06-05
For algo v2

jquesada2016:
This PR should close most issues currently open for `<For />`.


jquesada2016:
@gbj I think this branch is ready for a review to merge. Let me know if you find any outstanding bugs, regardless, this `<For />` implementation is in a much more usable state than the current one. Also please let me know if you would like me to do any chores.

gbj:
@jquesada2016 I've thrown everything I can at this and it looks good to me. I also went ahead and fixed #1168 which was relatively straightforward (having to do with the way the `<Each/>` calculates its opening node in the `clear: true` case being off by one comment node in debug mode).

Thanks so much for your work. This was a massive effort with a really good result.

PR #1154, opened 2023-06-06
feat: register server functions automatically

gbj:
This removes the need to explicitly `::register()` server functions on most platforms (including Linux, macOS, iOS, FreeBSD, Android, and Windows). If you're using server functions on a WASM server platform, you will still need to register server functions, by calling `::register_explicit()`. Otherwise, you can simply remove this from your code—the `#[server]` macro will automatically register server functions for you.

I'm opening this as an alternative to #613, since enough time has passed that I found it easier to start from the current state of things rather than updating that one.

This definitely seems to work for the Leptos examples. @Demonthos I'd appreciate you taking a look at it, as I'm less confident that I updated the default implementation correctly. And huge thanks to you for the initial implementation, which I leaned on pretty heavily here.


Demonthos:
Sorry, I didn't have a chance to look at this before it was merged. I am currently out of town, so I don't have a development setup. From what I can see the default implementation looks good

niclas-ahden:
We've had this in production for an hour now without issue. We have 12 server functions and have processed ~10k calls to them so far.

Thanks @gbj, it's exciting to see the pace you're moving at!

Issue #1156, opened 2023-06-07
Nested component Transitions render their fallback view on page refresh


PlGGS:
**Describe the bug**
Nested component Transitions render both their fallback as well as their content views on a page refresh

**Leptos Dependencies**
```toml
leptos = { git="https://github.com/leptos-rs/leptos", default-features = false, features = [
	"serde", "tracing",
] }
leptos_meta = { git="https://github.com/leptos-rs/leptos", default-features = false }
leptos_axum = { git="https://github.com/leptos-rs/leptos", optional = true }
leptos_router = {  git="https://github.com/leptos-rs/leptos", default-features = false }
```

**To Reproduce**
1. Create two components with transitions for rendering async content
2. Nest one of the components within the other
3. Build and watch project
4. Refresh the page to see both the fallback and the content of the inner component's Transition

**Expected behavior**
I expected both components to render their content as normal

**Screenshots**
My inner component: (Ignore the unnecessary braces warning. That's a separate issue, because those braces are  necessary)
![image](https://github.com/leptos-rs/leptos/assets/20197872/6592712c-d6e1-4b94-8eb5-2f7a5cc48417)

DOM before refresh:
![image](https://github.com/leptos-rs/leptos/assets/20197872/e01e1158-fc04-4fc6-9dba-74e260d8b9a4)

DOM after refresh:
![image](https://github.com/leptos-rs/leptos/assets/20197872/434e4965-776f-4f26-a308-75c56e43c986)

It goes from looking like this:
![image](https://github.com/leptos-rs/leptos/assets/20197872/76a9ca47-54e5-452a-a952-d90dfc70150f)

To this:
![image](https://github.com/leptos-rs/leptos/assets/20197872/2fe81474-9ea5-4beb-893f-658aa0b07df2)

Thanks!

PlGGS:
Here's what the DOM looks like with my current workaround: <Transition fallback=move || view! {cx, **{}** }>
![image](https://github.com/leptos-rs/leptos/assets/20197872/cb2cbe6f-0cd6-4a19-a92f-320104f2dc6b)

PlGGS:
Update: This was apparently only an issue as of Firefox version 112.0.2. I just tried this in Chromium and then updated Firefox, and they both worked without a problem.

PlGGS:
Nevermind, it's now only occuring on this other nested component in my project:
[...]

Issue #1117, opened 2023-05-29
bug: Contents of `Transition` are not removed when navigating to another route (`PartiallyBlocked`)


Skyliegirl33:
The contents of the `Transition` for `/auth/settings` is not cleared upon navigation, but only if SSR was done on `/auth/settings`.

Example reproduction, `auth_context.user` is just any resource:

```rust
#[component]
pub fn App(cx: Scope) -> impl IntoView {
  provide_meta_context(cx);
  
  view! {
    cx,
    <Routes>
      <Route path="" view=|cx| view!{ cx,
          <div>
            <p>"Hello!"</p>
            <A href="/auth/settings">"Settings"</A>
          </div> }
          ssr=SsrMode::PartiallyBlocked
      />
      <Route path="auth/settings"
          view=move |cx| view!{ cx,
          <Transition fallback=|| ()>
              {move || {
              match auth_context.user.read(cx) {
                  Some(Ok(Some(_))) => view! { cx,
                  <div>
                    <h1>"Settings"</h1>
                    <A href="/">"Home"</A>
                  </div> }.into_view(cx),
                  None => ().into_view(cx),
                  _ => view! { cx, <Redirect path="/"/> }.into_view(cx)
              }
              }}
          </Transition>
          }
          ssr=SsrMode::PartiallyBlocked
      />
    </Routes>
  }
}
```

gbj:
As far as I can tell this is caused by the fact that there's no intervening component at all, but that the `Transition` is directly inside the `Route` view, which I was able to resolve and will roll into #1157.

PR #1157, opened 2023-06-07
feat: clone <Suspense/> children instead of calling again (closes #398) 

### LINKED ISSUE #398: Suspense should cache children to avoid rerendering them if you fall back a second time

gbj:
Currently `<Suspense/>` and `<Transition/>` rerender their children each time they resolve. This is bad for rendering efficiency and causes issues like this https://github.com/leptos-rs/leptos/issues/398#issuecomment-1581144454, in which creating and reading resource inside a `<Suspense/>` immediately creates an infinite loop.

The changes in #1135 should make it possible to clone `<Suspense/>` instead of rerendering. We already move people toward reading resources reactively inside `<Suspense/>`, so that should continue to work.

TODO 
- [x] ErrorBoundary/Suspense interaction seems broken
- [x] Check that earlier memory leak hasn't reoccurred (shouldn't)
- [x] Double-check that examples are using `<Suspense/>` reactively correctly

### DIFF
--- a/leptos/src/suspense_component.rs
+++ b/leptos/src/suspense_component.rs
@@ -119,7 +119,8 @@ where
                                         current_id.clone(),
                                     );
                                     DynChild::new({
-                                        let orig_children = orig_children(cx).into_view(cx);
+                                        let orig_children =
+                                            orig_children(cx).into_view(cx);
                                         move || orig_children.clone()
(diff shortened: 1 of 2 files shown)


gbj:
I would like to roll #1117 and #1156 into this too.

PR #1165, opened 2023-06-09
Replace preempted with version in Resource

### LINKED ISSUE #1124: Resource output will not correspond to the input if first request arrives after second request
Resource output will not correspond to the input if first request arrives after second request, here's a small example:

```rust
#[server(Foo, "/api")]
async fn server_foo(with_delay: bool) -> Result<bool, ServerFnError> {
    use tokio::time::{sleep, Duration};

    if with_delay {
        sleep(Duration::from_secs(5)).await;
    }

    Ok(with_delay)
}

async fn fetch_server_foo(with_delay: Option<bool>) -> Option<bool> {
    if let Some(with_delay) = with_delay {
        server_foo(with_delay).await.ok()
    } else {
        None
    }
}

#[component]
fn HomePage(cx: Scope) -> impl IntoView {
    let (with_delay, set_with_delay) = create_signal(cx, None);
    let resource = create_resource(cx, with_delay, fetch_server_foo);

    let on_click_with_delay = move |_| {
        set_with_delay(Some(true));
    };

    let on_click_without_delay = move |_| {
        set_with_delay(Some(false));
    };

    view! { cx,
        <button on:click=on_click_with_delay>"With delay"</button>
        <button on:click=on_click_without_delay>"Without delay"</button>
        { move || with_delay().map(|delay| if delay {
            view! { cx, <p>"Input has delay"</p> }
        } else {
            view! { cx, <p>"Input has no delay"</p> }
        }) }
        <Suspense fallback=move || view!{ cx, <p>"Loading.."</p> }>
        { move || resource.read(cx).flatten().map(|delay| if delay {
            view! { cx, <p>"Output had delay"</p> }
            } else {
                view! { cx, <p>"Output had no delay"</p> }
            }) }
        </Suspense>
    }
}
```

You can also run it at [this repo](https://github.com/prscoelho/server-fn-delay)

Reproducing: click with delay, then click without delay.

Expected: it should show loading until you click on without delay, and it should show "Input has no delay", "Output had no delay". 
Result: it shows loading until the slowest (outdated) resource finishes, and it shows "Input has no delay", "Output had delay".

PaulWagener:
Closes #1124

I've also added a unit test that tests for this issue but it doesn't run correctly because of some microtask issue I don't fully understand.

### DIFF
--- a/leptos_reactive/src/resource.rs
+++ b/leptos_reactive/src/resource.rs
@@ -211,7 +211,7 @@ where
         fetcher,
         resolved: Rc::new(Cell::new(resolved)),
         scheduled: Rc::new(Cell::new(false)),
-        preempted: Rc::new(Cell::new(false)),
+        version: Rc::new(Cell::new(0)),
         suspense_contexts: Default::default(),
         serializable,
     });
@@ -348,7 +348,7 @@ where
         fetcher,
         resolved: Rc::new(Cell::new(resolved)),
         scheduled: Rc::new(Cell::new(false)),
-        preempted: Rc::new(Cell::new(false)),
+        version: Rc::new(Cell::new(0)),
         suspense_contexts: Default::default(),
         serializable: ResourceSerialization::Local,
     });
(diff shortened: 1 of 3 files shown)


gbj:
Thanks, this looks good! I made a couple very small changes so the test runs correctly and it passes.

PR #1167, opened 2023-06-10
feat: basic TODO app; no bells, no whistles

jdevries3133:
I was wishing this was present when trying to learn from the examples in the repo – just a basic client-side TODO app without any server-side code or isomorphism.

I feel like this specific implementation probably stinks – I'm totally new to both Leptos and Rust, so happy to take any feedback & no hard feelings if there isn't interest in pulling an example like this into the repo. I fully appreciate my naive perspective here :)

### DIFF
--- a/examples/todo_app_basic/src/main.rs
+++ b/examples/todo_app_basic/src/main.rs
@@ -1,0 +1,86 @@
+use leptos::*;
+
+fn main() {
+    mount_to_body(|cx| view! { cx, <App /> });
+}
+
+#[derive(Eq, Hash, PartialEq, Clone, Debug)]
+struct TodoItem {
+    id: u32,
+    name: String,
+}
+
+#[derive(Copy, Clone)]
+struct TodoGenerator {
+    next_id: u32,
+}
+
+impl TodoGenerator {
+    fn new() -> Self {
+        TodoGenerator { next_id: 1 }
+    }
+    fn get_todo(&mut self, name: String) -> TodoItem {
+        let next_id = self.next_id;
+        self.next_id += 1;
+        TodoItem { name, id: next_id }
+    }
+}
+
+#[component]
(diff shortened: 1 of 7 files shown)


phillipbaird:
Would Vec.retain() be a simpler way to implement delete_todo?

jdevries3133:
>  Would Vec.retain() be a simpler way to implement delete_todo?

Yeah, that's a definite improvement, thank you!

jdevries3133:
Maybe this could be called `todo_component`? I think that not everyone is familiar with the MVC pattern, and this is line-for-line akin to a todo list app that you might see taught to React developers. Nonetheless, my feelings won't be hurt if you think this doesn't fit! I think you are right that the same ideas are represented already in other examples.

examples/todo_app_basic/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[derive(Eq, Hash, PartialEq, Clone, Debug)]
+struct TodoItem {
+    id: u32,
+    name: String,
+}
+
+#[derive(Copy, Clone)]
+struct TodoGenerator {
```
flosse, line 14: This should be a `struct TodoId` that has a `fn next(&self)` or so but you don't want to have a `TodoGenerator`.

gbj:
@jdevries Sorry for my slow response here. Thanks for your contribution. We have created a new `projects` directory to house a wider variety of projects without the same maintenance burden as examples — Would you be interested in updating your PR and adding it to `projects` instead? I totally understand if not, it's been a while :-)

gbj:
@jdevries Sorry for my slow response here. Thanks for your contribution. We have created a new `projects` directory to house a wider variety of projects without the same maintenance burden as examples — Would you be interested in updating your PR and adding it to `projects` instead? I totally understand if not, it's been a while :-)

PR #1169, opened 2023-06-10
doc: previews to backup CodeSandbox

jdevries3133:
I experience sporradic unavailability with the code sandbox MicroVM embed that we're using in the book. It's very nice when it works, but I was getting an error where I couldn't read the examples at all. This drops the text of all MicroVM examples into the book itself behind a `<preview>` so that readers can still view the source even if CodeSandbox is wonky.


jdevries3133:
I tore through these pretty quickly; I will definitely give these a second look when I have more time & I expect I probably did at least one mis-paste, so please look closely if you review this before I get a chance to take a second look here lol!

gbj:
Thanks for doing all this — I finally had time to check them and only found one mistake.

PR #1173, opened 2023-06-12
fixed removes being wrong and leading to panics sometimes

jquesada2016:
This PR fixes a bug when removing items that would almost always cause moves to be off by the number of removals


gbj:
Was this marked draft on purpose, or is it good to go?

jquesada2016:
@gbj sorry about that, I didn't notice. It's good ta gooooo

PR #1181, opened 2023-06-13
fixed bug when clearing which ignores further additions

jquesada2016:
This branch is intended to fix a bug which causes added items to be ignored when clearing in the `<For />` component.


jquesada2016:
@gbj got it, it was the `Range`mode when clearing that was set to also delete the opening marker node.

gbj:
Awesome, thanks!

PR #1183, opened 2023-06-14
Use config for serde_qs deserialization

hchockarprasad:
Currently `from_form_data` implementation tries to convert `UrlSearchparams` formatted query string to Type T using 'serde_qs::from_str' which couldn't convert encoded square brackets to nested types of T.

This PR fixes that by replacing `serde_qs::from_str`  with  `serde_qs::Config` as it provides the needed functionality.

Example
```
#[derive(Serialize, Deserialize)]
struct Address {
    door: String,
    street: Option<String>
}

#[derive(Serialize, Deserialize)]
struct Data {
    id: u32,
    address: Address
}

let formData = "id=5&address[door]=248"; // UrlSearchParams encoded

serde_qs::from_str<T>(formData); // throws error
serde_qs::Config::default()::deserialize_str<T>(formData) // this works
```

### DIFF
--- a/router/src/components/form.rs
+++ b/router/src/components/form.rs
@@ -692,6 +692,6 @@ where
             web_sys::UrlSearchParams::new_with_str_sequence_sequence(form_data)
                 .unwrap_throw();
         let data = data.to_string().as_string().unwrap_or_default();
-        serde_qs::from_str::<Self>(&data)
+        serde_qs::Config::default().deserialize_str::<Self>(&data)
     }
 }
--- a/router/src/components/form.rs
+++ b/router/src/components/form.rs
@@ -692,6 +692,6 @@ where
             web_sys::UrlSearchParams::new_with_str_sequence_sequence(form_data)
                 .unwrap_throw();
         let data = data.to_string().as_string().unwrap_or_default();
-        serde_qs::Config::default().deserialize_str::<Self>(&data)
+        serde_qs::Config::new(5, false).deserialize_str::<Self>(&data)
     }
 }
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -375,7 +375,7 @@ where
         // decode the args
         let value = match Self::encoding() {
             Encoding::Url | Encoding::GetJSON | Encoding::GetCBOR => {
-                serde_qs::from_bytes(data)
+                serde_qs::Config::new(5, true).deserialize_bytes(data)
                     .map_err(|e| ServerFnError::Deserialization(e.to_string()))
             }
             Encoding::Cbor => ciborium::de::from_reader(data)


gbj:
I've gotten myself fairly confused trying to test how this is supposed to work in forms in practice.

Here's my attempt at reconstructing the example you gave so I can test it with an `<ActionForm/>`. Trying to use this to submit the form gives 
```
    Err(
        Deserialization(
            "missing field `address`",
        ),
    ),
```
Against the current main branch it gives 
```
"missing field `data`",
```
so clearly the PR is an improvement!

I'm sure I'm doing something wrong here, but I'd like to be able to know how it's supposed to work!

(Also, can you run `cargo fmt` on your branch and push it?)

```rust
use leptos::*;
use leptos_meta::*;
use leptos_router::*;
use serde::{Serialize, Deserialize};

#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct Address {
    door: String,
    street: Option<String>
}

#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct Data {
    id: u32,
    address: Address
}

#[server(EchoData, "/api")]
pub async fn echo_data(data: Data) -> Result<Data, ServerFnError> {
    Ok(data)
}

#[component]
pub fn HomePage(cx: Scope) -> impl IntoView {
    let action = create_server_action::<EchoData>(cx);

    let form_data = "data%5Bid%5D=3&data%5Baddress%5Bdoor%5D%5D=16&data%5Baddress%5Bstreet%5D%5D=Eden";
    let method1 = serde_qs::from_str::<EchoData>(form_data);
    let method2 = serde_qs::Config::default().deserialize_str::<EchoData>(form_data);

    view! { 
        cx,
        <pre>{format!("{method1:?}")}</pre>
        <pre>{format!("{method2:?}")}</pre>
        <pre>
            {move || format!("{:#?}", action.value().get())}
        </pre>
        <ActionForm action>
            <input type="number" name="data[id]"/>
            <input type="text" name="data[address[door]]"/>
            <input type="text" name="data[address[street]]"/>
            <input type="submit"/>
        </ActionForm>
    }
}

#[component]
  provide_meta_context(cx);
  
  view! {
    cx,
    <Router>
        <Routes>
            <Route path="" view=HomePage/>
        </Routes>
    </Router>
  }
}
```

gbj:
> ```
> <input type="number" name="data[id]"/>
> <input type="text" name="data[address][door]"/>
> <input type="text" name="data[address][street]"/>
> <input type="submit"/>
> ```

👏 Beautiful, thank you. Can confirm this syntax is broken on main and correct in the PR, without breaking the other, non-nested examples, so I think this is good to go. Thanks for finding it an fixing it, and now I have a good example for the docs. Much appreciated.

PR #1185, opened 2023-06-14
Chore: l021 was outdated version 0.2.1 becomes 0.3.0.

martinfrances107:


gbj:
`l021` is here in `dev-dependencies` specifically in order to benchmark the old reactive system (in Leptos 0.2.1) against the new reactive system (whatever the current version is) and others in `leptos_reactive/benches`. Whether any of these should still be here is another question, given that I think I've migrated them all over to using Criterion in the root `benchmarks` directory instead.

martinfrances107:
> in order to benchmark the old reactive system (in Leptos 0.2.1)

On the roadmap issue - I have just raised the spectre that "version 0.3  needs  either a hot fix,  or to be marked as outdated - do not use".

I am not familiar enough with  0.2 to comment but the chances of such problems only grows with time?

Do you have a grasp on the 0.2.1 user base? That always complicates the discussion.

gbj:
> On the roadmap issue - I have just raised the spectre that "version 0.3  needs  either a hot fix,  or to be marked as outdated - do not use".
> 
> I am not familiar enough with  0.2 to comment but the chances of such problems only grows with time?
> 
> Do you have a grasp on the 0.2.1 user base? That always complicates the discussion.

I think there's some misunderstanding here. What I'm saying is that `l021` here is an alias, used only as a dev dependency, so that the performance of the current reactive system can be benchmarked (by me, or anyone else interested) against the previous version. I do not expect that anyone is currently using 0.2.1, and I can't imagine why anyone would other than simply not having had time to update an old project. 

The intention of including this benchmark was to make sure we're not introducing performance regressions relative to the previous version by having benchmarks against it.

martinfrances107:
Sorry for the confusion 

I guess I analyse the situation very differently 

So running my mental decision making process in strict order of importance.  A then B then C

A) Is 2.01 known to be insecure/unmaintained?

B)  Does it have hot fix support for emergent issues.

C) Is their a performance known performance regression..

What I was trying to question was "Is asking C even appropriate?" 

Should we insert a friendly note in the README saying "Don't panic - just don't use 0.2 anymore"?

gbj:
This conversation seems to me to be collapsing a few separate topics:
1. **Are issues in old minor of Leptos addressed with new patch releases?** No. My assumption would be that previous versions of small FOSS library are not maintained independently of the library itself, i.e., there's no "long-term support" for Leptos 0.2 that involves releases patches for bugs or issues in Leptos 0.2. I can't speak for all library maintainers; speaking for myself, I simply wouldn't have the time or capacity to do this. (The 0.3.1/ouroboros situation was unusual in that 0.3.0 was the latest version so needs to be fixed; if 0.4.0 had already been released, I would not have gone back and released a 0.3.1.)

2. **Should a new version of 0.3.x be released that replaces `ouroboros`?** Yes, and this has been done. 

3. **Should `l021` in this benchmark setting be updated to `0.3.x`?** No, in that case we'd be benching the current reactive system against itself, which wouldn't be helpful. 

Is your concern at this point about 1, 3, both, or something else?

martinfrances107:
Sorry for the noise .. lets keep things as they stand.

PR #1190, opened 2023-06-15
Removed clippy warnings - merged "if"s

martinfrances107:
The change in indentation makes the PR hard to review

so I will discuss the change in conversational language

Two "if"'s checks were merged into one "if"

this

-        if let Some(expr) = node.value() {
-            if let syn::Expr::Tuple(tuple) = expr {

becomes

+        if let Some(Tuple(tuple)) = node.value() {


gbj:
Excellent, thanks!

PR #1196, opened 2023-06-15
Reorder `<For />` `apply_diff` steps

jquesada2016:
This PR fixes more edge cases with `apply_diff` by reordering steps and simplifying `unpack_moves`.


gbj:
My only request — could you let me know the edge cases, or add them to `test-bench`, so we continue to have a good set of cases to make sure there are no regressions? Obviously with the additional PRs we hadn't thought up all the possible issues, so adding them as we discover them is good for regression testing in the future.

gbj:
Excellent! This is exactly what I wanted.

jquesada2016:
I'll go ahead and change `<Test /> to take an iterator of vecs so we can test all the cases more thoroughly.

Issue #1209, opened 2023-06-18
ErrorBoundary doesn't work with SSR on initial request


M1cha:
**Describe the bug**
When you have an ErrorBoundary and a resource like in ssr_modes or ssr_modes_axum and the resource returns an error while being rendered on the server, you just get an empty view.

**Leptos Dependencies**
main branch at 3c6748b30d8f3bb7fa606f57ebdb9e514fc4b2cc

**To Reproduce**
1. Edit `examples/ssr_modes_axum/src/app.rs`:
```patch
diff --git a/examples/ssr_modes/src/app.rs b/examples/ssr_modes/src/app.rs
index f85d431e..8711eed6 100644
--- a/examples/ssr_modes/src/app.rs
+++ b/examples/ssr_modes/src/app.rs
@@ -168,13 +168,7 @@ pub struct PostMetadata {
 #[server(ListPostMetadata, "/api")]
 pub async fn list_post_metadata() -> Result<Vec<PostMetadata>, ServerFnError> {
     tokio::time::sleep(std::time::Duration::from_secs(1)).await;
-    Ok(POSTS
-        .iter()
-        .map(|data| PostMetadata {
-            id: data.id,
-            title: data.title.clone(),
-        })
-        .collect())
+    Err(ServerFnError::ServerError("well, well, ...".to_string()))
 }
```

2. Build and run using `cargo leptos watch`. 
3. Load the site. You'll see `Loading posts..` for 1s and then nothing.

**Expected behavior**
Render the fallback view of the ErrorBoundary after 1s.

**Additional context**
- It seems to work correctly when the client renders the view. Either via a local resource or because the page gets reloaded by the client(through the Router).
- This seems to be the case for both axum and actix.

gbj:
I'm a bit confused — There's no `<ErrorBoundary/>` on the home page, where `list_post_metadata` is read, so there's nothing to catch the `Err`.

https://github.com/leptos-rs/leptos/blob/3a570dc0d9d9f8856f97e8b77d655b567801356f/examples/ssr_modes_axum/src/app.rs#L47-L62

There *is* an `<ErrorBoundary/>` on the post page, and it does correctly catch the `Err`.

Is this a case of a confusing example/misreading across the two components, or am I missing something?

M1cha:
I have the issue in a different project and tried to provide a minimal example, but failed to do it correctly :P 
[...]

PR #1210, opened 2023-06-19
docs: document `<ErrorBoundary/>`/`<Suspense/>` relationship

gbj:
See issue #1209: hydration errors when `<ErrorBoundary/>` is outside `<Suspense/>` vs. inside `<Suspense/>`. This does not fix the issue (which is hard) but adds a note in the docs, and makes sure examples are consistent with it.

### DIFF
--- a/leptos/src/error_boundary.rs
+++ b/leptos/src/error_boundary.rs
@@ -28,6 +28,24 @@ use leptos_reactive::{
 /// }
 /// # });
 /// ```
+/// 
+/// ## Interaction with `<Suspense/>`
+/// If you use this with a `<Suspense/>` or `<Transition/>` component, note that the 
+/// `<ErrorBoundary/>` should go inside the `<Suspense/>`, not the other way around,
+/// if there’s a chance that the `<ErrorBoundary/>` will begin in the error state. 
+/// This is a limitation of the current design of the two components and the way they 
+/// hydrate. Placing the `<ErrorBoundary/>` outside the `<Suspense/>` means that 
+/// it is rendered on the server without any knowledge of the suspended view, so it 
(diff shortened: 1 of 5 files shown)


gbj:
> Would it be possible to implement either a compile-time check or a runtime warning for this?

Good idea. I was able to add a runtime warning for at least the simplest case (ErrorBoundary with Suspense or Transition as direct child.) I could do this for a more general case (any Suspense under any ErrorBoundary) but I'm worried it would catch too many false positives. I've found that people love warnings that help them fix a problem themselves, and hate warnings they consider false positives or want to be able to disable, so maybe this is a happy medium.

In theory it would be awesome to have this as a compile error but doing fully-typed component trees is beyond us for now.

leptos/src/error_boundary.rs
```diff
@@ -64,7 +64,25 @@ where
...
     // Run children so that they render and execute resources
-    let children = children(cx).into_view(cx);
+    let children = children(cx);
+
+    #[cfg(all(debug_assertions, feature = "hydrate"))]
+    {
+        use leptos_dom::View;
+        if children.nodes.iter().any(|child| {
+            matches!(child, View::Suspense(_, _))
+            || matches!(child, View::Component(repr) if repr.name() == "Transition")
```
M1cha, line 74: would  this produce a false positive when the user names a custom component `Transition`?

gbj: Yes. At that point, I honestly don't know what to say.

M1cha: could we use downcast_ref like already done in other places within leptos?

M1cha: I agree

Issue #1212, opened 2023-06-19
Panic calling with_untracked on memo


lpotthast:
Using Leptos 0.3.1, when calling `with_untrackd` on a Memo, I get a runtime-downcast error. Calling `get_untracked` instead works fine, but needlessly copies my data.

A simple 
```
    let memo = create_memo(cx, move |_| 42);
    let copied_out = memo.with_untracked(|val| *val);
```
reproduces it for me.

It seems that `get_untracked` and `with_untracked` call `try_with_no_subscription` with different types for ´T´.

See:
```rust
impl<T: Clone> SignalGetUntracked<T> for Memo<T> {
    fn get_untracked(&self) -> T {
        with_runtime(self.runtime, move |runtime| {
            let f = move |maybe_value: &Option<T>| maybe_value.clone().unwrap();
            match self.id.try_with_no_subscription(runtime, f) {
                Ok(t) => t,
                Err(_) => panic_getting_dead_memo(
                    #[cfg(any(debug_assertions, feature = "ssr"))]
                    self.defined_at,
                ),
            }
        })
        .expect("runtime to be alive")
    }
```
```rust
impl<T> SignalWithUntracked<T> for Memo<T> {
    fn with_untracked<O>(&self, f: impl FnOnce(&T) -> O) -> O {
        // Unwrapping here is fine for the same reasons as <Memo as
        // UntrackedSignal>::get_untracked
        with_runtime(self.runtime, |runtime| {
            match self.id.try_with_no_subscription(runtime, |v: &T| f(v)) {
                Ok(t) => t,
                Err(_) => panic_getting_dead_memo(
                    #[cfg(any(debug_assertions, feature = "ssr"))]
                    self.defined_at,
                ),
            }
        })
        .expect("runtime to be alive")
    }
```

While `get_untracked` uses `&Option<T>`(whose unwrapping could probably be `.expect("invariant: must have already been initialized") instead`), `with_untracked` uses `&T`, leading to the downcast inside the `Signal` implementation,
```rust
let value = value
            .downcast_ref::<T>()
            .ok_or_else(|| SignalError::Type(std::any::type_name::<T>()))
            .expect("to downcast signal type");
```
, failing.

It seems that in memo.rs:248, we could just replace this
```rust
match self.id.try_with_no_subscription(runtime, |v: &T| f(v)) {
```
with this?
```rust
[...]

PR #1213, opened 2023-06-19
fix: memo with_untracked

lpotthast:
Closes https://github.com/leptos-rs/leptos/issues/1212


lpotthast:
- Should we add a comment at the top of the file stating that the value is internally stored as `Option<T>`?
- Saw that the reference forwarding happened in two places and created a small function creating the correct closure. Can inline that if you want.

gbj:
> * Should we add a comment at the top of the file stating that the value is internally stored as `Option<T>`?

Great idea! Always good to have things like this live in comments instead of only my memory, so I added an explanation.

> * Saw that the reference forwarding happened in two places and created a small function creating the correct closure. Can inline that if you want.

Lovely, thanks.

This all looks good and much appreciated. Thank you for taking the time.

PR #1219, opened 2023-06-21
tailwind_csr_trunk has maintenance issues

martinfrances107:
It will eventually fail and become buggy.

examples/hackernews_axum pulls in leptos submodules using this pattern

```
leptos_axum = { path = "../../integrations/axum", optional = true }
leptos_meta = { path = "../../meta", default-features = false }
leptos_router = { path = "../../router", default-features = false }
```

Happiness, as the leptos version transitions from 3, 4, 5 the examples track the current version..

This PR just updates tailwind_csr_trunk to follow that convention.

Previously tailwind_csr_trunk, was stuck at version 2, it has dropped behind already ( current  leptos is 3.0.1 ).

### DIFF
--- a/examples/tailwind_csr_trunk/Cargo.toml
+++ b/examples/tailwind_csr_trunk/Cargo.toml
@@ -4,18 +4,14 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-leptos = { version = "0.2", features = [
-  "serde",
-  "csr",
-] }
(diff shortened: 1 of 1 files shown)


gbj:
Thanks. This may be the case with other examples that people have contributed, in which I overlooked the way dependencies were defined in `Cargo.toml`, something I'll have to be more careful about in the future. (In fact, I just fixed it in `leptos-tailwind-axum` this morning.)

PR #1227, opened 2023-06-23
change: migrate to `nightly` and `csr` features rather than `stable` and `default-features = false`

### LINKED ISSUE #567: Use a `nightly` feature flag instead of `stable`
We had some good discussion on the leptos discord about the current `stable` feature flag. I'll try to reproduce the conversation here.

# Problem
Let's say you have this dependency graph:
```
my-crate
 |
 |--- foo
 |     |---- leptos (features = ["stable"])
 |
 |--- bar
       |---- leptos
```

Because features are additive and `foo` used `stable`, the whole build tree now builds as `stable`. If `bar` used nighly features in it's code, now the build fails. 

The key is in this line of the cargo book: 
> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features

Currently this is not the case since enabling the `stable` feature flag actually disables functionality. 

# Possible Solution
Remove `stable` and introduce a feature flag called `nightly` that is on by default and switch all conditional compilation from `#[cfg(not(feature = "stable"))]` to `#[cfg(feature = "nightly")]`

This is a breaking change since everybody that either uses `stable` or has disabled default features has to change their `Cargo.toml` in order to keep using nightly features.

gbj:
Closes #567.

Ideally, I would have done this from 0.0.1, but I was younger and more foolish then :smile:. 

Ideally, Cargo features should be additive, not subtractive or mutually exclusive, i.e., each feature should enable additional functionality in the crate.

We do use a few mutually-exclusive features (`csr`, `ssr`, and `hydrate`). This is because they enable different behavior or enable certain optimizations for the different platforms, in a better way than the alternatives. (For example, using `target_arch = wasm32` to detect the browser would fail to support the growing number of server-side WASM environments; using a generic View<T> that could be either a DOM element or an SSR version wouldn't let us detect SSR mode for compile-time SSR optimizations; etc.)

However, the `stable` feature was just a bad idea. The crate was originally nightly-only, and I added the `stable` feature when asked to support stable. This causes some downstream issues for libraries due to feature unification.

This PR reverses the current setup so that `nightly` is now an opt-in feature, which basically enables the function-call syntax for signal getters and setters. 

While we're breaking things, it also removes the default `csr` feature from the `leptos` crate, which now works the same way as `leptos_router` and `leptos_meta`: you should opt into `csr`, `ssr`, or `hydrate`.

This means that `leptos = "0.x"` should now be something like `leptos = { version = "0.x", features = ["csr", "nightly"] }` It also means all the `default-features = false` can be dropped from `Cargo.toml` for setups using SSR + hydration, as Leptos will no longer default to CSR.

Todos
- [x] Finish checking examples
- [x] Update docs


gbj:
Migration:
- If you're using CSR, add the `csr` feature to `leptos` (same as `leptos_router` and `leptos_meta`
- If you're using `stable`, drop the `stable` feature -- it's now the default
- If you're using `nightly`, add the `nightly` feature to `leptos`, `leptos_router`, and `leptos_meta`
- You can remove most `default-features = false` on Leptos dependencies, as that mainly served to disable the default `csr` feature when you were using `ssr`/`hydrate`, which is no longer necessary

PR #1243, opened 2023-06-28
fix(counters_stable): intermittent closed target errors (#1240)

### LINKED ISSUE #1240: test(counters_stable): intermittent closed target errors
**Describe the bug**
Intermittent `Error: locator.click: Target closed` errors result in flaky tests.

While the tests eventually pass with retries, it would be better to eliminate the flaky behavior, if possible.

**Leptos Dependencies**

N/A

**To Reproduce**

1. Run `cargo make test-e2e` (Notice the number of flaky tests)

**Expected behavior**
It would be best if there were no flaky tests.

**Screenshots**

N/A

**Additional context**

The **Add 1000 Counters** test can be made to fail consistently by adding assertions between calls to `ui.addOneThousandCounters()`.

```js
test.describe("Add 1000 Counters", () => {
  test("should increment the total count by 1K", async ({ page }) => {
    const ui = new CountersPage(page);
    await ui.goto();

    // FIXME: Uncomment to make the test fail consistently

    await ui.addOneThousandCounters();
    // await expect(ui.total).toHaveText("0");
    // await expect(ui.counters).toHaveText("1000");

    await ui.addOneThousandCounters();
    // await expect(ui.total).toHaveText("0");
    // await expect(ui.counters).toHaveText("2000");

    await ui.addOneThousandCounters();
    await expect(ui.total).toHaveText("0");
    await expect(ui.counters).toHaveText("3000");
  });
});
```

agilarity:
READY FOR REVIEW

Fix #1240

The solution is to wait before clicking a button.

**Two Strategies**

1. Wait before each click attempt (used by most buttons)
2. Synchronize navigation and the wait for a button (used only by Add 1000 Counters button)

All buttons except the **Add 1000 Counters** button use strategy **1**. It is not clear why this is a requirement. Switching strategies breaks the test!?

### DIFF
--- a/examples/counters_stable/e2e/playwright.config.ts
+++ b/examples/counters_stable/e2e/playwright.config.ts
@@ -14,11 +14,11 @@ export default defineConfig({
   /* Run tests in files in parallel */
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
-  forbidOnly: !!process.env.CI,
+  forbidOnly: !process.env.DEV,
   /* Retry on CI only */
-  retries: process.env.CI ? 10 : 10,
+  retries: process.env.DEV ? 0 : 10,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 1,
+  workers: process.env.DEV ? 1 : 1,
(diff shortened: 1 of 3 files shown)


gbj:
Very odd! But if it works it works. This is not my area of expertise.

PR #1249, opened 2023-06-29
Add fallback support for workspace in get_config_from_str

afiqzx:
Current `get_config_from_str` and `get_configuration` will return an error when used in workspace environment.

This PR attempts to add support for root Cargo.toml in workspace environment.
Since it will always search for `[package.metadata.leptos]` first it shouldn't break existing examples (I have not really check each one of them).

Feel free to close this PR if you think there's a better alternative than just a slapping a fallback.


gbj:
Thanks — this looks good to me!

PR #1259, opened 2023-07-01
Minor: Ran cargo clippy --fix and reviewed changes.

martinfrances107:
A few .clone() calls  removed.


gbj (commented):
~~formatting~~

Oops that was supposed to be a commit message haha.

PR #1262, opened 2023-07-01
Added watch

### LINKED ISSUE #1177: Add on(...) / watch(...)
SolidJS has 

```js
createEffect(
  on(a, (value, prevValue) => { ... })
);
```

and Vue has

```js
watch(a, (value, prevValue)  => { ... })
```

SolidJS-USE actually provides a `watch` that is just a shorthand for `createEffect(on(...))`.

I kind of hacked a `watch` together in leptos-use but wouldn't it make sense to have sth like this in core leptos? IMO ideally like the `watch` from Vue where the returned function removes the watcher.

If you agree I'd like to try my hand at this.

maccesch:
Fixes #1177

Should I also add sth to the book?

### DIFF
--- a/leptos_reactive/src/runtime.rs
+++ b/leptos_reactive/src/runtime.rs
@@ -1,11 +1,12 @@
 #![forbid(unsafe_code)]
+
 use crate::{
     hydration::SharedContext,
     node::{NodeId, ReactiveNode, ReactiveNodeState, ReactiveNodeType},
     AnyComputation, AnyResource, Effect, Memo, MemoState, ReadSignal,
     ResourceId, ResourceState, RwSignal, Scope, ScopeDisposer, ScopeId,
-    ScopeProperty, SerializableResource, StoredValueId, Trigger,
-    UnserializableResource, WriteSignal,
+    ScopeProperty, SerializableResource, SpecialNonReactiveZone, StoredValueId,
+    Trigger, UnserializableResource, WriteSignal,
 };
 use cfg_if::cfg_if;
 use core::hash::BuildHasherDefault;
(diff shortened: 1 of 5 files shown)


leptos_reactive/src/runtime.rs
```diff
@@ -681,6 +712,81 @@ impl RuntimeId {
...
+                let callback_value =
+                    Some(self.untrack(wrapped_callback.clone()));
+
+                prev_callback_value.replace(callback_value);
+
+                prev_deps_value.replace(Some(deps_value));
+            }
+        };
+
+        let id = self.create_concrete_effect(
```
maccesch, line 769: I did reuse `Effect` here because it really is an enhanced version of that

leptos_reactive/src/scope.rs
```diff
@@ -349,6 +318,27 @@ impl Scope {
...
         instrument(level = "trace", skip_all,)
     )]
+    pub(crate) fn remove_scope_property(&self, prop: ScopeProperty) {
+        _ = with_runtime(self.runtime, |runtime| {
+            let scopes = runtime.scopes.borrow();
+
+            if let Some(scope) = scopes.get(self.id) {
+                let mut scope = scope.borrow_mut();
+                if let Some(index) = scope.iter().position(|p| p == &prop) {
+                    scope.swap_remove(index);
```
maccesch, line 328: This assumes that the order of scope properties doesn't matter.

gbj:
This looks good to me! I don't think it needs to be added to the book for now, I will do an "advanced reactivity" chapter or something at some point.

PR #1275, opened 2023-07-03
leptos axum extract with state

sjud:
Pr following up from discord discussion.


gbj:
Thanks — This looks great to me. I think this is a good solution, at least for now!

PR #1278, opened 2023-07-04
test(counters_stable): add wasm testing

agilarity:
READY FOR REVIEW

Supports #210

Add wasm browser tests for the `counters_stable` example.

This PR covers 80% of e2e test scenarios.

### Covered Scenarios (8/10)

- [x] Add 1000 Counters › should increase the number of counters
- [x] Add Counter › should increase the number of counters
- [x] Clear Counters › should reset the counts
- [x] Decrement Counter › should decrease the total count
- [x] Increment Counter › should increase the total count
- [x] Remove Counter › should decrement the number of counters
- [x] View Counters › should see the title
- [x] View Counters › should see the initial counts

### Scenarios Not Covered (2/10)

I did not cover the following scenarios because I do not know how to trigger the input box change from a wasm test.

- Enter Count › should increase the total count
- Enter Count › should decrease the total count

The following test code, gleaned from two Sycamore tests, correctly changed the input value but failed to trigger the change to the total count.

```rust
  let input = ...
  input.set_value(count.to_string().as_str());
  let event = Event::new("input").unwrap();
  input.dispatch_event(&event).unwrap();
```
Note, the `Event` and `EventTarget` features were enabled.


gbj:
This is awesome!

> The following test code, gleaned from two Sycamore tests, correctly changed the input value but failed to trigger the change to the total count.
> 
> ```rust
>   let input = ...
>   input.set_value(count.to_string().as_str());
>   let event = Event::new("input").unwrap();
>   input.dispatch_event(&event).unwrap();
> ```
> 
> Note, the `Event` and `EventTarget` features were enabled.

Interesting. Off the top of my head, I can imagine two scenarios:
1) This has to do with the way we handle event delegation clashing with the attempt to manually dispatch an event. The way to test this would be to change the handler from `on:input` to `on:input:undelegated` and see if it works then.
2) It's a browser security thing. I think there are some events the browser prevents you from manually firing to prevent spoofing user input. I have no idea if this would be one of them.

agilarity:
> 1. This has to do with the way we handle event delegation clashing with the attempt to manually dispatch an event. The way to test this would be to change the handler from `on:input` to `on:input:undelegated` and see if it works then.

This resolved the problem.

PR #1307 adds the missing Enter Count scenarios.

Issue #1263, opened 2023-07-02
no method named `line` found for struct `proc_macro::LineColumn` in the current scope


leefordjudes:
**Describe the bug**
This is my first try:
i installed following, 
cargo install --locked trunk	(version 0.17.1)
cargo install --locked wasm-bindgen-cli  (version 0.2.87)
cargo install --locked cargo-leptos (version 0.1.11)

rustup toolchain install nightly
rustup default nightly
rustup target add wasm32-unknown-unknown

and create leptos project as like steps given in book url: `https://leptos-rs.github.io/leptos/02_getting_started.html`

and run using `trunk serve`,
i got error, 

![image](https://github.com/leptos-rs/leptos/assets/13608248/bf5f065c-28c6-40e0-b2ff-c5ccaa23bdca)

![image](https://github.com/leptos-rs/leptos/assets/13608248/711ad6eb-17a0-48a7-80f4-00caae644718)

![image](https://github.com/leptos-rs/leptos/assets/13608248/b449a86b-6888-4828-9bbd-db749c47c70c)

![image](https://github.com/leptos-rs/leptos/assets/13608248/92e96393-fe24-4c06-9575-2ee9ccbfac4a)

also given:
`cargo update`

**Leptos Dependencies**

[package]
name = "leptos-tutorial"
version = "0.1.0"
edition = "2021"

[dependencies]
leptos = { version = "0.4.1", features = ["csr", "nightly"] }

**To Reproduce**
simply create new leptos project like i mentioned above

**Expected behavior**
should run

nguiard:
Hi! I just ran into exactly this on linux a few hours ago while trying to reproduce another issue... Right now to be honest I have no idea what causes it, but I managed to get around it with a "cleanup", and it's no longer happening...

I suggest you try:
- cargo clean
- rustup update
- removing the `Cargo.lock` file
- (if you use `sccache`, also clear the cache (typically by removing the content of the sccache cache dir)

Maybe you can try a `trunk serve` between each of these steps, to identify which steps solves it (it seems that some combination of this solved it for me but have no idea what step(s) exactly, and anyway it would be nice to pinpoint it so that it doesn't happen to other people)

luoxiaozero:
might to access the line field instead of the method
https://doc.rust-lang.org/proc_macro/struct.LineColumn.html

gbj:
What is your `rustc --version`?

[...]

PR #1290, opened 2023-07-05
Fix over-eager parens

bugeats:
Hi I don't know anything about anything, but this fixes builds on my system.


gbj:
It looks like this is trying to address #1263 on your own machine. What you actually need to do is `rustup update nightly && cargo update` and you'll be fine. This unstable API changed slightly on a recent nightly, so what we have updated to the newer syntax.

PR #1292, opened 2023-07-06
fixed example for error handling

webmstk:
type="number" prevents user from typing letters (at least in Chrome)

Nice lib!


gbj:
Thanks for taking the time to make a PR. I mostly use Firefox so hadn't noticed this behavior in Chrome. I notice that it's still useful for catching empty strings, but definitely confusing for someone who tries to type some letters and has the browser prevent them!

PR #1294, opened 2023-07-06
Seanaye/feat/server wasm

seanaye:
Draft PR (still very much WIP) just to show progress of serverside WASM integration.

Goals:
* Support all SSR modes with streaming and server actions
* Target deno deploy and cloudflare workers


seanaye:
Thanks! will do. I realized I should probably implement `Service` from tower instead of creating my own handler trait. Haven't had much time to work on this recently but I will get there eventually!

Issue #1274, opened 2023-07-03
`<For />`:  when removing elements, getting "panicked at 'called `Option::unwrap()` on a `None` value'" at each.rs:853, in some conditions


nguiard:
**Describe the bug**

When clearing multiple elements of a Vec used in a `<For />`, for some combinations of removed elements, I get 
```
start-axum.js:569 panicked at 'called `Option::unwrap()` on a `None` value', [...]/leptos_dom/src/components/each.rs:853:57
```

I have a hard time producing a simple reproduction for now, working on it... But as it seems to point to a specific line in leptos, I thought you might have an idea.

In my app (still trying to make a nice todo app), the problem happens only when removing some specific subsets of elements. 
- with 3 elements in the list, I'm unable to produce this message
- with 4 elements in the list however (indexes [0, 1, 2, 3]), the error happens only when removing elements [0, 2] (at the same time, using `Vec::retain`)
- with 5 elements in the list, it seems to happens with a lot of combinations, apparently *all* of these involving removing the first element (index 0) *and not* removing the second (index 1). But also not in all of the situations involving these conditions produce the bug... (removing indices [0, 2, 3, 4] works fine for example, but not removing [0, 2, 4])

As I said I have a hard time producing a minimal example right now, but I'll work on it when I can...
Also, I *can* work around this issue by looping over the elements in the vec, and calling `update` sequentially for each one I want to remove, removing only this one (but yeah, this ain't very pretty nor declarative, and probably not performant at all)...

In any case, thanks for the great work on leptos, loving it :)

Edit: sorry, I forgot to mention, I'm on the latest commit at the time of writing (66f54e7)

godstail:
I also had this panic, but in my case it was when one of the item's key wasn't stable and got changed with each rerender.
Slightly different line number in my case, 849 (prob. different version):
```let each_item = children[move_.from].take().unwrap();```

**Comment by nguiard** [comment]
Hmm interesting... In my case, as far as I can tell, the items' keys are actually stable...

**Comment by gbj** [comment]
@jquesada2016 This is caused by a situation in which the diff causes multiple moves. 

Going from `[0, 1, 2, 3]` to `[1, 3]` yields the diff
```
Diff {
    removed: [
        DiffOpRemove {
            at: 0,
        },
        DiffOpRemove {
            at: 2,
        },
    ],
    moved: [
        DiffOpMove {
            from: 1,
            len: 2,
            to: 0,
            move_in_dom: false,
        },
    ],
[...]

Issue #1289, opened 2023-07-05
Rendered content of `<For/>` stops updating when underlying signal changes


LoganWalls:
**Describe the bug**
I am trying to render a dynamic list of elements using `<For/>`. In this example, I have a search box that searches over the names of countries, and a `<For/>` that renders each matched country as a button:

```rust
let all_items = create_resource(cx, || (), fetch_items);
let visible_items = ||{
  // ...
  filtered_copy_of_all_items
}

// ...
view!{cx, <For
    each=visible_items
    key=|item| item.id
    view=move |cx, item| {
      view! {
        cx,
        <button>{item.data.key}</button>
      }
    }
  />
}
```

The items in `all_items` are fixed (it's a resource that is fetched only once), and each `Item` has an id field that is a unique `usize` which never changes.

An example that I can easily reproduce is searching for "Canada". The contents of the `<For/>` update correctly for at least two updates. But at some point the `<For/>` stops updating. As you can see in this gif, even though I change the query to "United", the results do not change at all. However, if I make a side-effect that console logs `visible_items`, I can see that the signal itself is updating correctly. For example, in the gif you can see that after I change the query to "United", `visible_items` contains the UK and US as one would expect.

**Leptos Dependencies**
```toml
leptos = { version = "0.4", features = ["nightly", "csr"] }
```

**To Reproduce**
1. Clone the example repository [here](https://github.com/LoganWalls/leptos-mwe)
2. Change directories into the cloned repository
3. Run `trunk serve`, and open the url it prints in your browser
4. Type "Canada" into the search box
5. Delete "Canada" and try typing another query like "United"

**Expected behavior**
I would expect the items rendered in the `<For each=visible_items />` tag to match the items in the `visible_items` signal.

**Screenshots**
![leptos](https://github.com/leptos-rs/leptos/assets/2934282/9d25200f-f259-4488-9326-9354ee7fe0c3)

Thank you in advance for your help!

gbj:
Thanks! This definitely appears to be path-dependent — i.e., if I paste in "Canada" and then start deleting letters, it works, but if I type "Canada", and then start deleting, it sticks. It is likely related to the issue with moving multiple items in a single change in #1274.

For your particular use case, by the way, where it's only 10 rows and often completely removing all the rows — I almost wonder if a plain iterator might work better anyway. 
```rust
move || {
    visible_items()
        .into_iter()
        .map(|item| {
            view! { cx, <button>{item.data.key}</button> }
        })
[...]

PR #1296, opened 2023-07-06
Rework diff functionality for Each component

g-re-g:
Hi! While reading through the code behind this guy: https://github.com/leptos-rs/leptos/issues/1289 I noticed that the `diff` makes some assumptions / calculations that may not be correct. In the example provided, if you type `C` then `a` notice the bottom two items `Cocos` and `Christmas` are out of order. This is with `add`s and `remove`s alone, no `move`s necessary. If you check the console you can see the data in the correct order. This results in not only a visual inconsistency but a panic on what i think is trying to reach an incorrect index. I think this PR as it stands solves that issue, and potentially any others with panicing / proper arrangement, but is of course not optimized. I don't have the time to finish this right now but wanted to leave it up in case it helps anyone else working on it. I will have time to pick this back up tomorrow.

The approach this PR is taking is to build a canonical list of mutations from `from` to `to` that will result in the correct `to` and from there perhaps combine moves and remove no-ops. Those are my next steps personally.

Feel free to remove this or whatever if it's off base or if draft PRs aren't welcome / too much noise!!

### DIFF
--- a/leptos_dom/examples/test-bench/src/main.rs
+++ b/leptos_dom/examples/test-bench/src/main.rs
@@ -55,6 +55,8 @@ fn view_fn(cx: Scope) -> impl IntoView {
             <hr/>
             <Test from=[1, 4, 3, 2, 5] to=[1, 2, 3, 4, 5]/>
             <Test from=[4, 5, 3, 1, 2] to=[1, 2, 3, 4, 5]/>
+            <Test from=[0, 1, 2, 3] to=[1, 3]/> // issue #1274
+            <Test from=[] to=[3, 9, 17] then=vec![3, 5, 7, 9, 17, 23]/> // issue #1297
         </ul>
     }
 }
--- a/leptos_dom/src/components/each.rs
+++ b/leptos_dom/src/components/each.rs
@@ -17,7 +17,7 @@ mod web {
     pub use wasm_bindgen::JsCast;
 }
 
-#[cfg(all(target_arch = "wasm32", feature = "web"))]
+// #[cfg(all(target_arch = "wasm32", feature = "web"))]
 type FxIndexSet<T> =
     indexmap::IndexSet<T, std::hash::BuildHasherDefault<rustc_hash::FxHasher>>;
 
@@ -499,9 +499,10 @@ where
 #[educe(Debug)]
 struct HashRun<T>(#[educe(Debug(ignore))] T);
 
-/// Calculates the operations need to get from `a` to `b`.
-#[cfg(all(target_arch = "wasm32", feature = "web"))]
+/// Calculates the operations needed to get from `from` to `to`.
+// #[cfg(all(target_arch = "wasm32", feature = "web"))]
 fn diff<K: Eq + Hash>(from: &FxIndexSet<K>, to: &FxIndexSet<K>) -> Diff {
+    // Optimization: Fast path if either `to` or `from` is empty
     if from.is_empty() && to.is_empty() {
         return Diff::default();
     } else if to.is_empty() {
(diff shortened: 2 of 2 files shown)


gbj:
@g-re-g Thank you for your work, and for taking the time to make this contribution!

@jquesada2016 This PR, as is, fixes issues #1274, #1289, and #1297. I'll add tests for #1274 and #1297 to the test bench file in this branch. Could you take a look at this and see what you think?

gbj:
I have not actually looked at the implementation so far, having just focused on setting up tests and checking against issues in the "does it work?" category, but ran the js-framework-benchmark against it. As you can see this is a very significant performance regression on the swap/remove/append tests, i.e., everything where it diffs between two lists that are not empty.

Screenshot 2023-07-07 at 7 43 47 AM

gbj:
Okay, yeah, that's more comfortable. Of course, checking the removes/moves is the hard part 😄 

Screenshot 2023-07-07 at 8 29 20 AM

g-re-g:
Similarly, I'm looking for  a good way to write / run tests for `#[cfg(all(target_arch = "wasm32", feature = "web"))]` mode if you know of any.

gbj:
> @gbj how do I setup / run those perf tests? That would be awesome to check as I make some changes!

This is the [`js-framework-benchmark`](https://github.com/krausest/js-framework-benchmark) If you follow the instructions in that repo to get set up, you can run the benchmarks. All I did was copy-paste the `leptos` directory in `frameworks/keyed` to `leptos-1296`, and change the `Cargo.toml` to point to my local copy. 

Then you can `npm run bench --framework keyed/leptos-1296 && npm run bench --framework keyed/leptos && npm run results` to run the benchmarks and generate results. This basically takes over your laptop for like 10 minutes, I pretty much close all my other tabs.

gbj:
So, here's an alternative version of the `diff` function that:
1) Runs correctly in the face of the issues linked above
2) Performs equivalently to the current version in the js-framework-benchmark

It does *not* include any of the additional optimizations for moving ranges, etc.

Essentially it still creates a large number of "move" operations when you have something like "remove item at index 2 of 1000," which requires the next 998 to shift left — but it tests whether they need to be moved in the DOM or not. Copying the 998 items in memory ends up being very cheap relative to the cost of any DOM operations, which is why it still gets book benchmark performance.

It's also 12kb smaller in the `-04` optimized binary for the benchmark, which is respectable.

I'll re-run the benchmark just to confirm.

@g-re-g I'm copying and pasting below rather than committing it just to avoid merge conflicts, but I'm fairly sure this is the way to go.

@jquesada2016 I'm fine with building some of the range optimizations back in, but to me this looks like one answer to your question of how to proceed.

```rust
/// Calculates the operations need to get from `a` to `b`.
fn diff<K: Eq + Hash>(from: &FxIndexSet<K>, to: &FxIndexSet<K>) -> Diff {
    if from.is_empty() && to.is_empty() {
        return Diff::default();
    } else if to.is_empty() {
        return Diff {
            clear: true,
            ..Default::default()
        };
    } else if from.is_empty() {
        return Diff {
            added: to
                .iter()
                .enumerate()
                .map(|(at, _)| DiffOpAdd {
                    at,
                    mode: DiffOpAddMode::Append,
                })
                .collect(),
            ..Default::default()
        };
    } else {
        let mut removes = vec![];
        let mut moves = vec![];
        let mut adds = vec![];
        let max_len = std::cmp::max(from.len(), to.len());

        for index in (0..max_len) {
            let from_item = from.get_index(index);
            let to_item = to.get_index(index);

            // if they're the same, do nothing
            if from_item != to_item {
                // if it's only in old, not new, remove it
                if from_item.is_some() && !to.contains(from_item.unwrap()) {
                    let op = DiffOpRemove { at: index };
                    removes.push(op);
                }
                // if it's only in new, not old, add it
                if to_item.is_some() && !from.contains(to_item.unwrap()) {
                    let op = DiffOpAdd {
                        at: index,
                        mode: DiffOpAddMode::Normal,
                    };
                    adds.push(op);
                }
                // if it's in both old and new, it can either
                // 1) be moved (and need to move in the DOM)
                // 2) be moved (but not need to move in the DOM)
                //    * this would happen if, for example, 2 items
                //      have been added before it, and it has moved by 2
                if let Some(from_item) = from_item {
                    if let Some(to_item) = to.get_full(from_item) {
                        let moves_forward_by =
                            (to_item.0 as i32) - (index as i32);
                        let move_in_dom = moves_forward_by
                            != (adds.len() as i32) - (removes.len() as i32);
                        let op = DiffOpMove {
                            from: index,
                            len: 1,
                            to: to_item.0,
                            move_in_dom,
                        };
                        moves.push(op);
                    }
                }
            }
        }

        let mut diff = Diff {
            removed: removes,
            items_to_move: moves.len(),
            moved: moves,
            added: adds,
            clear: false,
        };

        apply_opts(from, to, &mut diff);

        diff
    }
}
```

jquesada2016:
@gbj yes! That's exactly what I was going to start doing yesterday, but didn't get around to starting.

There are essentially 2 diffs we need to keep track of:

1. the in-memory ordering on `children`
2. the DOM ordering

When I re-wrote this algorithm, I frankly was optimizing for the second DOM case, and neglected the first one, this is where we're getting the inconsistancies and glitches. What you're doing is totally correct. If you have 1,000 items, and you add a new item to index 1, you necessarily need to shift 999 elements to the right in order to keep things consistant. We then need to apply the optimiztions on top of this first DIFF to optimize DOM movements, which is the really expensive part.

gbj:
Oops looks like `optimize_moves` in `apply_opts` actually broke "Swap Rows" there by shifting one of the two move operations to be `move_in_dom: false`, but not actually swapping the two? I'm not sure quite what's going on. But here's the benchmark result if I remove `apply_opts`. i.e., identical performance to the present, but correct moves.

I'm not sure how best to re-enable the various optimizations, none of which apply to this benchmark, but I'd be inclined to actually merge my version for now as it closes those issues, then add some of the optimizations back in when you have time and make sure there aren't regressions. 

Any further thoughts? @g-re-g thanks for localizing this to `diff` and inspiring me to work on it some more.

Screenshot 2023-07-07 at 11 06 28 AM

jquesada2016:
@gbj go ahead and disable them, almost none of them apply to the benchmark, as you said. They are also incompatible with the new aproach, so I'll need to adapt them back in later, which we can do in later PR's :)

g-re-g:
@gbj i think the only thing left is to merge groups of `move`s together so items that move as a group are in one move command. Happy to proceed with this PR however works for y'all. Feel free to ask me to do it, do it here in this PR, close this PR and open another, whatever works best!

gbj:
> @gbj i think the only thing left is to merge groups of `move`s together so items that move as a group are in one move command. Happy to proceed with this PR however works for y'all. Feel free to ask me to do it, do it here in this PR, close this PR and open another, whatever works best!

If you could add merging groups of moves together that would be awesome!

g-re-g:
@gbj added `group_adjacent_moves` and a couple tests for it. Gonna mark this as "Ready for review" when the tests pass. Lemme know if there's anything else / more tests you want me to put in here!

gbj:
This looks awesome, and examples pass with flying colors. I'll let the CI run of course, then merge. Thanks so much!

PR #1307, opened 2023-07-07
test(counters_stable/wasm): enter count

agilarity:
READY FOR REVIEW

Supports #210

Add wasm tests for Enter Count feature.

This PR covers 100% of e2e test scenarios.

### E2E Scenarios

- [x] Add 1000 Counters › should increase the number of counters
- [x] Add Counter › should increase the number of counters
- [x] Clear Counters › should reset the counts
- [x] Decrement Counter › should decrease the total count
- [x] Enter Count › should increase the total count
- [x] Enter Count › should decrease the total count
- [x] Increment Counter › should increase the total count
- [x] Remove Counter › should decrement the number of counters
- [x] View Counters › should see the title
- [x] View Counters › should see the initial counts

_Thank you @gbj. The handler change from `on:input` to `on:input:undelegated` enabled the input to work as expected._

### DIFF
--- a/examples/counters_stable/Cargo.toml
+++ b/examples/counters_stable/Cargo.toml
@@ -18,6 +18,7 @@ pretty_assertions = "1.3.0"
 [dev-dependencies.web-sys]
 features = [
     "Event",
+    "EventInit",
     "EventTarget",
     "HtmlElement",
     "HtmlInputElement",
--- a/examples/counters_stable/Cargo.toml
+++ b/examples/counters_stable/Cargo.toml
@@ -16,5 +16,11 @@ wasm-bindgen-test = "0.3.37"
 pretty_assertions = "1.3.0"
 
 [dev-dependencies.web-sys]
-features = ["Element", "HtmlElement", "XPathResult"]
+features = [
+    "Event",
+    "EventTarget",
+    "HtmlElement",
+    "HtmlInputElement",
+    "XPathResult",
+]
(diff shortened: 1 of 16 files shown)


gbj:
Thank you, as always, for your wonderful work!

Apologies for pushing directly into your PR but I thought it would be easier to show than to tell.

It looks like you were able to confirm my hypothesis that the issue with dispatching the synthetic `input` event had to do with the event delegation. I thought it would be ideal to solve this issue so that there was an example of how to test events like this without needing to change the example code itself. Luckily, this turned out to be fairly simple.

Most native DOM events (including `input`) bubble by default, which means that the event will bubble up through the DOM to the event listener, which is delegated to the window. Leptos automatically detects events that *don't* bubble, and uses undelegated event listeners for those. However, using `new Event()` (or `Event::new()`) explicitly [defaults to not bubbling](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event#values), whether or not an event of the same type fired by an actual DOM element would bubble. However, it's possible to use `EventInit` to provide a dict of options, setting bubbling to `true`. 

So in the commit I added I edited the synthetic event dispatched in `enter_count` so that it bubbles. Now the test works with the original example, with event delegation intact!

Thank you, again. This is a really excellent model to show people for how they can implement e2e testing of a Leptos app using multiple different testing methods.

agilarity:
Nice work! Please merge.

Issue #1295, opened 2023-07-06
Rename Leptos config options to kebab-case via serde, instead of replacing characters


filipdutescu:
**Is your feature request related to a problem? Please describe.**

My app is using a bigger config file, which contains a member of the `leptos::LeptosOptions` type. The config is serialized as a `.toml`, which adheres to the `kebab-case` naming convention. Since `leptos::LeptosOptions` does not use `#[serde(rename_all = "kebab-case")]`, if I want to deserialize it, I would need to specify all Leptos options I set under the snake_case.

**Describe the solution you'd like**

Add `#[serde(rename_all = "kebab-case")]` to the `leptos::LeptosOptions` struct. This can be easily done by modifying the [this line](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L25).

```rust
// ...
#[derive(TypedBuilder, Debug, Clone, serde::Deserialize)]
#[serde(rename_all = "kebab-case")]
pub struct LeptosOptions {
// ...
```

This would also lead to not needing to [replace '-' characters in the *entire* config file](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L180).

**Describe alternatives you've considered**

Duplicating the Leptos config options in a struct I own, using the rename method and implementing `From<T>` to allow conversion to `leptos::LeptosOptions`. This seems however a bit meaningless, since Leptos *already* uses kebab-case, just not in a manner in which it could satisfy my use case. 

**Additional context**

Example of my `config.toml`:

```toml
log-level = "Debug"
application-address = "0.0.0.0:3000"

# ...

[leptos]
output-name = "my-awesome-app"
# ...
```

My `config.rs`:

```rust
#[derive(Deserialize)]
#[serde(rename_all = "kebab-case")]
pub(crate) struct Config {
    #[serde(default = "default_application_address")]
    pub application_address: String,

    #[serde(
        default = "default_log_level",
        deserialize_with = "deserialize_log_level"
    )]
    pub log_level: Level,

    pub leptos: LeptosOptions,
}
```

filipdutescu:
Yes, I'd love to contribute! Should I have any concerns in mind while implementing this? Ie do any tests, other areas of serialization depend on the current behaviour?

gbj:
[...]

PR #1308, opened 2023-07-08
feat(leptos-config): kebab-case via serde's rename_all

filipdutescu:
Currently, `leptos::LeptosOptions` are deserialized (and serialized) to `snake_case`, since, by default, `serde` uses the name of the field as-is. The options given via `.toml` files are given using `kebab-case`. This behaviour leads to problems if you wish to use `leptos::LeptosOptions` as the type of a field in your own config, which uses the `kebab-case` syntax, as it will not provide any values to that field.

These changes switch out the previous mechanism of manually replacing the `-` character to the `_` character in order for serialization to work. In its place, it uses `serde`'s `#[serde(rename_all = ...)]` macro to handle the naming convention.

In case other naming conventions are used, a workaround would be to define a user-owned mirror struct of `leptos::LeptosOptions`, which would contain the required fields, deserialize it as desired and' convert it to the latter via, for example, the `From<T>` trait.

Closes: #1295


filipdutescu:
should I also add defaults for `serde` deserialization? same as the `config-rs` ones

gbj:
This looks good to me, overall. 

Two things:
1. Could you revert the changes to `examples/todo_app_sqlite/Todos.db`? I assume this was just due to testing that example but the sqlite database changes got committed.

> should I also add defaults for `serde` deserialization? same as the `config-rs` ones
2. Yes, defaults would be great! I think the default values for fields that can have them are listed in the `typed-builder` macro annotations on each field.

filipdutescu:
Hopefully that should be it. Please let me know if I should make further changes! :grin:

gbj:
Looks good to me!

PR #1309, opened 2023-07-08
feat: use lazy thread local for regex

tqwewe:
Uses `thread_local` `Lazy` for storing regex instances to avoid parsing the same regex multiple times on the client.


gbj:
Beautiful, thanks so much. Enjoy your Saturday!

PR #1313, opened 2023-07-09
update warnings to remove mention of csr as a default feature

Heliozoa:
Hi,
thanks for your work on Leptos.

I noticed these warnings while working on my project. Since `csr` is no longer a default feature as of 0.4.0, I figured the warnings should be updated.


gbj:
Thanks!

PR #1325, opened 2023-07-11
Bump indexmap to version 2

g-re-g:
Not sure if y'all are interested but indexmap had a major version release a couple weeks ago:
https://github.com/bluss/indexmap/blob/master/RELEASES.md

Notably it includes a version bump of its underlying hashmap library hashbrown:
https://github.com/rust-lang/hashbrown/blob/master/CHANGELOG.md

Also notably it now requires rust 1.64 or later where as the previously specified highest version of 1.9.2 required rust 1.56.1 or later.

Just a little housekeeping PR, I'm not trying to advocate for anything. Feel free to delete this PR if it's not needed/wanted.

### DIFF
--- a/leptos_dom/Cargo.toml
+++ b/leptos_dom/Cargo.toml
@@ -14,7 +14,7 @@ drain_filter_polyfill = "0.1"
 educe = "0.4"
 futures = "0.3"
 html-escape = "0.2"
-indexmap = "1.9"
+indexmap = "2"
 itertools = "0.10"
 js-sys = "0.3"
 leptos_reactive = { workspace = true }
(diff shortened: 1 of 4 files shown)


gbj:
This looks great. I checked `cargo tree` and it looks like `indexmap` gets pull in in two places: via Leptos itself in a few crates (excellent, great, updated) and via `reqwest`, which is only used in the rare circumstance of server function clients that are not running in the browser (i.e., a desktop app using `server_fn`). I think in this case having duplicate copies of `indexmap` (1.9 and 2.0) should be fine until `reqwest` updates.

PR #1339, opened 2023-07-13
build: run tasks from workpace or member directory

agilarity:
READY FOR REVIEW

Add the ability to run `ci` checks and tests from either the workspace root or a member directory.

## Benefits

- Support test driven development workflow
- Enable crate level `ci` workflows

## Changes 

- Move common tasks to the `cargo-make/main.toml`
- Extend all workspace members from `cargo-make/main.toml`
- Move project specific tasks to `leptos`
- Move project specific tasks to `leptos_macro`
- Replace `check-all-release` with `check-release`

### DIFF
--- a/cargo-make/main.toml
+++ b/cargo-make/main.toml
@@ -13,12 +13,6 @@ command = "cargo"
 args = ["+nightly", "check-all-features"]
 install_crate = "cargo-all-features"
 
-# FIXME: Should this duplicate of check-all be removed?
-[tasks.check-all-release]
-command = "cargo"
-args = ["+nightly", "check-all-features"]
-install_crate = "cargo-all-features"
-
 [tasks.test]
 alias = "test-all"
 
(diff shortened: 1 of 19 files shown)


cargo-make/main.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[tasks.check]
+alias = "check-all"
+
+[tasks.check-all]
+command = "cargo"
+args = ["+nightly", "check-all-features"]
+install_crate = "cargo-all-features"
+
+# FIXME: Should this duplicate of check-all be removed?
```
gbj, line 16: `check-all-release` was *supposed* to be the equivalent of `cargo check --release`, as opposed to `check-all` running the debug check. The release flag must have dropped out inadvertently at some point.

agilarity: I pushed up the change.

PR #1343, opened 2023-07-13
fix: `server_fn` `rustls` feature shouldn't pull in `default-tls`

### LINKED ISSUE #1341: feature option for `rustls` instead of `openssl-sys`
When cross compiling, for instance for musl or another architecture, the `openssl-sys` crate always produces problems.  
You can usually get away with this, when you add the `openssl-sys` crate yourself and set the `vendored` feature explicitly.
However, as far as I can see right now, `openssl-sys` is pulled in only by `reqwest` from the `server_fn` package, which is just using default features after I took a look at the `Cargo.toml`.

A nice solution would be to either use `rustls` by default, or provide it as an optional feature, which then gets rid of the `openssl-sys` dependency. If you want `rustls` or the "real" `openssl-sys` by default is of course up to the maintainers of this project.

Getting rid of it should be pretty easy:  
Instead of using `reqwest` with default features, it could be changed to something like this, depending on the needed features of course:

```toml
reqwest = { version = "0.11", default-features = false, features = ["async-compression", "brotli", "json", "rustls-tls"] }
```

This would make the cross compilation (especially for musl targets) a lot easier.  
A nice bonus is, that the project is one step further in the direction of "rust only" ;-)

An alternative is, as described above already, if the user imports 
```toml
openssl-sys = { version = "0.9", features = ["vendored"] }
```
manually with the `vendored` feature.

gbj:
Fixes #1341.

### DIFF
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -13,19 +13,19 @@ cfg-if = "1"
 leptos_dom = { workspace = true }
 leptos_macro = { workspace = true }
 leptos_reactive = { workspace = true }
-leptos_server = { workspace = true}
+leptos_server = { workspace = true }
 leptos_config = { workspace = true }
 tracing = "0.1"
 typed-builder = "0.14"
-server_fn = { workspace = true}
+server_fn = { workspace = true }
 web-sys = { version = "0.3.63", optional = true }
 wasm-bindgen = { version = "0.2", optional = true }
 
 [dev-dependencies]
-leptos = { path = "."}
+leptos = { path = "." }
 
 [features]
-default = ["serde"]
+default = ["serde", "default-tls"]
 template_macro = ["leptos_dom/web", "web-sys", "wasm-bindgen"]
 csr = [
   "leptos_dom/web",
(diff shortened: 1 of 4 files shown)


sebadob:
You are so quick :D

sebadob:
For some reason this still seems to pull in `hyper-tls`, coming from `reqwest`, but I currently have no idea why.
All other options do work fine and I do the same thing with `reqwest` in quite a few other projects to get rid of the `openssl-sys` dependency.

I will try to figure out where this might come from.

gbj:
You'd need to do `default-features = false` on Leptos, as I set it to enable `default-tls` by default.

sebadob:
Yes I did, but it still gets pulled in somehow. I guess it's my bad.

As soon as I remove the `default-tls` from `default = ["serde", "default-tls"]` in `leptos/Cargo.toml` the `openssl-sys` dependency is gone.

It is a bit strange, because I imported leptos like `leptos = { path = "../../leptos", default-features = false }`, but yeah I guess I screwed it up somehow with the features. So I think this is good, yes. :)

gbj:
> As soon as I add either `leptos_meta` or `leptos_router` to the list, the `native-tls` is pulled in again.
Of course, that makes perfect sense. I will just remove it from the defaults, as this would be a huge ecosystem issue — any library you pulled in that depended on Leptos would break it.

Issue #1334, opened 2023-07-12
CSP Headers / unsafe-inline vs nonce's


gbj:
### Discussed in https://github.com/leptos-rs/leptos/discussions/1333

<div type='discussions-op-text'>

<sup>Originally posted by **sebadob** July 12, 2023</sup>
I am currently evaluating Leptos for a bigger upcoming project.  
I do have the prod ready template almost fully set up, before I can actually build some components and I guess, if it turns out nicely, I can provide this template back to the community,  It builds up on the axum starter with a few more additions to make it prod ready.

The last step I am taking a look at now is the CSP settings.  
I usually keep `unsafe-inline` for CSS for practical reasons, but for the rest I want to get rid of all unsafe stuff.

Usually, I do this with inserting `nonce`s on the server side and set a proper CSP for each request to be able to set `strict-dynamic` and a `nonce`.

So, after all this talking:  
**Is there a way to get access to a custom `index.html` with out of order streaming, into which I could insert `nonce`s in all places with a templating engine, or does leptos maybe even provide a way to insert nonces into all inline scripts it injects?**

The last value I needed was `wasm-unsafe-eval`, which I guess is the normal way of working with wasm currently. Leptos is now my first attempt with a wasm framework.</div>

@sebadob This is so clearly a good idea that I've just converted it into an issue so I can track it here.

Here's my understanding of the various places a nonce would need to be inserted:
1. Loading the JS/WASM module for hydration
https://github.com/leptos-rs/leptos/blob/e9665b34e5f5f926343d5be97115d3c90a20f507/integrations/utils/src/lib.rs#L73
2. The initial "pending resources" script tags, for out-of-order and in-order streaming
https://github.com/leptos-rs/leptos/blob/e9665b34e5f5f926343d5be97115d3c90a20f507/leptos_dom/src/ssr.rs#L234-L238 https://github.com/leptos-rs/leptos/blob/e9665b34e5f5f926343d5be97115d3c90a20f507/leptos_dom/src/ssr_in_order.rs#L132-L136
3. Script to insert an HTML fragment during out-of-order streaming
https://github.com/leptos-rs/leptos/blob/e9665b34e5f5f926343d5be97115d3c90a20f507/leptos_dom/src/ssr.rs#L335-L340
4. Script to resolve a resource value
https://github.com/leptos-rs/leptos/blob/e9665b34e5f5f926343d5be97115d3c90a20f507/leptos_dom/src/ssr.rs#L688-L695

[...]

PR #1348, opened 2023-07-14
feat: add support for adding CSP nonces

gbj:
Addresses #1334.

### DIFF
--- a/integrations/actix/Cargo.toml
+++ b/integrations/actix/Cargo.toml
@@ -19,3 +19,6 @@ serde_json = "1"
 parking_lot = "0.12.1"
 regex = "1.7.0"
 tracing = "0.1.37"
+
+[features]
+nonce = ["leptos/nonce"]
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -728,6 +728,8 @@ fn provide_contexts(
     provide_context(cx, res_options);
     provide_context(cx, req.clone());
     provide_server_redirect(cx, move |path| redirect(cx, path));
+    #[cfg(feature = "nonce")]
+    leptos::nonce::provide_nonce(cx);
 }
 
 fn leptos_corrected_path(req: &HttpRequest) -> String {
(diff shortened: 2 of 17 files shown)


gbj:
So I've got this working for a super simple demo. Toggling the `nonce` feature on `leptos_actix` adds or removes a nonce. No need for app code to change in either case—if no nonce is present this works correctly, un-nonced, and if a nonce is present it injects the CSP meta tag and correct nonces.

I just need to get the Axum and Viz integrations up to parity with Actix.

I'm generating nonces in what I *think* is a legitimate way: filling 16 bytes (128 bits) with randomness generated by a `rand::ThreadRng`, then using `base64` to encode it in a URL-safe way. But I'm not at all an expert here.

```rust
#[component]
    let csp = use_nonce(cx).map(|nonce| {
        view! { cx,
            <Meta
                http_equiv="Content-Security-Policy"
                content=format!("script-src 'nonce-{nonce}' 'unsafe-eval'")
             />
        }
    });

    view! { cx,
      {csp}
      <script nonce=use_nonce(cx)>"console.log('Hello, world!');"</script>
    }
}
```

sebadob:
I just tested it and it is such a cool feature - I'm loving it!
This is quite a big boost for the security I would say.

And the way it is implemented now, it is almost trivial to set up even without any custom extractors / handlers / whatever.

Really cool! Thank you very much!

The 128 bit is a toally fine value as well, since they are re-generated with each request anyway. Anything longer would have a bigger performance impact.

sebadob:
btw, for the documentation later on, a good default CSP would be something like this:

```
default-src 'self'; script-src 'strict-dynamic' 'nonce-{}'; style-src 'nonce-{}';
```

The `strict-dynamic` for the `script-src` is pretty nice. It basically means "If you are calling another inline script from a nonce'd script, accept it too, even without a nonce.

gbj:
Oops my example created a hydration error due to the way the meta tag was created. This should work better
```rust
#[component]
    provide_meta_context(cx);

    view! { cx,
        <p>"Test"</p>
        <Meta
            http_equiv="Content-Security-Policy"
            content=move || {
                use_nonce(cx)
                    .map(|nonce| {
                        format!(
                            "default-src 'self'; script-src 'strict-dynamic' 'nonce-{nonce}' \
                            'wasm-unsafe-eval'; style-src 'nonce-{nonce}';"
                        )
                    })
                    .unwrap_or_default()
            }
        />
        // manually insert nonce during SSR on inline script 
        <script nonce=use_nonce(cx)>"console.log('Hello, world!');"</script>
        // leptos_meta <Style/> and <Script/> automatically insert the nonce
        <Style>"body { color: blue; }"</Style>
    }
}
```

I'll merge this once the CI passes again.

sebadob:
I thought this hydration error was because you needed to do some leftover work with axum and I did my test with axum actually.
Sorry, I should have commented this anyway. I saw an assertion panic on my side.

sebadob:
And yes, with your new example, the error is gone! yay

gbj:
All good, I'd initially been ignoring it too! :smile:

PR #1349, opened 2023-07-14
use cfg_attr for conditional derives

CircuitSacul:
Rather than using cfg_if to re-define structs with and without sqlx::FromRow, we can use cfg_attr to conditionally derive sqlx::FromRow. This simplifies the sqlite todo examples.


gbj:
Thanks so much. I know “PRs welcome ;-)” is a meme obnoxious maintainer response, but it genuinely does help a LOT when people notice something like this and put in the work so I can just check it and merge.

PR #1365, opened 2023-07-16
fix: update link to example code in book testing page

BakerNet:
The current link here `404`s - this just fixes the link in the book markdown


gbj:
Thanks!

Issue #1355, opened 2023-07-14
Hot reload failing


sebastian:
**Describe the bug**

When changing node attributes hot reloading crashes.
Here is the simplest app I can reproduce the error with:

```
#[component]
pub fn App(cx: Scope) -> impl IntoView {
    provide_meta_context(cx);

    view! {
        cx,
        <Stylesheet id="leptos" href="/pkg/app.css" />

        <Router>
            <div class="change a class here">
                    <Routes>
                        <Route path="" view=|_cx| view! { cx, "Hello" }/>
                    </Routes>
            </div>
        </Router>
    }
}
```

When changing the classes in the outermost div, the hot reloader crashes with:

```
TypeError: undefined is not an object (evaluating 'child.node.setAttribute')
```

The problem occurs at [`child.node.setAttribute`](https://github.com/leptos-rs/leptos/blob/main/leptos_hot_reload/src/patch.js#L77C18-L77C18) because `node` is `undefined`.

This seems to happen because as the `childAtPath` function recursively traverses the DOM, it ends up [at a call](https://github.com/leptos-rs/leptos/blob/main/leptos_hot_reload/src/patch.js#L367) where it constructs the element as `{children: ...}` which then eventually gets returned as the child. This element doesn't however have a `node`-property set.

I believe this is what's happening anyway?

Strangely the error does not occur when the `<Stylesheet id="leptos" href="/pkg/kleio.css" />` is commented out 😬

**Leptos Dependencies**

```toml
leptos = { default-features = false, features = ["serde", "nightly"], git="https://github.com/leptos-rs/leptos.git", branch="main" }
leptos_axum = { default-features = false, git="https://github.com/leptos-rs/leptos.git", branch="main" }
leptos_meta = { default-features = false, features=["nightly"], git="https://github.com/leptos-rs/leptos.git", branch="main"  }
leptos_router = { default-features = false, features=["nightly"], git="https://github.com/leptos-rs/leptos.git", branch="main"  }
```

**To Reproduce**
1. Change the classes associated with the div and save
2. Monitor the console in the browser. It will, if like on my end, show an exception
3. Comment out the stylesheet include statement and try again. Now the changes should work?

**Expected behavior**
I would expect the hot reload to find the node to patch and apply the patches.

**Screenshots**
![CleanShot 2023-07-14 at 22 06 34@2x](https://github.com/leptos-rs/leptos/assets/1089/cd03cc56-627e-4c81-9ca0-863e66790b54)

[...]

PR #1368, opened 2023-07-17
Hot reload bug-fix

sebastian:
I believe this will fix [the bug I reported previously](https://github.com/leptos-rs/leptos/issues/1355). If nothing else, to the best of my abilities I am now unable to reproduce the bug.

TL;DR: 
In one of the instances where you created a tree walker, you instantiated it with a filter. The filter was however incorrect. It returned a boolean true/false, where the spec states it needs to be one of 3 enum values, `NodeFilter.FILTER_{ACCEPT,REJECT,SKIP}`.

The first commit, which I am happy to remove if you want me to(!), is a reformatting pass of the patch javascript file. It had a mix of indentation styles and tabs and spaces.

### DIFF
--- a/leptos_hot_reload/src/patch.js
+++ b/leptos_hot_reload/src/patch.js
@@ -1,386 +1,393 @@
 console.log("[HOT RELOADING] Connected to server.");
 function patch(json) {
-	try {
-		const views = JSON.parse(json);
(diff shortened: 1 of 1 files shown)


gbj:
That is awesome, and thank you so much! Looks good to me.

PR #1377, opened 2023-07-18
feat: add `create_query_signal` for url parameter synced signals

tqwewe:
Adds a new `create_query_signal` function to leptos_router which creates a signal synchronized with a specific URL query parameter.

```rust
#[component]
pub fn SimpleQueryCounter(cx: Scope) -> impl IntoView {
    let (count, set_count) = create_query_signal::<i32>(cx, "count");
    let clear = move |_| set_count(None);
    let decrement = move |_| set_count(Some(count().unwrap_or(0) - 1));
    let increment = move |_| set_count(Some(count().unwrap_or(0) + 1));

    view! { cx,
        <div>
            <button on:click=clear>"Clear"</button>
            <button on:click=decrement>"-1"</button>
            <span>"Value: " {move || count().unwrap_or(0)} "!"</span>
            <button on:click=increment>"+1"</button>
        </div>
    }
}
```
A full example can be found in `examples/counter_url_query`.

I have not added an example to the code docs of this function as I'm not sure what the best way to do so is while keeping it runnable.

Improvments are welcome. Thanks


gbj:
I just tried adding the example in the doc comment. Components (like the one you pasted into the PR!) often work well because nothing tries to run them so it just checks that they compile for all the feature sets. But we'll see if I did it properly :smile:

gbj:
Thanks for your patience while I screwed around with the CI setup... And thanks for the contribution!

PR #1379, opened 2023-07-19
ci: lint with clippy

agilarity:
READY FOR REVIEW

Add clippy to the `lint` task.

### Changes

- Support conditional compilation with `cargo hack`
- Fix most violations
- Defer resolution of [incorrect clone implementation](https://rust-lang.github.io/rust-clippy/master/index.html#/incorrect_) errors
- Add clippy to `ci` flow

### No errors on warnings

- I did not find a way to pass clippy options with `cargo hack`

_See [Clippy Lints](https://rust-lang.github.io/rust-clippy/master/index.html) for rule details._

### DIFF
--- a/cargo-make/lint.toml
+++ b/cargo-make/lint.toml
@@ -1,9 +1,10 @@
-[tasks.pre-clippy]
-env = { CARGO_MAKE_CLIPPY_ARGS = "--all-targets --all-features -- -D warnings" }
-
 [tasks.check-style]
-dependencies = ["check-format-flow", "clippy-flow"]
+dependencies = ["check-format-flow", "clippy-each-feature"]
 
 [tasks.check-format]
 env = { LEPTOS_PROJECT_DIRECTORY = "../" }
 args = ["fmt", "--", "--check", "--config-path", "${LEPTOS_PROJECT_DIRECTORY}"]
+
+[tasks.clippy-each-feature]
+command = "cargo"
+args = ["hack", "clippy", "--all", "--each-feature", "--no-dev-deps"]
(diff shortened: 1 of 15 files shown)


martinfrances107:
By way of review ... all the changes of this form are 

-  any(debug_assertions, features = "ssr"),
+ any(debug_assertions, feature = "ssr"),

Wow - that is a really good find. 

When resolved, I think the library users experience will be improved.
A subtle library bug like has the potential to cause lots frustration.

My experience with wide ranging subtle changes to the code base like this -- 
It will,  with time, uncover lots  of other problems lurking here are down-stream in application code.

So,  politely, I am think the emphasis here is all wrong.. that bugfix should be promoted to a headline issue, lead to an  isolated PR .. and trigger a version bump.

That way IF unexpected issue blow up ... and someone is running git bisect at 2AM on a rainy Tuesday - this otherwise seeming innocuous change will stand out.

I am not one to nominate other people for work ... What I am saying is if other reviewers think it is a good idea.. then send me a message and I will do the separation

gbj:
@martinfrances107 In this specific case those feature flags are definitely an unfortunate copy-paste issue, but in every case what they are doing is enabling `tracing` instrumentation. There should not be any effects on actual logic as there would be in many cases with features like this.

However, I think it's still a fair point to keep it as a separate commit. In this case, I'd propose that I merge these as separate commits rather than squashing, so that the `tracing` change is its own commit. Does that sound reasonable?

martinfrances107:
> There should not be any effects on actual logic as there would be in many cases with features like this.

That is a good point ...

> I'd propose that I merge these as separate commits rather than squashing,

As you see fit... anything I can do to help out .. just shout.

gbj:
Thanks to both of you for all your work cleaning up the codebase!

Issue #1382, opened 2023-07-19
panicked at 'called `Option::unwrap()` on a `None` value when building for release mode


tqwewe:
**Describe the bug**
My app thus far works great in debug mode with no errors, but when building to release, I get a panic in the console:
```js
panicked at 'called `Option::unwrap()` on a `None` value', /home/ari/.cargo/registry/src/index.crates.io-6f17d22bba15001f/leptos_dom-0.4.2/src/components/dyn_child.rs:349:38
```

The code that panics in dyn_child.rs is:
https://github.com/leptos-rs/leptos/blob/c1c49ce53be5c85cbb3505bfd8b7a86dbc14fd60/leptos_dom/src/components/dyn_child.rs#L347-L352

Could this be as simple as removing the `unwrap` and doing an `if let Some` check instead?

**Leptos Dependencies**

```toml
leptos = { version = "0.4", features = ["nightly"] }
leptos_axum = { version = "0.4" }
leptos_meta = { version = "0.4", features = ["nightly"] }
leptos_router = { version = "0.4", features = ["nightly"] }
leptos_dom = { version = "0.4" }
```

**To Reproduce**
I'm not sure :(

**Expected behavior**
This should not panic, as there's no error when running in debug mode.

**Screenshots**

![image](https://github.com/leptos-rs/leptos/assets/16362377/35f82ad9-b2e5-4cde-9eea-42c1974c297f)

**Additional context**
-

gbj:
So generally speaking this suggests some other hydration mismatch error. 

For context:
https://github.com/leptos-rs/leptos/blob/c1c49ce53be5c85cbb3505bfd8b7a86dbc14fd60/leptos_dom/src/components/dyn_child.rs#L344-L352 that comment points to https://github.com/leptos-rs/leptos/blob/c1c49ce53be5c85cbb3505bfd8b7a86dbc14fd60/leptos_dom/src/ssr.rs#L484-L507 So if the `DynChild` (= dynamic child, and in this case specifically dynamic text — think some `move || some_string_here()`) was hydrated properly, it *should* also have a previous sibling which is a comment node.

There's nothing harmful about not removing it at all if it is there, but the fact that it's not found suggests something else has gone awry.

In your PR to remove the unwrap, does every dynamic string field in the page appear to update correctly after hydration? I know it can be hard to narrow these things down...

tqwewe:
[...]

PR #1383, opened 2023-07-19
fix: remove unwrap when removing previous sibling

tqwewe:
I don't understand really any of the code here, but in case the fix is as simple as adding a `if let Some` check to avoid the panic, then this PR does that. If not, then feel free to close this PR.

Fixes https://github.com/leptos-rs/leptos/issues/1382


gbj:
Closing this in favor of #1506.

PR #1386, opened 2023-07-19
feat: create a <AnimatedShow> component (closes #1370)

### LINKED ISSUE #1370: `<ShowAnimate>` or similar components
I have used svelte quite a lot in the past and one thing, that is really great about it, is the built in animations.
Using these, you can get something "pretty" really quickly.

I now that leptos has no opinion about css, but a basic wrapper component built in would be really nice.
Since I don't know if leptos even wants to include such things, I created an issue and not a PR directly.
I am using components like these a lot to provide a smoother UX and the most basic one is already created, so you could actually just copy & paste it maybe.

The following code is based on the `start-axum` template and you can copy & paste it into the `app.rs` to test it.
The idea is, that you can have easy unmounting animations with it while still providing your own css classes or solutions.
I have not written an inline style because some people might not want to allow `unsafe-inline` for css in their CSP (which actually is an issue for svelte too and they do have an open issue to fix that).

The `<ShowAnimate>` component from this code could maybe be implemented into leptos directly, if you want to go that route.
There is the possibility for other basic default transitions in the future, which could be realized this way too. Just the same way that svelte does it, which is actually awesome.

```rust
use leptos::*;
use leptos::leptos_dom::helpers::TimeoutHandle;
use leptos_meta::*;

#[component]
    provide_meta_context(cx);

    // add the following content to the `main.scss`:
    //
    //body { font-family: sans-serif; text-align: center; }
    //@keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
    //@keyframes fadeOut { from { opacity: 1; } to { opacity: 0; } }
    //.fadeIn1000 { color: green; animation: 1000ms fadeIn forwards; }
    //.fadeOut1000 { color: red; animation: 1000ms fadeOut forwards; }

    let is_hover = create_rw_signal(cx, false);

    view! { cx,
        <Title text="AnimateMe" />

        <Stylesheet id="leptos" href="/pkg/start-axum.css"/>

        <div
            on:mouseenter=move |_| is_hover.set(true)
            on:mouseleave=move |_| is_hover.set(false)
        >
            "Hover Me"
        </div>

        <ShowAnimate
            when=is_hover
            show_class="fadeIn1000"
            hide_class="fadeOut1000"
            hide_delay=core::time::Duration::from_millis(1000)
        >
            "Here I Am!"
        </ShowAnimate>
    }
}

#[component]
pub fn ShowAnimate(
    cx: Scope,
    #[prop(into)] when: MaybeSignal<bool>,
    show_class: &'static str,
    hide_class: &'static str,
    hide_delay: core::time::Duration,
    children: ChildrenFn,
) -> impl IntoView {
    let handle: StoredValue<Option<TimeoutHandle>> = store_value(cx, None);
    let cls = create_rw_signal(
        cx,
        if when.get_untracked() {
            show_class
        } else {
            hide_class
        },
    );
    let show = create_rw_signal(cx, when.get_untracked());

    create_effect(cx, move |_| {
        if when.get() {
            // clear any possibly active timer
            if let Some(h) = handle.get_value() {
                h.clear();
            }

            cls.set(show_class);
            show.set(true);
        } else {
            cls.set(hide_class);

            let h = set_timeout_with_handle(
                move || {
                    show.set(false);
                },
                hide_delay,
            )
                .expect("set timeout");
            handle.set_value(Some(h));
        }
    });

    on_cleanup(cx, move || {
        if let Some(h) = handle.get_value() {
            h.clear();
        }
    });

    view! { cx,
        <Show when=move || show.get() fallback=|_| ()>
            <div>
                <div class=move || cls.get()>
                    {children(cx)}
                </div>
            </div>
        </Show>
    }
}

```

I can open a PR about this, if you like this idea.

sebadob:
This is an implementation for https://github.com/leptos-rs/leptos/issues/1370 

Renamed from the original `<ShowAnimate>` idea to `<ShowHide>`, because it is not about animations by definition. I would use it for animations, but there may be other use cases. Actually, it is only a component with a delayed unmount from the DOM and optional CSS classes on change.


sebadob:
So, the CI was failing before, because it tried to set a timeout in a non-wasm32 context.
It makes sense, that it fails in that case, but since this is my very first PR on this project, I don't really get it, why it even tried it.

The timeout is being set inside a `create_effect`, which only runs on the client, and therefore with wasm32 only anyway (right.. ?). My questions then is why did it even complain in the first place that this function is not available in a non wasm context?

How I got around it now is by setting `#[cfg(target_arch = "wasm32")]` right before I would set the timeout even inside the `create_effect`. This is not that big of a deal, but then clippy complained about the not used `hide_delay` variable which I needed to get a way around as well. 
This solution is working now and I ran `cargo make check && cargo make test && cargo make check-examples` without a single issue on my localhost.

However, like I needed to get around the not used `hide_delay` now is not the cleanest approach and I don't really like it.
The other solution would be to simply ignore the test case in the docs comment, which I don't like either.

Any suggestions about this?

gbj:
So re: CI, you're correct that `create_effect` only runs when the `csr` or `hydrate` features are enabled, but it's not gated on `target_arch = "wasm32"` itself. That's probably a mistake on my part, but I think outside the scope here. For the sake of allowing examples to run in doctests I'd suggest either
1. making the example into a component, which will cause it to compile but not run
```rust
#[component]
fn ExampleComponent(cx: Scope) -> impl IntoView {
 // the example
}
```
2. adding a `cfg!` check that's commented out with a `#`, which won't display
```rust
# if !cfg!(target_arch = "wasm32") {
// the example
# }
```

gbj:
Re: naming—I'd suggest either `<AnimatedShow/>` or `<DelayedShow/>`, as either of these would capture the purpose in a clearer way than `<ShowHide/>`. We do also have `<AnimatedRoutes/>` and `<AnimatedOutlet/>`, so I'd suggest that pattern, personally.

sebadob:
Changing it into a component in the docs test is a nice solution, worked out perfectly.
And I renamed it into `AnimatedShow` as you suggested and did another merge from master.

All 3 `cargo make`'s are fine again locally.

sebadob:
I guess we could update the CONTRIBUTING.md a bit to provide some more help and information for new contributors.

For instance, my cargo uses stable by default and not nighty. And my IDE does run `rustfmt` on save, which is why I never care about executing `cargo fmt` manually and I have never done it with a special config.

We could do some more checks locally before actually waiting for the pipelines to fail.
What I did now to make sure it (hopefully) works out fine this time:

- `cd leptos` (or whereever you modified code)
- `cargo +nightly fmt -- --config-path ..`

- `cargo +nightly make check`
- `cargo +nightly make test`
- `cargo +nightly make check-examples`
- `cargo +nightly make --profile=github-actions ci`

if you modified an example:
- `cd examples/your_example`
- `cargo +nightly fmt -- --config-path ../..`
- `cargo +nightly make --profile=github-actions verify-flow`

PR #1387, opened 2023-07-19
Template refactor + snapshot tests

vldm:
Hi, a month ago i had discussed the idea of refactoring template and view macro in discord.

The main points were:
- removing code duplication, by using single "visitor"
- splitting code ssr/client view macro code.

This PR consist of two major commits:
1. simple refactor of view/template by splitting view.rs into separate modules, all common functional are kept in separate modules as well.
2. Added test that will "snapshot" current macro expansion, so any future improvements would not be left unattended.

There also last commit that remove unique id in `thread_local` for `template` macro, which as i understand is redundant, since macro expand in separate `{ block }`, so names should not conflict.
This change is important for "snapshot" tests. And i think avoid random (and other side effects) is generally good practice in macro expansion.

If you insist i could split any of this change in separate PR.

_As for refactoring, there is still room for future improvements, since methods were left untouched. My plan was to use specialized "visitors" for each common routine, so we can avoid most of repetition. But currently i have not a lot of capacity and this can took some time from me, so i decided to separate implemented part and create pr._

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,22 +26,22 @@ members = [
 exclude = ["benchmarks", "examples"]
 
 [workspace.package]
-version = "0.4.5"
+version = "0.4.6"
 
 [workspace.dependencies]
-leptos = { path = "./leptos", version = "0.4.5" }
-leptos_dom = { path = "./leptos_dom", version = "0.4.5" }
-leptos_hot_reload = { path = "./leptos_hot_reload", version = "0.4.5" }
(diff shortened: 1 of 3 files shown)


gbj:
Sorry I just screwed something up with git while trying to merge in some other changes to the view macro so the CI could run on this. I'll figure it out.

vldm:
@gbj It's okay, if you need, i can update it to the newest codebase.

I just waited for some feedback, want to be sure that you find this refactoring useful.

PR #1400, opened 2023-07-20
assign input event to the web_sys::InputEvent

Bechma:
small adjustment in order to get the correct event from the input event listener

### DIFF
--- a/leptos_dom/src/events/typed.rs
+++ b/leptos_dom/src/events/typed.rs
@@ -248,7 +248,7 @@ generate_event_types! {
   formdata: Event, // web_sys does not include `FormDataEvent`
   #[does_not_bubble]
   gotpointercapture: PointerEvent,
-  input: Event,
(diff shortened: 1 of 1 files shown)


gbj:
This is not necessarily correct; handling it correctly would be possible but would require an approach that's different from what we're currently doing.

> For [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements that accept text input (type=text, type=tel, etc.), the interface is [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent); for others, the interface is [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event).

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event)

PR #1402, opened 2023-07-20
fix typo for WrapsChildren

icub3d:
Enjoying the book! Found what I believe to be a typo.


gbj:
Nice catch! It would be cool if there were a feature to let you use the wrong name for a component and have it compile, but alas. Maybe Leptos v2, powered by ChatGPT :smile:

Issue #73, opened 2022-11-12
Classes with numbers are invalid


jquesada2016:
The following valid class name does not compile because of the `25`:

```rust
view! { cx,
  <span class-opacity-25=true />
}
```

gbj:
Thanks. This is a limitation of `syn-rsx` at the moment and I'll look into opening a PR over there.

gbj:
By the way, I believe `class-opacity25` *is* valid (i.e., without the dash). The issue is not the numbers per se, that `syn-rsx` parses each dash-separated piece of the attribute name as a `syn::Ident`, and `25` is an invalid Rust ident.

gbj:
So this is fixed by a0a66b75ddbbd9e49bdbf338499a96b8a7e8a1e3, and the newly-available syntax for the original case is
```rust
view! { cx,
  <span class=("opacity-25", true)/>
}
```

PR #1416, opened 2023-07-22
Feat: Update rstml to v0.11.0

vldm:
It's marked as feature, since in rstml v0.11.0 i added support of numbers in dash seperated attributes

https://github.com/leptos-rs/leptos/issues/73
https://github.com/rs-tml/rstml/issues/9

Currently during parse it follows SGML rule:
ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").
https://www.w3.org/TR/html4/types.html#type-cdata

So special symbols are still forbidden in attribute names.

### DIFF
--- a/leptos_macro/src/view.rs
+++ b/leptos_macro/src/view.rs
@@ -1225,7 +1225,7 @@ fn attribute_to_tokens(
         };
         let undelegated_ident = match &node.key {
             NodeName::Punctuated(parts) => parts.last().and_then(|last| {
-                if last == "undelegated" {
+                if last.to_string() == "undelegated" {
                     Some(last)
                 } else {
                     None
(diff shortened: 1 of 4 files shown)


leptos_macro/src/view.rs
```diff
@@ -1246,19 +1246,17 @@ fn attribute_to_tokens(
         let event_type = if is_custom {
             event_type
         } else if let Some(ev_name) = event_name_ident {
-            let span = ev_name.span();
-            quote_spanned! {
-                span => #ev_name
+            quote! {
```
vldm, line 1249: I have replaced `quote_spanned` with `quote`, since `quote`/ `quote_spanned` already preserve the `Span` of original tokens, and it's `_spanned` version only needed when new tokens are created.

https://docs.rs/proc-quote/latest/proc_quote/macro.quote_spanned.html#hygiene

gbj: Excellent, thanks.

gbj:
Excellent! So if I'm understanding correctly, this means that we can now support 
```
class:bg-red-500=true
```
which would previously be
```
class=("bg-red-500", true)
```
but still support the alternate syntax for bizarre Tailwind situations 
```
class=("why_oh[why_50px_why]_help_me", true)
```

PR #1422, opened 2023-07-23
(extremely WIP) Islands/partial hydration

gbj:
Got very basic partial hydration/islands working. 

- goal: remove as much WASM binary size + hydration cost as possible
- of course for now this will only work with MPA-style navigation (think Astro), with reloads from server on each navigation (because the WASM bundle no longer knows how to render anything that isn't marked as an island)
- but we can later add something like Solid's experimental hybrid/islands routing, where you do a client-side fetch to grab the new HTML partial from the server (partial because based on nested routes), while maintaining the state of the islands currently on the page if they are the same

initial observations
- SSR benchmarks at about 30-40% for the server-only parts of "islands" mode apps, because it does not need to generate hydration keys at all for anything except islands
- in exchange for ~4kb of additional initial binary size to handle the logic to hydrate each island separately, you get the ability to scale the size of "server components" (non-island components) infinitely — in the example I add what amounts to 15kb additional WASM binary size in normal mode, but it adds 0kb in island mode because they're not in the island
- for most apps, where most content is not interactive or can be structured not to be, I'd expect pretty large binary size drops
-  this doesn't handle props or island children yet at all... this is just TBD
- passing context from island to island will depend on 0.5

conclusion so far: extremely worthwhile, will continue pursuing

https://github.com/leptos-rs/leptos/assets/286622/fedfe7c2-e6f5-43c6-8012-ab7305d31963


gbj:
Clearly I'm just happily chugging along on this. A few status updates copied and pasted below from recent Discord conversation, saved here for posterity.
### Hackernews Demo

https://leptos-hackernews-islands.fly.dev/ — 49.6kb of WASM, 4.6kb JS (both brotli-encoded sizes), 70-80% reduction over WASM binary size of traditionally-hydrated version

### Serializing Props
Got support working for serializing/deserializing props to islands, so that islands can receive both children and props from the server. Note also the ease of using server-only APIs (like `std::fs`) directly in server components. This whole demo is just
```rust
#[component]
    // this would panic if it ran in the browser
    // but because this isn't an island, it only runs on the server
    let file = std::fs::read_to_string("./src/is_this_a_server_component.txt").unwrap();
    let len = file.len();

    view! {
        <p>"The starting value for the button is the file's length."</p>
        <Island value=len>
            // `file` is only available on the server
            // island props are projected in, so we can nest
            // server-only content inside islands inside server content etc.
            <p>{file}</p>
        </Island>
    }
}

#[island]
pub fn Island(#[prop(into)] value: RwSignal<usize>, children: Children) -> impl IntoView {
    view! {
        <button on:click=move |_| value.update(|n| *n += 1)>
            {value}
        </button>
        {children()}
    }
}
```

https://github.com/leptos-rs/leptos/assets/286622/92dce3b7-4f96-401b-8833-0de397be168c

### Overview

With this islands feature on, components are only rendered on the server, by default. You can opt into small “islands” of interactivity in this “ocean” of server rendered, non-interactive HTML. 

This allows reducing the WASM binary size by a lot (say ~80% for a typical demo app), making it interactive much faster on loading. 

It also allows you to treat most components as “server components” and use server-only APIs (like std::fs in the example above). 

Adding islands themselves is relatively easy. 

Passing server-rendered children inside the island is slightly more complex but super important because it means you can project server-only content through islands without them needing to know how to render it. 

Passing props to islands from the server was the last missing piece. This involves serializing them into the DOM and then reading them back out — each individual island can be hydrated without needing to walk through the whole state of the app. 

Passing props in this way is also important to the now-plausible goal of something much like the promise of React server components — the user experience of a client-rendered single page app but with most of it actually rendered on the server

### Roadmap

Multi-page app (MPA) + islands
- [x] Get basic islands/partial hydration
- [x] Children of islands and nested islands
- [x] Passing props to islands

Hybrid routing
- [ ]Fetch new HTML page from server and apply it to DOM, without page refresh (see Astro View Transitions newly unveiled)
- [ ] Fetch HTML partials from the server instead, based on nesting routing, and only mutate those small areas of the page. Basically swapping Outlet per Outlet as needed (see eg SolidJS Movies app demo)
- [ ] Maintain the state of individual islands, changing their props as needed. For example, you can be typing in a search bar and swapping out content around it without losing the focus state. (SolidJS movies demo, RSCs)

PR #1423, opened 2023-07-23
Fix warning message about updating a signal after it has been disposed

jasonrhansen:
* Remove "either" because it didn't make sense in the sentence
* Change "cause cause" to "not cause"


gbj:
Thanks!

PR #1427, opened 2023-07-24
feat: allow feature-name flexibility when using server functions

gbj:
The current server-function implementation includes `#[cfg(feature = "ssr")]` in its codegen, which means that it actually checks an `ssr` feature in your application code, not whether an `ssr` feature is enabled for `server_fn_macro` itself. This means that your application is forced to use `ssr` as a feature name and not, for example, `server`. 

This PR instead checks the feature when the proc-macro runs, rather than emitting it into the generated code, which means that it will check which features are enabled on `server_fn_macro`, not in your application. This means you can name your features whatever you want — say `server` and `client` instead of Leptos's go-to `ssr` and `hydrate` — and it will work.

I've tested this by running the `todo_app_sqlite` example as is against the new code, and by replacing the `ssr` and `hydrate` features with `server` and `client`, and it compiles and runs identically in both cases.

@Demonthos tagging you both FYI, in case it requires anything on Dioxus's end, and in case you can see any issues with this.

### DIFF
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -46,6 +46,7 @@ ssr = [
   "leptos_macro/ssr",
   "leptos_reactive/ssr",
   "leptos_server/ssr",
+  "server_fn/ssr",
 ]
 nightly = [
   "leptos_dom/nightly",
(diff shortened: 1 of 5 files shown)


Demonthos:
This looks like a better approach for Dioxus as well!

PR #1430, opened 2023-07-24
Binary search dom stuff

g-re-g:
Since these are all static lists of things, theoretically it should be faster to sort them and then use a binary search rather than `.find` or sequential pattern matches. Binary search should be around O(log n) and find should be around O(n).

### DIFF
--- a/leptos_macro/src/view.rs
+++ b/leptos_macro/src/view.rs
@@ -20,32 +20,19 @@ enum TagType {
     Math,
 }
 
+// Keep list alphabetized for binary search
 const TYPED_EVENTS: [&str; 126] = [
-    "afterprint",
-    "beforeprint",
-    "beforeunload",
-    "gamepadconnected",
-    "gamepaddisconnected",
-    "hashchange",
-    "languagechange",
-    "message",
-    "messageerror",
-    "offline",
-    "online",
-    "pagehide",
-    "pageshow",
-    "popstate",
-    "rejectionhandled",
-    "storage",
-    "unhandledrejection",
-    "unload",
+    "DOMContentLoaded",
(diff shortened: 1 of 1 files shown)


leptos_macro/src/view.rs
```diff
@@ -127,28 +138,20 @@ const TYPED_EVENTS: [&str; 126] = [
...
-    "paste",
-    "fullscreenchange",
-    "fullscreenerror",
-    "pointerlockchange",
-    "pointerlockerror",
-    "readystatechange",
-    "visibilitychange",
 ];
 
+const CUSTOM_EVENT: &str = "Custom";
```
g-re-g, line 153: Used in a few places, figured it might be nice to have it as a const.

leptos_macro/src/view.rs
```diff
@@ -1831,23 +1831,13 @@ fn is_custom_element(tag: &str) -> bool {
...
-            | "img"
-            | "input"
-            | "link"
-            | "meta"
-            | "param"
-            | "source"
-            | "track"
-            | "wbr"
-    )
+    // Keep list alphabetized for binary search
```
gbj, line 1834: Is this actually an improvement? I would've thought that matching on a set list of names would be faster than the search here. I may be wrong, of course. 

ETA: Same applies to the SVG and MathML searches. Although now that I think about it a little more, I guess it's an `O(n)` conditional here (`if name == X else if name == Y else if name == Z`) so a binary search is better, you're right.

gbj (approved):
A merge issue messed up the Makefiles for a couple recent examples but if you rebase the CI should be fine. This all makes sense to me.

PR #1432, opened 2023-07-24
feat: add `noscroll` attribute to `Form`, `ActionForm`

mforsb:
This already works for `a` tags,
 https://github.com/leptos-rs/leptos/blob/daeb47e72e7e0e384fa33db7af2df88da21ace68/router/src/components/router.rs#L418
but unless I missed something there was no equivalent for forms.


gbj:
Awesome, and thanks so much. This looks great.

PR #1435, opened 2023-07-25
Template refactor + snapshot tests

gbj:
This was #1387 but I somehow borked it.


gbj:
@vldm I haven't had time to look at this in depth yet, but did get it all merged and up to date so that everything passes. The snapshot tests here will be very helpful (although they'll change slightly with 0.5.0 with the removal of `cx`/`Scope`)

I'll take a look at it in more depth soonish.

PR #1444, opened 2023-07-26
Convenient event handling with slots

rkuklik:
Hello,

apart from fixing missing reexport I noticed two things:

1. Many struct structs and enums are missing Debug impl
2. There isn't a straightforward way to bundle events with handlers into collection

Is there a reason for not adding Debug? And if not, can I sprinkle in few `#[derive(Debug)]`s?

And as for what I mean by 2, I would like to create some collection to hold different event listeners to be used later:

```rust
#[slot]
pub struct Input {
    #[prop(default = "text".into())]
    itype: TextProp,
    // Event listeners with different types, like input and click
    listeners: Vec<EventListener>
}

#[component]
pub fn FancyForm(cx: Scope, input: Vec<Input>) -> impl IntoView {
    // Attach event listeners to inputs
}
```

However, only way to use EventListener with different types (as far as I know) is to create an enum holding all possible variants. Can I modify generate_event_types macro to also generate this enum, or would that be counterproductive?

```rust
// Something like this
pub enum EventListener {
    Abort(abort, Box<dyn FnMut(UiEvent) + 'static>),
    Input(input, Box<dyn FnMut(Event) + 'static>),
    // And so on, with possible helper IntoEventListener trait
}
```

### DIFF
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -166,8 +166,8 @@ pub use leptos_dom::{
     },
     html, log, math, mount_to, mount_to_body, nonce, svg, warn, window,
     Attribute, Class, CollectView, Errors, Fragment, HtmlElement,
-    IntoAttribute, IntoClass, IntoEventHandler, IntoProperty, IntoStyle,
-    IntoView, NodeRef, Property, View,
+    IntoAttribute, IntoClass, IntoProperty, IntoStyle, IntoView, NodeRef,
+    Property, View,
 };
 
 /// Types to make it easier to handle errors in your application.
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -166,8 +166,8 @@ pub use leptos_dom::{
     },
     html, log, math, mount_to, mount_to_body, nonce, svg, warn, window,
     Attribute, Class, CollectView, Errors, Fragment, HtmlElement,
-    IntoAttribute, IntoClass, IntoProperty, IntoStyle, IntoView, NodeRef,
-    Property, View,
+    IntoAttribute, IntoClass, IntoEventHandler, IntoProperty, IntoStyle,
+    IntoView, NodeRef, Property, View,
 };
 
 /// Types to make it easier to handle errors in your application.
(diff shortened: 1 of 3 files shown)


gbj:
Yeah this looks fine to me. You have a `// TODO` about `Debug` and `Clone` implementations on the enum — Makes sense to me to include these. And thanks for adding the missing `Debug` ones.

The odd spacing in the event names is to correctly CamelCase them in the enum variants, right?

gbj:
The 

> I didn't add `Clone` impl because I would have to constrain the boxed closure to be Clone, which can't be done. So I either have to create a supertrait, or use `fn(Event)`. I haven't used function pointers by themselves, are there any potential future problems they might create if I switch to them? I could also use something like Rc for the closure, but that might be overkill.

Makes perfect sense. `fn(Event)` generally won't work, as function pointers aren't closures so you can't do something like move a signal into them. No need for `Clone` here I don't think.

> I will add `IntoEventHandler` and add some methods, but am bit unsure whether and how to support this in `view!`. My original intention with this PR was to be able to store multiple different event handlers and then attach them with something like:
> 
> ```rust
> view! { cx, <input type="text" on:{/* collection iterating over (event, handler) */} />}
> ```
> 
> But I am unsure if this is something you would want to add.

I don't think there's a way to do it directly in the view macro, but the signature of `HtmlElement::on` (and also `View::on`) could be changed to take something that `impl IntoEventHandler`. For `impl FnMut(<E as EventDescriptor>::EventType) + 'static` this would work the same as current `on`, for collections it would iterate over them. It might not be possible to implement it both for `FnMut` and for `IntoIterator` though, Rust tends to be conservative about conflicting implementations like this.

> NOTE: I also wanted to sprinkle in a few `Debug`s, but there are 36 structs and enums missing `Debug`, so I will add them in a separate PR.

Sounds good.

gbj:
Yeah so for `on`, what do you think of this:

Create an additional trait (`EventCallback` or something, I don't know) with `fn add_event(self, element: HtmlElement<_>)` and implement it for
1) `impl FnMut(E::EventType) + 'static`
2) `EventHandler` (the enum)
3) a couple of common collections (say `Vec<EventHandler>` and `[EventHandler; N]` or anything else you can think of) 

For 1 and 2, it would work exactly like `on` does currently. For 3, it would just iterate over the items and do what `on` does.

Then `on` can take something that `impl EventCallback`. This maintains backwards compat with the current `on` but allows additional types too (`EventHandler` and its collections)

gbj:
Okay, sounds good.

PR #1452, opened 2023-07-27
Added smarter functions for routing with custom handlers

benwis:
Added functions to let custom handlers mind the SsrMode of each route


gbj:
I think the CI issue here is just that this is set up as a PR to `main` but it's assuming the `effect-cleanups` branch which is 0.5 — I think you probably want to make a PR to that branch instead.

benwis:
> I think the CI issue here is just that this is set up as a PR to `main` but it's assuming the `effect-cleanups` branch which is 0.5 — I think you probably want to make a PR to that branch instead.

Ugh, I'm so dumb. I didn't merge with that because it had hundreds of changes, now I realize that's cause my search/replace edited all the book and docs.

PR #1468, opened 2023-07-29
Document the magic number 128 as FILTER_SHOW_COMMENT.

martinfrances107:
This is just a minor nudge, seen whilst reading the code.

'128' make not sense to me until I spent a few minutes googling.

Elsewhere in the code the   https://developer.mozilla.org/ website is used a golden reference for explaining HTML document concepts.


gbj:
Excellent, thanks very much!

PR #1469, opened 2023-07-29
Minor: doclink to Error needed FQDN.

martinfrances107:
Fixed warning  seen whilst running 

```bash
Cargo doc
```

### DIFF
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 /// This is a result type into which any error can be converted,
 /// and which can be used directly in your `view`.
 ///
-/// All errors will be stored as [`Error`].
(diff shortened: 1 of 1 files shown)


gbj:
Hm, interesting. I think it's *supposed* to refer to the `Error` struct defined immediately below. Is it perhaps ambiguous because I see to be importing `thiserror::Error` for no obvious reason?

https://github.com/leptos-rs/leptos/blob/c1991858081c0e8dd7aed590f44ce8b62ebf57fd/server_fn/src/error.rs#L8-L14

martinfrances107:
>  think it's supposed to refer to the Error struct defined immediately below

Opps  I also now  think you are correct 

I will fix this up... one sec..

martinfrances107:
fixed

gbj:
Thanks!

PR #1470, opened 2023-07-29
Update README.md

randommm:
This caused some confusion in my head. I think this was forgotten here since Leptos now is in version 0.4.x.


gbj:
Thanks!

PR #1489, opened 2023-08-03
Fix js scoping

mateusvmv:
The javascript scopes are altering global variables. This creates an issue with the "open" and "close" methods of the javascript window object. This PR simply changes the var declarations to let, which utilize sane scoping rules.


gbj:
Thanks. The `let` version makes it impossible to send multiple `<Suspense/>` fragments because the `let` variables are already declared and can't be declared again, which is why they were `var` in the first place. But then I realized that using an IIFE here with `let` is the proper solution so I just added those function wrappers.

I am genuinely bad at JavaScript (hence creating a Rust frontend framework) but I think this is right now.

PR #1493, opened 2023-08-04
fix: github links

markcatley:


markcatley:
I also noticed `.github/FUNDING.yml` and wondered if that needed to change. I don't know much about how that works so I didn't do anything.

gbj:
Thanks! No, the funding one is correct I think.

PR #1498, opened 2023-08-04
Tell where the source is

stappersg:
Prevent that contributors need to search
which of several git repositories of the Leptos project contains the source of what they are feeling to need changing.


stappersg:
_For What it is Worth:_

This patch should avoid that others have to make same effort has I had to make for #1497.

PR #1510, opened 2023-08-08
feat: support multiple class expressions

tqwewe:
Splits classes by whitespace, preventing a runtime error from occurring when using multiple classes in the tuple syntax `class=("foo bar", maybe)`.

I based this PR off the `effect-cleanups` branch, but if it's preferred to be into `main` directly then I can change it to be so.

This should probably be tested some, though I thought I'd push what I have as a start. It removed runtime errors for me.


gbj:
The `.split_ascii_whitespace()` should be done in `leptos_macro`, I think, not in `leptos_dom`, i.e., it should be done at compile time rather than at runtime so that it doesn't incur any overhead in the default cause (passing only a single class).

PR #1531, opened 2023-08-12
tracing component property

luoxiaozero:
Use for tracing component property.

截屏2023-08-13 01 45 16

Returns a json value when the property is serializable.  If serialization is not possible, the reason for the failure is returned.


luoxiaozero:
Another feature of this PR that is not implemented is to track updates to the signal value.
I came up with two ideas:
1.  When leptos_dom::tracing_props, create an effect to track updates to the signal value.
2.  Modify the tracing of the leptos_reactive library and add the value of the signal when the message is created and updated.

luoxiaozero:
The second one feels better, in addition to implementing props for tracking the component, we can also track the signals created inside the component.

gbj:
This is really cool!

PR #1532, opened 2023-08-12
on_cleanup misorder? in fetch examples

MrNossiom:
Hi :wave:,

while reading examples online, I saw this piece of code. Shouldn't the on_cleanup register be before we wait to fully fetch the API ?

If this is a real issue, I can see if this error has been reproduced in others examples and add a commit to finish this PR.

Thanks


gbj:
It seems to me like you're very obviously correct! Thanks for noticing. The only examples using this same pattern should be the other hackernews examples so if you want to add it to those, that's fantastic.

MrNossiom:
Done!

PR #1548, opened 2023-08-15
Update autoreload websocket connection to work outside of localhost

rabidpug:
Currently we have to access the site served by `cargo leptos watch [--hot-reload]` via localhost, as the websocket connection on the client is tied to the server configuration.

This PR updates the client-side websocket connection to use window.location.protocol/hostname and an additional optional configuration reload_external_port to allow more control over how the client connects to the server. 

This allows us to access via IP other than 127.0.0.1/localhost or a domain via reverse proxy (that can be wss)


gbj:
Thanks very much!

PR #1564, opened 2023-08-21
Remove Clone requirement for slots in Vec

Senzaki:
This PR removes the requirement for `S: Clone` for when rendering a list of slots `S`.

It is currently required because the `view!` macro is building an array of `S`, then calling `to_vec()` (which obviously requires `S: Clone`). This can be a problem when the slot contains `ChildrenFn` (i.e. `Box<dyn Fn(Scope) -> Fragment`), as this type does not implement `Clone`.

This PR removes this requirement by using the `vec!` macro directly instead.

Take this code for example:
```rust
use leptos::{component, mount_to_body, slot, view, ChildrenFn, CollectView, IntoView, Scope};

#[slot]
pub struct MenuItems {
    pub menu_item: Vec<MenuItem>,
}

#[slot]
pub struct MenuItem {
    pub children: ChildrenFn,
}

#[component]
pub fn Menu(cx: Scope, menu_items: MenuItems) -> impl IntoView {
    view! {
        cx,
        <ul>
            {
                menu_items
                    .menu_item
                    .into_iter()
                    .map(|item| view! {
                        cx,
                        <li>{ (item.children)(cx) }</li>
                    })
                    .collect_view(cx)
            }
        </ul>
    }
}

fn main() {
    mount_to_body(|cx| {
        view! {
            cx,
            <Menu>
                <MenuItems slot>
                    <MenuItem slot>"hello"</MenuItem>
                    <MenuItem slot>"world"</MenuItem>
                </MenuItems>
            </Menu>
        }
    });
}
```

With the current version of leptos, it fails to compile with the following message:
```
error[E0277]: the trait bound `MenuItem: Clone` is not satisfied
note: required by a bound in `std::slice::<impl [T]>::to_vec`
```

With this PR, it build properly.

### DIFF
--- a/leptos_macro/src/view/client_builder.rs
+++ b/leptos_macro/src/view/client_builder.rs
@@ -76,18 +76,18 @@ pub(crate) fn fragment_to_tokens(
     let tokens = if lazy {
         quote! {
             {
-                ::leptos::Fragment::lazy(|| [
+                ::leptos::Fragment::lazy(|| ::std::vec![
                     #(#nodes),*
-                ].to_vec())
+                ])
                 #view_marker
             }
         }
     } else {
         quote! {
             {
-                ::leptos::Fragment::new([
+                ::leptos::Fragment::new(::std::vec![
                     #(#nodes),*
-                ].to_vec())
+                ])
                 #view_marker
             }
         }
--- a/leptos_macro/src/view/server_template.rs
+++ b/leptos_macro/src/view/server_template.rs
@@ -86,9 +86,9 @@ pub(crate) fn fragment_to_tokens_ssr(
     });
     quote! {
         {
-            leptos::Fragment::lazy(|| [
+            leptos::Fragment::lazy(|| ::std::vec![
                 #(#nodes),*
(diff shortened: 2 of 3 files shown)


Senzaki:
The CI fail does not seem to be a regression, it might be an ICE because of the version of rust nightly, as it also occurs on the main branch.

Running the following commands fail on both `main` and on my branch ` slot-vec-without-clone`:
```bash
cd leptos_macro/example
cargo +nightly doc
```

It fails in the `bumpalo` dependency, which has not published a new version since May (so it mustn't be this crate's fault either?)

Senzaki:
I have also confirmed it works with a previous version of rust nightly (`cargo +nightly-2023-08-15 doc`), so it is indeed a rust nightly ICE.

gbj:
Thanks for the PR! I'll try to review it this week. Agreed that this is an ICE on most recent nightly, affecting several branches here. I can run CI locally if need be.

gbj:
This looks good to me. Thanks very much.

PR #1579, opened 2023-08-24
feat: start adding some Rustier interfaces for reactive types

gbj:
This library has traditionally had a style that's more idiomatic/familiar to JS devs coming from React or Solid than to typical Rust:
```rust
let (count, set_count) = create_signal(cx, 0);
let double_count = create_memo(cx, move |_| count() * 2);
```
as opposed to
```rust
let count = RwSignal::new(0);
let double_count = Memo::new(move |_| count() * 2);
```
Historically this has been for three reasons
1. `create_signal`'s tuple return doesn't lend itself well to a `::new()` style; it's not returning a struct, after all.
2. The addition of a `cx` variable has made `::new()` seem odd.
3. I largely ported it from SolidJS and its idioms, which are all that classless JS `createSignal()` and `useSomething()` style.

But there are some real advantages to using structs with associated functions and methods:
1. Idiomatic Rust.
2. Better discoverability in docs (easier to find `Memo::new()` than `create_memo` if you're on the page for the `Memo` struct)
3. Reduced duplication of the generic types on functions
4. Easier to create variants like `Resource::once()` rather than increasingly-long function names like `create_resource_once`

**I do not plan to remove the current style or make any breaking changes.** I do plan to start adding more Rust-y idioms for all the same APIs.

Still to do:
- [ ] `Signal::new` (`MaybeSignal::new`, `MaybeProp::new`)
- [x] `StoredValue::new`
- [x] `Effect::run()` (name?)
- [x] `Selector`?
- [ ] others that can get the same treatment?

And there's a lot of beauty in actual idiomatic Rust, no?
<details>
<summary>A Rustier example</summary>

```rust
pub fn counter(initial_value: i32, step: u32) -> impl IntoView {
    let count = RwSignal::new(Count::new(initial_value, step));
    let double_count = Memo::new(move |_| count.get() * 2);

    div()
        .child((
            button()
                .on(ev::click, move |_| count.update(Count::clear))
                .child("Clear"),
            button()
                .on(ev::click, move |_| count.update(Count::decrease))
                .child("-1"),
            p().child(("Value: ", move || count.get().value(), "!")),
            p().child(("Double Value: ", double_count, "!"))
        ))
        .child(
            button()
                .on(ev::click, move |_| count.update(Count::increase))
                .child("+1"),
        )
}

#[derive(Debug, Clone)]
pub struct Count {
    value: i32,
    step: i32,
}

impl Count {
    pub fn new(value: i32, step: u32) -> Self {
        Count {
            value,
            step: step as i32,
        }
    }

    pub fn value(&self) -> i32 {
        self.value
    }

    pub fn increase(&mut self) {
        self.value += self.step;
    }

    pub fn decrease(&mut self) {
        self.value += -self.step;
    }

    pub fn clear(&mut self) {
        self.value = 0;
    }
}
```
</details>

### DIFF
--- a/examples/counter_without_macros/src/lib.rs
+++ b/examples/counter_without_macros/src/lib.rs
@@ -3,7 +3,7 @@ use leptos::{ev, html::*, *};
 /// A simple counter view.
 // A component is really just a function call: it runs once to create the DOM and reactive system
 pub fn counter(initial_value: i32, step: u32) -> impl IntoView {
-    let (count, set_count) = create_signal(Count::new(initial_value, step));
+    let count = RwSignal::new(Count::new(initial_value, step));
 
     // the function name is the same as the HTML tag name
     div()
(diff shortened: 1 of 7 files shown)


rakshith-ravi:
Seems like a relatively simple change. Is there any way I can help?

gbj:
@rakshith-ravi Thanks for the offer — I think it's all set for now, but if there are other APIs you spot that could benefit from a Rust-ier interface feel free to open a PR!

PR #1580, opened 2023-08-24
fix: nightly warning in server macro for lifetime

markcatley:
On the latest lifetime we're getting the following warning in the server macro:
 warning: `&` without an explicit lifetime name cannot be used here
   --> src/login.rs:19:1
    |
 19 | #[server(Login, "/api")]
    | ^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #115010 <https://github.com/rust-lang/rust/issues/115010>
    = note: this warning originates in the attribute macro `server` (in Nightly builds, run with -Z macro-backtrace for more info)

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -281,7 +281,7 @@ pub fn server_macro_impl(
         }
 
         impl #struct_name {
-            const URL: &str = if #fn_path.is_empty() {
+            const URL: &'static str = if #fn_path.is_empty() {
                     #server_fn_path::const_format::concatcp!(
                     #fn_name_as_str,
                     #server_fn_path::xxhash_rust::const_xxh64::xxh64(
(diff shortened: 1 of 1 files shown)


gbj:
Personally, in our case I think this compiler warning is extremely silly. But we will do what we need to please rustc 🙂

As in, the compiler is claiming that this
```rust
const PREFIX: &str = "/api";
```
needs to be this
```rust
const PREFIX: &'static str = "/api";
```
because the lifetime of associated constants has never been formally defined as 'static. Fair enough, but... it's a string literal.

gbj:
Rereading the issue after a cup of coffee I see that they are actually talking about exactly this situation — apparently an associated string literal currently desugars to 
```rust
struct Foo;
impl<'a> Foo {
    const STR: &'a str = "hello, world";
}
```
That makes no sense to me at all, but whatever.

PR #1596, opened 2023-08-26
Callback proposal

rambip:
This is a proposal to include callbacks in leptos 5.0.

It is heavily inspired by leptonic, see [here](https://github.com/lpotthast/leptonic/blob/f8a270ae5512561a92f5deca057f7999940de11b/leptonic/src/callback.rs
)

It has 2 main goals: 
- creating a stable way to design components that need special interactivity given as a prop. Currently, there are multiple approaches to write the same code, a problem framework like Yew don't have, thank's to callbacks.
- removing any boxed closure from the perspective of the user: if it works like a function, the user should be able to write a function.

### DIFF
--- a/leptos_dom/src/callback.rs
+++ b/leptos_dom/src/callback.rs
@@ -5,21 +5,22 @@
 //! You can always create a callback from a closure, but the prefered way is to use `prop(into)`
 //! when you define your component:
 //! ```
+//! # use leptos::*;
+//! # use leptos::leptos_dom::{Callback, Callable};
 //! #[component]
-//! fn MyComponent(#[prop(into)] render_number: Callback<i32, String>) {
+//! fn MyComponent(
+//!     #[prop(into)] render_number: Callback<i32, String>,
+//! ) -> impl IntoView {
 //!     view! {
 //!         <div>
 //!             {render_number.call(42)}
 //!         </div>
 //!     }
 //! }
-//! ```
-//!
-//! That way, you can create it from a closure directly:
-//! ```
+//! // now you can use it from a closure directly:
 //! fn test() -> impl IntoView {
 //!     view! {
-//!         <MyComponent render_number = |x| x.to_string()>
+//!         <MyComponent render_number = |x: i32| x.to_string()/>
 //!     }
 //! }
 //! ```
@@ -41,13 +42,20 @@
 //! All callbacks type defined in this module are [Clone] but not [Copy].
 //! To solve this issue, use [StoredValue]; see [StoredCallback] for more
 //! ```
-//! let callback: Callback<i32, String> = Callback::new(|x| x.to_string);
-//! let stored_callback = store_value(callback);
-//! view! {
-//!     <div>
-//!         {move || callback.call(1)}
-//!         {move || callback.call(42)}
-//!     </div>
+//! # use leptos::*;
+//! # use leptos::leptos_dom::{Callback, Callable};
+//! fn test() -> impl IntoView {
+//!     let callback: Callback<i32, String> =
+//!         Callback::new(|x: i32| x.to_string());
+//!     let stored_callback = store_value(callback);
+//!
+//!     view! {
+//!         <div>
+//!             // `stored_callback` can be moved multiple times
+//!             {move || stored_callback.call(1)}
+//!             {move || stored_callback.call(42)}
+//!         </div>
+//!     }
 //! }
 //! ```
 //!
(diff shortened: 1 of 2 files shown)


leptos_dom/src/callback.rs
```diff
@@ -0,0 +1,245 @@
...
+/// ```
+///
+/// Note that in this scenario, you should use a generic component instead.
+/// You should use callback mainly for optional generic props.
+///
+/// # Cloning
+/// The default [Callback] type can be cloned cheaply but cannot be copied.
+/// If you want to use the callback in a lot of closures, you will have to clone it each time.
+/// To avoid that, see [StoredCallback]
+#[derive(Clone)]
```
gbj, line 42: This should be switched to a manual `Clone` implementation; deriving `Clone` means that `Callback` is only clone if `In` and `Out` are clone, when in fact a manual implementation lets you just clone the inner `Rc`.

leptos_dom/src/callback.rs
```diff
@@ -0,0 +1,245 @@
...
+{
+    fn call(&self, input: In) -> Out {
+        self.with_value(|cb| cb.call(input))
+    }
+}
+
+/// a callback type that
+/// - can be cloned
+/// - is `Send` and `Sync` if the input type is
+#[derive(Clone)]
```
gbj, line 92: See note above about Clone.

gbj (commented):
Yes! This is great. It let me pretty quickly adapt `<Show/>` to the following "show with optional fallback"
```rust
#[component]
    view! {
        <ShowOptFall when=|| { 5 < 3 } fallback=|_| view! { <p>"YES"</p> }>
            <p>"NO"</p>
        </ShowOptFall>
        <ShowOptFall when=|| { 5 < 3 }>
            <p>"NO"</p>
        </ShowOptFall>
        <ShowOptFall when=||{ 5 > 3 }>
            <p>"YES"</p>
        </ShowOptFall>
    }
}

#[component]
pub fn ShowOptFall<W>(
    /// The components Show wraps
    children: Box<dyn Fn() -> Fragment>,
    /// A closure that returns a bool that determines whether this thing runs
    when: W,
    /// A closure that returns what gets rendered if the when statement is false
    #[prop(optional, into)]
    fallback: Option<ViewCallback<()>>,
) -> impl IntoView
where
    W: Fn() -> bool + 'static,
{
    let memoized_when = create_memo(move |_| when());

    move || match memoized_when.get() {
        true => children().into_view(),
        false => match fallback.as_ref() {
            Some(fallback) => fallback.call(()).into_view(),
            None => ().into_view(),
        },
    }
}
```

The only (small) drawback here, in my mind, is the need for a unit struct/empty tuple as the argument, so `fallback=|_| ...` (Unless I've misunderstood something.)

Overall this seems very good to me.

gbj:
P.S. The failing tests etc. are just because of the examples not having imports and so on. When it's ready I can go in and clean those up as well as the docs.

rambip:
> Yes! This is great. It let me pretty quickly adapt `<Show/>` to the following "show with optional fallback"
> 
> ```rust
> #[component]
>     view! {
>         <ShowOptFall when=|| { 5 < 3 } fallback=|_| view! { <p>"YES"</p> }>
>             <p>"NO"</p>
>         </ShowOptFall>
>         <ShowOptFall when=|| { 5 < 3 }>
>             <p>"NO"</p>
>         </ShowOptFall>
>         <ShowOptFall when=||{ 5 > 3 }>
>             <p>"YES"</p>
>         </ShowOptFall>
>     }
> }
> 
> #[component]
> pub fn ShowOptFall<W>(
>     /// The components Show wraps
>     children: Box<dyn Fn() -> Fragment>,
>     /// A closure that returns a bool that determines whether this thing runs
>     when: W,
>     /// A closure that returns what gets rendered if the when statement is false
>     #[prop(optional, into)]
>     fallback: Option<ViewCallback<()>>,
> ) -> impl IntoView
> where
>     W: Fn() -> bool + 'static,
> {
>     let memoized_when = create_memo(move |_| when());
> 
>     move || match memoized_when.get() {
>         true => children().into_view(),
>         false => match fallback.as_ref() {
>             Some(fallback) => fallback.call(()).into_view(),
>             None => ().into_view(),
>         },
>     }
> }
> ```
> 
> The only (small) drawback here, in my mind, is the need for a unit struct/empty tuple as the argument, so `fallback=|_| ...` (Unless I've misunderstood something.)
> 
> Overall this seems very good to me.

Oh you're right I didn't think about that.
Maybe I could implement a specific `From` for `F: Fn() -> View` ?

rambip:
Ok now I'm happy with the docs.
I agree that the ignored argument in `callback = ViewCallback::new(|_| view!{})` could create confusion for the user, but I don't know what to do about it.

Issue #1600, opened 2023-08-27
ActionForm follows redirects from leptos_actix::redirect and tries to deserialize the result causing an error


JonRCahill:
**Describe the bug**
If a server action performs a redirect and is called from an ActionForm submit, then the ActionForm will correctly redirect you to the destination page but it will also:
-  follow the redirect 
- make a new request to the destination 
- attempt to deserialize the result and raise an error

**Leptos Dependencies**

```toml
leptos = { version = "0.5.0-beta" }
leptos_meta = { version = "0.5.0-beta" }
leptos_actix = { version = "0.5.0-beta", optional = true }
leptos_router = { version = "0.5.0-beta" }
```

have also tried with the current version on main.

**To Reproduce**
1. Have 2 routes
2. On the first roue have an ActionForm with a single submit button and have a server function as it's action
3. Handle any errors from the ActionForm action and log them to the console 
4. The server function has a redirect to the other route
5. Click the button within the ActionForm
6. Observe in the console an error logged similar to: `Error: error deserializing server function results: expected value at line 1 column 1`
7. Observe in the network tab the additional fetch to the destination page

Example code:

```
use leptos::*;
use leptos_meta::*;
use leptos_router::*;

#[component]
pub fn App() -> impl IntoView {
    // Provides context that manages stylesheets, titles, meta tags, etc.
    provide_meta_context();

    view! {
        // injects a stylesheet into the document <head>
        // id=leptos means cargo-leptos will hot-reload this stylesheet
        <Stylesheet id="leptos" href="/pkg/leptos_start.css"/>

        // sets the document title
        <Title text="Welcome to Leptos"/>

        // content for this welcome page
        <Router>
            <main>
                <Routes>
                    <Route path="" view=HomePage/>
                    <Route path="/another-page" view=AnotherPage/>
                </Routes>
            </main>
        </Router>
    }
}

/// Renders the home page of your application.
#[component]
fn HomePage() -> impl IntoView {
    let something = create_server_action::<SomeThing>();
    create_effect(move |_| {
        match something.value().get() {
            Some(Ok(success)) => {
                log::info!("Success: {}", success);
            }
            Some(Err(error)) => {
                log::error!("Error: {}", error);
            }
            _ => {}
        };
    });

    log::info!("Home Page");

    view! {
        <h1>"Welcome to Leptos!"</h1>
        <ActionForm action=something>
            <input type="hidden" name="should_work" value="true" />
            <button type="submit">
                "Redirect"
[...]

PR #1604, opened 2023-08-28
Don't try to parse as JSON the result from a server function redirect

JonRCahill:
When a server function redirects using `leptos_actix::redirect` or `leptos_axum::redirect` the result of that redirect will be a page not JSON, but the `ActionForm` `on_response` still attempts to parse it as JSON. This change just adds a guard to not attempt to parse the body of a redirected response.

This should address the error part of #1600


gbj:
Looks good, thanks.

PR #1610, opened 2023-08-28
fix: add docs on #[server] functions

### LINKED ISSUE #1606: API with both POST and GET Methods
**Describe the bug**

Defining both POST and GET methods will return only the GET method.

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.5.0-beta", features = ["nightly"] }
leptos_meta = { version = "0.5.0-beta", features = ["nightly"] }
leptos_router = { version = "0.5.0-beta", features = ["nightly"] }
leptos_axum = { version = "0.5.0-beta" }
```

**To Reproduce**

Using the [`leptos-rs/start-axum-workspace`](https://github.com/leptos-rs/start-axum-workspace) change the `server/scr/main.rs` to:

```rust
#[server(MyFun, "/api", "GetJson", "balance")]
pub async fn balance() -> Result<usize, ServerFnError>{
    Ok(0)
}
#[server(MyFun2, "/api", "Url", "balance")]
pub async fn balance_post() -> Result<usize, ServerFnError>{
    Ok(1)
}

#[tokio::main]
async fn main() {
...
    let app = Router::new()
        .route("/api/*fn_name", post(handle_server_fns))
        .route("/api/*fn_name", get(handle_server_fns)) // Add a `get` as well
        .leptos_routes(&leptos_options, routes, || view! {  <App/> })
        .fallback(file_and_error_handler)
        .with_state(leptos_options);
...
}
```

I've followed the instructions in #1311.

**Expected behavior**

Now do a `curl` using `GET`/`POST`.
It should return `1` for `POST` and `0` for `GET`.

But it does not. The `POST` method is overwritten with the `GET` method somehow:

```bash
> curl -X GET http://localhost:3000/api/balance
0
> curl -X POST http://localhost:3000/api/balance
0 # <----- This should be `1`, right?
```

realeinherjar:
- adds explainer in the `leptos_macros/src/lib.rs` for the requirement that server functions should have unique paths even for different HTTP methods.
- adds a new section in the book chapter on server functions explaining the 4th argument with a more relaxed language than the docs
- Since we have already a section on the 3rd argument and now a section on the 4th, why not add a section also for the 2nd argument? Hence, this was added as well.

Closes #1606.

### DIFF
--- a/docs/book/src/server/25_server_functions.md
+++ b/docs/book/src/server/25_server_functions.md
@@ -74,12 +74,12 @@ There are a few things to note about the way you define a server function, too.
 
 You can optionally define a specific URL prefix to be used in the definition of the server function.
 This is done by providing an optional 2nd argument to the `#[server]` macro.
-By default the URL prefix will be `/`, if not specified.
+By default the URL prefix will be `/api`, if not specified.
 Here are some examples:
 
 ```rust
-#[server(AddTodo)]         // will use the default URL prefix of `/`
-#[server(AddTodo, "/api")] // will use the URL prefix of `/api`
+#[server(AddTodo)]         // will use the default URL prefix of `/api`
+#[server(AddTodo, "/foo")] // will use the URL prefix of `/foo`
 ```
 
 ## Server Function Encodings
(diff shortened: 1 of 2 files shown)


docs/book/src/server/25_server_functions.md
```diff
@@ -105,6 +117,28 @@ In other words, you have two choices:
...
+
+```rust
+#[server(MyServerFnType, "/api", "Url", "hello")]
+```
+
+will generate a server function endpoint at `/api/hello` that accepts a POST request.
+
+> ** Can I use the same server function endpoint path with multiple encodings?**
+>
+> Yes, but server functions must have unique paths.
```
realeinherjar, line 137: Maybe this should be a "No"?

gbj:
Thanks. I've made a commit with some small clarifications/changes for accuracy, for example, on the main branch the default path is `/api`.

PR #1612, opened 2023-08-29
fix(macro/params): clippy warning

Maneren:
I got a `clippy` warning `clippy::redundant_closure_for_method_calls` on the code generated by the `Params` derive macro. This is the fix suggested by `clippy`, which fixes the issue.


gbj:
Thanks!

PR #1613, opened 2023-08-30
Configuration for Hot-Reloading Websocket Protocol and enable ENV PROD selection

Indrazar:
1. Allows the PROD selection for LEPTOS_ENV, previously it was locked to DEV at all times.
2. Adds a new configuration: LEPTOS_RELOAD_WS_PROTOCOL which must be either `ws` or `wss`.
   - cargo-leptos can only perform hot-reloading over non-secure websockets so `ws` is used by default, but this option allows `wss` to be selected when the site is behind a reverse https proxy. The default of `ws` is also useful when your server binary supports https without a reverse proxy.
3. Updates the tests to verify that LEPTOS_ENV: PROD and LEPTOS_RELOAD_WS_PROTOCOL: WSS can be selected if set in the environment variables.

### DIFF
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -11,14 +11,17 @@ fn autoreload(nonce_str: &str, options: &LeptosOptions) -> String {
         Some(val) => val,
         None => options.reload_port,
     };
+    let protocol = match options.reload_ws_protocol {
+        leptos_config::ReloadWSProtocol::WS => "'ws://'",
+        leptos_config::ReloadWSProtocol::WSS => "'wss://'",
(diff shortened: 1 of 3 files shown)


Indrazar:
Please note that actually selecting ENV: PROD at this time has some very strange behavior likely due to it being always DEV previously. This will likely need to be fixed before a release occurs.

gbj:
Thanks. This all looks good. It's a little unclear to me what the `PROD` setting actually does at this point because it looks like the websocket code is looking for a `LEPTOS_WATCH` env variable in any case. @benwis or @aperepel any insight on this?

PR #1614, opened 2023-08-30
docs: Derived signals - Clarified derived signals 

martinfrances107:
I love these docs .. they are well work the read...

This is just a minor thing that broke my concentration 

The double use of the word "**per**" in a sentence made no sense to me 

It read like the units of acceleration meters per second per second...


gbj:
Fair enough, thanks :-)

PR #1617, opened 2023-08-30
doc(book,deployment): update reference to binary in dockerfile

SadraMoh:
It seems like the generated server binary was renamed from `leptos_website` to `leptos_start` at some point.

This PR reflects this change in the example `Dockerfile` under the deployments section in the leptos book.


gbj:
Thanks; to be clear the binary name is whatever the name of your Cargo project is, I think, but `leptos_website` was just from the Leptos website's own Dockerfile so `leptos_start` makes more sense here in any case.

PR #1619, opened 2023-08-31
feat: implement simple spread attributes

mrvillage:
As discussed in Discord, here's a simple PR for some basic spread attributes support for items implementing `IntoIterator<Item= (&'static str, Attribute)>`.

Essentially, you can do the following:
```rs
let attrs = ["value", 1.into_attribute()];
view! { <input {..attrs} /> }
```
and it'll spread the attributes from the iterator into the component as if they were applied normally.

I've tested it and it seems to work correctly, this is my first PR to leptos as well so no idea if I did any of this correctly but please let me know. :)

### DIFF
--- a/leptos_dom/src/html.rs
+++ b/leptos_dom/src/html.rs
@@ -63,7 +63,9 @@ cfg_if! {
 use crate::{
     ev::EventDescriptor,
     hydration::HydrationCtx,
-    macro_helpers::{IntoAttribute, IntoClass, IntoProperty, IntoStyle},
+    macro_helpers::{
+        Attribute, IntoAttribute, IntoClass, IntoProperty, IntoStyle,
+    },
     Element, Fragment, IntoView, NodeRef, Text, View,
 };
 use leptos_reactive::Oco;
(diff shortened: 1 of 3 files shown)


gbj:
Wow this is great, thanks so much. I should've done this ages ago. 

This will also allow me to add something to `<Component/>`s like an `attr:` syntax and also the spread syntax, and interpret both of those as something like "pass a prop named `attributes` which is a `Vec<&'static str, Attribute>` to the component.

So you can have
```rust
#[component]
pub fn Input(attrs: Vec<(&'static str, Attribute)>) -> impl IntoView {
  view! {
    <input class="fancy-input" {...attrs}/>
  }
}

view! { 
  <Input attr:type="text" attr:placeholder="Hello!"/>
}
```
I'll do that in a separate PR of course.

mrvillage:
Thank you for letting me do it! :) And that syntax sounds great, if you'd like I can make a PR for that or something similar, whatever you'd like (it's the exact reason I wanted spread props in the first place so I'm more than happy to implement it :) ).

gbj:
If you want to work on it, that would be awesome! You may have time to get to it sooner than I do.

mrvillage:
Awesome! I'll get something worked out. :)

PR #1623, opened 2023-09-01
docs(book): fix wrong variable name

Gaareth:
I believe the variable should be named `once` not `a`, as only `once` is used in the view macro below.


gbj:
Thanks!

PR #1626, opened 2023-09-01
docs: cleanup

Banzobotic:
Cleaned up whitespace issues with examples in the book. Also added some missing brackets to a function.

I also noticed that `<input type=` messes up the rendering of the examples since type is a rust keyword. I thought of two ways to fix this but both have drawbacks. Adding a zero width space to `type` fixes the issue but would obviously cause problems if someone copy and pastes the code. Another option would be simply removing cases of `type="text"` as this is the default anyway, however this wouldn't fix the issues with `type="checkbox"` or `type="submit"`.

I think that the `type="number"` in the Error Handling section examples should be removed since it prevents typing letters so it makes it more awkward to test the error handling shown off in that section.

### DIFF
--- a/docs/book/src/02_getting_started.md
+++ b/docs/book/src/02_getting_started.md
@@ -65,7 +65,7 @@ And add a simple “Hello, world!” to your `main.rs`
 use leptos::*;
 
 fn main() {
-    mount_to_body(|| view! {  <p>"Hello, world!"</p> })
+    mount_to_body(|| view! { <p>"Hello, world!"</p> })
 }
 ```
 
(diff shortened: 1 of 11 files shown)


gbj:
Thanks very much, this is really helpful!

I agree with your point about the `<input type="number">`, I mostly work in Firefox which I think behaves differently from Chrome here; I remember someone making this point elsewhere and it's fair.

Not sure what to do about `<input type="text">` and syntax highlighting. It's one of the drawbacks of an "RSX"/HTML-mixed-into-Rust syntax like this, for sure.

PR #1628, opened 2023-09-02
feat: Add dynamically resolved attributes

mrvillage:
This is a followup to #1619, everything seems to work as expected when I did some testing, however `cargo test` fails. I think the failures are just it not adjusting the snapshots for the new expected output (I'm new here and have no idea how to fix that, if anyone has any advice that'd be wonderful!).


gbj:
Thanks for this. A couple things:
- Could you give an example of usage? I'm not sure exactly how this is supposed to work.
- Naming: I don't love the `DynAttrs`/`dyn_attrs`/`dyn:` naming. Maybe something like `additional_attributes` or just `attributes`, and `attr:` in the view? (Although I recognize the potential for name conflicts with other user-defined props here)
- I think the reason the tests are failing is that this line will *always* add at least `.dyn_attrs(vec![])` to every component
```rust
    let dyn_attrs = quote! { .dyn_attrs(vec![#(#dyn_attrs),*]) };
```
I'd suggest checking if `dyn_attrs.is_empty()` and just using `quote! {}` in that case so that this does nothing if you don't add additional attributes.

mrvillage:
Agh! I keep forgetting to include examples sorry! I put one to show off the syntax below:
```rs
#[component]
pub fn Input(#[prop(dyn_attrs)] attrs: Vec<(&'static str, Attribute)>) -> impl IntoView {
    view! { <input type="text" {..attrs} /> }
}

<Input dyn:value="hello" dyn:label=label />
```

I'm happy to change the naming if you'd like, I decided to build the naming off the `dyn:` prefix to show that they're dynamic attributes, if you'd rather me used `attr:` and/or `additional_attributes` then let me know!
Also the `.is_empty` check seems to have helped, but it still seems to be having snapshot issues with generated code since it changed the format slightly, not sure what the correct way to address that is sorry.

gbj:
Okay, so I went ahead and renamed to `#[prop(attrs)]` and `attr:`, while leaving the trait and method with the same name. And I think I should have fixed that test -- it was just a question of updating the error message to include `attrs` as a valid prop type, which is perfect.

I removed the import of `use leptos::DynAttr`, which was the other big diff in the snapshot test. If someone is using `use leptos::*`, it's imported automatically, and if not the compiler gives a useful error saying it should be imported. I'm open to putting this back in, or a different way of handling the need to import that trait, though.

gbj:
Oh, I should add, here's a current example:
```rust
#[component]
    view! {
        <Input attr:value="hello" attr:label="foo" />
        <Input attr:type="number" attr:value="0" />
    }
}

#[component]
pub fn Input(
    #[prop(attrs)] attrs: Vec<(&'static str, Attribute)>,
) -> impl IntoView {
    view! {
        <input {..attrs} />
        <pre>{format!("{attrs2:#?}")}</pre>
    }
}
```

mrvillage:
Sounds good, thank you! :)

Issue #1629, opened 2023-09-02
`create_resource` + `<Router>` + `<Transition>` infinite loop bug


sebadob:
**Describe the bug**
Since the `v0.5.0-beta2` (I guess, I never had this problem before) I am experiencing a very annoying bug.
You need the following situation:
1. SSR setup with axum (don't know if it happens with actix too)
2. Leptos Router set up
3. Fetch data in multiple resources from ServerFn's
4. Read these resources in the same `<Transition>` block
5. Finally, get stuck in an inifite loop which crashes the browser when you switch to that route during CSR

**Leptos Dependencies**
Use the `start-axum` example as the base with leptos `v0.5.0-beta2`

**To Reproduce**
I created a minimal example based on the `start-axum` template which creates this bug for me every time.
Basically, just paste the code into the `app.rs`. Everything is perfectly fine when you just hit F5 / refresh multiple times on both routes. My browser fully crashes each time, when I render the page and then navigate to `Page 1` after the hydration on the client side has finished.

You can see it easily if you just pull up a sys monitor and notice that 1 core is being utilized 100% all of the time, until the browser notifies me that the script is broken and I should kill it manually. During this time, everything inside the Transition block is being re-rendered as fast as the CPU can. This does not have any impact on the server side (I tested this with running the backend on another host), this happens only inside the browser.

As soon as you out-comment one of the resource `.get()`s, this does not happen any more.  
Another solution is to wrap a second `<Transition>` inside the first one. I don't know if this is the basic idea, that each resource read should be wrapped in its own transition or not, I never had this problem before. I am not sure if it is linked to the beta2 release.

```rust
use leptos::*;
use leptos_router::*;

#[server]
pub async fn access_server() -> Result<String, ServerFnError> {
    Ok("Hi from Server".to_string())
}

#[server]
pub async fn access_server_second() -> Result<String, ServerFnError> {
    Ok("Hi No. 2 from Server".to_string())
}

#[component]
pub fn App() -> impl IntoView {
    view! {
        <Router fallback=|| "err".into_view()>
            <nav>
                <A href="">"Page 1"</A>
                <A href="/page2">"Page 2"</A>
            </nav>
            <Routes>
                <Route path="" view=|| view! { <Page1/> } />
                <Route path="/page2" view=|| view! { <Page2/> } />
            </Routes>
        </Router>
    }
}

#[component]
pub fn Page1() -> impl IntoView {
[...]

PR #1642, opened 2023-09-03
fix: memoize Suspense readiness to avoid rerendering children/fallback

gbj:
Digging into issue #1629 I found that the switch between children and fallback is being re-triggered every time the Suspense count changes, not only when it switches between loading and readiness. This is a problem if multiple resources are read. I think I've fixed this by just memoizing that check.


sebadob:
I just tested it and this fixes #1629 for me.

gbj:
Excellent, thanks for confirming that.

PR #1647, opened 2023-09-03
Fix Issue in `normalized_call_site` Function when Using `cargo leptos`

s20055232:
While attempting to build my app, I performed the following steps:
1. cargo leptos new --git https://github.com/leptos-rs/start
2. cd project
3. cargo leptos watch

I discovered an issue with the implementation of the `normalized_call_site` function. After removing the parentheses `()`, my code started working as expected. Does this help?
Screenshot 2023-09-03 at 6 43 59 PM


s20055232:
You're right, thank you.

PR #1648, opened 2023-09-03
fix: custom events on components

liquidnya:
**This PR fixes the following bug**
I was using events on components like in the chapter [Parent-Child Communication](https://leptos-rs.github.io/leptos/view/08_parent_child.html#3-use-an-event-listener) of the book:
```rust
<ButtonC on:click=move |_| set_toggled.update(|value| *value = !*value)/>
```
But I was using custom events like this:
```rust
<Modal header="Hello, World!" id="test-modal" on:hidden.bs.modal=|_: Event| log!("modal closed")>
    "This is a test message."
</Modal>
```
Which resulted into a compiler error on `0.5.0-beta2` and also on commit 4a43983f4ed4472d8c34f51a069d4f2c5b4a40e0:
rustc 1.74.0-nightly (4e78abb43 2023-08-28)
```
error[E0423]: expected value, found struct `leptos::leptos_dom::ev::Custom`
```
This is because the macro was generating
```rust
.on(
    ::leptos::leptos_dom::ev::undelegated(
        ::leptos::leptos_dom::ev::Custom,
    ),
    |_: Event| ::leptos_dom::console_log(
        &format_args!("modal closed").to_string(),
    ),
)
```
instead of
```rust
.on(
    ::leptos::leptos_dom::ev::undelegated(
        ::leptos::leptos_dom::ev::Custom::new("hidden.bs.modal"),
    ),
    |_: Event| ::leptos_dom::console_log(
        &format_args!("modal closed").to_string(),
    ),
)
```
**Use case**
My use case for using custom events on components is that I am using the Bootstrap 5.3 [Modal](https://getbootstrap.com/docs/5.3/components/modal/) component.
When I am using this component within my leptos application, the external component is sending a custom event `hidden.bs.modal` whenever the modal is closed and it is very convenient to add the event handler onto the component instead of using a callback property on the component and adding the event handler on the `<div class="modal fade" ...>` inside the component.

### DIFF
--- a/leptos_macro/src/view/mod.rs
+++ b/leptos_macro/src/view/mod.rs
@@ -522,17 +522,7 @@ pub(crate) fn event_from_attribute_node(
 
     let handler = attribute_value(attr);
 
-    #[allow(unused_variables)]
-    let (name, name_undelegated) = parse_event(&event_name);
-
-    let event_type = TYPED_EVENTS
-        .binary_search(&name)
-        .map(|_| (name))
-        .unwrap_or(CUSTOM_EVENT);
-
-    let Ok(event_type) = event_type.parse::<TokenStream>() else {
-        abort!(attr.key, "couldn't parse event name");
-    };
+    let (event_type, _, name_undelegated) = parse_event_name(&event_name);
 
     let event_type = if force_undelegated || name_undelegated {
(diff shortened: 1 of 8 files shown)


gbj:
Looks good, thanks! Always good to remove duplicated logic, especially when it's wrong...

PR #1649, opened 2023-09-04
feat: Static Site Generation

mrvillage:
This PR adds support for upfront static site generation as well as incremental static site generation, alongside public access to the build API to enable runtime builds as people need.
# API
Static site generation is built off of two key components, a new `<StaticRoute />` component (used in the same manner as the `<Route />` component) and the `leptos_router::build_static_routes` function to build all static routes. The two can be combined as shown below.

This will define a new route as a static route, in order to be a static route all parent routes must be static in order to ensure all parent sections of the UI are intended to be static routes (for example, it wouldn't make sense to allow one page of a dashboard to be a static route while the rest are dynamic, in that case the browser would do a full refresh and end up using whatever static defaults were in use for the user's profile picture or other more dynamic components of the dashboard).
```rs
// `mode` defaults to `StaticMode::Upfront`, the only difference is `Incremental` will fall back
// to attempting to build and cache a page for that route instead of redirecting to the 404 page
<StaticRoute mode=StaticMode::Incremental path="/static/:id" view=StaticIdView static_params=static_params=move || Box::pin(async { get_data().await.unwrap() }) />

#[server]
pub async fn get_data() -> Result<StaticParamsMap, ServerFnError> {
    // the map expects a vector of items for each route parameter in the route's path,
    // if any are missing it'll panic noting the path and the missing parameter name
    // when building the route, it will run through all possible combinations of the values
    // provided for all the parameters in the path and then attempt to build a page for every one
    let mut map = StaticParamsMap::new();
    map.insert("id".to_string(), vec![(1).to_string(), (2).to_string()]);
    Ok(map)
}
```
In the main function of the server, the following is required:
```rs
let (routes, static_data_map) = generate_route_list_with_ssg(|| view! { <App /> }).await;
build_static_routes(&conf.leptos_options, || view! { <App /> }, StaticRenderContext::default(), &routes, &static_data_map).await.unwrap();
```
This also comes with support for adding a nice 404 page, this can be done with two simple steps:
- Add the `not_found_path` item to your leptos config (in your `Cargo.toml`)
- Add a `<StaticRoute />` using that path
# To Do
- [x] Allow option for the client-side router to fully rebuild the entire page when routing to a static route (enables static deployment to a CDN without a server involved).
  - Use `rel="external"` on `a` tags to static pages to be fully re-fetched.
- [x] Have a way for the `static_data` prop to only be passed on the server, at present server-only APIs won't work (current thought is an `ssr:` prefix).
- [ ] Figure out documentation.
# Possible Future Improvements
- TTL on static routes to essentially just enable framework-side caching.
- Ability to purge a path from the cache.
- Static components, essentially enables islands to function without a server to back the rendering, rather all possible combinations are pre-rendered up front.


gbj:
Okay, finally had some time to sit down and play around with this. Overall, this looks really good.

Thoughts:

**Implementation**: Looks good!

**API**: It took me a little while to wrap my head around, but basically it seems like `static_data` returns a map between each named param and all its possible values, and this is the primary functionality here: using this set of server-generated static routes to generate pages. If rendering a page depends on some server data (I want to render a blog post, so I pull the blog post from the DB) is the way to do this is using Suspense and a server function as usual? I wonder if renaming the `static_data` prop to `static_params` might help here.

I think the `methods` prop on the `StaticRoute` component can be dropped, this should really only be GET, right?

**Bugs**: I did notice a couple bugs/unclear behaviors:
- If I define `Upfront` routes at, say, `/upfront/:id`, and I give `vec![1.to_string(), 2.to_string()]` as the `static_data` param, if I access another route like `/upfront/3` I just get an `Internal Server Error` rendered in the body. What's the 404/fallback behavior here?
- If I define `Incremental` routes at at, say, `/inc/:id`, and I give `vec![1.to_string(), 2.to_string()]` as the `static_data` param, I can still access `/inc/3` and `/inc/4` and they generate reasonable data. However, if I then change the template and restart the server, if I visit `/inc/1` and `/inc/2` they update to the new template... but if I revisit `/inc/3` or `/inc/4` they serve the old HTML. i.e., incremental generation for a value of a param that's not in the `static_data` works for creating an initial HTML file (unlike `Incremental`, which gives the error above), but doesn't update it.

***

I think this is really good work and will be much appreciated. Especially combined with the islands stuff this is pretty much a full static site generator, so... well done!

gbj:
One additional thought: it's useful to be able to mix static and non-static nested routes for things like layouts. Is it possible to mix them or would the parent route have to be a StaticRoute with an empty params map?

mrvillage:
Awesome! Gonna just write my response in bullet points so it doesn't become a convoluted paragraph:
- Yup! `static_data` is just the function that'll return the `StaticParamsMap` that'll be used to resolve routes, to get data like the full blog post for example you'd just do the same thing as now since the full route is just rendered as if it was in `Async` mode, so functionally the only difference is it's rendered exactly once instead of on every request without any other environment changes. If you want me to rename it to `static_params` I'm happy to do so!
- I just dropped `methods` yeah, it should only ever be GET.
- Oh I see what I screwed up, so it currently assumes you've provided a 404 path in your `Cargo.toml`, I just fixed it so that it'll at least send the text "Not Found" instead of "Internal Server Error" if it can't find the path.
- Oh that's weird, hmm. The way it does it now it generates everything from `static_data` every time you call build, I added a `purge_all_static_routes` function to purge all generated files (they're now `.static.html` files instead of just `.html` files to allow for figuring out which files were generated and which were placed there by the user).

mrvillage:
Also for the mixing, right now I have it that it requires all parents to be static, reason being it's meant to refetch the entire HTML page when it makes a route change (in the future I'd love to add static HTML partials to go nicely with islands but that felt a bit out of scope given I started this when islands weren't a thing and islands are only experimental) and if the user marks the parent as static then that's their assertion that "this route doesn't have dynamic things that could get screwed up by making it static. It does however allow for dynamic children of static routes, but not the other way around (i.e. I can set the main layout part of my site to be static, then have some children be fully static, like the landing page, but then others have some dynamic stuff, like a pricing page that'll check for active sales when rendering). I'm tried but I'm not sure how to get the client-side router to do a full HTML refresh then re-mount itself on route change though, so would love some help on that! :)

gbj:
I ran into two small issues while testing:
1) It was generating all routes up front, whether `Upfront` or `Incremental`
2) Routes that began with `/`, when joined to the `target/site` or similar, were ending up as eg `/static/1` rather than `target/site/static/1` so I trimmed this from the front

srid:
Is there a demo app that uses this? Would love to take a look.

mrvillage:
There's not but I should probably make one! If you copy the couple snippets in the description you should be able to get it up and running instantly in any of the examples.

srid:
> There's not but I should probably make one!

Would be nice have a template repository with Github Actions / Github Pages deployment event. Like https://github.com/srid/emanote-template

lukexor:
Has anyone tried to use this on routes hat don't need `static_params`? The docs say the prop is optional, but if you exclude it, Rust can't infer the correct type. e.g.

```
<StaticRoute path="" view=Homepage />
```

```
  --> src/portfolio.rs:47:22
   |
47 |                     <StaticRoute path="" view=Homepage />
   |                      ^^^^^^^^^^^ cannot infer type of the type parameter `S` declared on the function `StaticRoute`
   |
help: consider specifying the generic arguments
   |
47 |                     <StaticRoute::<_, _, &str, S> path="" view=Homepage />
   |                                 +++++++++++++++++

error[E0283]: type annotations needed
   --> src/portfolio.rs:47:22
    |
47  |                     <StaticRoute path="" view=Homepage />
    |                      ^^^^^^^^^^^ cannot infer type of the type parameter `S` declared on the function `StaticRoute`
    |
    = note: multiple `impl`s satisfying `_: std::ops::Fn<()>` found in the following crates: `alloc`, `core`:
            - impl<A, F> std::ops::Fn<A> for &F
              where A: std::marker::Tuple, F: std::ops::Fn<A>, F: ?Sized;
            - impl<Args, F, A> std::ops::Fn<Args> for std::boxed::Box<F, A>
              where Args: std::marker::Tuple, F: std::ops::Fn<Args>, A: std::alloc::Allocator, F: ?Sized;
note: required by a bound in `leptos_router::StaticRoute`
   --> /home/luke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/leptos_router-0.5.0/src/components/route.rs:182:8
    |
155 | pub fn StaticRoute<E, F, P, S>(
    |        ----------- required by a bound in this function
...
182 |     S: Fn() -> Pin<Box<dyn Future<Output = StaticParamsMap>>> + 'static,
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `StaticRoute`
help: consider specifying the generic arguments
    |
47  |                     <StaticRoute::<_, _, &str, S> path="" view=Homepage />
    |                                 +++++++++++++++++
```

PR #1656, opened 2023-09-05
fix: compare path components to detect active link in router

### LINKED ISSUE #1655: is_active detection in router's A component incorrect
**Describe the bug**
I stubmled across this when I added a condition style class with the `aria-current=page` attribute. In some cases, more than one link in the navigation bar was highlighted. This happens when there are two (or more) routes like `/foo/test` and `/foo/test_bar`. This is because the latter one starts with the first one. This is what is checked to determine `is_active` [here](https://github.com/leptos-rs/leptos/blob/495862e9f99d94ebb30d0b065a9d650e6c15fe2e/router/src/components/link.rs#L136C2-L136C2).

**Leptos Dependencies**
```toml
leptos = { version = "0.4.10", features = ["csr", "nightly"] }
leptos_meta = { version =  "0.4.10", features = ["csr", "nightly"] }
leptos_router = { version = "0.4.10", features = ["csr", "nightly"] }
```

**To Reproduce**
1. Create two routes, `/foo/test` and `/foo/test_bar`.
2. Use the `aria-current=page` attribute to conditionally change the text color of the link in the navigation bar.
3. Click on the link to `/foo/test_bar`.
4. Both links are considered active and change their color.

**Expected behavior**
`/foo/test` and `/foo/test_bar` should be considered different routes that cannot be active at the same time.

flo-at:
With this chage, the two path fragments are compared by their components, not just as strings.
This should fix  #1655.
I'm note sure if this covers each and every corner case though.

### DIFF
--- a/router/src/components/link.rs
+++ b/router/src/components/link.rs
@@ -133,7 +133,8 @@ where
                         if exact {
                             loc == path
                         } else {
-                            loc.starts_with(&path)
+                            std::iter::zip(loc.split('/'), path.split('/'))
+                                .all(|(loc_p, path_p)| loc_p == path_p)
                         }
                     })
                 })


gbj:
Tested it out with different scenarios and this looks good to me.

For comparison, here's Solid Router, which is what Leptos Router is ported from. Looks like they've switched from `exact` to `end` but would have the same issue
```js
  const isActive = createMemo(() => {
    const to_ = to();
    if (to_ === undefined) return false;
    const path = normalizePath(to_.split(/[?#]/, 1)[0]).toLowerCase();
    const loc = normalizePath(location.pathname).toLowerCase();
    return props.end ? path === loc : loc.startsWith(path);
  });
```
([source](https://github.com/solidjs/solid-router/blob/79a2b577564772b8248aa75c5300dcb61e650e59/src/components.tsx#L226))

Here's React Router, similar in many ways. Here they get around it by checking if the last character is `/` (so, for example, `/foo` and `/foo_bar` don't match)
```js
    let isActive =
      locationPathname === toPathname ||
      (!end &&
        locationPathname.startsWith(toPathname) &&
        locationPathname.charAt(toPathname.length) === "/");
```
([source](https://github.com/remix-run/react-router/blob/4b494b935d62cd1244fe5c091db920d3f0315e9e/packages/react-router-dom/index.tsx#L661-L665))

The iterator-based solution is nicer, I think, than either of these. I couldn't find any edge cases, but I'm sure we'll get some nice bug reports whenever they appear 😄 

Thanks very much.

flo-at:
Thanks for the quick response and merge!

> Here's React Router, similar in many ways. Here they get around it by checking if the last character is `/`

That was my first thought, too. But then I came up with the example I put into the issue which still wouldn't work.

Keep up the great work! I'm finally not hating front-end development anymore thanks to leptos :smile:.

PR #1660, opened 2023-09-06
feat: islands

gbj:
## Overview

This PR adds an `experimental-islands` feature that opts you into a rendering mode that's different from the traditional server-rendered/client-hydrated model described [here](https://leptos-rs.github.io/leptos/ssr/22_life_cycle.html), in which the entire page requested is rendered to HTML on the server for the first request, and then runs on the client to hydrate this server-rendered HTML, and subsequent navigations take place by rendering pages in the browser.

With this islands feature on, components are only rendered on the server, by default. Navigations follow the traditional/multi-page-app (MPA) style of navigating by fetching a new HTML page from the server. The name "islands" comes from the concept of the "islands architecture," in which you opt into small “islands” of interactivity in an “ocean” of server-rendered, non-interactive HTML.

This allows reducing the WASM binary size by a lot (say ~80% for a typical demo app), making the time-to-interactive for a page load much faster. It also allows you to treat most components as “server components” and use server-only APIs, because those plain components will never need to be rendered in the browser.

Importantly, the same size optimizations apply not only to WASM binary size, but to
1) hydration markers in HTML 
2) data for resources

Hydration markers are only added for HTML that is part of an island. Likewise, resource data is only serialized and sent to the client for resources that are read inside an island. If a resource is read in a Suspense that's not in an island, that resource data will not be serialized and sent to the client.

Taken together, this means that page loads can be much faster, as the overall HTML file sent is significantly shorter.

## Using Islands

When you activate the `experimental-islands` feature, every `#[component]` is server-only by default. This "default to server" behavior is important: you opt into shipping code to the client, rather than opting out. You can opt into client-side interactivity for any given component by changing from `#[component]` to `#[island]`—the two macros are otherwise identical.

Everything that is included inside an island will be compiled to WASM and shipped to the browser. So the key to really benefiting from this architecture is to make islands as small as possible, and include only the minimal required amount of functionality in islands themselves.

Remember, only code included in an island itself is compiled to WASM. This means:
1. `children` can be provided from a server `#[component]` to an `#[island]` without the island needing to be able to hydrate them/needing to compile them to WASM.
2. props can be given from the server to an island.

It's probably easiest to illustrate with an example from our `hackernews_islands_axum` demo. 

https://github.com/leptos-rs/leptos/blob/bfbb6ffcfb219ec2be082d0d7c868e326f67c001/examples/hackernews_islands_axum/src/routes/story.rs#L91-L115

Here, the `<Toggle/>` island manages only an extremely simple process: it toggles a single signal that hides or shows some other content.

It's used from our `Comment` component like this:
https://github.com/leptos-rs/leptos/blob/bfbb6ffcfb219ec2be082d0d7c868e326f67c001/examples/hackernews_islands_axum/src/routes/story.rs#L126C1-L134

The `comment.comments` data is projected into the `children` of the `Toggle`, but it is not known by the `Toggle` itself. This means that none of the resource data needs to be serialized.

## Present Limitations 

A few noteworthy limitations, at the moment:
1. Context does work between islands, but I have not figured out a good story for how to use `leptos_router` hooks yet, so things like `use_query_map` won't work inside islands at the moment. (They do work in regular server components.)
2. `children` are completely opaque in islands. You can't iterate over children; in fact they're all bundled into a single `<leptos-children>` HTML element. 
3. Similarly, `children` need to be used in the HTML rendered on the server. You'll notice the semi-awkward `display: none` rather than `<Show>` used to show/hide the children in the toggle. This is because the children aren't serialized at all, other than as HTML: if that HTML isn't present in the DOM, even if hidden, it is never sent and not available to the client at all.

## Future Work

There's lots that can be done to make the routing feel more "SPA-like," in the following order:
- Fetch new HTML page from server and apply it to DOM, without page refresh (see Astro View Transitions recently unveiled)
- Fetch HTML partials from the server instead, based on nesting routing, and only mutate those small areas of the page. Basically swapping Outlet per Outlet as needed (see e.g., SolidJS Movies app demo)
- Maintain the state of individual islands, changing their props as needed. For example, you can be typing in a search bar and swapping out content around it without losing the focus state. (SolidJS movies demo, RSCs)

## API Stability

I've labeled this feature `experimental-islands`. I'm fairly confident at this point that islands won't require additional breaking changes in 0.5 itself. However, the islands APIs may change, particularly to accommodate the routing concerns as they develop, and the `leptos_router` and other hooks mentioned above.

The `experimental-` on the feature is intended as a "user, beware." I reserve the right to break parts of the API that are covered by this feature.

## A Fun Demo

Just to close with a fun little demo: If you've ever tried to use `std::fs` in a component before, you'll know it panics in the browser. Not so in an island! Here we read from a file on the server side and pass the data directly into an island in the browser as the starting value for an interactive button.

This demo is silly on its own, but it points to some extremely powerful possibilities.

```rust
#[component]
    // this would panic if it ran in the browser
    // but because this isn't an island, it only runs on the server
    let file = std::fs::read_to_string("./src/is_this_a_server_component.txt").unwrap();
    let len = file.len();

    view! {
        <p>"The starting value for the button is the file's length."</p>
        // `value` is serialized and given to the island as a prop
        <Island value=len>
            // `file` is only available on the server
            // island props are projected in, so we can nest
            // server-only content inside islands inside server content etc.
            <p>{file}</p>
        </Island>
    }
}

#[island]
pub fn Island(#[prop(into)] value: RwSignal<usize>, children: Children) -> impl IntoView {
    // because `RwSignal<T>` implements `From<T>`, we can pass in a plain 
    // value and use it as the starting value of a signal here
    view! {
        <button on:click=move |_| value.update(|n| *n += 1)>
            {value}
        </button>
        {children()}
    }
}
```

https://github.com/leptos-rs/leptos/assets/286622/92dce3b7-4f96-401b-8833-0de397be168c


g2p:
There's no place where islands are documented at the moment, but this PR is linked from the release notes, so I'm going to give a few hints here.  To enable islands:

Add the `experimental-islands` feature to the leptos dependency (as well as the integration dependency, like leptos_axum).

More importantly, replace the call to `leptos_dom::mount_to_body` (at the end of the lib.rs hydrate function) with `leptos::leptos_dom::HydrationCtx::stop_hydrating();`.

Symptoms if you don't do that: when trying to navigate a link, most of the work to render the new view client-side happens, but the DOM is never updated with the new state, and the app appears stuck.

g2p:
Maybe the component code could complain about trying to render a component at the client side (through mount_to or otherwise).  Something in leptos_dom will discard the result anyway.

PR #1661, opened 2023-09-06
Remove unnecessary `move ||`?

jmadotgg:
Why do we need the `move ||` here? Doesn't the outer closure already make sure that `errors` is up to date?  I could not notice any difference between the two running this example locally ... I'm also quite new to Leptos though.

### DIFF
--- a/docs/book/src/view/07_errors.md
+++ b/docs/book/src/view/07_errors.md
@@ -143,7 +143,7 @@ fn App() -> impl IntoView {
                         // we can render a list of errors
                         // as strings, if we'd like
                         <ul>
-                            {move || errors.get()
+                            {errors.get()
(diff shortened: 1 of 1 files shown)


gbj:
`errors` here is a signal, so this means that it will rerender if the set of errors contained in that signal changes. `fallback=|errors|` is a function that takes that signal -- it's not a reactive closure.

So, good thinking, but I think the book is correct!

(It won't make a difference in this particular example if the set of errors doesn't change, but an ErrorBoundary can catch multiple different errors from multiple different sources)

jmadotgg:
Thank you, now it's clear to me. So the `move ||` basically makes sure, that additional errors can be shown when the fallback is already beiing displayed.

PR #1669, opened 2023-09-07
`Rc` backed `ChildrenFn`

Baptistemontan:
Currently components only accept `Box` backed children, it makes sense for `Children` and `ChildrenFnMut` but makes working with `ChildrenFn` difficult as it can't be cloned, so the solution most user would take is to put it in a `Rc` but it introduce more code and unecessary allocations.

This PR change `ChildrenFn` to be `Rc` backed, so component taking `ChildrenFn` as an argument can now clone it directly with minimal cost.

### DIFF
--- a/leptos/src/children.rs
+++ b/leptos/src/children.rs
@@ -25,6 +25,7 @@ impl<F> ToChildren<F> for Children
 where
     F: FnOnce() -> Fragment + 'static,
 {
+    #[inline]
     fn to_children(f: F) -> Self {
         Box::new(f)
     }
@@ -34,6 +35,7 @@ impl<F> ToChildren<F> for ChildrenFn
 where
     F: Fn() -> Fragment + 'static,
 {
+    #[inline]
     fn to_children(f: F) -> Self {
         Rc::new(f)
     }
@@ -43,6 +45,7 @@ impl<F> ToChildren<F> for ChildrenFnMut
 where
     F: FnMut() -> Fragment + 'static,
 {
+    #[inline]
     fn to_children(f: F) -> Self {
         Box::new(f)
     }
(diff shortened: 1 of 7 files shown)


leptos/src/suspense_component.rs
```diff
@@ -61,14 +61,14 @@ pub fn Suspense<F, E, V>(
...
     /// Children will be displayed once all `async` [`Resource`](leptos_reactive::Resource)s have resolved.
-    children: Box<dyn Fn() -> V>,
+    children: Rc<dyn Fn() -> V>,
 ) -> impl IntoView
 where
     F: Fn() -> E + 'static,
     E: IntoView,
     V: IntoView + 'static,
 {
-    let orig_children = Rc::new(children);
```
gbj, line 71: Nice! Clearly I'd already encountered the same issue and just... added another `Rc<_>` instead of dealing with it :smile:

gbj (approved):
Yes this looks great to me, and I love that it seems to work on all the current examples, so any actual changes to user code should be very minimal.

leptos/src/suspense_component.rs
Baptistemontan: Yeah this is clearly the kind of situation this PR is trying to solve, I just came across this one but there is maybe other built in components that could take advantages of this. I might look into it

leptos/src/children.rs
```diff
@@ -0,0 +1,58 @@
...
+#[doc(hidden)]
+pub trait ToChildren<F> {
+    fn to_children(f: F) -> Self;
+}
+
+impl<F> ToChildren<F> for Children
+where
+    F: FnOnce() -> Fragment + 'static,
+{
+    fn to_children(f: F) -> Self {
```
Baptistemontan, line 28: Would `#[inline]` be a good idea here (and for other impl)? As it's just a wrapper function it may be nice

gbj: Yes I think that makes sense.

PR #1670, opened 2023-09-08
Remove (most) syn 1 dependencies

blorbb:
Most of the syn 1 dependencies have been removed. The main culprits are `proc-macro-error` with the default `syn-error` feature, and `educe`. An upstream crate was updated to also remove default features from proc-macro-error. Educe was not used much, so manual implementations were added (I believe they should be identical implementations, but not 100% sure).

Currently, the last remaining dependency on syn 1 is [attribute-derive](https://github.com/ModProg/attribute-derive), again from proc-macro-error default features. They have recently removed proc-macro-error entirely on their git repo, but this has not been added in a new release yet. Not sure whether this change will be released as a patch or minor version bump, so further updates may be required once they make the new release.


gbj:
Great, thanks! Yes, these manual implementations instead of educe look right to me.

PR #1673, opened 2023-09-08
Router version bump

martinfrances107:
Saw a few outdated package while reading the code

-cached = { version = "0.44.0", optional = true }
+cached = { version = "0.45.0", optional = true }

-lru = { version = "0.10", optional = true }
+lru = { version = "0.11", optional = true }

NB gloo_net needs updating but that would take bigger refactoring as the raw response is now private.


gbj:
Thanks very much!

PR #1674, opened 2023-09-08
chore: removed indentation error while building book

martinfrances107:
when I run mdbook build

```
An error occured:
Indentation error on line:

  - [A Basic Component](./view/01_basic_component.md)
```

I am not sure if this is a interporability issue
or if mdbook has started to rejected indentation flaw.

But I find I need to use 4 spaces, where as 2 spaces we previously ok.


gbj:
Weird. I just tried updating `mdbook` and the two-space version still works. What `mdbook --version` do you have? (Mine is `0.4.34`)

martinfrances107:
Ah nuts,  the problem is a my end... I was on 0.0.28.

 forced a reinstall and now I am a happy bunny.

PR #1678, opened 2023-09-08
Feedback Wanted — feat: add `.write()` and `.read()` for signals

gbj:
## The Problem 

Historically, we have not been able to directly expose (immutable or mutable) references to the values of signals. This is fine for `Copy` types or values you're willing to clone, because you can access their values using getters:
```rust
let (a, set_a) = create_signal(1);
let (b, set_b) = create_signal(2);
let c = move || a() + b();
```
However, for values that are more expensive to clone, this is not ideal:
```rust
let (first, set_first) = create_signal("John".to_string());
let (middle, set_middle) = create_signal("Roberts".to_string());
let (last, set_last) = create_signal("Smith".to_string());
// clones all three Strings unnecessarily
let full_name = move || format!("{} {} {}", first(), middle(), last());
```
The `.with()` function allows you to work on references to signal values, but quickly leads to deep nesting, which has implications on both formatting and code readability
```rust
// this is kind of gross
let name = move || {
    first.with(|first| {
        middle.with(|middle| last.with(|last| format!("Full name is {first} {middle} {last}")))
    })
};
```
It would be very nice to be able to write something like this, instead:
```rust
let name = move || {
    let first = first.read(); // returns a reference, without cloning
    let middle = middle.read();
    let last = last.read();
    format!("{first} {middle} {last}")
};
```

## Proposed Solution

This PR introduces `SignalRead` and `SignalWrite` traits that allow you to access the inner values, either mutably or immutably, using the same API as a `RefCell`. (This includes reorganizing how signal values are stored to make that possible, which had previously been impossible.)
```rust
let (value, set_value) = create_signal(vec![0, 1, 2]);
let value_ref: Ref<'_, Vec<i32>> = value.read();
let value_ref_mut: RefMut<'_, Vec<i32>> = set_value.write();
```
These work reactively: `.read()` registers the access with the nearest effect or memo, just like `.get()` or `.with()` does. `.write()` notifies subscribers that the value has changed, and reruns any effects accordingly.

> Timing this for `.write()` is slightly complicated. You know that the values are done updating at the end of the `.update()` closure or the `.set()` call. But because `.write()` returns a `RefMut`, the signal value has not actually been updated by the end of `.write()`, so you can't run effects yet with the new value. Currently, I have this scheduled with a `queueMicrotask()`. Alternately, implementing a wrapper for `RefMut` that notifies on or after a `DerefMut` could work.

### Pros
- It's nice to be able to work directly with references (or `std` library reference-like RAII types, in any case.)
- Reduced nesting of `.with()` closures
- Encourages better practices: rather than simply `.get()`ing a `String` for convenience when you only need a reference, you can `.read()` it with the same ergonomics
- Small reduction in WASM binary size because of rewritten signal storage system

### Cons
- Exposes users to the same set of possible panics you get with `RefCell`. This example was there to show the types, but it actually panics: the `.read()` guard is still held as an immutable reference when you try to borrow the same value mutably with `.write()`. Users need to be careful when doing things like reading and writing in the same lexical scope, or holding a `RefMut` across a `.await`.
```rust
let (value, set_value) = create_signal(vec![0, 1, 2]);
let value_ref: Ref<'_, Vec<i32>> = value.read();
let value_ref_mut: RefMut<'_, Vec<i32>> = set_value.write();
```
- Uses our own bespoke arena implementation, which is shared thread-locally across runtimes, rather than a single slotmap per runtime. This means we would need to do some testing to make sure that there are no memory leaks and no leakage of signal values across runtimes (i.e., across requests) on the server. I'm not too worried about this but replacing a well-known crate with your own solution is always somewhat risky.

### DIFF
--- a/examples/counter/src/lib.rs
+++ b/examples/counter/src/lib.rs
@@ -1,4 +1,5 @@
-use leptos::*;
+use leptos::{SignalWrite, *};
+use std::cell::{Ref, RefMut};
 
 /// A simple counter component.
 ///
@@ -12,12 +13,20 @@ pub fn SimpleCounter(
 ) -> impl IntoView {
     let (value, set_value) = create_signal(initial_value);
 
+    let something: Ref<'_, i32> = value.read();
+
+    spawn_local(async move {
+        let mut something_else: RefMut<'_, i32> = set_value.write();
+        async {}.await;
+        *something_else = 30;
+    });
+
     view! {
         <div>
             <button on:click=move |_| set_value(0)>"Clear"</button>
-            <button on:click=move |_| set_value.update(|value| *value -= step)>"-1"</button>
+            <button on:click=move |_| *set_value.write() -= step>"-1"</button>
             <span>"Value: " {value} "!"</span>
-            <button on:click=move |_| set_value.update(|value| *value += step)>"+1"</button>
+            <button on:click=move |_| *set_value.write() += step>"+1"</button>
         </div>
     }
 }
--- a/leptos_reactive/Cargo.toml
+++ b/leptos_reactive/Cargo.toml
@@ -8,6 +8,7 @@ repository = "https://github.com/leptos-rs/leptos"
 description = "Reactive system for the Leptos web framework."
 
 [dependencies]
+elsa = "1"
 slotmap = { version = "1", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde-lite = { version = "0.4", optional = true }
--- a/leptos_reactive/src/arena.rs
+++ b/leptos_reactive/src/arena.rs
@@ -1,0 +1,161 @@
+use elsa::FrozenVec;
+use std::{
+    any::Any,
+    cell::{Cell, Ref, RefCell, RefMut},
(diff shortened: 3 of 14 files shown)


blorbb:
imo the possibilities of panics outweigh the ergonomics here. especially on a more complicated project, if references are being held in all sorts of places, it can be difficult to track what is causing a panic. The existing `with` method is great because the reference is only active in one place. The issue of nested `with`s could probably be solved with a simple macro as well (actually I might make one).

if you go ahead with this, maybe call the methods `borrow` and `borrow_mut` instead, to make it clearer that it is using a RefCell.

mscofield0:
I like the macro approach @blorbb proposed, for the same reasons he proposed them.

jquesada2016:
Yeah, a macro might very well be preferrable, because `RefCell` is hell to debug, especially when you can have signals that can trigger effects in often times unpredictable ways. It's nearly impossible to hold your apps' signal tree in your head.

gbj:
Closing as "noped by community" but it's good to know it's available for the future, if we ever want to go in this direction 😄

PR #1688, opened 2023-09-10
fix: broken `mount_to_body` in CSR mode

gbj:


hayley:
I added the following to my `Cargo.toml` and can confirm that it fixed the blank page render issue for me.

`leptos = { git = "https://github.com/leptos-rs/leptos", features = ["csr", "nightly"], branch="fix-csr"}`

Thanks for the quick turnaround!

Issue #1682, opened 2023-09-09
`with!` macro for signals


blorbb:
**Is your feature request related to a problem? Please describe.**

Using `signal.with` is cumbersome and takes a lot of writing.
This is especially true for using multiple signals, e.g:
```rust
let first = create_rw_signal("Greg".to_string()); let middle = ...; let last = ...;
let name = move || first.with(|first| middle.with(|middle| last.with(|last| format!("{first} {middle} {last}"))));
```

Making it easier to use `with` would also encourage less cloning.

**Describe the solution you'd like**

A `with!` macro that makes it a little easier to use `Signal::with`.

Not sure about the syntax, maybe something like this:
```rust
with!(|first, middle, last| format!("{first} {middle} {last}"))
```
That would expand to the same thing as the manual implementation above.

The identifiers `first` `middle` and `last` are captured from the environment, and can be used in the closure's body with the same identifier.

**Pros**: Closure-like syntax reflects the closure used in `Signal::with`. The variables that are defined in the context of the closure are clear. Might also be able to be formatted with rustfmt? (not sure how it works with macros)
**Cons**: It might be too closure-like, it's not clear that first, middle and last are captured from the environment, rather than arguments of a closure.

An alternative could be:
```rust
with!(first, middle, last; format!("{first} {middle} {last}"))
```
Where the captured idents and the body of the closure are separated by a `;`.

**Pros**: It is clearer that this is not an ordinary closure.
**Cons**: don't really like how this looks.

A corresponding macro for `Signal::update` could also be made.

For both of these syntaxes, only identifiers would be accepted (so for something like `settings.big_thing.with(|thing| ...)`, you would need to `let thing = settings.big_thing` first).
I don't think this is much of an issue, as you could do something like
```rust
let something = {
    let thing = settings.big_thing;
    with!(|thing| ...)
}
```

**Describe alternatives you've considered**

[`.write()` and `.read()` for signals](https://github.com/leptos-rs/leptos/pull/1678). I personally don't like the idea of having `Ref` or `RefMut`s that could be used anywhere, as it can lead to panics.

**Additional context**
N/A

I can work on this feature if you like the idea.

gbj:
I *believe* that this is fairly easy to implement as a simple, declarative `macro_rules!` macro, and I quite like it. I agree with you that this syntax is preferable if its possible:
```rust
with!(|first, middle, last| format!("{first} {middle} {last}"))
```
[...]

PR #1693, opened 2023-09-11
with! macros

blorbb:
PR for #1682 

I wasn't able to run the doctests due to some errors with the `nix` crate - if you could run them and lmk of any changes needed that would be great. Not really sure how they are supposed to be set up, I copied some sections from the `view!` macro docs. 

Also the doc links to `StoredValue::with_value` and `update_value` aren't working for me (can't click it in VSC), but rustdoc doesn't give any warnings for it.


gbj:
Thanks so much for this. I think it will be really, really helpful for people. 

Doctests passed with flying colors. I agree that there's some oddness happening with `StoredValue` method links in VSC but cargo doc generates them correctly. I did have to fix the links to `with_value!` and `update_value!` macros from `with!` and `update!` and I have no idea why -- I just do what works with rustdoc.

PR #1694, opened 2023-09-11
fix: clippy lints

fundon:
```
cargo clippy --tests -- -Dclippy::all -Dclippy::pedantic
```


fundon:
hi @gbj, Any suggestions?

agilarity:
Thank you for working to improve the quality of this code base.

I know from personal experience that introducing a new set of lints can be challenging (See PR #1379). It is very easy to end up with a PR that is too big.

Here are my suggestions to split up this PR.

1. Reduce the scope of change to one crate or directory
2. Fix one type of lint violation at a time
3. Write meaningful commit messages
    - Include the scope in the commit message
    - Reference the lint in the commit message (_Consider using the [Clippy Lint](https://rust-lang.github.io/rust-clippy/master/index.html) name_)
    - See [How to Write Better Git Commit Messages](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/) for advice and examples
4. Open one PR for each fix
5. Link each PR to an issue
    - This could be a top level initiative like "**add pedantic lints**" or a supporting task like "**fix(leptos_reactive): redundant_closure_for_method_calls**."

The above suggestions are my own. I will, of course, defer to @gbj.

leptos/src/show.rs
```diff
@@ -51,8 +51,9 @@ where
 {
     let memoized_when = create_memo(move |_| when());
 
-    move || match memoized_when.get() {
-        true => children().into_view(),
-        false => fallback().into_view(),
+    if memoized_when.get() {
```
gbj, line 54: dropping the closure here is a behavior change that introduces a bug — whether `children` or `fallback` is shown will not be reactive any more

gbj:
Thank you for the large amount of work you've put into this.

It is hard to review, because it includes changes across 134 files without explanations, including
- very useful and appreciated fixes to doc links
- stylistic changes I'm not sure I agree with ("an `SSRed` Leptos app")
- breaking changes that affect user/app code (to `get_configuration`, for example)
- things that change behavior/introduce bugs (noted as separate comment)
- lots of suggested clippy improvements

I agree that separate PRs would be much more manageable: for example, I'd merge a PR that fixes all those broken or incorrectly-formatted links/references to types in doc comments instantly. But I do not have time to thoroughly review a PR of this size right away.

fundon:
@agilarity @gbj, thank you for your advice.

I will follow your suggestions. I'll cut them up into smaller ones.

PR #1696, opened 2023-09-12
Update possibly deprecated docs for `component` macro

ChristopherPerry6060:
Someone more familiar would be kind to double check this for me? I believe these docs are out of date. it clears up some *signal* (get it?) to noise ratio for the `component` attribute.

Feel free to kick out 125f17c9e21221e25924211542af32092bede0e3 if it seems excessive.


gbj:
Can't believe this has survived in the docs this long, honestly, as it hasn't been true in a couple releases. For clarity: the `Props` type simply doesn't need to be imported or used at all, we used some trait tricks to make it entirely implicit. Any mention of it can basically be removed.

The numbering for that list needs to be adjusted now though; removing section 3 means it skips from 2 to 4.

leptos_macro/src/lib.rs
```diff
@@ -466,6 +466,11 @@ pub fn template(tokens: TokenStream) -> TokenStream {
 ///    recognizes that a particular tag is a component, not an HTML element. It's important to be aware
 ///    of this when using or importing the component.
 ///
+/// 3. A function with the signature `fn MyComponent() -> impl IntoView` does *not* necessarily *need* to
```
ChristopherPerry6060, line 469: @gbj Would you prefer I delete this commit then? This would take spot three. I thought it was worth noting, but can see why it might not belong in the docs.

ChristopherPerry6060:
I did fix the numbering issue here by just making them into bullets. I noticed there was a mention of Scope, which I think you managed to correct on main already so I conflicted. Would be good to squash this unless there is anything else to correct.

Documentation really can be a pain to keep up with 😮‍💨 , as I am sure you would agree.

gbj:
While reviewing I found one additional inaccuracy w.r.t. component generics, which have additional features/flexibility now -- made that change and I'll let the CI cook and then merge this. Thanks for your work.

PR #1703, opened 2023-09-13
Callbacks: Manual Clone and Debug impl; Visibility fix

lpotthast:
Great to see, that this is now part of Leptos :). This PR mainly adds manual Clone and Debug implementations.

The manual `Clone` impl removes the `Clone` type  bounds on the generic `In` and `Out` types. All callback types can be cloned by simply cloning the inner `Rc`. The generic types are not important here.

The Debug impls are debatable, as they are not strictly needed. But I would say it's good practice to supply users with a Debug impl for ALL custom types a library provides, as this is somewhat expected. (The ability to derive Debug on your types without problems.) If not provided, users would always need to add custom `Debug` impls for all structs in which they want to store some `Callback` type and have a Debug impl on that struct. 
 
Also adds a missing `pub` modifier to construct `ViewCallback::new`.

### DIFF
--- a/leptos_dom/src/callback.rs
+++ b/leptos_dom/src/callback.rs
@@ -64,7 +64,7 @@
 
 use crate::{AnyElement, ElementDescriptor, HtmlElement, IntoView, View};
 use leptos_reactive::StoredValue;
-use std::{rc::Rc, sync::Arc};
+use std::{fmt, rc::Rc, sync::Arc};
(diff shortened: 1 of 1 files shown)


gbj:
Thanks, I always forget about manual `Clone` impls and it always bites me. And I agree re: `Debug` implementations, so thanks for those.

PR #1706, opened 2023-09-13
impled LeptosRoutes for &mut ServiceConfig in integrations/actix

cosmobrain0:
the Actix integration provides the `leptos_routes` method for the App type, but when you need to configure the App with a ServiceConfig (such as when trying to integrate Actix with Shuttle.rs), the `leptos_routes` method is not available. This PR fixes that.


gbj:
Thanks!

PR #1725, opened 2023-09-15
feat: Allow component names to be paths

mrvillage:
Simply put, this allows you to do `<module::Component />` on top of the existing `<div />` and `<Component />` style syntaxes.


gbj:
Great idea! I believe we technically do allow namespaced element names like `<svg::a>` to disambiguate between HTML and SVG elements in that very specific situation, so I believe this still does need to check the last path segment to see if it has an uppercase name if it's going to behave correctly.

gbj:
Thanks!

PR #1729, opened 2023-09-15
fix: relax bounds on `LeptosRoutes`

### LINKED ISSUE #1727: Relax bounds on `leptos_axum::LeptosRoutes`
https://github.com/leptos-rs/leptos/blob/afa67726c1f153c29e6e180098e7b5518e108a24/integrations/axum/src/lib.rs#L1289

`leptos_axum::LeptosRoutes` is bounded by `LeptosOptions: FronRef<S>`. I *think* that bound should only exist on the implementation and not the trait definition. If someone wants to use the integration they must supply LeptosOptions within the router

I deleted that line on my local repo and CI is cooking without issues (so far). Bounds should try to only exist where absolutely needed. Everyone will likely use the default impl for `Router<S>` which is already bounded by `LeptosOptions`

Writing the integration for Routes without using that trait, it would just be very convenient to remove
a few instances of LeptosOptions from signatures if possible. I could also just use LeptosOptions, but I do not think I benefit if I just want to do full SSR (I could be wrong though).

Would be interested to understand why / if that bound is actually needed for some edge case. Otherwise, I can submit a PR?

ChristopherPerry6060:
fixes #1727

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1286,7 +1286,6 @@ where
 /// having to use wildcards or manually define all routes in multiple places.
 pub trait LeptosRoutes<S>
 where
-    LeptosOptions: FromRef<S>,
     S: Clone + Send + Sync + 'static,
 {
     fn leptos_routes<IV>(


gbj:
Sweet!

PR #1734, opened 2023-09-16
chore(server_fn): improve docs

fundon:


server_fn/src/default.rs
```diff
@@ -64,6 +64,8 @@ inventory::collect!(DefaultServerFnTraitObj);
 ///     }
 /// }
 /// ```
+///
+/// [`actix-web`]: <https://docs.rs/actix-web/>
```
gbj, line 68: This doesn't seem to produce a working link to `actix-web`'s docs when I run `cargo doc` locally.

fundon: Fixed. The `lib.rs` file also has a link to actix-web.

gbj:
Thanks!

PR #1735, opened 2023-09-16
chore(leptos_hot_reload): apply lints suggestions

fundon:


leptos_hot_reload/src/node.rs
```diff
@@ -38,6 +39,13 @@ pub enum LAttributeValue {
 }
 
 impl LNode {
+    /// # Errors
+    ///
+    /// Will return `Err` if parsing the view fails.
+    ///
+    /// # Panics
```
gbj, line 46: Is this true? It checks for `.len() == 1` before `.pop().unwrap()` so I don't think this can ever panic. pedantic clippy false positive?

I'm open to a better version of it of course.

PR #1740, opened 2023-09-16
Make `Oco::Owned` change internally into `Counted` on first clone

DanikVitek:


DanikVitek:
Inner mutability is tough

PR #1744, opened 2023-09-17
Callback clone fix

lpotthast:
Argh.. I forgot the generic Out parameter in the manual clone implementations. How could we miss this? This means that 0.5.0-rc2 breaks for everyone trying to clone any `Callback` or `SyncCallback` and may warrant a bugfix 0.5.0-rc3 release!
Added some tests for cloning the various callback types all using a non-clone for their generic types. Should have done that in the first place.


lpotthast:
@gbj Can we merge this?

gbj:
Thanks very much — looks good, and thanks for the regression tests too!

PR #1756, opened 2023-09-19
Nix Shell

pelasgus:
Includes trunk; Installing trunk via cargo on Nix breaks after a few gcs


tsimouris:
There are currently some issues with the current configuration; will get on it tomorrow

rambip:
you can already use `nix develop` if you have flakes.
Just add it to `flake.nix` in `mkShell`

You could also add `cargo make`

benwis:
Yeah, I mean I'm not really against adding a nix-shell file, but it seems like most Nix projects I see are using the flake.nix file, and @rambip has a workaround for nix-shell. @pelasgus does that work for you?

benwis:
I'm going to close this for now, as it seems like most Nix projects use flakes.

PR #1762, opened 2023-09-20
Support default values for annotated server_fn arguments

g2p:
This allows form submission with checkbox inputs to work.
For example:

    let doit = create_server_action::<DoItSFn>();
    <ActionForm action=doit>
      <input type="checkbox" name="is_good" value="true"/>
      <input type="submit"/>
    </ActionForm>

    #[server(DoItSFn, "/api")]
    pub async fn doit(#[server(default)] is_good: bool) -> Result<(), ServerFnError> {}

If is_good is absent in the request to the server API,
`Default::default()` is used instead.

This is the approach suggested by https://github.com/leptos-rs/leptos/pull/1745#issuecomment-1724680903

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -84,7 +84,7 @@ pub fn server_macro_impl(
     let fn_path = fn_path.unwrap_or_else(|| Literal::string(""));
     let encoding = quote!(#server_fn_path::#encoding);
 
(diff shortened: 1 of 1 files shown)


gbj:
Looks good, thanks! (Failing CI was due to my mistake on the main branch earlier, have tested locally and this works well.)

Issue #1119, opened 2023-05-29
Blocking resource view rendered in template


ogesaku:
I'm trying to create an SEO-friendly html response with blog post content using a `create_blocking_resource`.

```rs
#[component]
fn Post(cx: Scope) -> impl IntoView {
    let post = create_blocking_resource(
        cx,
        || (),
        |_| async move {
            get_post(0)
                .await
                .map(|data| data.ok_or(PostError::PostNotFound))
                .map_err(|_| PostError::ServerError)
                .flatten()
        },
    );
    let post_view = move |cx| {
        post.with(cx, |post| {
            post.clone().map(|post| {
                view! { cx,
                    <h1>{&post.title}</h1>
                    <p>{&post.content}</p>
                    <Title text=post.title/>
                    <Meta name="description" content=post.content/>
                }
            })
        })
    };

    view! { cx,
        <Suspense fallback=|| "Loading post...">
            {post_view(cx)}
        </Suspense>
    }
}
```

In the browser, all looks well. When I enter the blog post page, the page is rendered with blog post content (fallback does not appear), but when I check the response using curl I can see that the blog post content is not rendered in its final destination:

```html
<!-- removed head, comments and scripts -->
<body>
  <!>Loading post...
  <template id="0-0-4-1-2-0-0f">
    <h1 id="_0-0-4-1-2-0-5">My first post</h1>
    <p id="_0-0-4-1-2-0-6">This is my first post</p>
  </template>
</body>
```

What's the best way to render a blocking resource?

gbj:
Are you familiar with the [`ssr` prop on a `<Route/>`](https://docs.rs/leptos_router/latest/leptos_router/fn.Route.html#optional-props)? There is now an [`SsrMode::PartiallyBlocked`](https://docs.rs/leptos_router/latest/leptos_router/enum.SsrMode.html) which combines out-of-order streaming with replacing the fallback HTML for suspended fragments that use blocking resources on the server rather than in the browser. Otherwise you can use `SsrMode::Async` or `SsrMode::InOrder` and both will wait for suspended chunks on the server entirely.

gbj:
> Is there usage for `create_blocking_resource` with other ssr modes?

In any SSR mode, `create_blocking_resource` prevents the server from sending any part of the HTTP response until any suspense that has read from that resource has loaded. This is useful for doing things like setting headers or `<head>` tags.

g2p:
[...]

PR #1765, opened 2023-09-21
Better document the interaction of SsrModes with blocking resources

g2p:
Meant to address users making the same mistake as
https://github.com/leptos-rs/leptos/issues/1119

### DIFF
--- a/docs/book/src/ssr/23_ssr_modes.md
+++ b/docs/book/src/ssr/23_ssr_modes.md
@@ -8,12 +8,12 @@ If you’ve ever listened to streaming music or watched a video online, I’m su
 
 Let me say a little more about what I mean.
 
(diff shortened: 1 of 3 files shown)


gbj:
Thanks! This is really helpful and looks accurate to me. `PartiallyBlocked` was added after this chapter was written, and I clearly didn't do a good job of explaining it. I appreciate the help.

gbj:
> By the way, I wonder if `PartiallyBlocked` might become the default. I don't see the point of having the expected HTML seemingly ready but hidden within an unrendered template.

It's purely a performance thing: `PartiallyBlocked` is really *only* useful in situations that do not support any JavaScript usage, including inline script tags, at the cost of an extra `O(n)` string replace on the server. I guess I'm open to argument in the other direction, but if you want to have that conversation we should do it in a separate discussion or issue.

PR #1776, opened 2023-09-23
`#[prop(default)]` uses `Default::default()`, unless specified.

k1rowashere:
`#[prop(default)]` now acts as `#[prop(default = Default::default())]`.
* with appropriate test changes


gbj:
What's the reasoning here? Isn't this what `#[prop(optional)]` already does?

k1rowashere:
It was just more intuitive to me. 
I didn't realize at first that `optional` worked on non-option types.
Sure, it is a duplication of features but I don't think it has a reason not to work.

gbj:
Fair enough. I'd suggest editing it so that `#[prop(default)]` without an argument is just interpreted as `#[prop(optional)]`, for the sake of reducing maintenance burden/not adding additional code.

k1rowashere:
Sadly, the `attribute_derive` crate doesn't provide attribute aliases, and it will error out if the attribute is not given a value, unless the `as_flag` method from `ConvertParsed` is overridden and returns a `Some(syn::Expr)` instead of `None`.

It might be possible to merge the implementations of `optional` and `default`, but I think it would be more cumbersome/ requires more code.

tqwewe:
+1 for this.
In the past I've done `#[prop(default)]` and it failed because I also assumed it would use `Default::default()`.

Perhaps in 0.6 `#[prop(into)]` could even be removed/deprecated in favor of having fewer `#[prop(...)]` variants - though not sure if people would agree with this

PR #1782, opened 2023-09-24
fix: improve rust-analyzer auto-completion

### LINKED ISSUE #1781: Rust Analyzer's auto-completion is weakened by the #[component] macro
## Describe the bug

The `#[component]` proc macro makes Rust Analyzer less capable at auto-completing code when a syntax error is present. Consider the following example code:

```rust
#[component]
pub fn Example(cx: Scope) -> impl IntoView {
    let // <-- Causes a syntax error

    let (value, set_value) = create_signal(cx, initial_value);

    set_value. // <-- Place your cursor after this dot and try auto-completing with Rust Analyzer.
}
```

When placing the cursor in the specified location, r-a does not give me auto-complete suggestions. However, if I remove the `#[component]` attribute, r-a is smart enough to provide suggestions, despite the syntax error in line 3.

## Explanation

I believe this issue occurs because of the way the component proc-macro is implemented. It uses the `syn` crate to parse the whole function body and then serializes it again. If an error occurs during parsing, the proc-macro will exit with only a compile error, and no actual code.

## How it could be better (I think)

Rust Analyzer can work with code that is emitted from a proc-macro even if it is prefixed or followed up with a compile error. In the case of this (relatively) simple attribute macro, it would be sufficient to emit the input token stream in the case of a parsing error. Since I've noticed you are using the `proc_macro_error` crate, this could be done with
```rust
proc_macro_error::set_dummy(input)
```
at the beginning of the component macro. This might even remove the necessity to output the error generated by `syn` for the function body.

See also:
* https://github.com/rust-lang/rust-analyzer/issues/10520
* https://github.com/rust-lang/rust-analyzer/issues/10468

## Rust Analyzer setup

I use the following Neovim configuration for Rust Analyzer. Notably, I have macro expansion enabled:
```lua
lspconfig.rust_analyzer.setup {
  cmd = { "rustup", "run", "stable", "rust-analyzer" },
  on_attach = on_attach,
  capabilities = capabilities,
  settings = {
    ["rust-analyzer"] = {
      procMacro = {
        enable = true,
      },
      checkOnSave = {
        command = "clippy",
      },
    },
  },
}
```

## Leptos Dependencies

```toml
leptos = "0.4.10"
```

pikaju:
This change fixes #1781. Please do try it out. You might have to restart rust-analyzer for it to "rebuild" the macros.

## HOWEVER

There is a caveat.

The situation is a bit more tricky than I initially thought, because the macros (at least the client macro) actually uses non-standard Rust syntax: Function parameter documentation comments (and potentially more, I'm not sure). Here's what happens:

image

1. The macro is fed the token stream with my extraneous `let`
2. The macro realizes that the syntax is broken and returns the token stream as is
3. The Rust compiler and rust-analyzer get token stream _with_ function parameter doc comments, and then complain

This is not a problem with valid syntax, as the behavior there is unchanged. However, it might confuse users, and they might remove the doc comments (thinking they are not allowed) before fixing the actual issue.

## Can we fix this?

I believe so. For this particular case, a simple fix would be to parse and re-serialize the function signature and body separately. This way, the macro can remove the doc comments before rust-analyzer sees them, even if there is a syntax error in the function body. Before I attempt this, I wanted to make sure you are fine with bigger changes?

I leave it up to you if the PR in its current state is an improvement or a regression (though I'm guessing the latter).

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -10,9 +10,10 @@ use quote::{format_ident, quote_spanned, ToTokens, TokenStreamExt};
 use syn::{
     parse::Parse, parse_quote, spanned::Spanned,
     AngleBracketedGenericArguments, Attribute, FnArg, GenericArgument, Item,
-    ItemFn, LitStr, Meta, Pat, PatIdent, Path, PathArguments, ReturnType, Stmt,
-    Type, TypePath, Visibility,
+    ItemFn, LitStr, Meta, Pat, PatIdent, Path, PathArguments, ReturnType,
+    Signature, Stmt, Type, TypePath, Visibility,
 };
+
 pub struct Model {
     is_transparent: bool,
     is_island: bool,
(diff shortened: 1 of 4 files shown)


pikaju:
Okay, I didn't expect this to fail any tests, to be honest

gbj:
Does this work consistently when you try it out? I tried it in a local example but found that it generally did not have any effect. However, my rust-analyzer is often also busted because I am switching back and forth between so many different PRs/examples/etc. so I wouldn't be surprised if it worked for you but not me, and I'd believe it was working if it does.

pikaju:
I also have a very bad experience with r-a (at least in this project), it often doesn't work at all, regardless of the macro. Maybe there is another issue, unrelated to what I'm trying to achieve here. My r-a logs are full of errors, perhaps those break it?

Just to clarify, did you:
1. Restart r-a (or better yet your whole editor) after checking out the commit (this is strictly necessary, otherwise it uses the old, cached version of the macro internally, at least that's what I found)
2. Wait a long time for it to _actually_ start up? In the counter example, I need to wait about 30 seconds before it even tries. SimpleQueryCounter took even longer...

If possible, I wouldn't take my word for it, because I'm new to Leptos, proc-macros and even Rust itself. I played around with my first proc-macro like 3 days ago.

I can't speak about consistency, but here are a couple of screenshots where I get auto-completions despite a syntax error:

image

image

blorbb:
not sure if this would work, but I've been working on my own proc macro too and managed to improve r-a by parsing a `TokenStream` instead of some syn structure if I don't need to care about what's inside. Instead of parsing a `syn::ItemFn`, could you make a custom AST that parses all the way until the body of the function, then just takes a `proc_macro2::TokenTree` (which should be the `Group` variant containing the entire function body)? Then this `TokenTree` can be inserted directly when expanding, syntax errors don't matter. This might be easier than cloning everything into a dummy.

pikaju:
> not sure if this would work, but I've been working on my own proc macro too and managed to improve r-a by parsing a `TokenStream` instead of some syn structure if I don't need to care about what's inside. Instead of parsing a `syn::ItemFn`, could you make a custom AST that parses all the way until the body of the function, then just takes a `proc_macro2::TokenTree` (which should be the `Group` variant containing the entire function body)? Then this `TokenTree` can be inserted directly when expanding, syntax errors don't matter. This might be easier than cloning everything into a dummy.

If I understand you correctly, that's basically what I'm doing, the dummy model only parses the function signature and leaves the body as a TokenStream (unparsed). The reason I have to parse the function signature is because of the error messages you'd otherwise get, see the original post.

blorbb:
ah my bad, I didn't read the code properly haha

gbj:
I can't see any downside to this, so I'm merging it in the hopes that it does increase the recoverability a bit!

Issue #1722, opened 2023-09-15
Fn: Into<Signal>


smessmer:
With the 0.5.0 changes to the reactive system, it's now easier than ever to convert a `Fn` into a `Signal`, just call `.into_signal()` on it. However, I think `.into()` would be even better. If `Fn` implemented `Into<Signal<_>>`, then those `Fn` could be passed directly to any components taking a `#[prop(into)] Signal<T>`.

`Fn: Into<MaybeSignal<_>>` probably makes sense too. Most of my components currently take `#[prop(into)] MaybeSignal<T>` so that call sites can decide whether to pass in a signal or a function.

gbj:
As far as I know this is impossible if signals can implement `Fn()`, because it conflicts with the global trait `impl From<T> for T`, as far as I know.

If you try to 
```rust
impl<T, F> From<F> for Signal<T> where F: Fn() -> T {}
```
you get a compiler error that says exactly that
```
error[E0119]: conflicting implementations of trait `From<signal_wrappers_read::Signal<_>>` for type `signal_wrappers_read::Signal<_>`
  --> /Users/gjohnston/Documents/Projects/leptos-main/leptos/leptos_reactive/src/signal_wrappers_read.rs:85:1
   |
85 | impl<T, F> From<F> for Signal<T> where F: Fn() -> T {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T> From<T> for T;

For more information about this error, try `rustc --explain E0119`.
```

gbj:
(To be clear it's a great idea and I would love it if the compiler would allow it, but it does not)

smessmer:
I see, that makes sense. Can we maybe change #[prop(into)] in some way that it allows passing in `IntoSignal` types on top of `Into` types then?

gbj:
Sure, I'd be happy to review a PR that adds something like `#[prop(into_signal)]`. I don't think it's possible to handle with the current `#[prop(into)]` as that just adds `.into()`.

rambip:
I don't know how to solve this issue, but I have a similar one:
For my callback proposal, I would like to:
1) convert a `Fn(A) -> B` to a `Callback<A, B>`
2) implementing `Fn(A) -> B` for `Callback<A, B>`

But it is not possible exactly for the same reason.

rambip:
[...]

PR #1795, opened 2023-09-27
moved callback to leptos root and improved design

rambip:
This is a small update on `callback.rs`.

The most noticeable change is the location of the code: it moved to `leptos` because it may be usefull outside of the browser.
Another change: the suppression of `HtmlCallback` because now any `HtmlElement` implements `Into<HtmlElement<AnyElement>>`.

I wanted to implement `Fn` for the `Callable` trait, but 1) I was not able to do it because of unconstrained type parameters and 2) the `call` method on `Callable` is the same as the call method on `Fn`, this is a bit dumb ...

It would be really cool if this change could be accepted for 5.0, just for the fact that `HtmlCallback` and `ViewCallback` will never exist.


gbj:
Thanks so much. I will definitely review before releasing 0.5.

gbj:
@rambip I'm curious to hear what you think of this one, which does implement `Fn` for `Callback`: https://github.com/leptos-rs/leptos/pull/1795/commits/ccef3845882f6f56b31251e5396db2b00a8a93e1

Using the `Tuple` trait here means you can call it like a normal function. This compiles and works as expected:
```rust
#[component]
fn MyComponent(#[prop(into)] render_number: Callback<(i32, i32), String>) -> impl IntoView {
    view! {
        <div>
            {render_number(42, 37)}
        </div>
    }
}

#[component]
    view! {
        <MyComponent render_number=|(x, y)| format!("x: {x} - y: {y}")/>
    }
}
```

rambip:
> @rambip I'm curious to hear what you think of this one, which does implement `Fn` for `Callback`: [ccef384](https://github.com/leptos-rs/leptos/commit/ccef3845882f6f56b31251e5396db2b00a8a93e1)
> 
> Using the `Tuple` trait here means you can call it like a normal function. This compiles and works as expected:
> 
> ```rust
> #[component]
> fn MyComponent(#[prop(into)] render_number: Callback<(i32, i32), String>) -> impl IntoView {
>     view! {
>         <div>
>             {render_number(42, 37)}
>             // does the same thing
>             {render_number.call((42, 37))}
>         </div>
>     }
> }
> 
> #[component]
>     view! {
>         <MyComponent render_number=|(x, y)| format!("x: {x} - y: {y}")/>
>     }
> }
> ```

I didn't manage to implement `Fn` for `Callback`, thank's ! I got an error saying it cannot disambiguate betwen `Callable::call` and `Fn::call` 🤔

Please implement it for `StoredValue<Callback>` too !

That said, It can be very counter-intuitive for users to write `Callback<(i32,) -> String>`; I already thought about that.
Packing arguments into a tuple and calling with a tuple seems to be better for me (meaning implementing `Fn` only for a 1-Tuple). In addition, if the user wants a callback with 0 arguments, he should use `Into<Signal>`, (see #1722) instead (maybe we should add that in the module docs too)

gbj:
I switched back to the more obvious `Fn` implementation.

`Fn` can't be implemented on `StoredCallback` unless `StoredCallback` is its own struct and not a type alias, for two reasons
1) orphan rules: `StoredValue` isn't implemented in `leptos`, it's in `leptos_reactive`, and we can't implement a trait from A on a struct from B in crate C
2) `StoredValue` already implements `Fn` IIRC

rambip:
> I switched back to the more obvious `Fn` implementation.
> 
> `Fn` can't be implemented on `StoredCallback` unless `StoredCallback` is its own struct and not a type alias, for two reasons
> 
> 1. orphan rules: `StoredValue` isn't implemented in `leptos`, it's in `leptos_reactive`, and we can't implement a trait from A on a struct from B in crate C
> 2. `StoredValue` already implements `Fn` IIRC

Could `Callback` be defined in `leptos_reactive` and be defined as `StoredValue<Box<dyn Fn>>` ?
That would be way simpler to use ...

rambip:
I wrote it !
The only thing I had to add is `In: 'static` and `Out: 'static` basically.
I don't know what that means exactly to be honnest but I don't think that's an issue.

gbj:
> But wasn't there a discussion about not storing each callback as a stored value for performance reasons? Did some preconditions change in 0.5 or am I wrong and this was generally deemed ok?

Enough people argued against me and the difference is small enough that I just gave in :smile:

PR #1799, opened 2023-09-28
Fix render_route error message and matching of non standard routes in…

### LINKED ISSUE #1798: render_route() functions fall down on wildcard routes
In the Axum integration, render_route fails to match routes like `/posts/:slug` because we're using the Uri(/posts/a-post), and the leptos router tries to compare the two.

benwis:
Fixed the render_route route matching for wildcard/param routes by getting the matched rule from Axum for the comparison to the Leptos Router. Should close #1798

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -653,7 +653,7 @@ where
         // 2. Find RouteListing in paths. This should probably be optimized, we probably don't want to
         // search for this every time
         let listing: &RouteListing =
-            paths.iter().find(|r| r.path() == path).unwrap_or_else(|_| {
+            paths.iter().find(|r| r.path() == path).unwrap_or_else(|| {
                 panic!(
                     "Failed to find the route {path} requested by the user. \
                      This suggests that the routing rules in the Router that \
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -7,7 +7,7 @@
 
 use axum::{
     body::{Body, Bytes, Full, StreamBody},
-    extract::{FromRef, FromRequestParts, Path, RawQuery, MatchedPath},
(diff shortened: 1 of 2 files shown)


integrations/axum/src/lib.rs
```diff
@@ -644,17 +644,21 @@ where
...
         // search for this every time
         let listing: &RouteListing =
-            paths.iter().find(|r| r.path() == path).expect(
-                "Failed to find the route {path} requested by the user. This \
+            paths.iter().find(|r| r.path() == path).expect(&format!(
+                "Failed to find the route {} requested by the user. This \
                  suggests that the routing rules in the Router that call this \
                  handler needs to be edited!",
-            );
+                path
```
OvermindDL1, line 660: Why was this moved out of the format string I'm curious?

benwis: Ah, because the original was not a format string, and was just saying {path} instead of a useful error message containing the route that was broken

OvermindDL1: Also, not from your code but just a prior-existing issue, `format` should not not not be used in `expect`, that means it's doing string formatting and allocation every single call, even when the expect is not being done, instead it should be in a closure, like via `.unwrap_or_else(|_| panic!("Failed to find the route {path} requested by the user. This suggests that the routing rules in the Router that call this handler needs to be edited!"))` that way the formatting operation isn't done unless it actually needs to be done

OvermindDL1: > Ah, because the original was not a format string, and was just saying {path} instead of a useful error message containing the route that failed to match

Uh, it shouldn't?  `format!("Blah: {path}")` will return a string of `Blah: /blorp` if path is `/blorp`.

OvermindDL1: Ah I misread, you added the format call, yeah format definitely doesn't go in an expect, use the unwrap_or_else I put above.  :-)

PR #1803, opened 2023-09-29
Set Content-Type header for all Responses to text/html;charset="utf-8"

benwis:
This sets the Content Type header for all non serverfn responses. Firefox will complain and show a Content Encoding error if you try to view page source without it. Plus it's good practive

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -773,8 +773,13 @@ async fn generate_response(
     if let Some(status) = res_options.status {
         *res.status_mut() = status
     }
+    // Set the Content Type headers on all responses. This makes Firefox show the page source
+    // without complaining
+    let headers = res.headers_mut();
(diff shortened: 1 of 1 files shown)


integrations/axum/src/lib.rs
```diff
@@ -773,8 +773,15 @@ async fn generate_response(
     if let Some(status) = res_options.status {
         *res.status_mut() = status
     }
+    // Set the Content Type headers on all responses. This makes Firefox show the page source
+    // without complaining
+    let headers = res.headers_mut();
+    headers.insert(
```
gbj, line 779: I think will overwrite a user-provided `CONTENT_TYPE` if there is one, right? I can't imagine why there would be one, to be fair—any page we're rendering is going to be HTML because of the boilerplate that gets inserted, and it's going to be UTF-8—but is it worth checking for any reason so we don't overwrite it?

PR #1804, opened 2023-09-29
Docs: DX improvements: add section about jetbrains `intellij-rust`

sebadob:
This PR does 2 things:

1. Added [mdbook-admonish](https://crates.io/crates/mdbook-admonish) to the book:
This is a really nice preprocessor for mdbook, which adds the same functionality as you have with Docusaurus for instance for adding notes and stuff.  
For instance, you can write a code block and annotate it with `admonish info` (like done in this PR) behind the the 'backtics' and you get a nicely formatted info box. The most used ones will probably be
- info
- warning
- danger
- example

but there are quite a few more.

This can be used in a lot of places in the whole documentation (if you decide that you like this and merge it ;-) ) and it will make it a lot nicer to read in certain places.

Once the preprocesser has been installed, there is not much you need to do about it. Maybe at some point in the future do an update or something, but it is really stable and I use it a lot already in other projects. Basically, you just need to do:

`cargo install mdbook-admonish` and then `admonish install path/to/your/book`

2. Added a note about the [intellij-rust](https://github.com/intellij-rust/intellij-rust) plugin in case of excluding the `component` and `server` macros from the macro expansion. These cause a lot of trouble for users which use either IDEA with the Rust plugin, CLion or RustRover.  
Currently, you cannot exclude macros dynamically, but they maintain a hardcoded list of these internally.  
I created a small PR in this project which will hopefully get merged soon and will then be included in the next release.
When this is the case, we might just update this docs section again, which takes only a few minutes anyway.


sebadob:
The intellij-rust plugin has been deprecated kind of "just now" and they moved the codebase over to RustRover.
However, I submitted a support request for this as well and just referred to the PR, since I don't know, if they would otherwise ignore the PR.

The note however is still valid and I reached out to them and provided everything they need to make it work with the Jetbrains products as well.

sebadob:
I keep getting E-Mails that the CI Examples run failed, but everything here is green, even inside the E-Mail from Github itself.

Do I miss something here? And if so, how / where can I see it or maybe execute on localhost?

sebadob:
I just added an update to the note about the RustRover support request and did another merge from master.  
I hope, that with the merge from the latest master, these E-Mail warning will go away (even though everything here is green).

gbj:
Sorry about that — for the sake of clarity, I think sometimes the "failed" message is really a "CI detected that it didn't need to check the examples and canceled that action." 

`mdbook-admonish` sounds great, and like it carries some more useful additional semantics than my tendency to just use `>` and insert an aside ranging from a sentence to four paragraphs :smile: I would welcome additional edits to the docs for that purpose, if there are any places you're thinking of!

Re: the macros by the way there's been some work (https://github.com/leptos-rs/leptos/pull/1782) done aimed at making the macro expansions a little more recoverable, although I can't guarantee it's done very much to improve the situation, so I'm hopeful this whole section can eventually be removed.

sebadob:
Yes I think the problem is a bit more on the Jetbrains side.  
Sometimes, the code completion gets stuck completely at the moment. Easy to fix with just out-commenting the macro on the function for 1 second, but annoying of course. I now have this open PR for the plugin itself and have the support request on the new RustRover platform. Let's hope that they listen in the near future.

The `view!` macro is totally fine though.

PR #1811, opened 2023-09-29
Remove extra angle brackets in generate_head_metadata

Sirius902:
`generate_head_metadata_separated` adds a set of angle brackets around the body tag as shown here https://github.com/leptos-rs/leptos/blob/2d634364a9a4087b64389e9cd54ecc337151aaeb/meta/src/lib.rs#L308-L308 That means `generate_head_metadata` introduces an extra set of angle brackets by interpolating that value like so `<{body}>`. This PR removes the extra set of angle brackets.


gbj:
Whoops! Thanks.

PR #1817, opened 2023-09-30
made show fallback optional

### LINKED ISSUE #1760: For the Show component, make the fallback property or default to an empty view
Very often, we have:
```
<Show when=... fallback=|| view! { "" }>
    ...
</Show>
```

It would be nice to make this attribute the default one.

maccesch:
Fixes #1760

### DIFF
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -230,8 +230,12 @@ pub use wasm_bindgen; // used in islands
     feature = "template_macro"
 ))]
 pub use web_sys; // used in islands
+
 mod children;
+mod view_fn;
 pub use children::*;
+pub use view_fn::*;
+
 extern crate self as leptos;
 
 /// A type for taking anything that implements [`IntoAttribute`].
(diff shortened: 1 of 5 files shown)


gbj:
Dare I mention that the same `Fallback` type should be reusable for `Suspense` and `Transition` too, unless I'm missing something? Would you like to try to include that in this PR? If not, we can handle it separately.

This seems obviously good to me, I don't know why I didn't do it a long time ago other than stubbornness and lack of creativity! 😄

maccesch:
Working on a project for a long time seems to make some amount of tunnel vision unavoidable. At least for me it's like that.

I renamed the struct to `ViewFn` because that seems to be describe it better to me. I didn't add this to `ErrorBoundary` because I don't think we should encourage irgnoring errors 😄 .

gbj:
> Working on a project for a long time seems to make some amount of tunnel vision unavoidable. 

This is 100% my experience too!

Thanks so much for taking the time — this will be such a nice quality-of-life improvement for everyone.

PR #1820, opened 2023-09-30
Portal

maccesch:
Previous discussion: https://github.com/Synphonyte/leptos-use/issues/18

This adds the `<Portal>` component. This resembles the Solid implementation quite closely.

I didn't implement the the special case for when the target is the `<head>` because for that we already
have leptos_meta.

Any feedback is highly welcome.


leptos/Cargo.toml
```diff
@@ -21,42 +21,40 @@ typed-builder-macro = "0.16"
 serde = { version = "1", optional = true }
 serde_json = { version = "1", optional = true }
 server_fn = { workspace = true }
-web-sys = { version = "0.3.63", optional = true }
+web-sys = { version = "0.3.63", features = ["ShadowRoot", "ShadowRootInit", "ShadowRootMode"] }
```
gbj, line 24: This change also pulled in a bunch of formatting changes to this `Cargo.toml` when you saved it, presumably due to an editor plugin with different formatting settings or something.

leptos/src/portal.rs
```diff
@@ -0,0 +1,76 @@
...
+            })
+        });
+    } else {
+        let _ = mount;
+        let _ = use_shadow;
+        let _ = is_svg;
+        let _ = children;
+    }}
+
+    ""
```
gbj, line 75: This should probably return `()` rather than `""`. I think if you just remove this return value it will return `()` implicitly, in fact.

gbj (commented):
Thanks, this looks really good! I appreciate the example, as I've never really understood what portals are for. 

I like that you added a test, as well.

gbj:
Thanks!

rakshith-ravi:
Ooooh. We portal now! Thanks a lot @maccesch

maccesch:
@gbj Thank you!

PR #1821, opened 2023-10-01
feat: directives

### LINKED ISSUE #1668: Directives in leptos
Currently we have to create custom components if we need to add/modify minor functionality for native elements like customized click events e.t.c.

We can have directives like solidjs.

Now we have to create a component wrapper and then pass every props to the native element via component props which makes our code bit messy.

maccesch:
Fixes #1668.

This isn't finished yet but I think it makes the idea clear.

You'd define your directives as functions.

```rust
// This doesn't take an attribute value
fn my_directive(el: HtmlElement<AnyElement>) {
    // do sth
}

// This requires an attribute value
fn another_directive(el: HtmlElement<AnyElement>, params: i32) {
    // do sth
}
```

Then in the view template:

```html
<!-- no attribute value -->
<div use:my_directive></div>

<!-- with value -->
<SomeComponent use:another_directive=5 />
```

Should I finish this? Should I do sth different?


leptos_dom/src/lib.rs
```diff
@@ -775,6 +775,67 @@ impl View {
...
+    /// second is the paramter that is provided in the attribute.
+    pub fn directive<T, P>(
+        self,
+        handler: impl DirectiveHandler<T, P>,
+        param: &P,
+    ) -> Self {
+        cfg_if! { if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
+            match &self {
+                Self::Element(el) => {
+                    let _ = el.element.on_mount(move |el| handler.call(el, param));
```
gbj, line 790: Note that `on_mount` will only ever run once, so if the element is unmounted and remounted, or moved, it won't run again. This may or may not be desirable, I'm not sure.

gbj:
This looks totally reasonable to me, and thanks for taking it on! The only question I have is whether implementing it using `NodeRef` and a `create_effect` makes more sense, as `on_mount` is a bit heavy in terms of its implementation.

leptos_dom/src/lib.rs
maccesch: Yeah I read that in the docs comment, but I wasn't sure either 😄

maccesch:
Ok I see. I guess in combination with the only-run-once of `on_mount` a `NodeRef` would be better indeed. I wonder if sth like a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) could be used to cover all the edge cases as well? Could maybe even be used to "lighten" the `on_mount` implementation?

maccesch:
So I did a bit more research how Solid does directives and they even say in the docs that it's just syntactic sugar for a node ref. They execute them before mounted.

So I think it makes sense to make our directives just syntactic sugar for

```rust
let the_ref = create_node_ref();

create_effect(move || {
    if let Some(the_ref) = the_ref.get() {
        directive_func(the_ref, possibly_some_param);
    }
});
```

That makes the implementation rather straight forward, is easy to understand for the user and we're 100% covered because we don't guarantee mounted and let users do whatever they want. Do you agree?

maccesch:
@gbj I think this is ready

gbj:
This looks really great! My only question was about whether it made sense to have `Directive` also generic over `E: ElementDescriptor` so that typed element could be used rather than only `HtmlElement<AnyElement>`. However, I can imagine it's also useful to have directives exist that can be used with any element, and the `view` macro wouldn't have an indicator as to whether it should convert them. For now, I think this is the right design, and people can downcast as needed in the directive, if they need a particular type.

Thanks for this really useful PR!

PR #1822, opened 2023-10-01
Fix template! cfg condition

Dragonink:
The current `#[cfg]` condition for the `template!` aliased as `view!` is not the negation of the condition for the real `template!`.
This causes an error when the compilation target is not `wasm32` but the `template_macro` feature is enabled, which happens with rust-analyzer in a workspace for example.

The fix is just to use the negation of condition for the real `template!`.


gbj:
Thanks very much! This is clearly right.

PR #1836, opened 2023-10-03
feat: support stored values in `with!` and `update!`

blorbb:
This adds support for using both signals and stored values in the `with!` and `update!` macros, without needing to use `with_value!` and `update_value!`.

(also modified the doctests a bit to match the doctests used on signals)

I'm not that great with traits, so feel free to change up the implementation if you see a better way :)

### DIFF
--- a/leptos_reactive/src/macros.rs
+++ b/leptos_reactive/src/macros.rs
@@ -17,8 +17,9 @@ pub(crate) use debug_warn;
 
 /// Provides a simpler way to use [`SignalWith::with`](crate::SignalWith::with).
 ///
-/// To use with [stored values](crate::StoredValue), see the [`with_value`](crate::with_value)
-/// macro instead.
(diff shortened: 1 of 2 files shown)


gbj:
Nice! Thanks!

leptos_reactive/src/macros.rs
```diff
@@ -269,3 +283,62 @@ macro_rules! update_value {
...
+/// This is a private module intended to only be used by macros. Do not access
+/// this directly!
+#[doc(hidden)]
+pub mod __private {
+    use crate::{SignalUpdate, SignalWith, StoredValue};
+
+    pub trait Withable {
+        type Value;
+
+        // don't use `&self` or r-a will suggest importing this trait
```
gbj, line 296: This is a really useful trick, which I didn't know about but which makes sense!

PR #1839, opened 2023-10-03
Don't print `=` on empty attribute when spreading

Baptistemontan:
If one attribute was `Attribute::Option(None)` when spreading the `=` is still printed.
This PR remove that equal sign.


Baptistemontan:
How I'm stupid, the idea was to not print it at all...

PR #1847, opened 2023-10-05
"#two" was being hidden. Added another # to unhide

he00741098:
Screenshot 2023-10-05 8 22 02 AM

Screenshot 2023-10-05 8 23 01 AM

### DIFF
--- a/docs/book/src/interlude_styling.md
+++ b/docs/book/src/interlude_styling.md
@@ -50,7 +50,7 @@ use stylers::style;
 #[component]
 pub fn App() -> impl IntoView {
     let styler_class = style! { "App",
-        #two{
+        ##two{
             color: blue;
         }
         div.one{


gbj:
Thanks! But this fix makes it appear as `two`, when it should be `#two` I think? I guess it's a peril of trying to display CSS inside Rust inside Markdown... 

Who knows, maybe a third `#`? Or some kind of escape?

he00741098:
image

That is strange. For some reason when I run `mdbook serve`, the `#` is still present

he00741098:
Looking back at this, perhaps you were confused by the 2 pictures I initially added. I was just trying to demonstrate that the line was hidden. I should have clarified that the two pictures were both of the actual book and the hide/unhide feature of mdbook.

gbj:
My bad! I just looked at your second image and thought it was after the fix, hadn't run mdbook myself. Thanks for this fix, I'll merge it.

PR #1849, opened 2023-10-05
Allow disposing of Signal & StoredValue

PaulWagener:
Small addition that allows for manual disposing of `StoredValue` and `Signal` types


gbj:
Nice, thanks!

PR #1850, opened 2023-10-05
Fix: use placeholder in comments for empty component names

flosse:
In some contexts the generated comments leads to invalid HTML/XML.
This quick fix uses `unknown` as placeholder to avoid the problem.
probably there are cleaner solutions, but it would be a first step.


gbj:
Thanks! I think `Fragment` is the only place that uses `""` as the component name so `<>` is probably the best default value to use here, rather than `unknown`.

flosse:
> so `<>` is probably the best default value to use here, rather than `unknown`.

done!

PR #1858, opened 2023-10-07
docs: clarify what "once per signal change" means

arcstur:
It took me some time to understand what running twice means for derived signals, I think this clarification can help.

If you don't think it's necessary, that's ok, I was able to figure it out while reading anyways haha.

A lot of lines are modified so that they keep around the same width, but I only added the parenthesis.


gbj:
Looks good to me, thanks!

PR #1866, opened 2023-10-08
fix: Add missing argument name to WrapsChildren component in section 3.9 of the book

johnnynotsolucky:
In 3.9, the [Manipulating children](https://leptos-rs.github.io/leptos/view/09_component_children.html#manipulating-children) section, the `WrapsChildren` example component was missing the argument name. This PR adds the name.


gbj:
Thanks!

PR #1893, opened 2023-10-13
Update hackernews README

erlend-sh:
Adding a brief mention of https://github.com/LemmyNet/lemmy-ui-leptos as a real-world example of the hackernews example it's based on.


gbj:
Happy to include it in [`awesome-leptos`](https://github.com/leptos-rs/awesome-leptos/) but it doesn't make sense to me in the README of an example.

PR #1895, opened 2023-10-14
Remove `Clone` requirement on `SignalWith` for `Resource`

Xendergo:
I'm assuming this is a mistake because the point of `with` is to not clone, and the code compiles just fine with the bound removed.


gbj:
Thanks!

PR #1896, opened 2023-10-15
feat: add `new` method to `NodeRef`

tqwewe:
Adds a `new` method to `NodeRef`.
This PR also implements `Default` for `NodeRef` (as was suggested by clippy).

Additionally removes the old `cx` being used in the docs for `create_node_ref`, and fixed some indentation there too.


gbj:
Looks good, thanks very much!

PR #1898, opened 2023-10-15
docs: update Dockerfile in deployment.md to work with cargo-leptos@0.2.0

quanhua92:
- cargo-leptos@0.2.0 changes the file structure under the target folder from `/app/target/server/release/` to `/app/target/release/`. https://github.com/leptos-rs/cargo-leptos/pull/159
- `cargo leptos new` now creates the project with the name `leptos-start` instead of `leptos_start`. So, we need to change the binary filename too.


gbj:
Thanks very much!

PR #1904, opened 2023-10-17
feat: optional named arguments for #[server] macro (closes #1621)

### LINKED ISSUE #1621: Implement optional named arguments for `#[server]` macro
In the coming Leptos v0.5, server macro will make all arguments to `#[server]` macro optional. This is great! However, since it uses positional arguments, to change function's endpoint (for example), one would need to specify all the arguments that come before:
```rust
#[server(MyServerFnType, "/api", "Url", "hello")]
pub async fn my_server_fn_type() /* ... */
```

To make this more convenient and less error-prone, the macro needs to take *named arguments*:
```rust
#[server(endpoint = "hello")]
pub async fn my_server_fn_type() /* ... */
```

safx:
Closes #1621.

- modified `ServerFnArgs.parse` to accept keyword arguments and positional arguments as well
- added default encoding in `server_impl` because the Rust compiler complained if we fed only `endpoint`
- added unit tests.
  - I couldn't compile leptos_macro/src/server.rs with `cargo test --features=ssr` in directory `leptos_macro`. So I enclosed it in cfg_if.
- update document

### DIFF
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -34,6 +34,7 @@ typed-builder = "0.16"
 trybuild = "1"
 leptos = { path = "../leptos" }
 insta = "1.29"
+serde = "1"
 
 [features]
 csr = []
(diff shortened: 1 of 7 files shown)


gbj:
Thank you for this gorgeous PR. I am so happy. It is a very useful new feature, which closes an issue, which comes in a PR with both tests and updated docs! Beautiful. I am so pleased to merge this, and thank you so, so much.

PR #1922, opened 2023-10-21
feat: Add local attribute for Await

### LINKED ISSUE #1567: Await local
There is `create_resource` and `create_local_resource` but there is no `AwaitLocal`. Wouldn't this make sense ?

`AwaitLocal` that does not require T to be serializable.

https://docs.rs/leptos/latest/leptos/fn.Await.html
https://docs.rs/leptos/latest/leptos/fn.create_local_resource.html

kerkmann:
This PR adds the ability to use `create_local_resource` instead of `create_resource`. This will run the create resource locally on the system and therefore its result type does not need to be `Seriaziable`.

Closes #1567

### DIFF
--- a/leptos/src/await_.rs
+++ b/leptos/src/await_.rs
@@ -2,7 +2,8 @@ use crate::Suspense;
 use leptos_dom::IntoView;
 use leptos_macro::{component, view};
 use leptos_reactive::{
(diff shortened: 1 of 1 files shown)


gbj:
Awesome! Thanks!

PR #1923, opened 2023-10-21
Add `replace` property to form component

### LINKED ISSUE #1914: Allow replace navigation option for forms

Hello, and thanks for this amazing library!

I've implemented something very closed to the search page presented in the [Form tutorial](https://leptos-rs.github.io/leptos/router/20_form.html), with the form submitted on each input. One thing that surprised me is that the window history gets pushed with the new location on each key press, which I find to be a bit much.

We can maybe add the `replace` property to `Form`, like we have for the `scroll` property. Or maybe we can allow specifying the whole `NavigationOptions`?

Not much, I've wondered if I could replace the history manually by overriding `on:submit` or `onsubmit`, but I haven't tried it.

koopa1338:
This adds a `replace` property to the form component to control `NavigationOptions` on submit.

Closes #1914

### DIFF
--- a/router/src/components/form.rs
+++ b/router/src/components/form.rs
@@ -61,7 +61,7 @@ pub fn Form<A>(
     noscroll: bool,
     /// Sets whether the page should replace the current location in the history when the form is submitted.
     #[prop(optional)]
-    noreplace: bool,
+    replace: bool,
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! The CI issue is unrelated. This looks good to me.

minijackson:
Thanks a lot!

PR #1925, opened 2023-10-22
Implement `IntoAttribute` for `TextProp`

SleeplessOne1917:
`TextProp` seems like something that should implement `IntoAttribute`.

Edit: Added `IntoView` implementation as well.


SleeplessOne1917:
@gbj I don't know why the CI step keeps cancelling.

gbj:
Looks like none of the examples build? Also some `cargo fmt` needed in other packages.
```
 error[E0412]: cannot find type `MaybeProp` in this scope
   --> /home/runner/work/leptos/leptos/leptos_dom/src/macro_helpers/into_attribute.rs:272:24
    |
272 | impl IntoAttribute for MaybeProp<TextProp> {
    |                        ^^^^^^^^^ not found in this scope
    |
help: consider importing this struct
    |
5   + use leptos_reactive::MaybeProp;
```

Could you clarify why `TextProp` needs to move between crates? I assume it has something to do with not being able to implement a trait from one crate on a type for another but could you explain?

SleeplessOne1917:
> Could you clarify why TextProp needs to move between crates? I assume it has something to do with not being able to implement a trait from one crate on a type for another but could you explain?

Your assumption is correct. In particular, I couldn't implement `From<T>` on `MaybeProp<TextProp>` without moving either `TextProp` or `MaybeProp`. Of the two, `TextProp` was easier to move without affecting anything else.

SleeplessOne1917:
@gbj I think the CI should succeed now. The changes I made in the router crate is because it was required to make `cargo +nightly make --profile=github-actions ci` pass.

rambip:
Why is `TextProp` not a `MaybeSignal<Oco>` then ?
That would be way simpler to `TextProp` and also `ViewFn` in terms of `MaybeSignal`

And you could even create it from `I: IntoSignal<Value=>`
(I'm working on the latter currently)

PR #1929, opened 2023-10-23
Clippy: use .first() [not .get(0)]

martinfrances107:
Minor change - Over time clippy is refining its default ruleset.

here is the relevant documentation 

https://rust-lang.github.io/rust-clippy/master/index.html#/get_first


gbj:
Thanks!

PR #1933, opened 2023-10-24
Fix clone with `Cow<'static, str>` and implement `IntoAttribute`

tqwewe:
This PR adds a missing implementation of `IntoAttribute` for `Cow<'static, str>`, and removes a clone when converting `Cow<'static, str>` into a view.

**:warning: Contains breaking change**

`Cow<'_, str>` no longer implements `IntoView`, now only `Cow<'static, str>` does.

The reason being: `Cow<'_, str>` requires a clone when converting to a view, since we don't know if the str is static.


gbj:
Could you split this in half?  The missing `IntoAttribute` part is great, but I can't merge the breaking part of it right now, which is frustrating because I think it's probably used very rarely in practice.

tqwewe:
No worries! I've made 2 new PRs:
https://github.com/leptos-rs/leptos/pull/1945
https://github.com/leptos-rs/leptos/pull/1946

PR #1935, opened 2023-10-24
feat: add `new` method to `Trigger`

tqwewe:
Adds a `new` method to `Trigger`.
This PR also implements `Default` for `Trigger` (as was suggested by clippy).

### DIFF
--- a/leptos_reactive/src/trigger.rs
+++ b/leptos_reactive/src/trigger.rs
@@ -24,7 +24,7 @@ impl Trigger {
     /// of notifying other reactive code of a change. This can be useful
     /// for when using external data not stored in signals, for example.
     ///
-    /// This is identical to [`create_rw_signal`].
+    /// This is identical to [`create_trigger`].
     ///
     /// ```
     /// # use leptos_reactive::*;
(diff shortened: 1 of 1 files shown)


leptos_reactive/src/trigger.rs
```diff
@@ -18,6 +18,47 @@ pub struct Trigger {
 }
 
 impl Trigger {
+    /// Creates a [`Trigger`](crate::Trigger), a kind of reactive primitive.
+    ///
+    /// A trigger is a data-less signal with the sole purpose
+    /// of notifying other reactive code of a change. This can be useful
+    /// for when using external data not stored in signals, for example.
+    ///
+    /// This is identical to [`create_rw_signal`].
```
gbj, line 27: should read "identical to [\`create_trigger\`]" right?

tqwewe: Ah sorry good catch, it should

tqwewe: All fixed!

gbj:
Thanks!

PR #1936, opened 2023-10-24
chore: cargo doc -- removed lint warnings.

martinfrances107:
Seen when running 
```
cargo doc
```

There were a couple of disambiguation issues hiding behind the list of minor issues 
scrolling off the page 

So 

[Route] becomes [struct@Route]

and 

[`extract`] becomes [`extract()`].


gbj:
Thanks very much! I really appreciate it.

PR #1946, opened 2023-10-25
fix!: remove clone in `Cow<'static, str>` `IntoView` impl

tqwewe:
Removes a clone when using a `Cow<'static, str>` in a view.

**:warning: Contains breaking change**

`Cow<'_, str>` no longer implements `IntoView`, now only `Cow<'static, str>` does.

The reason being: `Cow<'_, str>` requires a clone when converting to a view, since we don't know if the str is static.

### DIFF
--- a/leptos_dom/src/lib.rs
+++ b/leptos_dom/src/lib.rs
@@ -1146,6 +1146,17 @@ impl IntoView for &'static str {
     }
 }
 
+impl IntoView for Cow<'static, str> {
+    #[cfg_attr(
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! I'll mark this with the 0.6 milestone. It's likely that 0.6 will include a rewritten renderer so this may become irrelevant, but I'll make sure the same issue doesn't apply. And if there are other breaking changes that require a new version before the new renderer, I'll include this then.

PR #1947, opened 2023-10-25
`IntoView` and `IntoAttribute` for `std::fmt::Arguments` improvements

tqwewe:
- Avoids an allocation when possible when using `std::fmt::Argumenets` in `IntoView`.
- Implements `IntoAttribute` for `std::fmt::Arguments`.


gbj:
Nice, thanks!

PR #1959, opened 2023-10-28
fix: router docs feature `stable` to `nightly`

ChristopherPerry6060:
* [suspect docs](https://docs.rs/leptos_router/0.5.2/leptos_router/#feature-flags)
* [feature flags](https://docs.rs/crate/leptos_router/0.5.2/features)

Likely missed when the `stable` flag was deprecated


gbj:
Clearly right, thanks!

PR #1966, opened 2023-10-30
fix: dead `params` attribute on `Params` derive

chrisp60:
Dead attributes like this can show up in docs, leading to confusion.

[implementation where it is not used](https://github.com/leptos-rs/leptos/blob/91f482992331fc9ce3379e39a4a69a94c4f9f09b/leptos_macro/src/params.rs#L16).


gbj:
Thanks. This is technically breaking so I will flag it for the next release.

PR #1973, opened 2023-11-02
fix: `leptos_router::params_map!`

### LINKED ISSUE #1971: `leptos_router::params_map!` remove from public API
# Issue

`params_map!` macro is in public API but is broken when used outside of the defining crate.

It is used in unit tests. I would assume it got `macro_exported` for internal use and leaked into public API.

I am not sure if it is meant to be public. Secondly, the first match arm is a compiler error when called.

https://github.com/leptos-rs/leptos/blob/d7ca5f2e961caa7fba8175c58afdd7b1be24a13c/router/src/history/params.rs#L89

# Solution
I can submit a PR for any of the following

* Fix the macro for external use (not sure if anyone would benefit as this has seemingly gone under the radar for quite a while).
* [hide it from the documentation](https://docs.rs/leptos_router/latest/leptos_router/index.html#macros).
* Remove from public API

# Images

image

 
image

image

ChristopherPerry6060:
fixes #1971 through reimplementation.

Comes with the benefit of knocking a crate out of the deps tree.

### DIFF
--- a/router/Cargo.toml
+++ b/router/Cargo.toml
@@ -14,7 +14,6 @@ leptos_integration_utils = { workspace = true, optional = true }
 leptos_meta = { workspace = true, optional = true }
 cached = { version = "0.45.0", optional = true }
 cfg-if = "1"
-common_macros = "0.1"
 gloo-net = { version = "0.2", features = ["http"] }
 lazy_static = "1"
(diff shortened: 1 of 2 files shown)


gbj:
Awesome! Thank you so much.

PR #1979, opened 2023-11-03
build: bump nix flake and add rust toolchain particulars

saikatdas0790:
- Bumped the nix flake lock to current to update rust toolchain to the latest
- Removed the Nix Rust overlay and instead moved the flake to use rustup and the corresponding `rust-toolchain.toml` file. `cargo make ci` was choking otherwise because it was unable to detect the rust toolchain

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,29 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1681202837,
-        "narHash": "sha256-H+Rh19JDwRtpVPAWp64F+rlEtxUWBAQW28eAi3SRSzg=",
+        "lastModified": 1694529238,
+        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "cfacdce06f30d2b68473a46042957675eebb3401",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
-    "flake-utils_2": {
-      "inputs": {
-        "systems": "systems_2"
-      },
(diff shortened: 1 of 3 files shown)


tqwewe:
I think the rust overlays in nix are worth keeping since they come with a few benefits over adding rust directly to the packages.

What exactly was the issue you were having with the `cargo make ci`?

saikatdas0790:
> I think the rust overlays in nix are worth keeping since they come with a few benefits over adding rust directly to the packages.
> 
> What exactly was the issue you were having with the `cargo make ci`?

`cargo make ci` for all of the examples is outputting this error

image

I imagine it's due to the Rust toolchain not being discovered at the expected location?

benwis:
Updating the rust toolchain is good, but I agree with @tqwewe I much prefer rust-overlay over rustup. I'd prefer to fix whatever the CI is doing, at the end of the day plenty of people install rust without rustup, so that shouldn't be a factor

saikatdas0790:
> Updating the rust toolchain is good, but I agree with @tqwewe I much prefer rust-overlay over rustup. I'd prefer to fix whatever the CI is doing, at the end of the day plenty of people install rust without rustup, so I don't want to require it

I believe `rustup` is the de-facto way for non-Nix users to install and use Rust. Hence, the change.

But, I'm happy to revert the changes to still use the Rust overlay if that's what this repo should be using. 
I'm kind of a noob at Nix still though and hence am clueless on how to fix the above. If you have suggestions on the way forward, happy to check out 🙂

benwis:
I should also ask about the other settings that the NixOS wiki recommends for rustup in a dev shell:
https://nixos.wiki/wiki/Rust

I'd probably say with that added on that it's simpler to maintain the current system. I'll see if I can figure out why the CI is broken today

benwis:
I'm going to close this for now. It seems like most of our Nix users are not using rustup, and the current one seems to be working alright. Thanks for bringing this to our attention

PR #1981, opened 2023-11-03
Deprecate the `path` argument in the relevant `LeptosRoutes` trait methods.

gibbz00:
```rust
let app = Router::new()
    .leptos_routes(&leptos_options, generate_route_list(app), app)
// ...
```
becomes 

```rust
let app = Router::new()
    .leptos_routes(&leptos_options, app)
// ...
```

Might be too much of a breaking change for such a subjective improvement. I'll refrain from sending in a PR updating `start_axum` until this change is deemed appropriate. 😊

### DIFF
--- a/examples/counter_isomorphic/src/main.rs
+++ b/examples/counter_isomorphic/src/main.rs
@@ -9,7 +9,7 @@ cfg_if! {
         use actix_files::{Files};
         use actix_web::*;
         use crate::counters::*;
-        use leptos_actix::{generate_route_list, LeptosRoutes};
+        use leptos_actix::LeptosRoutes;
 
         #[get("/api/events")]
         async fn counter_events() -> impl Responder {
@@ -41,9 +41,7 @@ cfg_if! {
             // Setting this to None means we'll be using cargo-leptos and its env vars.
             // when not using cargo-leptos None must be replaced with Some("Cargo.toml")
             let conf = get_configuration(None).await.unwrap();
-
             let addr = conf.leptos_options.site_addr;
-            let routes = generate_route_list(|| view! { <Counters/> });
 
             HttpServer::new(move || {
                 let leptos_options = &conf.leptos_options;
@@ -52,7 +50,7 @@ cfg_if! {
                 App::new()
                     .service(counter_events)
                     .route("/api/{tail:.*}", leptos_actix::handle_server_fns())
-                    .leptos_routes(leptos_options.to_owned(), routes.to_owned(), || view! { <Counters/> })
+                    .leptos_routes(leptos_options.to_owned(), Counters)
                     .service(Files::new("/", site_root))
                     //.wrap(middleware::Compress::default())
             })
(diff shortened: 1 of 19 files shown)


gbj:
I'm open to making a change this like in the future. As far as I can tell it would break ~100% of Leptos-Axum apps so of course it's not changeable until the next version. I am also not sure the churn of breaking everything is worth the benefit of saving 20 characters or so, but I'm open to being wrong.

gibbz00:
Spot on. Thought about mentioning that it could alternatively be added to the 0.6 milestone, but seems like you're way ahead of me :) Another option could also be to hold on until a major version is released that would require fixing ~100% of the packages regardless.

I'll make sure to fix the examples as well as the starter crate tomorrow for completeness’ sake.

gibbz00:
Hey again, 

I went ahead to made the changes consistent in the `viz` and `actix` routers, as well as opening up fix PRs in the respective starter repos. More so to not keep an incomplete PR lying around, and not to as a means to push for it being merged right away. I would be fully understanding if you choose to close this whenever, given previous discussions ✌️

benwis:
The reason to do this is to allow people to pass in some routes, but not all routes, for automatic route generation in LeptosRoutes. Moving it inside saves props, sure, but also removes the ability to handle them seperately

gibbz00:
Hi @benwis!

Could you give me an example for when/how it would be useful? You're probably on to something, it's more that I lack the experience to understand when that could be utilized 🙈

benwis:
Sure, let's say you wanted to apply an Axum middleware like TraceLayer, CompressionLayer, or some kind of auth thing, to specific routes in your app, kinda like so
```rust
let router = Router
.route("/compressme", get(handler))
.layer(CompressionLayer::new())
```
because LeptosRoutes will automatically find all routes in your Leptos app, and you can't have conflicting routes, you first have to exclude the route from general handling
```rust
let routes = generate_route_list_with_exclusions("/compressme")
// then pass the routes to LeptosRoutes via the path param
```

gibbz00:
Ahh, well put.

`leptos_routes_with_handler()` still exists. Would it be possible to use it for that use case?

https://github.com/leptos-rs/leptos/blob/a8e25af5233bb014d3cee85e4e9be8b3e4586de9/integrations/axum/src/lib.rs#L1437-L1445
https://github.com/leptos-rs/leptos/blob/a8e25af5233bb014d3cee85e4e9be8b3e4586de9/examples/session_auth_axum/src/main.rs#L91-L93

benwis:
It has a different usecase, it's meant mostly to let people provide their own context. Unfortunately, at least for the idea of middleware support, I don't think you could apply that. You'd have to do per handler middleware instead of layer middleware. Something to think about though

benwis:
In the interest of maintaining flexibility and now with server fn routes being added to the macro, we'll need to preserve the path var. Thanks for making the PR and having the conversation!

PR #2001, opened 2023-11-07
Make leptos_reactive benches run

patefacio:
The leptos_reactive benchmarks do not build/run. This makes it so they run. I'm not sure about the 1021::* bits and I replaced with `use leptos::*` which woks but may be benchmarking something not desired.


gbj:
Thanks — I think I'm just going to remove these in #2014, as they duplicate the benchmarks already present in the `benchmarks` directory and test against a very old version of Leptos (0.2.1 is the `l021`) and a proposed crate `reactive-signals`.

Issue #2005, opened 2023-11-08
Cache busting for WASM and JS


xbb:
When the wasm/js assets are cached by the browser and if there is any change, you have to force reload the page to see them.

Implementing cache busting would solve this.

I've looked at the discussion here https://github.com/leptos-rs/leptos/discussions/761 and tried experimenting with the [html_parts_separated](https://github.com/leptos-rs/leptos/blob/cb7e07496a3b9ed8154bcbaa5ba39e6c9a09c70b/integrations/utils/src/lib.rs#L54) function

This below works as an example (by appending the files modified time as query-string, but it could be the file hash or whatever), but surely it's not the best place to do that.

Even better if it could be configurable at runtime when initializing the server.

```patch
diff --git a/integrations/utils/src/lib.rs b/integrations/utils/src/lib.rs
index 702e18fc..f4281499 100644
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -100,6 +100,38 @@ pub fn html_parts_separated(
     } else {
         "() => mod.hydrate()"
     };
+
+    use std::path::PathBuf;
+
+    let get_timestamp = |path: PathBuf| -> String {
+        path.metadata()
+            .ok()
+            .and_then(|m| m.modified().ok())
+            .and_then(|m| m.duration_since(std::time::UNIX_EPOCH).ok())
+            .map(|e| e.as_millis().to_string())
+            .unwrap_or_default()
+    };
+
+    let js_output_url = {
+        let ts = get_timestamp(
+            PathBuf::from(&options.site_root)
+                .join(&pkg_path)
+                .join(&output_name)
+                .with_extension("js"),
+        );
+        format!("/{pkg_path}/{output_name}.js?{ts}")
+    };
+
+    let wasm_output_url = {
+        let ts = get_timestamp(
+            PathBuf::from(&options.site_root)
+                .join(&pkg_path)
+                .join(&wasm_output_name)
+                .with_extension("wasm"),
+        );
+        format!("/{pkg_path}/{wasm_output_name}.wasm?{ts}")
+    };
+
     let head = format!(
         r#"<!DOCTYPE html>
             <html{html_metadata}>
@@ -107,8 +139,8 @@ pub fn html_parts_separated(
                     <meta charset="utf-8"/>
                     <meta name="viewport" content="width=device-width, initial-scale=1"/>
                     {head}
[...]

PR #2006, opened 2023-11-08
This adds cache busting to wasm and javascript artifacts

dgsantana:
As per @xbb implementation on issue #2005, this adds cache busting as an option.


xbb:
Thanks for the pull request but please wait, I don't think cache busting via query string is the way to go.

It's ok for development but not for production. I should have mentioned it in the feature request!

Found a better solution though, I have updated the issue.

Issue #2004, opened 2023-11-08
tailwind_axum example broken when trying to set LEPTOS_SITE_PKG_DIR env var for cache busting


wt:
**Describe the bug**
In the tailwind_axum example, setting the env var `LEPTOS_SITE_PKG_DIR` to anything other than "pkg" causes site to fail at runtime with errors on JS console.

**Leptos Dependencies**

Head revision of the main branch.

**To Reproduce**
1. Change to root of repo
2. cd examples/tailwind_axum
3. LEPTOS_SITE_PKG_DIR=blah cargo leptos watch
4. after service is started, open http://localhost:3000 in browser
5. CSS doesn't load as the app is trying to load it from /pkg/tailwind.css (see app.rs)

**Expected behavior**
The CSS is loaded and works.

**Additional context**
If I manually override the setting the App object to load the css from the right location, I then get other errors in the js console:

    tailwind.js:1 Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.

FWIW, I was experimenting with trying to implement cache busting by generating a random string to insert into the LEPTOS_SITE_PKG_DIR. Also, it seems very add that the App doesn't seem to know where to the load the CSS from if it isn't the default. I am not sure how to fix this. Should the config be plumbed into the App? Is it already available somehow?

gbj:
I don't see a step where you changed
```
        <Stylesheet id="leptos" href="/pkg/tailwind.css"/>
```
to
```
        <Stylesheet id="leptos" href="/blah/tailwind.css"/>
```
If I do the rest of your steps and point the `Stylesheet` component to the correct URL, it works as expected.

> Also, it seems very add that the App doesn't seem to know where to the load the CSS from if it isn't the default. I am not sure how to fix this. Should the config be plumbed into the App? Is it already available somehow?
Not sure if there's a typo or something, but I can't figure out what this means, sorry.

wt:
Let me try again.

Running something like `LEPTOS_SITE_PKG_DIR=blah cargo leptos watch` should not require someone to change the source of the App component to match in order for the site to work.

Is there some way for the App component to know what the pkg dir is so that the stylesheet path doesn't need to be changed for that to work? The app should just know that the stylesheet needs to be sourced from the appropriate dir.

[...]

PR #2021, opened 2023-11-14
Make the tailwind_axum example work with non-default pkg dir.

wt:
The tailwind_axum example would not run correctly if built with LEPTOS_SITE_PKG_DIR env var not set to "pkg" during build. Now it works.

### DIFF
--- a/examples/tailwind_axum/src/app.rs
+++ b/examples/tailwind_axum/src/app.rs
@@ -6,9 +6,11 @@ use leptos_router::*;
 pub fn App() -> impl IntoView {
     provide_meta_context();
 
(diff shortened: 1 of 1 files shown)


gbj:
This is in response to #2004, for anyone else reading who wants the context.

I think it makes sense either to do this for all the relevant examples, or to add a utility to do it and use that. It feels odd to me to do it for just a single example. Would you be interested in taking it a little further to do one of those?

wt:
Agree about doing it for all examples instead of only one. I may not get around to this any time soon. I figured it'd be better to have at least on example rather than none.

Honestly, I think it would be better tho have a way to plumb certain config identically into both the wasm and non-wasm sides. That may be even more valuable.

PR #2024, opened 2023-11-15
fix document book input name query

datewu:
I believe it is a typo in The document book section [form]( https://leptos-rs.github.io/leptos/router/20_form.html).

The input name attribute should be `q` for `let search = move || query().get("q").cloned().unwrap_or_default();` working.


gbj:
Thanks!

PR #2029, opened 2023-11-15
fix: use `create_effect` for `<Portal/>` to avoid hydration issues (closes #2010)

### LINKED ISSUE #2010: <Portal /> can cause hydration issues
**Describe the bug**
If a `<Portal />` component is used with SSR and does not have any surrounding `<Suspense />`, `<Show />`, etc., hydration will fail on the portal.

**Leptos Dependencies**

```toml
leptos = { version = "0.5.2", features = ["nightly"] }
leptos_axum = { version = "0.5.2", optional = true }
leptos_meta = { version = "0.5.2", features = ["nightly"] }
leptos_router = { version = "0.5.2", features = ["nightly"] }
```

**To Reproduce**
Here's an example:

```rs
#[component]
    let (show, set_show) = create_signal(false);
    create_effect(move |_| set_show(true));

    view! {
        <Show when=show>
            <Portal>
                <h1>"This mounts properly."</h1>
            </Portal>
        </Show>

        <Portal>
            <h1>"This does not mount."</h1>
        </Portal>
    }
}
```
error_example

[Here's a link to a repo where I set up this example for quick reference.](https://github.com/CorvusPrudens/leptos-portal-issue)

**Expected behavior**
While I don't have a deep understanding of the framework, I think the above should simply work in both cases. This is useful in certain situations where you want a portal to always be present, not just transient like a tooltip.

gbj:
This would close #2010. 

`create_render_effect` runs immediately/synchronously during rendering. `create_effect` doesn't run until the tick after it's created.

This means that if used during hydration, `create_render_effect` runs during the hydration process and `create_effect` runs after hydration.

`<Portal/>` was causing a bug during hydration because its elements were created inside a render effect, which meant they happened during hydration, and when they went to create elements thought they should be looking them up in the existing HTML in the DOM rather than creating them afresh.

This fixes the issue by switching to `create_effect`, so a `<Portal/>` only creates elements after hydration is done and there is no error.

@maccesch This does have the effect of slightly changing execution order: Portals whose `mount` is `None` will end up being placed after all other children in the body, *not* just after the children in the body as it exists at the moment they're created. My assumption is that portals are usually used with `Some(mount)` so this is not a big deal. Does this sound right to you?

An alternative would be to temporarily suspend hydration during the creation of a Portal. I haven't explored this as this is a one-line fix otherwise.

### DIFF
--- a/leptos/src/portal.rs
+++ b/leptos/src/portal.rs
@@ -29,13 +29,13 @@ pub fn Portal(
 ) -> impl IntoView {
     cfg_if! { if #[cfg(all(target_arch = "wasm32", any(feature = "hydrate", feature = "csr")))] {
         use leptos_dom::{document, Mountable};
-        use leptos_reactive::{create_render_effect, on_cleanup};
+        use leptos_reactive::{create_effect, on_cleanup};
         use wasm_bindgen::JsCast;
 
         let mount = mount
             .unwrap_or_else(|| document().body().expect("body to exist").unchecked_into());
 
-        create_render_effect(move |_| {
+        create_effect(move |_| {
             let tag = if is_svg { "g" } else { "div" };
 
             let container = document()


maccesch:
Ah I didn't know that about `create_render_effect`! 

No, all of this sounds good to me. Usually you're not worried where in the `body` your portal contents end up in. And as AFAIK other frameworks put it at the end of the `body`too.

PR #2031, opened 2023-11-15
examples: add CSR with server functions example (closes #1975)

### LINKED ISSUE #1975: Add full Leptos native support with flags or example/guide for CSR apps with Server functions
There has been an appetite including my own to have a template/example or an easier way to set up CSR applications with Server functions.
Would it be possible to have this implemented in future/future releases?

gbj:
@Innominus I put together this example of CSR with server functions. I'm sure it's not perfect but it's enough to demonstrate the idea. It handles generating the `index.html` file for you, using the exact same template used by server-rendered examples, just with an empty `<body>` tag, and client-renders the app instead of doing any server rendering.

I'm hoping you and some of the others interested might be able to provide feedback on the example, and anything that's not clear.

It is set up to run with `cargo leptos` and supports the live-reloading features built into that, and so on. (It doesn't support `--hot-reload` because that's based on some added marker comments that are only added in SSR mode at the moment, but I think that could be added pretty easily in the future).

I based it on the regular Axum example of this so there are probably some unnecessary dependencies etc. left over.


Innominus:
@gbj apologies for the very late reply because you've already merged this. I took it out for a spin and it works great. Is there a way to pass compilation flags down to what you've got compiling the wasm bundle? (I think it's wasm-pack?) Similar to when you compile with Trunk and you can set flags for wasm-opt.

The example is great btw, will be able to use this soon, thank you!

PR #2033, opened 2023-11-16
Minor: stop using std::fmt, instead used core::fmt.

martinfrances107:
This is a coding standard, I have adopted in my own libraries, and well this is open source, so I want to start using this practice in modules that I use on a regular basis too..

It follows a recent trend in the rust community.

It is a aspiration, rather than an achievable goal - One day someone will try and run my library of an IOT device 
and I will have to answer question on the lines of

What portions of the library can I still use in a "no_std" environment.

So the is a defensive change .. and to drop std::fmt and use core::fmt  where possible 

I hope this helps.


gbj:
Huh, that's interesting. There is no plausible path to `no_std` for Leptos, but seems like  good practice overall!

PR #2035, opened 2023-11-16
Have fetch example conform to docs guidance around using <ErrorBoundary> and <Transition> in conjunction

MoonKraken:
The docs state that when used in conjunction, `<ErrorBoundary>` should be inside `<Transition>`, not the other way around. This didn't have a functional effect on the fetch example because it happens to use `create_local_resource` which guarantees that the resource will be loaded on the client side. https://docs.rs/leptos/latest/leptos/fn.ErrorBoundary.html#interaction-with-suspense

But I was working on an application that uses `create_resource`, and I happened to be using this example as a model of using `<ErrorBoundary>` and `<Transition>` in conjunction. When server side rendering is a possibility, reversing them seems to cause an issue where the `<ErrorBoundary>` fallback doesn't render properly.

So I thought maybe we should use them in the recommended order in this example.

### DIFF
--- a/examples/fetch/src/lib.rs
+++ b/examples/fetch/src/lib.rs
@@ -89,15 +89,15 @@ pub fn fetch_example() -> impl IntoView {
                     }
                 />
             </label>
-            <ErrorBoundary fallback>
-                <Transition fallback=move || {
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! That makes sense to me, for sure.

PR #2044, opened 2023-11-19
docs: move to table of features for `leptos`

chrisp60:
This is a cosmetic change and may be opinionated. I find tables like this to be more readable.

For some it can be hard to parse without visual organization. No issue at all if it does not fall into line with your view for the docs.

I would not mind unifying the other crates if merged

here it is rendered on my local machine without any styles that docs.rs might add.
image

### DIFF
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -75,28 +75,6 @@
 //! - **Routing**: the [`leptos_router`](https://docs.rs/leptos_router/latest/leptos_router/) crate
 //! - **Server Functions**: the [`server`](crate::leptos_server) macro, [`create_action`], and [`create_server_action`]
 //!
-//! # Feature Flags
-//! - `nightly`: On `nightly` Rust, enables the function-call syntax for signal getters and setters.
-//! - `csr` Client-side rendering: Generate DOM nodes in the browser
-//! - `ssr` Server-side rendering: Generate an HTML string (typically on the server)
-//! - `hydrate` Hydration: use this to add interactivity to an SSRed Leptos app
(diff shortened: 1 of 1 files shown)


leptos/src/lib.rs
```diff
@@ -140,6 +118,37 @@
...
+//! | serde                     | **(De)Serialization**: use [serde] for [Resource]s.           | Yes     |
+//! | serde-lite                | **(De)Serialization**: use [serde-lite] for [Resource]s.      | No      |
+//! | rkyv                      | **(De)Serialization**: use [rkyv] for [Resource]s.            | No      |
+//! | miniserde                 | **(De)Serialization**: use [miniserde] for [Resource]s.       | No      |
+//! | tracing                   | **Logs & Telemetry**, [tracing] support for [component]s.     | No      |
+//! | default-tls[^2]           | **Networking**: Use default native TLS support.               | No      |
+//! | rustls[^2]                | **Networking**: Use [rustls].                                 | No      |
+//! | template_macro            | **Templating**: Enables the legacy [template] macro.          | No      |
+//!
+//! [^1]: At least one **rendering** feature must be enabled. An Isomorphic crate may have
```
gbj, line 140: This rewritten version sounds misleading to me. Exactly one should be enabled, not more than one.

gbj:
I like this overall, and thank you for it! I would prefer to keep the original wording about using **one** of `csr`, `ssr`, or `hydrate`.

chrisp60:
Hey! I see your recommendations and you are correct. I'm a little overwhelmed with work at the moment but plan to revisit and correct this when I have the time. Thanks for the feedback. 

I am going to move this to a draft for now.

PR #2047, opened 2023-11-20
Remove hyper: Use methods in viz and axum.

martinfrances107:
hyper 1.0 has been released 

-hyper = "0.14.23"
+hyper = "1"

Conventionally  0.x releases are considered pre-releases.
It seems hyper is acting in this vein.

here is the relevant blogs post, which is linked in the release notes. 

[stability-promise](https://hyper.rs/contrib/vision/#stability-promise)

### DIFF
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -13,7 +13,7 @@ axum = { version = "0.6", default-features = false, features = [
 ] }
 futures = "0.3"
 http = "0.2.8"
-hyper = "0.14.23"
+hyper = "~1"
 leptos = { workspace = true, features = ["ssr"] }
 leptos_meta = { workspace = true, features = ["ssr"] }
 leptos_router = { workspace = true, features = ["ssr"] }
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -13,7 +13,6 @@ axum = { version = "0.6", default-features = false, features = [
 ] }
 futures = "0.3"
 http = "0.2.8"
-hyper = "~1"
 leptos = { workspace = true, features = ["ssr"] }
 leptos_meta = { workspace = true, features = ["ssr"] }
 leptos_router = { workspace = true, features = ["ssr"] }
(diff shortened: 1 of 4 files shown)


martinfrances107:
This is a experimental PR.

At two places in the codebase "A" is converted to "B" -  where

A) let body = body::to_bytes(body).await.unwrap_or_default();

----------------------

B) let body = body
        .collect::<Vec<_>>()
        .await
        .into_iter()
        .filter_map(|result| result.ok())
        .collect::<Vec<Bytes>>()
        .concat()
        .into();

The question becomes is this the best use of the new API to convert all
stream chunks into a HTTTP Body while minimising the number of copies/clones.

My current thoughts into_iter() is not free !

and

collect.cocat().into() is cumbersome.

martinfrances107:
Just making notes for myself in public 

There is good discussion here about the various ways or iterating over a Result type.   The strategies to handle the error type.


https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

martinfrances107:
Cool, we now have a base line -- a change that passes tests

I am still worried about the performance implications here.

Regarding clones and copies .. I am not sure this is  bad, indifference or a good change 

Please be aware I am still thinking about this ... I will ask on discord in the morning if there are any standard ways the gauge a suspected performance regressions.

martinfrances107:
To justify that last change 

hyper is a low level API  and we are now accessing that functionality through the viz module, or through the axum module. Modules on which we directly depend.

From a maintenance perspective this is a good change  less dependencies to track.

dgsantana:
You can check my PR #2082, it moves everything to http 1.0 and axum 0.7. It uses the http-body-util to get the bytes in a easier way. Right now all examples, except for the one which depends on axum_js_fetch is converted. This examples also requires an update on gloo-net, for which I made a PR, let's see if it gets merged.

PS: For Viz, you can't move to http 1.0, because Viz hasn't update yet to hyper 1 and http 1. We need to wait a bit. Maybe make a PR for it?

martinfrances107:
Thanks, StreamBody and BoxBody  -- leads to a much cleaner solution

> PS: For Viz, you can't move to http 1.0, because Viz hasn't update yet to hyper 1 and http 1. We need to wait a bit. Maybe make a PR for it?

I will help out on that where possible.

PR #2049, opened 2023-11-20
updated axum session to latest 0.9 in examples

genusistimelord:
I have updated the example to use the latest version of AxumSession


gbj:
Thanks!

Issue #2007, opened 2023-11-08
Leptos Book Enhancement Proposals - for discussion


diversable:
I've been reading through the docs, and I think there's a few items we could add/expand in the docs / book that I think would be helpful for newcomers to Leptos.

Specifically, what I would suggest adding to the book includes:

1. a section on the burgeoning ecosystem of crates / projects around Leptos, such as leptos-use, leptos-query, leptonic, leptos-i18n, etc. I think it would help make these projects more discoverable and help newcomers at the same time by including links and brief descriptions of the projects in the book, perhaps as the last chapter. 

2. possibly(?) a section on using Leptos as the UI along with the Tauri desktop/mobile app framework. Until the work on custom renderers is complete, Tauri might be the best way of producing non-web apps with Leptos (imho) - it would be nice to alert newcomers to this possibility, along with relevant links and perhaps even a 'quickstart with Leptos/Tauri' page to get users up and running quickly.

3. move the info from the appendix on DX into a new section that comes after the "getting started" section. You could call this new section "Getting Ready for Leptos Development" or something like that. Why? I think it would be helpful to surface DX improvements up front to help put Leptos' 'best foot forward' -  it'd be really nice for new users to benefit from better DX while they're initially getting started with Leptos

4. create a new section on "Contributing to Leptos" at the end of the book - providing users a few different options of how to meaningfully contribute to Leptos, such as feature and docs enhancements to Leptos itself, and including links to ecosystem crates that are looking for new contributors, might help take some pressure off the core maintainers. We could create subsections on Leptos internals in this section, which might include things like "How the Reactive System Works", and maybe how other internal modules work, which would be a great way of onboarding new maintainers to Leptos.

5. Move the section on "Optimizing WASM Binary Size" from an appendix into the section (or a subsection) of "Deployment", as that's where optimizing binary size is most relevant...

I'll leave it at just those suggestions for now, as the above represents a good chunk of work on the book already. 

I'd like to get some discussion going on this topic to see if there's any interest in making some modifications to the book, using the above as a starting point for discussions. Once the community has a chance to chime in to provide feedback and further direction, I can start implementing the requested changes.

chrisp60:
I believe the largest chance for improvement and growth of the docs is just dealing with this in small chunks. This repo has a lot to cover and moves quickly, that leaves a lot of opportunity for docs to become a burden on maintainers.

[...]

PR #2052, opened 2023-11-21
Book updates

diversable:
In this (larger) commit, several things have been updated in the book as discussed in [issue #2007](https://github.com/leptos-rs/leptos/issues/2007). 

1) the DX section has been moved right after the "getting started" section
2) a new section has been added to "getting started" that's all about describing the Leptos community, and where to find the leptos-related crates ie. in the "Awesome Leptos repo
3) this last commit was the biggest one: I have drawn out the inherent structure of the book and explicitly split the book into 2 parts: part 1 on CSR and part 2 on SSR - this structure was already implicit in the text, I've just made it a little more explicit so that people feel like they've accomplished something in part 1 (CSR) before moving onto part 2 (SSR). 

This last commit required editing a few sections to make the explicit separation work. I hope you like it, and that it wasn't too much of a large change in one pull request.  

(I used to be an English academic & professional editor, before being struck by a debilitating disability and eventually moving into tech; I couldn't help but add a bit more structure to the text)

I've also added a few words here and there in response to questions about the book on Discord - mostly trying to clarify the differences between the CSR and SSR modes.


docs/book/src/csr_wrapping_up.md
```diff
@@ -0,0 +1,26 @@
+# Wrapping Up Part 1: Client-Side Rendering
+
+So far, everything we’ve written has been rendered almost entirely in the browser. When we create an app using Trunk, it’s served using a local development server. If you build it for production and deploy it, it’s served by whatever server or CDN you’re using. In either case, what’s served is an HTML page with
+
+1. the URL of your Leptos app, which has been compiled to WebAssembly (WASM)
+2. the URL of the JavaScript used to initialized this WASM blob
```
reedwoodruff, line 6: `initialize` instead of `initialized`

docs/book/src/getting_started/README.md
```diff
@@ -2,19 +2,35 @@
...
 
-If you don’t already have it installed, you can install Trunk by running
+We’ll introduce `cargo-leptos` in Part 2 of this book, which is all about working with the full power of Leptos in its Full-stack, SSR mode.
+
+```admonish note
+If you're coming from the Javascript world and terms like Client-Side Rendering (CSR) and server-side rendering (SSR) are unfamiliar to you, the easiest way to understand the difference is by analogy:
+
+CSR is similar to working with React (or a 'signals'-based framework like SolidJS), and focuses on producing a client-side UI which you can use with any tech stack on the server.
+
+SSR is like working with a full-stack framework like Next.js in the React world (or Solid's "SolidStart" framework) - SSR helps you build sites and apps using JS on both the client and the server.
```
reedwoodruff, line 17: Maybe change from `SSR is like working with...` to `SSR is similar to working with...` to avoid using "like" twice in a row and to parallel the previous CSR item.

Also it might be confusing to say that "SSR helps you build sites and apps using JS on both the client and the server" when we aren't really concerned with using JS on either the client or the server in Leptos. I understand this section is directed toward JS devs, but maybe this could be reworded to be more language-agnostic -- something like:
"SSR lets you build apps which are rendered on the server and sent to the client, potentially making client/server communication more streamlined." I'm not sure, something like that.

docs/book/src/getting_started/README.md
```diff
@@ -95,3 +111,10 @@ Now run `trunk serve --open` from the root of the `leptos-tutorial` directory.
 Trunk should automatically compile your app and open it in your default browser.
 If you make edits to `main.rs`, Trunk will recompile your source code and
 live-reload the page.
+
+
+Welcome to the world of UI development with Rust and WebAssembly (WASM), powered by Leptos and Trunk!
+
+---
+
+Now before we get started building your first real UI's with Leptos, there's a couple of things you might want to know to help make your experience with Leptos just a little bit easier.
```
reedwoodruff, line 120: `there's a couple of things` -> `there are a couple of things`
or `there're a couple of things` 🤔 lol

reedwoodruff (commented):
A couple of suggestions :+1:

docs/book/src/getting_started/README.md
reedwoodruff: Looks great to me! :+1:

docs/book/src/getting_started/README.md
reedwoodruff: No problem. And no judgement here!

docs/book/src/getting_started/README.md
```diff
@@ -0,0 +1,120 @@
+# Getting Started
+
+There are two basic paths to getting started with Leptos:
+
+1. Client-side rendering with [Trunk](https://trunkrs.dev/) - a great option if you just want to make a snappy website with Leptos, or work with a pre-existing server or 3rd-party API's.
+2. Full-stack / Universal server-side rendering with [`cargo-leptos`](https://github.com/leptos-rs/cargo-leptos) - a great option for building a CRUD-style website or a web-app with Rust powering both frontend and backend. Universal apps rely on either Actix-web or Axum on the server (so you may want to keep their docs handy, too), or you can even go Universal WASM/WASI with the Leptos and Axum integration (more on these options later).
+
+In Part 1 of this book, we'll start with Client-side rendering  Leptos sites and building reactive UI's using `Trunk` to serve our JS+Wasm bundle to the browser.
```
gbj, line 8: small one but I usually go for "client-side" lowercases and "JS and WASM" or "JS & WASM"

there's also a double space between "rendering" and "Leptos"

docs/book/src/getting_started/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+We’ll introduce `cargo-leptos` in Part 2 of this book, which is all about working with the full power of Leptos in its Full-stack, SSR mode.
```
gbj, line 10: likewise, I'd lowercase full-stack

docs/book/src/getting_started/README.md
```diff
@@ -0,0 +1,120 @@
...
+cargo init leptos-tutorial
+```
+
+`cd` into your new `leptos-tutorial` project and add `leptos` as a dependency
+
+```bash
+cargo add leptos --features=csr,nightly
+```
+
+> **Note**: This version of the book reflects the Leptos 0.5 release. The CodeSandbox examples have not yet been updated from 0.4 and earlier versions.
```
gbj, line 45: This Note is no longer accurate or necessary, since I've updated them! Can be removed.

docs/book/src/ssr/README.md
```diff
@@ -1,21 +1,19 @@
-# Server Side Rendering
+# Part 2: Server Side Rendering
 
-So far, everything we’ve written has been rendered almost entirely in the browser. When we create an app using Trunk, it’s served using a local development server. If you build it for production and deploy it, it’s served by whatever server or CDN you’re using. In either case, what’s served is an HTML page with
+The second part of the book is all about how to turn your beautiful UI's into full-stack, "Universal" Rust + Leptos powered websites and applications.
```
gbj, line 3: "UI's" should be "UIs" IMO but not a big deal

docs/book/src/ssr/README.md
```diff
@@ -1,21 +1,19 @@
...
 
-1. the URL of your Leptos app, which has been compiled to WebAssembly (WASM)
-2. the URL of the JavaScript used to initialized this WASM blob
-3. an empty `<body>` element
+As you read in the last chapter, there are some limitations to using Client-Side Rendered Leptos apps - over the next few chapters, you'll see how we can overcome those limitations
+and get the best performance and SEO out of your Leptos apps.
 
-When the JS and WASM have loaded, Leptos will render your app into the `<body>`. This means that nothing appears on the screen until JS/WASM have loaded and run. This has some drawbacks:
+Let's see how we can use the full power of Leptos and Rust on the server to make your next cutting-edge application!
```
gbj, line 8: > to make your next cutting-edge application!

this feels a bit marketing-y to me

gbj (commented):
Thanks for all your work! I left a bunch of notes in comments while reading through it. 

Higher-level notes:
- I think the restructure makes a lot of sense.
- Thanks for the added content, this is really great.
- I tend to avoid "universal" as a term for SSR/hydration. I understand it's used fairly widely, but to me it conveys both too little information ("universal" doesn't tell me "oh, this renders to HTML on a server") and claims too much (there's no sense in which actually "universal"). You may have a broader experience/perspective than I do and if you think it's a useful term for communicating what's going on to people than I'll yield to your judgment. (Also, maybe I just have too much painful baggage from trying to set up Angular Universal in early 2020 :-p) 

I really appreciate your energy for this work. Thanks again!

docs/book/src/getting_started/README.md
```diff
@@ -2,21 +2,28 @@
...
 
-In Part 1 of this book, we'll start with Client-side rendering  Leptos sites and building reactive UI's using `Trunk` to serve our JS+Wasm bundle to the browser.
+2. **Full-stack, server-side rendering (SSR) with [`cargo-leptos`](https://github.com/leptos-rs/cargo-leptos)** - SSR is a great option for building CRUD-style websites and custom web apps if you want Rust powering both your frontend and backend.
+With the Leptos SSR option, your app is rendered to HTML on the server and sent down to the browser; then, WebAssembly is used to instrument the HTML so your app becomes interactive - this process is called 'hydration'. On the server side, Leptos SSR apps integrate closely with your choice of either [Actix-web](https://docs.rs/leptos_actix/latest/leptos_actix/index.html) or [Axum](https://docs.rs/leptos_axum/latest/leptos_axum/index.html) server libraries, so you can leverage the those communities' crates to help build out your Leptos server.
```
reedwoodruff, line 9: "so you can leverage ~~the~~ those communities'"

gbj:
Awesome! Thanks, I will merge this and then move the book over to the new repo.

For context re: "isomorphic": I am ultimately fine with any term here that you and others decide is reasonable. I tend to use "isomorphic" in the specific sense of "having the same shape," so I say server functions are "isomorphic" because they have the same "shape" when called on either the server or client -- likewise with SSR + hydration, the view tree/tree of components and elements has the same shape on the server and client.

But mostly I'm just fond of it because I've been reading ancient Greek for a long time and so the name makes perfect sense to my brain (isos = equal/same, morphikos = relating to a shape). I recognize that I've had an... unusual path here and it may not make the same sense to other devs 😅

PR #2054, opened 2023-11-21
Use new with macro in book pages, update formatting for create effect page

jollygreenlaser:
See title


jollygreenlaser:
It seems these work with the vars inside the `{}`, not sure on the style guide regarding that. Found another place that can use `with`.

jollygreenlaser:
Not sure if it should be a separate PR but I also ran this thru `leptosfmt` and updated the context type for log.

PR #2057, opened 2023-11-22
example session-auth-axum: Minor changes for ergonomics

j0lol:
I've always hated the get_todos function and I wanted to change it badly. Added a .env file containing the db url for sqlx-cli, and cleaned up with leptosfmt.

### DIFF
--- a/examples/session_auth_axum/.env
+++ b/examples/session_auth_axum/.env
@@ -1,0 +1,1 @@
+DATABASE_URL=sqlite://Todos.db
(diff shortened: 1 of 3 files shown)


gbj:
Thanks! I believe this just needs a `cargo fmt` in the example directory, which is the failing CI check.

j0lol:
whoops! ill fix that

j0lol:
fixed & squashed the commit :) i dont have github notifications so i missed this

Issue #1815, opened 2023-09-30
ci(leptos): run ci on change instead of check


agilarity:
PROPOSED: _Let me know if you would like me to submit a PR_

**Is your feature request related to a problem? Please describe.**
Some breaking changes are detected only when the daily CI Examples workflow runs.

**Describe the solution you'd like**
- Run CI Examples when a leptos change is detected
- Do not run CI Examples when a leptos change is detected

**Describe alternatives you've considered**
1. Run Check Examples on leptos change (current behavior)
2. Run both Check Examples and CI Examples on leptos change
3. Run CI Examples on leptos change (proposed)

**Additional context**
- This change will provide earlier detection of breaking changes, making it easier to troubleshoot
- This change will minimize the number of breaking changes that make it into the `main` branch

gbj:
Sorry for my delayed reply to this. Yes, I think this is a good idea, and I'd appreciate a PR very much.

PR #2061, opened 2023-11-22
ci(leptos): run ci on change instead of check

agilarity:
Closes: #1815

**Key Changes**

- Trigger the CI Examples workflow when the Leptos source is changed
- Add the CI Stable Examples workflow that runs when the Leptos source is changed


gbj:
Thank you!

agilarity:
You're welcome.

PR #2064, opened 2023-11-22
Add instruction to install trunk to examples/README.md

hpepper:
Add the install step to the README.md so the user can follow the readme, without using the [leptos book](https://leptos-rs.github.io/leptos/02_getting_started.html)


gbj:
Thanks!

PR #2066, opened 2023-11-22
Fix WASI builds trying to use JS for `spawn_local`

### LINKED ISSUE #2056: `create_resource` panics when server is compiled to wasm32-wasi
**Describe the bug**

I am using SSR and building the server as a [Spin](https://github.com/fermyon/spin) application.  This means the server is compiled to `wasm32-wasi` rather than native code.

When I tried to run a variant of [the `counter_isomorphic` example](https://github.com/leptos-rs/leptos/tree/main/examples/counter_isomorphic), the server panicked on the `create_resource` call (based on https://github.com/leptos-rs/leptos/blob/1cd6603da01d4e556473fce588fe9afa0dba34c2/examples/counter_isomorphic/src/counters.rs#L102), with the message "cannot call wasm-bindgen imported functions on non-wasm targets', /home/ivan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/js-sys-0.3.65/src/lib.rs:5970:9".

The relevant part of the backtrace is:

```
           6: 0x1fe6d0 - <unknown>!js_sys::global::get_global_object::h905e1441bf8294a8
           7: 0x1fe6b7 - <unknown>!js_sys::global::h98e3a07765e5d726
           8: 0x1fe3a0 - <unknown>!wasm_bindgen_futures::queue::Queue::new::h6d3bb2f62074b61c
           9: 0x1fe462 - <unknown>!wasm_bindgen_futures::task::singlethread::Task::spawn::hf198b6ecb05f055e
          10: 0x18dae - <unknown>!std::thread::local::LocalKey<T>::with::h472ebb6959128837
          11: 0x2ba3b - <unknown>!<leptos_reactive::memo::Memo<T> as leptos_reactive::signal::SignalWith>::try_with::hb826f2fa273592e0
          12: 0x33bff - <unknown>!leptos_reactive::resource::ResourceState<S,T>::load::hcb00091d53adc33a
          13: 0x1e295 - <unknown>!<leptos_reactive::effect::EffectState<T,F> as leptos_reactive::effect::AnyComputation>::run::h7382e9f0cd086cde
          14: 0x1f2326 - <unknown>!leptos_reactive::runtime::Runtime::update_if_necessary::h7ec72211ebe3fde0
          15: 0x1a035 - <unknown>!std::thread::local::LocalKey<T>::with::h7e93c221acbbee39
          16: 0x354f1 - <unknown>!leptos_reactive::resource::create_resource_helper::h909cdad642bfb8f2
          17: 0x341cb - <unknown>!leptos_reactive::resource::create_resource_with_initial_value::hf6fa4f40ea300293
          18: 0x33eda - <unknown>!leptos_reactive::resource::create_resource::h28393832763b21dd
```

The issue appears to arise in the `leptos-reactive` `spawn_local` function (https://github.com/leptos-rs/leptos/blob/1cd6603da01d4e556473fce588fe9afa0dba34c2/leptos_reactive/src/spawn.rs#L78), where if the target architecture is `wasm32` it uses `wasm_bindgen_futures::spawn_local`, which depends on access to JavaScript globals (https://github.com/rustwasm/wasm-bindgen/blob/bb12058e1facd26c5c647282a125088535be577e/crates/futures/src/queue.rs#L88).  In Spin SSR, the server code _is_ Wasm but JavaScript is _not_ present, hence, panic.

**Leptos Dependencies**

```toml
leptos = { version = "0.5" }
leptos_integration_utils = { version = "0.5.2", optional = true }
leptos_meta = { version = "0.5" }
leptos_router = { version = "0.5" }
```

The server environment was Spin 2.0.1, although this is of course not an official integration, and is very much experimental at this point!

**To Reproduce**

1. [Install the Spin CLI.](https://developer.fermyon.com/spin/v2/install)
2. If you don't already have it, install the Rust `wasm32-wasi` target
3. Clone https://github.com/itowlson/spin-leptos-investigation and checkout out the `counter-isomorphic` branch
4. Run `cargo leptos build --release` (this builds the client side Wasm and the assets; we ignore the native binary)
5. Run `spin up --build` (this builds the app with the `ssr` feature, and runs it)
6. Open localhost:3000 in the browser

**Expected behavior**

Expected to see Incr and Decr buttons and a counter that went up and down

**Additional context**

1. I understand that what I am doing is completely experimental, and I am very much finding my way.  It's entirely possible I've gone about this the wrong way and the problem is all on my end!

2. I had a crack at fixing it by adding the following clause at the top of the `spawn_local` function:

```
        if #[cfg(all(target_arch = "wasm32", target_os = "wasi"))] {
            futures::executor::block_on(fut)
        }
        else if #[cfg(target_arch = "wasm32")] {
        // ... rest as before ...
```

This appeared to work and I am now seeing a working page.  I don't know if it's a reasonable approach or if it could cause problems in more sophisticated apps, and I appreciate you may be reluctant to take a fix that is presumably not hit by testing.  But I'd be delighted to send a PR if you're open to it.

itowlson:
Fixes #2056.

This is a change to the original fix proposed in the issue.  After talking to colleagues working on upstream WASI, the recommendation was to poll the future only _once_, on the basis that if the future were pending at that point, `futures::executor::block_on` would block forever, and it was better to panic with a specific message than to do that.

The caveat here is that `spawn_local` is a very low level function.  My fix works for the case I ran into, which was `create_resource`.  But if `spawn_local` is used elsewhere for a future that is truly long-running and non-blocking (such as a WASI Preview 2 outbound HTTP request) then this will fail.  Hopefully in that case the message will at least point us to the fix that's needed.  I don't think this is an issue as yet, but I don't know how `spawn_local` really gets used in SSR, so advice is welcome!

The risk on top of that caveat is that, as far as I know, the Leptos test suite doesn't currently build for WASI, and I haven't tried to tackle that as part of this PR.  So this new case is extremely Works On My Machine.  I am not sure how to move forward with testing on WASI I'm afraid.

### DIFF
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -66,6 +66,7 @@ miniserde = ["leptos_reactive/miniserde"]
 rkyv = ["leptos_reactive/rkyv"]
 tracing = ["leptos_macro/tracing"]
 nonce = ["leptos_dom/nonce"]
+spin = ["leptos_reactive/spin"]
 experimental-islands = [
   "leptos_dom/experimental-islands",
   "leptos_macro/experimental-islands",
--- a/leptos_reactive/Cargo.toml
+++ b/leptos_reactive/Cargo.toml
@@ -26,6 +26,7 @@ bytecheck = { version = "0.7", features = [
 rustc-hash = "1"
 serde-wasm-bindgen = "0.5"
 serde_json = "1"
+spin-sdk = { git = "https://github.com/fermyon/spin", tag = "v2.0.1", optional = true }
 base64 = "0.21"
 thiserror = "1"
 tokio = { version = "1", features = [
@@ -76,6 +77,7 @@ serde-lite = ["dep:serde-lite"]
 miniserde = ["dep:miniserde"]
 rkyv = ["dep:rkyv", "dep:bytecheck"]
 experimental-islands = []
+spin = ["ssr", "dep:spin-sdk"]
 
 [package.metadata.cargo-all-features]
 denylist = ["nightly"]
--- a/leptos_reactive/src/spawn.rs
+++ b/leptos_reactive/src/spawn.rs
@@ -75,7 +75,10 @@ where
     F: Future<Output = ()> + 'static,
 {
     cfg_if! {
-        if #[cfg(target_arch = "wasm32")] {
+        if #[cfg(all(target_arch = "wasm32", target_os = "wasi", feature = "ssr", feature = "spin"))] {
+            spin_sdk::http::run(fut)
+        }
+        else if #[cfg(target_arch = "wasm32")] {
             wasm_bindgen_futures::spawn_local(fut)
         }
         else if #[cfg(any(test, doctest))] {
--- a/router/src/history/mod.rs
+++ b/router/src/history/mod.rs
@@ -39,8 +39,8 @@ impl BrowserIntegration {
         let loc = leptos_dom::helpers::location();
         LocationChange {
             value: loc.pathname().unwrap_or_default()
-                + &loc.search().unwrap_or_default()
-                + &loc.hash().unwrap_or_default(),
+                + loc.search().unwrap_or_default().as_str()
+                + loc.hash().unwrap_or_default().as_str(),
             replace: true,
             scroll: true,
             state: State(None),
--- a/router/src/matching/resolve_path.rs
+++ b/router/src/matching/resolve_path.rs
@@ -62,7 +62,7 @@ fn normalize(path: &str, omit_slash: bool) -> Cow<'_, str> {
 #[doc(hidden)]
 pub fn join_paths<'a>(from: &'a str, to: &'a str) -> String {
     let from = remove_wildcard(&normalize(from, false));
-    from + &normalize(to, false)
+    from + normalize(to, false).as_ref()
 }
 
 fn begins_with_query_or_hash(text: &str) -> bool {


gbj:
Wait, how does this work for the `create_resource` case? Doesn't it just panic if you have a resource that actually does any async work? If the `Future` is guaranteed to always already be `Ready` then there's no real point in using a resource, is there?

itowlson:
Okay, with a bare Tokio runtime, and reusing the standard `ssr` implementation of `spawn_local`, it crashed with a message about needing to be in a LocalSet.  When I tried adding a LocalSet, I got a "condvar wait not supported" somewhere deep inside WASI.  I'll appeal to the Fermyon folks working on the Wasm internals for ideas, but in the meantime, @diversable, you mentioned you were working on an Axum WASI template - have you run across the `spawn_local` problem and if so how did you tackle it?

itowlson:
@gbj Well, I don't know if this is going to meet with your approval, but after the various dead ends discussed above, the only solution I have found is, I'm afraid, Spin-specific.  It introduces a Spin SDK dependency into Leptos, which feels nasty outside of the integration crates; but it is optional and is brought in only via a `spin` feature gate.  If there's a better way to manage this, do let me know!

After discussions with the Spin team, the longer goal is to move towards using a Tokio implementation inside the Spin SDK, instead of the hand-rolled executor.  This would be beneficial for Spin too, of course, but from the Leptos point of view it would hopefully remove the need to special-case (and would remove that ugly dependency).  I don't have a time frame for this, though, and can't make any promises.

In the meantime, I've tested this with resources that are both sync and async on the server, and it has worked in both cases.  Given the lack of automated coverage, if there's any other testing you'd like me to do, let me know.

The apparently unrelated changes to the string concatenation operation in `leptos_router` is because of a weird quirk in `smartstring` that makes `String + &String` not work _anywhere_ in an application that uses `smartstring` (https://github.com/bodil/smartstring/issues/7).  This affects Spin builds because the Spin router uses `routefinder` which uses `smartstring`.

gbj:
Cool! It looks like the `String + &String` thing is very limited, doesn't cause any breaking changes, and is very easy to work around, so fine by me. I am totally fine introducing a Spin-specific feature flag here, given that it's an optional dependency.

Thanks so much for your work on this.

itowlson:
Thanks!  I pushed an update to fix the CI error if you wouldn't mind re-approving to run the workflows.

itowlson:
I think the remaining CI error is a flake installing jq.

Issue #1952, opened 2023-10-26
Creating a component causes signals in props to run


luxalpa:
**Describe the bug**
When a component is created within a view and has a signal as props - even if it does not actually use it - it will run the signal once on component creation.

**Leptos Dependencies**
```toml
leptos = { version = "0.5.2", features = ["nightly"] }
leptos_meta = { version = "0.5.2", features = ["nightly"] }
leptos_actix = { version = "0.5.2", optional = true }
leptos_router = { version = "0.5.2", features = ["nightly"] }
leptos_reactive = { version = "0.5.2", features = ["nightly"] }
```

**To Reproduce**
Running the following code, even just on actix server, will cause the log message to show. Furthermore, due to the `resource::get` in that derived signal, we get a warning in the browser that we are reading from a resource outside of Suspense.
```rs
use leptos::*;

#[component]
pub fn App() -> impl IntoView {
    let resource = create_blocking_resource(|| {}, |_| async move { 100 });
    let signal = (move || {
        logging::log!("signal");
        resource.get() // <- this runs
    })
    .into_signal();

    view! {
        <OtherComponent signal/>
    }
}

#[component]
fn OtherComponent(signal: Signal<Option<i32>>) -> impl IntoView {}
```

**Expected behavior**
The signal should only run when it being subscribed to in a reactive scope.

**Additional context**
Running the code with `env_logger` and `RUST_LOG=debug` gives the output below, which hints that it might be due to the tracing.

```
[2023-10-26T14:50:27Z DEBUG tracing::span] provide_context;
[2023-10-26T14:50:27Z DEBUG tracing::span] provide_context;
[2023-10-26T14:50:27Z INFO  tracing::span] <App />;
[2023-10-26T14:50:27Z DEBUG leptos_reactive::resource] create_blocking_resource; ty=i32 signal_ty=()
[2023-10-26T14:50:27Z DEBUG leptos_reactive::effect] Effect::run(); defined_at=C:\Users\Luxalpa\.cargo\registry\src\index.crates.io-6f17d22bba15001f\leptos_reactive-0.5.2\src\resource.rs:244:5 ty=()
[2023-10-26T14:50:27Z INFO  tracing::span] <OtherComponent />;
signal
[2023-10-26T14:50:27Z DEBUG tracing::span] use_context;
[2023-10-26T14:50:27Z DEBUG tracing::span] use_context;
[...]

PR #2072, opened 2023-11-24
Change behavior of prop tracing to avoid evaluating lazy values

gbj:
The issue is #1952: Passing a lazy type like a derived `Signal` or a `Memo` as a component property now causes them to be eagerly evaluated before they're actually used. This is because the `Serialize` implementation for signals/memos reads their value. This is surprising because users do not expect these values to be read simply by passing through a component, and it changes the behavior/execution order of the program.

Switching to `Debug`, which doesn't read the inner value of a memo/signal, resolves the issue, but changes exactly what is logged by this tracing integration.

@luoxiaozero I'd be curious to hear your thoughts on this, since it was your contribution initially. Is switching to `Debug` here compatible with your use case, or do we need to try some further autoref specialization to create a special implementation for reactive types that doesn't use `Serialize`?

### DIFF
--- a/leptos_dom/src/macro_helpers/tracing_property.rs
+++ b/leptos_dom/src/macro_helpers/tracing_property.rs
@@ -1,5 +1,3 @@
-use wasm_bindgen::UnwrapThrowExt;
-
 #[macro_export]
 /// Use for tracing property
 macro_rules! tracing_props {
@@ -11,8 +9,9 @@ macro_rules! tracing_props {
         );
     };
     ($($prop:tt),+ $(,)?) => {
+        #[cfg(any(debug_assertions, feature = "ssr"))]
         {
-            use ::leptos::leptos_dom::tracing_property::{Match, SerializeMatch, DefaultMatch};
+            use ::leptos::leptos_dom::tracing_property::{Match, DebugMatch, DefaultMatch};
             let mut props = String::from('[');
             $(
                 let prop = (&&Match {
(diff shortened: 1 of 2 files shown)


luoxiaozero:
1. I have a question, why do lazy types (derived Signal or Memo) implement the Serialize trait, is it used internally by leptos, for user convenience serialization, or both.

2. For development tools, only deserialization is required. Debug can be used as a temporary solution for the moment, since Debug may not be stable in the future. https://doc.rust-lang.org/std/fmt/trait.Debug.html#stability

gbj:
> I have a question, why do lazy types (derived Signal or Memo) implement the Serialize trait, is it used internally by leptos, for user convenience serialization, or both？

Solely for convenience: So that you can, for example, do
```rust
#[derive(Serialize)]
struct MyData {
  foo: RwSignal<String>,
  bar: Memo<i32>,
  baz: bool
}
```

I think the *best* solution here might actually be to add a `tracing` feature to `leptos_dom`, and gate the `tracing_props` on that. Then, we wouldn't be serializing props in a way that's unexpected by users, but in a way they could opt into or out of.

luoxiaozero:
Totally agree, but could there be a more explicit name for the tracing feature, such as `trace-component-props`?  Because I am [leptos-devtools](https://github.com/luoxiaozero/leptos-devtools) in the implementation of a component tree display and the component of signal properties.  In the future, I might implement monitoring signal value changes.  Then I can add a `trace-signal` featrue for leptos_reactive, and finally a `devtools` featrue for lepots.  To enable all DevTools-related features.  (Just my initial thoughts, of course)

gbj:
Yes, this makes perfect sense to me. I'm going to close this and open a different PR that keeps the `Serialize` approach but feature-gates component prop tracing. Thanks for your input.

PR #2084, opened 2023-11-28
Improve macro hygiene

blorbb:
improved macro hygiene by using more fully qualified paths. Main thing was swapping trait methods called with `.` for fully qualified trait methods, so that users don't always need to import `leptos::*` / `leptos_router::*` for macros to work.

Didn't bother as much for fully qualified std traits since no one is working in no_std anyways

(minor bit: also reformatted some `quote_spanned!` bits to follow convention (no space before `=>`) and remove redundant variable bindings)

might need to change up some `#[allow(unused_braces)]` if incorrect warnings appear

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -415,7 +415,7 @@ impl ToTokens for Model {
                 };
                 let children = if is_island_with_children {
                     quote! {
-                        .children(::std::boxed::Box::new(move || ::leptos::Fragment::lazy(|| vec![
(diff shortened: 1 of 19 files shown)


blorbb:
updated the snapshots just to test if everything else works, I can't tell if they actually pass tho (what are the snapshots testing?)

gbj:
Cool, thanks! The snapshots are there to allow us to refactor macro internals without breaking macro output. Thanks for updating them for your new output. The fact that the CI on all the examples and all the crates passes suggests that changing the output doesn't break any behavior, so updated snapshots + all green otherwise here is perfect.

I think this effectively closes #227 as well.

PR #2089, opened 2023-11-28
Fix `aria-current` attribute value

alexisfontaine:
#1953 modified the `aria-current` attribute of the router link component to an invalid value.

As a result, the CSS selector `[aria-current=page]` is no longer working in the latest release `0.5.3`.


gbj:
Apologies, good catch, and thanks.

PR #2093, opened 2023-11-30
fix: add support for placing attributes on server functions

tlowerison:
Adding instrumentation to server functions is not straightforward (currently requires calling a different instrumented function). This commit takes all attributes placed on the server function (any attributes between `#[server]` and the source function body) and places them on both the generated front end and back end functions. If those attributes are only desirable on one of the targets (wasm or server side), they can be conditionally included using `cfg_attr`.


gbj:
Thanks! This looks great, and much appreciated.

PR #2094, opened 2023-12-01
docs: fix a typo

cyril-marpaud:


gbj:
The book is in the process of migrating to another repo -- I still need to delete it from this one but it is now at https://github.com/leptos-rs/book

PR #2099, opened 2023-12-07
Added documentation for server fns

rakshith-ravi:
We have a `#[warn(missing_docs)]` in our code and server_fns always throws an warning for that. This should fix it. Do let me know if I'm missing something


gbj:
Thanks!

PR #2106, opened 2023-12-13
add nix-darwin support, update flake

Gentle:
the last time this was updated is a long time ago and also it now supports nix on OSX :)


gbj:
Thanks!

PR #2107, opened 2023-12-13
Axum removed the axum::Server re-export

Gentle:
axum::Server is not re-exported anymore, this fixes compiling the doctests


Gentle:
Pull #2082 is better

PR #2113, opened 2023-12-16
Support additional context for route generation

NiklasEi:
Using Leptos as a static site generator currently has two main steps:
1. generating routes
2. building static routes

The second step allows passing additional context (e.g. global state) via `build_static_routes_with_additional_context` that can then be used in components, but the first step doesn't. This means all components that get evaluated during step 1 will fail to get the additional context, while others that are only evaluated during 2 will be able to use it.

This PR adds support for additional context during step 1

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -886,7 +886,7 @@ pub fn generate_route_list<IV>(
 where
     IV: IntoView + 'static,
 {
-    generate_route_list_with_exclusions_and_ssg(app_fn, None, || {}).0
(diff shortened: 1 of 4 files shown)


benwis:
Interesting, this definitely seems like a nice fix. I wonder why the actix check failed

gbj:
The failing checks are all failing on clippy on `leptos_reactive` (or another `leptos` crate), I think because of some disagreement between Rust versions over whether the `type_name_of_val` feature is stabilized. Hopefully fixed in https://github.com/leptos-rs/leptos/commit/a4bd7c4274bf7958e45ae2592f1937ce7b244706 

This looks good in any case. Thanks so much!

Issue #1806, opened 2023-09-29
Serving multiple frontends with the same server leads to only one being served


sbihel:
**Describe the bug**
I am trying to create a monolith that serves two web apps (`app1` and `app2`). When I spin up the server, the first web app that's opened (be it either `app1` or `app2`) is rendered correctly, but then when the other app is opened, it is rendered as the other app.

This is using hydration on the client and SSR on the server. If hydration is replaced with CSR on the client and SSR disabled on the server -- then the render happens twice for each app (because it seems hydration is still enabled), with one of the renders being correct and other not.

**Leptos Dependencies**
I am using `v0.5.0-rc3`.

**To Reproduce**
Here is a minimal reproduction example, with commands used in the README: https://github.com/sbihel/leptos-multiple-apps

**Expected behavior**
Each app renders as itself.

**Additional context**
An example use case for this could be to have a monolith with a user dashboard and an admin dashboard, where the data may be sensitive enough that you want to separate them whilst making it easier to share code. This may not be supported, and it may not even be a good idea. But considering the end result is confusing, and the fact `cargo-leptos` mentions that "_several projects can be defined and one package can be used in several projects_", I thought this might be of interest.

sbihel:
Ah yes I forgot to mention that, they're listening to different ports. And the reason why they have a router each is that I couldn't make it happen without it. As in, it's working properly, `localhost:3000` displays "app1" and `localhost:3001` displays "app2"

gbj:
Oh okay, I understand. Yes, `leptos_router` caches the set of possible route branches in a thread-local to avoid recalculating them on every single request.
https://github.com/leptos-rs/leptos/blob/870808e63f140a725b964400eb19a7f36ae6fbc0/router/src/components/routes.rs#L269-L289

So I guess if you have two separate apps using `leptos_router` being served on two separate ports but from the same server (and therefore being served from the same thread), then yes, this would be a problem.

I'm open to suggestions for a good solution.

sbihel:
I'm not sure if that's the issue, I tried the wrap the branches hashmap with another hashmap using a router ID but the result is the same. Or maybe it is only part of the issue -- I imagine the runtimes management in `thread_local` can cause issues?

PR #2135, opened 2023-12-25
Namespace caches and branches with new router IDs

sbihel:
I had another go #1806 and it turned out to be easier than I previously thought to solve naively. But I'm not sure this is an ideal solution, maybe the IDs shouldn't be UUIDs, or maybe the caches should be held by the objects directly.

I'm also finding it hard to add a unit/integration test, so I would welcome a suggestion on this as well.

### DIFF
--- a/router/Cargo.toml
+++ b/router/Cargo.toml
@@ -20,7 +20,6 @@ linear-map = { version = "1", features = ["serde_impl"] }
 once_cell = "1.18"
 regex = { version = "1", optional = true }
 url = { version = "2", optional = true }
-uuid = { version = "1", features = ["v4"] }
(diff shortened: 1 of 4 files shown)


gbj:
Rather than pulling in `uuid` maybe just a simple static `AtomicUsize` or something that increments every time you need to get a new router ID?

And rather than nested `HashMap`s, maybe using a `(AtomicUsize, String)` as the key of a single `HashMap`?

sbihel:
Thank you for the suggestions! Using tuple keys for the hashmap introduced a couple of clones, but I think it's still better

gbj:
Looks great, thanks so much!

PR #2139, opened 2023-12-29
feat: Add owning memos to allow memos that re-use the previous value

pheki:
Currently the memo function:
- Only receives a reference to the previous value when running
- Is required to return an owned value, even if it didn't change

This means that:
- The returned type needs to be allocated and copied one time per source signal change, something that may be very common for slices

This PR implements a Proof Of Concept proposing a new, more primitive, "raw memo", that allows the memo function to receive the previous value as owned, allowing it to be re-used by the memo, like effects (in fact, part of the implementation was taken from there).

"Normal" memos were changed to be implemented by using the raw memo, so there's no need to create a new reactive node. It should be backwards compatible. I haven't checked if there's any performance / binary size impact (benchmarks seems to be broken).

Unfortunately, I don't think the `create_slice` functions can take advantage of the raw memos with their current signature, but the `slice!` macro should be able to.

Related discussion: https://github.com/leptos-rs/leptos/discussions/1034#discussioncomment-5898913

# Main use-case: slices

The most prominent use-case is for slices[1][2][3], which can then be implemented to only clone if the value has changed. In [1], being forced to clone on every "source signal" change is actually noted as a major drawback. ([1](https://book.leptos.dev/15_global_state.html#option-3-create-a-global-state-struct-and-slices) [2](https://book.leptos.dev/view/04b_iteration.html#option-3-memoized-slices) [3](https://docs.rs/leptos/latest/leptos/macro.slice.html))

You can use to only clone when the value has changed:

```rust
let token = create_raw_memo(move |old_token| {
    state.with(move |state| {
        if let Some(token) = old_token.filter(|old_token| old_token == &state.token) {
            (token, false)
        } else {
            (state.token.clone(), true)
        }
    })
});
```

You can even re-use the allocation:

```rust
let token = create_raw_memo(move |old_token| {
    state.with(move |state| {
        let is_different = old_token.as_ref() != Some(&state.token);
        let mut token = old_token.unwrap_or_else(String::new);

        if is_different {
            token.clone_from(&state.token);
        }
        (token, is_different)
    })
});
```

### DIFF
--- a/leptos_reactive/src/memo.rs
+++ b/leptos_reactive/src/memo.rs
@@ -93,7 +93,62 @@ where
     })
 }
 
-#[allow(missing_docs)] // TODO
+/// Like [`create_memo`], `create_owning_memo` creates an efficient derived reactive value based on
+/// other reactive values, but with two differences:
+/// 1. The argument to the memo function is owned instead of borrowed.
+/// 2. The function must also return whether the value has changed, as the first element of the tuple.
+///
+/// All of the other caveats and guarantees are the same as the usual "borrowing" memos.
+///
+/// This type of memo is useful for memos which can avoid computation by re-using the last value,
+/// especially slices that need to allocate.
+///
+/// ```
+/// # use leptos_reactive::*;
+/// # fn really_expensive_computation(value: i32) -> i32 { value };
+/// # let runtime = create_runtime();
+/// pub struct State {
+///     name: String,
+///     token: String,
+/// }
+///
+/// let state = create_rw_signal(State {
+///     name: "Alice".to_owned(),
+///     token: "abcdef".to_owned(),
+/// });
+///
+/// // If we used `create_memo`, we'd need to allocate every time the state changes, but by using
+/// // `create_owning_memo` we can allocate only when `state.name` changes.
+/// let name = create_owning_memo(move |old_name| {
+///     state.with(move |state| {
+///         if let Some(name) =
+///             old_name.filter(|old_name| old_name == &state.name)
+///         {
+///             (name, false)
+///         } else {
+///             (state.name.clone(), true)
+///         }
+///     })
+/// });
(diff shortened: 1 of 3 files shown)


gbj:
This seems good! Might I suggest "owning" memo or something instead of "raw"? 

Re: defensively taking the BorrowMut twice: It seems to me that the only situation in which that could happen would be if something *during the memo function* caused the memo itself to rerun, right? i.e., it would have to do something like setting a signal inside it, that would trigger an effect, that would read from the memo, which will still be marked Check at that point so will rerun?

In this case `.take()`-ing the value out of the Memo will also cause a panic because `Memo::with()` always runs the memo function to generate a value if it doesn't exist, so this would still panic.

That is just off the top of my head, I don't have a good test case for either of those situations. In any case it is 100% a bug in the user's code so I wouldn't worry about it too much.

pheki:
> This seems good! Might I suggest "owning" memo or something instead of "raw"?

Yes! A lot better :)

> Re: defensively taking the BorrowMut twice: ...

Oh, sorry I didn't make it clear, I was gauging interest so I mostly copied that from the [effect implementation](https://github.com/leptos-rs/leptos/blob/16cf3c4eafd660ab68a7a8dff1e06096f061ffea/leptos_reactive/src/effect.rs#L353) (introduced on https://github.com/leptos-rs/leptos/commit/faa4dc4fa385c8df9c16070f129fd614939cc0ba) and assumed it would be similar for Memos. From you explanation it looks like it's not, so I simplified it.

The test was also mostly adapted from `leptos_reactive/tests/slice.rs`, but I now updated to match the use-case better.

Also added some docs, I think it's ready for review now!

pheki:
Hi! Any feedback for this?

gbj:
So sorry for the delay. This dropped off my radar after I looked at it the last time. It looks great now! Thanks for your work.

pheki:
ty!!!

Issue #2154, opened 2024-01-02
leptos_router: Feature: Support for trailing slashes.


NfNitLoop:
**Is your feature request related to a problem? Please describe.**

Relative URL resolution in plain HTML (`<a>`, not `<A>`) works by appending a relative path to the path in the URL bar, somewhat like when working with relative filesystem paths.  But in URLs the trailing `/` character is significant.  If it is present, the current URL is treated as a "directory", within which the URL is resolved. Otherwise, the last path part is treated as a "file" and is stripped before doing relative resolution.

So, without a trailing slash, visiting the URL `http://localhost:3000/foo`, an `<a href="bar.txt">` resolves to the URL `http://localhost:3000/bar.txt`.

But *with* a trailing slash, visiting the URL `http://localhost:3000/foo/`, the same link resolves as `http://localhost:3000/foo/bar.txt`.

So a trailing slash is a meaningful element of the route.  But leptos_router (or, maybe it's the actix implementation?) seems to be normalizing away trailing slashes.  Both `<Route path="/foo">` and `<Route path="/foo/">` end up serving the same path.

This is a problem because I'm hoping to gradually port an old codebase to Leptos. I want to keep the same URL structure for backward compatibility (i.e.: so I don't break old links). The relative links that I want to support are part of user input that can't be easily updated to use the `<A>` tags.

**Describe the solution you'd like**

SvelteKit has a nice way of handling trailing slashes.  See: <https://kit.svelte.dev/docs/page-options#trailingslash>

You can set a `trailingSlash` option that determines how the server and client handle trailing slashes.  You can configure this differently for different routes, and sub-routes inherit the parent behavior by default.

I'd love something similar in Router. Preferably I'd be able to set it once at the very top level on `<Router>` and just apply it to all of my routes by default.  Then I could override particular sub-routes w/ the same setting on `<Route>`.

**Describe alternatives you've considered**

1) Changing my URLs to have some `/foo/home` page instead of serving from `/foo/`.  

❌  Breaks link compatibility.
⚠️  Would still need to manually redirect from `/foo` to `/foo/home`.  If we could just redirect from `/foo` to `/foo/` that would be much better.

2) Forking Leptos router and adding this functionality for myself.  (😅)  

Not sure whether I can commit to doing it but if you'd like someone to help with this feature I might be interested enough to work on it.

benwis:
[...]

PR #2163, opened 2024-01-05
router: Retain trailing slashes in paths

skirsdeda:
This addresses client-side part of #2154 leaving path matching trailing slash insensitive (to be compatible with current behavior). For example `/*path/` would still match both `foo` and `foo/`. But when clicking on a link `foo/`, trailing slash would be retained.

I think it's sensible to keep matching compatible until server-side for this is figured out.

Since path resolution implementation in router is copied from solidjs router and that one retains trailing slashes, I just fixed regexp recreations in rust code to fully match ones from solidjs router:
https://github.com/solidjs/solid-router/blob/main/src/utils.ts#L5:
```js
const trimPathRegex = /^\/+|(\/)\/+$/g;
//
export function normalizePath(path: string, omitSlash: boolean = false) {
  const s = path.replace(trimPathRegex, "$1");
  //
}
```
(note that trailing slash is retained (match group `$1`) but excessive trailing slashes are trimmed)
https://github.com/solidjs/solid-router/blob/main/src/utils.ts#L37:
```js
export function joinPaths(from: string, to: string): string {
  return normalizePath(from).replace(/\/*(\*.*)?$/g, "") + normalizePath(to);
}
```
(matches wildcard pattern optionally from right and removes it along with all trailing slashes in `from` path)

### DIFF
--- a/router/src/components/routes.rs
+++ b/router/src/components/routes.rs
@@ -636,7 +636,10 @@ fn create_routes(
     }
     let mut acc = Vec::new();
     for original_path in expand_optionals(&route_def.path) {
-        let path = join_paths(base, &original_path);
+        // compat: trim_end_matches ensures that routes with trailing slash still match ones with no trailing slash
+        let path = join_paths(base, &original_path)
+            .trim_end_matches('/')
+            .to_string();
         let pattern = if is_leaf {
             path
         } else {
--- a/router/src/matching/resolve_path.rs
+++ b/router/src/matching/resolve_path.rs
@@ -51,7 +51,14 @@ fn has_scheme(path: &str) -> bool {
 
 #[doc(hidden)]
 fn normalize(path: &str, omit_slash: bool) -> Cow<'_, str> {
-    let s = path.trim_start_matches('/').trim_end_matches('/');
+    let s = path.trim_start_matches('/');
+    let trim_end = s
+        .chars()
+        .rev()
+        .take_while(|c| *c == '/')
+        .count()
+        .saturating_sub(1);
+    let s = &s[0..s.len() - trim_end];
     if s.is_empty() || omit_slash || begins_with_query_or_hash(s) {
         s.into()
     } else {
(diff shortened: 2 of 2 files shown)


NfNitLoop:
Testing this locally, with:

```rust
    let paths = generate_route_list(leptos_app::App);
    dbg!(&paths);
    cfg.leptos_routes(options.clone(), paths, leptos_app::App);
    cfg.app_data(Data::new(options));
```

I still see routes with their trailing slashes trimmed.  I bet this is leptos-actix.  I'm forking this branch and adding on some changes to leptos-actix as well to confirm.  🤞

skirsdeda:
@NfNitLoop I think that's expected, try removing this line: https://github.com/leptos-rs/leptos/pull/2163/files#diff-31f96b333e0c4d9391387d676c81a652b1a975d984ee51b3e2cbd1d4ecf8de8cR641

NfNitLoop:
> try removing this line

That did the trick! I think we should not trim there, because that path gets returned by `leptos_router::generate_route_list_inner()` to the integrations.  Without the trailing `/` you can't know the user's intent. 

The good news is that after removing that line, the leptos-actix integration does the "right thing" (IMO) and also keeps the trailing slash.  I wrote some unit tests to show this behavior at the two layers it passes through:

https://github.com/skirsdeda/leptos/compare/trailing_slashes...NfNitLoop:leptos:trailing_slashes?expand=1

> leaving path matching trailing slash insensitive

This has an unfortunate side-effect.  Now with `<Route path="/foo/">`,

* `/foo` on the server gives a 404. 
* `/foo/` on the server works.
* After hydration, `/foo` on the client does not give a 404. 🤔

IMO you really want your client and server behavior to match exactly or you'll end up with a lot of headaches.

I'm going to continue learning the codebase and see how I can make them align.  May also play around w/ some ideas in #2154 on top of this branch. 😊

skirsdeda:
Closing this in favor of #2172.

NfNitLoop:
Thanks for this work, though!  It was a good start for building the full feature, and a primer for me for diving into the Leptos codebase!

PR #2165, opened 2024-01-06
router: Fix router forgetting about state (fixes #2164)

### LINKED ISSUE #2164: Route `State` is being lost after Backward and/or Forward navigation
**Describe the bug**
`NavigateOptions` allows you to set a `state` for the current route. However, if the user navigates using the browsers forward and backward buttons this state is being lost. This is even true when just using JavaScript to navigate (i.e. completely avoiding the use_navigate/use_location). Just having the router on the page triggers this issue.

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.5.4", features = ["nightly"] }
leptos_meta = { version = "0.5.4", features = ["nightly"] }
leptos_actix = { version = "0.5.4", optional = true }
leptos_router = { version = "0.5.4", features = ["nightly"] }
leptos_reactive = { version = "0.5.4", features = ["nightly"] }
```

**To Reproduce**
Go to a page with leptos router on it, then run the following javascript in the console. Make sure that the line with pushState moves to a *different* route that is still handled by the router.

```js
  async function sleep() {
      await new Promise(resolve => setTimeout(resolve, 100));
  }
  async function do_navigate() {
      await sleep();
      
      history.pushState(123, 'title', '/about');
      await sleep();
      
      console.log("State from JS (before):", history.state);
      await sleep();
      
      history.back();
      await sleep();
      
      history.forward();
      await sleep();
      
      console.log("State from JS:", history.state);
  }
  do_navigate();
```

For more indepth examples (WithRouter, WithRouterWithJS and NoRouterJS which has the correct behaviour), see this gist: https://gist.github.com/luxalpa/c498a64e72f64d8ad293d47f6b66d2bf

**Expected behavior**
The state that's associated with the route should still be available after navigating backwards and forwards (that's kinda the point).

luxalpa:
I am pulling the "State" field in a similar manner to the other location fields within `BrowserIntegration::current()`. I also changed the State serialization for `None` from `UNDEFINED` to `NULL` as the latter is the default in browsers and should be more consistent. I am not sure what the reason was to originally use `undefined` and `State(None)`.

### DIFF
--- a/router/src/history/mod.rs
+++ b/router/src/history/mod.rs
@@ -37,13 +37,19 @@ pub struct BrowserIntegration {}
 impl BrowserIntegration {
     fn current() -> LocationChange {
         let loc = leptos_dom::helpers::location();
+        let state = window()
+            .history()
+            .and_then(|h| h.state())
+            .ok()
+            .and_then(|s| (!s.is_null()).then_some(s));
+
         LocationChange {
             value: loc.pathname().unwrap_or_default()
                 + loc.search().unwrap_or_default().as_str()
                 + loc.hash().unwrap_or_default().as_str(),
             replace: true,
             scroll: true,
-            state: State(window().history().and_then(|h| h.state()).ok()),
+            state: State(state),
         }
     }
 }
--- a/router/src/history/mod.rs
+++ b/router/src/history/mod.rs
@@ -43,7 +43,7 @@ impl BrowserIntegration {
                 + loc.hash().unwrap_or_default().as_str(),
             replace: true,
             scroll: true,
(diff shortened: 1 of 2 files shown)


gbj:
Thanks! I've never used `state` myself, it was just part of the original port of the router, so I can guarantee there was not a good reason. Most of the `state` stuff was `// TODO` for a long time so apologies if you run into additional issues.

gbj:
Failing tests are due to new clippy warnings outside this PR caused by new nightly clippy. I addressed those in #2169 — would you mind rebasing onto main so the CI will run again?

PR #2167, opened 2024-01-06
Fixed tests for `server_fn`

rakshith-ravi:
I have no clue if I'm doing the right thing

### DIFF
--- a/leptos_macro/tests/server.rs
+++ b/leptos_macro/tests/server.rs
@@ -3,7 +3,8 @@ use cfg_if::cfg_if;
 
 cfg_if! {
     if #[cfg(not(feature = "ssr"))] {
-        use leptos::{server, server_fn::Encoding, ServerFnError};
+        use leptos::{server, server_fn::{codec, ServerFn}, ServerFnError};
+        use std::any::type_name;
 
         #[test]
         fn server_default() {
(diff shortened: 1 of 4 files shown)


rakshith-ravi:
I'd specifically like some eyeballs on the way I'm comparing the path of the server endpoint. Given that there's a generated number attached to auto-generated endpoint paths, I didn't want to hard-code it. So I wanted to just check if the path begins with the given set of characters. Let me know if there's a better solution.

CC @benwis @gbj

benwis:
Overall it looks good, is it possible to avoid the numeric ranges by using `.contains()` or a similar helper method?

rakshith-ravi:
Yeah I was thinking of that too. What's the easiest way to say "give me all characters except the last N characters". If only rust had negative ranges, I'd have done `[..-32]` (or whatever the number is).

The only (highly inefficient solution) I can think of is:

```rust
path.chars().rev().skip(N).rev().collect()
```

Any better solutions?

gbj:
Oh thanks this is really helpful, and already showed me a bug. Could you actually revert the changes to the test examples that add `/` to the endpoint? i.e., rather than changing to `"/my_path"`, leave as `"my_path"` which currently works. Then the tests will fail for now, and I will have to fix it so that everyone's code doesn't break! This should be easy enough for me to fix in the `#[server]` macro.

rakshith-ravi:
Okay, so I did that and added a leading slash myself in the macro, but some tests have a leading /, so I those failed. Do you want me to do a conditional check?

gbj:
So, currently the macro determines `path` as follows (`server_fn_macro/src/lib.rs` line 380ish) 
```rust
    // note that all of this is emitted: the condition is at `const` evaluation time, 
    // not at macro expansion time
    let path = quote! {
        if #fn_path.is_empty() {
            #server_fn_path::const_format::concatcp!(
                #prefix,
                "/",
                #fn_name_as_str,
                #server_fn_path::xxhash_rust::const_xxh64::xxh64(
                    concat!(env!(#key_env_var), ":", file!(), ":", line!(), ":", column!()).as_bytes(),
                    0
                )
            )
        } else {
            #server_fn_path::const_format::concatcp!(
                #prefix,
                #fn_path
            )
        }
    };
```
I just played around with it a little and if you change to this it should be correct.
```rust
    // if the proc_macro2 Literal for and endpoint starts with a `/`, or it's empty, don't modify it
    let fn_path_starts_with_slash = fn_path.to_string().starts_with("\"/");
    let fn_path = if fn_path_starts_with_slash || fn_path.to_string() == "\"\""
    {
        quote! { #fn_path }
    } else {
        // if we have a plain endpoint name without a `/`, add the `/` before it
        quote! { concat!("/", #fn_path) }
    };
    let path = quote! {
        if #fn_path.is_empty() {
            #server_fn_path::const_format::concatcp!(
                #prefix,
                "/",
                #fn_name_as_str,
                #server_fn_path::xxhash_rust::const_xxh64::xxh64(
                    concat!(env!(#key_env_var), ":", file!(), ":", line!(), ":", column!()).as_bytes(),
                    0
                )
            )
        } else {
            #server_fn_path::const_format::concatcp!(
                #prefix,
                #fn_path
            )
        }
    };
```

rakshith-ravi:
Yup. That was the conditional I was referring to. Also, damn it! I wanted to feel like I wrote the code on my own and feel like I contributed to leptos 🤣

gbj:
I mean... I wrote a bug and you caught it by updating the tests, that's clearly better work than I was doing :smile:

rakshith-ravi:
Haha that's one way of looking at it 😅 

Also,

> ```rust
> path.chars().rev().skip(N).rev().collect()
> ```

Any thoughts on this?

rakshith-ravi:
Counterpoint to me feeling it's inefficient: It's just a test. Efficiency shouldn't matter, I suppose?

benwis:
I think it’s fine, not worth fussing too much over

rakshith-ravi:
Okay done. Now regardless of whether a user adds a slash or forgets to add one, we'll handle both cases

PR #2168, opened 2024-01-06
Added serde-lite implementation

rakshith-ravi:
Does this suffice? Do we need anything else?

This seems too simple to be it. There's gotta be more to it

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -75,7 +75,7 @@ pub fn server_macro_impl(
     let prefix = prefix.unwrap_or_else(|| Literal::string(default_path));
     let fn_path = fn_path.unwrap_or_else(|| Literal::string(""));
     let input_ident = input.unwrap_or_else(|| syn::parse_quote!(PostUrl));
-    let input = codec_ident(server_fn_path.as_ref(), input_ident);
+    let input = codec_ident(server_fn_path.as_ref(), input_ident.clone());
     let output = output.unwrap_or_else(|| syn::parse_quote!(Json));
     let output = codec_ident(server_fn_path.as_ref(), output);
     // default to PascalCase version of function name if no struct name given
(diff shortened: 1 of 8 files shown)


gbj:
I'd actually been picturing `SerdeLite` (or a similar name) as its own encoding option, parallel to `Json`, so you could choose between them on a per-server-function basis. I think the steps that would need to happen are
- add a feature to enable it
- duplicate the `Json` implementation and swap to `serde_lite` traits
- add the `serde_lite` intermediate steps
- update the `#[server]` macro so that it uses the `serde_lite` derive instead of the `serde` derive on the server fn type if you're using `serde_lite`

server_fn_macro/src/lib.rs
```diff
@@ -308,15 +306,15 @@ pub fn server_macro_impl(
...
+        "MultipartFormData" => quote! {},
+        "SerdeLite" => quote! {
+            Clone, #server_fn_path::serde_lite::Serialize, #server_fn_path::serde_lite::Deserialize
+        },
+        _ => quote! {
             Clone, #server_fn_path::serde::Serialize, #server_fn_path::serde::Deserialize
-        }
+        },
     };
     let serde_path = (!input_is_multipart && !input_is_rkyv).then(|| {
```
rakshith-ravi, line 319: Okay, what I've just pushed won't compile. I wanted to understand what this is for and why we need it

gbj: Okay, great! I just got to this file, everything else looks great to me.

As for why it doesn't compile: is it this kind of an error, when trying to use it in an example, about not finding `SerdeLite`?
```
error[E0412]: cannot find type `SerdeLite` in this scope
  --> src/todo.rs:58:35
   |
58 | #[server(AddTodo, "/api", input = SerdeLite, output = SerdeLite)]
   |                                   ^^^^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
1  + use crate::server_fn::codec::SerdeLite;
   |
1  + use server_fn::codec::SerdeLite;
   |

error[E0412]: cannot find type `SerdeLite` in this scope
  --> src/todo.rs:58:55
   |
58 | #[server(AddTodo, "/api", input = SerdeLite, output = SerdeLite)]
   |                                                       ^^^^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
1  + use crate::server_fn::codec::SerdeLite;
   |
1  + use server_fn::codec::SerdeLite;
```


`codec_ident` has a magic check for idents it recognizes as being internal to the `server_fn` crate, which (unfortunately?) needs to be kept in sync with the actual set.

https://github.com/leptos-rs/leptos/blob/d8514c153eae4628eafa45511970fbde43d890bc/server_fn_macro/src/lib.rs#L834-L843

However this also reveals a bug in the parsing for `input` and `output` because they don't actually support full paths:
```rust
// should work but doesn't at present
#[server(AddTodo, "/api", input = server_fn::codec::SerdeLite, output = server_fn::codec::SerdeLite)]
```


****

Other background:

This `derives` expression describes the additional derives we need to add to the server fn type.

Recall that a
```rust
my_server_fn(foo: String, bar: u32)
```
is converted to a
```rust
struct MyServerFn {
  foo: String, 
  bar: u32
}
```
This is so that arguments can be packed into the struct, serialized, sent to the server, deserialized on the server, and unpacked from the struct, then the original fn is called with them as the arguments.

This means we need to derive some traits on the struct:
- for the `serde` encodings (`serde_json`, `ciborium` and so on), we need to add `#[derive(Serialize, Deserialize)]`
- for `rkyv`, TODO
- for `serde_lite`, we need to derive the `serde_lite` `Serialize` and `Deserialize` traits (I think?)

Using the `#server_fn_path` bit is just so that we can use the reexported `serde` or `serde_lite` dependencies.

gbj:
Have tested and can confirm this actually works! I think this branch is either missing a commit or two from the other branch, or I need to go fix a few things over there, but they're in my work not yours... so I'm actually just going to merge this in and then will clean up what I need to over on the other branch. Thanks so much for doing the work on this.

DXist:
`ServerFnError` doesn't implement `serde_lite::Deserialize` when `serde-lite` feature is enabled

PR #2171, opened 2024-01-07
fix: change input type in error_boundary example

webmstk:
Chrome (mac) doesn't allow type non-digits in numeric inputs, so I removed `type=number` from input to make example more correct in this case.


gbj:
First: I very much appreciate the spirit of the PR, which I think is that the example is unrealistic, and it is much better to enforce constraints like this by doing them in built-in HTML rather than in user-land code, for many reasons. I 100% agree with that.

But I also notice that while you refer to Chrome, this constraint is not the same in other browsers. Firefox and Safari, at least in the versions I have installed, both display `type="number"` chrome (up and down arrows to change number, etc.) but *don't* enforce the validation; if I type or paste `foo` into them, `foo` appears in the input and the failed parsing triggers the `<ErrorBoundary/>`

In fact in Chrome, on an invalid copy and paste the `input` event still fires and tries to set the signal to a number, and fails to parse it, leaving a blank space in the UI and a blank value. In this case triggering the `<ErrorBoundary/>` is actually the only indication of an invalid input other than the browser just doing nothing!

So ironically assuming that the browser will enforce these pretty obvious constraints has the potential to make your UX worse because the behavior is incosistent across browsers.

I'd be open to either
1) adding an explanatory comment along those lines, saying that this is ensuring a consistent validation experience across browsers
2) creating a more realistic example if you can think of one: i.e., rather than loosening the browser constraint, creating a more realistic situation where we might have an error on the Rust side

webmstk:
Ok, let's stick with an exisiting example, I don't think it is a big deal.
I'll make a new pr if I come up with more realistic example :)

PR #2172, opened 2024-01-08
Better handling for trailing slashes. (#2154)

NfNitLoop:
This adds a trailing_slash option to `<Router>` and `<Route>`.

By default, this option is backward compatible with current Leptos
behavior, but users can opt into two new modes for handling trailing
slashes.


NfNitLoop:
Things still WIP:

- [ ] A few TODOs
- [ ] Figure out a way to test redirects in unit tests.  (They're working for me in Actix locally.)
- [ ] Are there any tests I'm forgetting to run?  `cargo +nightly test` doesn't work for me on `main`.  So I've been running it in the `leptos/`, `router/` and `integrations/actix/` subdirectories.

NfNitLoop:
Link to #2154, please, GitHub!  😊

router/src/extract_routes.rs
```diff
@@ -11,7 +11,7 @@ use std::{
 
 /// Context to contain all possible routes.
 #[derive(Clone, Default, Debug)]
-pub struct PossibleBranchContext(pub(crate) Rc<RefCell<Vec<Branch>>>);
+pub struct PossibleBranchContext(pub Rc<RefCell<Vec<Branch>>>);
```
NfNitLoop, line 14: TODO: I wasn't happy with making this public.  It was a quick hack for `test_unique_route_ids()`.

I think I can revert this if I move the tests from a standalone `testing/foo.rs` file to a test module. Happy to follow your suggestions here.

router/tests/extract_routes.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    assert!(branches
+        .iter()
+        .flat_map(|branch| &branch.routes)
+        .map(|route| route.id)
+        .all_unique());
+}
+
+/// This is how [`generate_route_list_inner`] gets its RouteDefinitions.
+/// But it doesn't expose it anywhere where we can easily test it, so here's a quick copy:
```
NfNitLoop, line 143: Also not happy with this copy/paste job.  I suppose this is another point toward making this a module test instead of a standalone test.  That might let me refactor this chunk out of generate_route_list_inner and test it directly w/o having to make it public.

rakshith-ravi:
Quick question - is this the same as #2154?

rakshith-ravi:
Wait, they're both by you. My bad. I'm sleep deprived. What's different in this than that, if you wouldn't mind explaining?

NfNitLoop:
@rakshith-ravi  #2163 isn't by me, it's by @skirsdeda. But we were discussing it as an incremental approach to this larger feature implementation in the feature request, so I provided some feedback over there and based these changes atop that work.

IMO we probably shouldn't merge #2163 in its current state. Without [10915f9](https://github.com/leptos-rs/leptos/pull/2172/commits/10915f9488b035f13292f4312f4234c8ef8caa9f) the front- and back-end routing are slightly different.

rakshith-ravi:
Yeah I got confused with the issue numbers for a second there. Like I said, super sleep deprived rn 😅 

> so I provided some feedback over there and based these changes atop that work.

Ahh, got it. Greg will have a better idea about this, but just my two cents - how about y'all put a PR on each other's branches and then have the complete slash handling as a single PR to main? That way, it'll be a feature complete PR.

On the flip side, it would mean a larger PR, so reviewing it might be difficult. Idk. Just putting both sides of the argument

NfNitLoop:
> how about y'all put a PR on each other's branches and then have the complete slash handling as a single PR to main?

This is the full PR.  Minus a couple TODOs and any suggestions I get from maintainers. 😊

router/tests/extract_routes.rs
```diff
@@ -0,0 +1,167 @@
...
+            "/baz/:name",
+            "/baz/:name/",
+            "/foo",
+            "/foo/",
+        ],
+    )
+
+    // TODO:
+    // Test we get a redirect from "/foo/" to "/foo"
+    // Test we get a redirect from "/bar" to "/bar/".
```
NfNitLoop, line 96: I've tested this locally in my browser, but I'm not immediately sure how to test it in a unit test. Suggestions welcome!  😊

benwis:
> > how about y'all put a PR on each other's branches and then have the complete slash handling as a single PR to main?
> 
> This is the full PR. Minus a couple TODOs and any suggestions I get from maintainers. 😊

This looks pretty solid to me, I've started the checks. I think we should add some mention to the interplay between the Axum and Actix routers

benwis:
Probably also want to run cargo clippy against your repo

NfNitLoop:
- [ ] Blocker:  In testing, I found that these changes break route matching for "" (leptos) and "/" (web servers).  I'm capturing that in some tests and will devise a fix but probably not until tomorrow night.

NfNitLoop:
Are all of these test failures in CI valid failures that I need to look into? Or are some just known to be flakey?

benwis:
> Are all of these test failures in CI valid failures that I need to look into? Or are some just known to be flakey?

@gbj is probably more familiar than I, but some of the examples test suites are failing now. Is this common?

gbj:
I'll have the CI rerun the failed tasks just in case. I would need to check out the branch and test out the examples myself to give a definitive answer, but I'm seeing failing e2e tests on several examples that do involve routing, so it's worth checking whether those examples actually work with the changes.

NfNitLoop:
OK, thanks. I hope to have more time to look into these tonight. Hopefully it's just the known issue with `"/"` and ""` routes breaking that I mentioned above.

maybe TMI but just to share: 

I've already looked into that a bit and found it was because of the changes to path normalization. The problem is that path normalization works the same everywhere … *except* root paths. For both Actix  and Axum, (and HTTP/HTML, AFAICT) root paths of "/" and "" are equivalent. So the error comes when Leptos specifies a root path of "", but the web server gives us a path of "/", which Leptos then can't doesn't find a match for (in this branch, given its more precise path matching.)  I've got a couple different fixes I'm testing out.

NfNitLoop:
OK, I've done local testing with Actix and Axum and have fixed my blocking issue. I bet that was the breaking issue for those integration tests, given they couldn't even render the first page of the site.🤞

I started down the path of trying to run the tests for myself locally but it looks like they require manual installation of some test tools and it's getting late here. If one of y'all can kick these off tonight/tomorrow I'll come back and check results here.  Thank you!  😊

benwis:
> OK, I've done local testing with Actix and Axum and have fixed my blocking issue. I bet that was the breaking issue for those integration tests, given they couldn't even render the first page of the site.🤞
> 
> I started down the path of trying to run the tests for myself locally but it looks like they require manual installation of some test tools and it's getting late here. If one of y'all can kick these off tonight/tomorrow I'll come back and check results here. Thank you! 😊

Looks promising, although it seems to have failed a cargo fmt check

NfNitLoop:
>  it seems to have failed a cargo fmt check

Weird, I would've sworn I ran `cargo fmt` before pushing my last commit lsat night.  🤷 

Thankfully that's an easy one to fix!  I assume/hope we'll squash merge this messy history away? 😅

NfNitLoop:
And merged? Just like that!   Thank you so much! I'll definitely be using this in my own project(s)!  😊

benwis:
Yep, the tests passed, it’s a useful feature, and the code’s solid. Thanks for contributing!

gbj:
@NfNitLoop @benwis Sorry to say this PR broke the router in a way that unfortunately was not caught by the existing CI setup. 

Two examples:
1) This valid approach for defining a param, and a fallback if the param was not provided, now panics at runtime when Axum tries to create duplicate routes
```
<Routes>
  <Route path="bar" view=|| view! { <Outlet/> }>
    <Route path="" view=|| view! { <p>"Empty"</p>}/>
    <Route path=":id" view=|| view! { <p>"ID"</p>}/>
  </Route>
</Routes>
2) Similarly, it breaks matching when there is a param at the root level, as in the `router` example in the repo or in #2201 

Given the breakages in existing code I'm going to have to revert this and yank the last two releases, since I unfortunately didn't catch it before then. If you can find a way to implement this that does not cause these issues, I will be happy to review a new PR.

PR #2194, opened 2024-01-17
fix: ci stopped detecting leptos or example changes

agilarity:
Fix the globstar patterns that specify directory changes.

Two checks were not successful.


gbj:
Looks good -- thanks so much!

PR #2199, opened 2024-01-18
Implement Into TextProp for signals

SleeplessOne1917:
I'm surprised the impl on line 1357 didn't already cover these.


gbj:
The CI is failing on updated nightly clippy warnings that came from outside your commit -- would you mind rebasing so the tests will run on your own code?

At a glance, I would have to assume that this implementation will conflict on `nightly` with the implementation for `Fn()` at line 1357 that you point out. These could be cfg-gated to only be used if that feature is not enabled.

SleeplessOne1917:
I have the most recent changes from main merged into this PR now. Is the `cfg` gate you have in mind enabling these only if it is not nightly?

gbj:
Correct. If you look at any of the failing checks here, they are along the lines of
```
error[E0119]: conflicting implementations of trait `From<signal_wrappers_read::Signal<_>>` for type `TextProp`
    --> /home/runner/work/leptos/leptos/leptos_reactive/src/signal_wrappers_read.rs:1368:1
     |
1357 | / impl<F, S> From<F> for TextProp
1358 | | where
1359 | |     F: Fn() -> S + 'static,
1360 | |     S: Into<Oco<'static, str>>,
     | |_______________________________- first implementation here
...
1368 | / impl<T> From<Signal<T>> for TextProp
1369 | | where
1370 | |     T: Into<Self> + Clone,
     | |__________________________^ conflicting implementation for `TextProp`
```
This is because, as you predicted, these implementations create a conflict with the blanket implementation for `Fn()`.

`Fn()` is only implemented for these types when the `nightly` feature is enabled, so you can add `#[cfg(not(feature = "nightly))]` to your additional implementations to avoid the conflict.

PR #2200, opened 2024-01-18
docs: `View::render_to_string` panic

chrisp60:
This caused me some trouble recently. Just thought it may be helpful


benwis:
Thanks for the doc comment, it can definitely be tricky

Issue #1594, opened 2023-08-26
Support for static site generation (SSG) and incremental static regeneration (ISR)


gbj:
## Is your feature request related to a problem? Please describe.

Currently, Leptos renders HTML on the server as part of a simple request/response cycle: someone makes an HTTP request for a certain page, Leptos responds by rendering that page on demand. This allows you to flexibly do things like "look this blog post up in a database" or "look up the preferences for the current user, they're logged in."

There are two kinds of pages that do not need on-demand rendering: 
1. Content pages, like a blog post, that are known at build time (which may or may not = compile time, but is a single discrete event; e.g., adding a row to a `posts` table or adding a `.md` file to a directory)
2. Static pages, like an About page, that are known at compile time 

Either of these sorts of pages can be rendered to HTML ahead of time, and served directly on request.
 
Leptos is quite efficient at rendering the second category of static pages. Because of the way that the `view` macro optimizes for server rendering, a large static page that just consists of HTML rendered by the `view` macro is converted, at compile time, into a big string, and this is served in response to a request. Whether it's faster or slower to serve it from a file will probably depend on the server, how deeply nested routes and components are, etc. In any case, it is already very fast.

But for the category of things like "read this Markdown file, convert it to HTML, and then serve it" there's a potential for a significant performance improvement by using static site generation or incremental static regeneration.

## Describe the solution you'd like 

Just to define these terms
- Static site generation (SSG) has a defined "build step" that builds the page for every possible option up front. The cost for any page request is O(1), as it just serves the pre-built HTML file.
- Incremental static regeneration (ISR) lets you build individual pages without needing to build the entire site up front. The cost for page requests amortizes to O(1): there is an initial O(n) rendering of the page the very first time it is requested, but this is cached as an HTML file which is what's served for every subsequent request.

The Leptos `<Route/>` component can, I think, be extended, possibly by extending `SsrMode`, to include `Static` and `Incremental`. `SsrModes` work on a per-route priority system, so in this case any path that contains a (nested or plain) route with `Static` will be statically generated; if there's no `Static`, but there is `Incremental`, it will be incrementally built and cached; then the current behavior.

## Additional requirements

**Path generation** `Static` would require some equivalent to [`getStaticPaths()`](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration), and async function that tells you which paths you need to build.

[...]

PR #2207, opened 2024-01-20
Make static rendered routes servable and provide context to static_params

NiklasEi:
This removes the `.static` from file names for static routes. Some cleanup methods no longer made any sense and were removed, too. With this change, the build output from static routes can be served as a static website without any file renaming (see discussion in #1594).

Additional context is provided for the static_params closures.


NiklasEi:
These changes work fine using only static routes and `StaticMode::Upfront`. I have not tested them with `Incremental` yet!

PR #2208, opened 2024-01-20
feat: `Default` for `LeptosOptions`, `ConfFile`

chrisp60:
Unsure of the intended behavior here as doc comments for the field state there is a default. It seems the default relies on `LEPTOS_OUTPUT_NAME`.
https://github.com/leptos-rs/cargo-leptos/blob/50dc2e74b7bbb8f98c2fae3844b707d48d6b4b6a/src/config/project.rs#L77 

Relying directly on the cargo env var seems like a sensible default.

Please feel free to close this if the current behavior is intended.


gbj:
That looks right to me; thanks!

Issue #1755, opened 2023-09-19
Support for axum extractors directly


Beiri22:
You are using macro magic to create server functions. I just wonder if it was possible to avoid an explicit `extract` function in favor of directly defining the extractors in the handler methode signature as axum does. Could that be possible?

Beiri22:
I mean, it's just a proposal. If you ask me, the axum syntax of extractors is pretty neat.

> Well, here's the question: What does it look like to call, from the client, a server function that takes arguments that only exist on the server? Are they all `Option<_>` and when you call it from the client you pass `None` and they become `Some(_)` when received on the server? To me, that's much more magical and unpredictable.

As we are talking of macros, for the client there should be a wrapper so that the client calls the function as he does right now.

> Doing it inline would also significantly increase the amount of macro "magic" going on, i.e., would shift more of the logic into a proc macro instead of a function call.

That might be the result, you're right. But on the other hand it would reduce boilerplate in favor of a clear and understandable syntax. You would, of course, need a few lines in the book to make the user understand the principle.

Beiri22:
As I see, the server macro is already a procedural macro. I'm unfortunately not too much into this topic; but is a procedural macro able to check if an argument is for example FromRequestParts?

gbj:
> is a procedural macro able to check if an argument is for example FromRequestParts?

No, unfortunately not; all proc macros do is transform a stream of tokens into a different stream of tokens, and then run before the compiler does anything having to do with types, so they have no access to type information like this.

> But on the other hand it would reduce boilerplate in favor of a clear and understandable syntax.

Here's the current form
```rust
#[server(QueryExtract, "/api")]
pub async fn query_extract() -> Result<String, ServerFnError> {
    // these are just some imports, not boilerplate
    // putting them inside the server fn body means they don't need to be available for WASM
    use axum::{extract::Query, http::Method};
    use leptos_axum::extract;

    // uses the exact same syntax as Axum handlers/extractors: 
    // typed arguments in an async function
    extract(|method: Method, res: Query<MyQuery>| async move {
            format!("{method:?} and {}", res.q)
        },
    )
    .await
    // pretty sure this can just be `?`
    .map_err(|e| ServerFnError::ServerError("Could not extract method and query...".to_string()))
}
```
If changed it would be something like this
```rust
#[server(QueryExtract, "/api")]
pub async fn query_extract(
    // the types are inline here because having them in a `use` outside the
    // server function body would mean they need to be compilable for WASM
[...]

PR #2209, opened 2024-01-20
add extractor_with_state function with better API than extract_with_state

rosarp:
I faced one issue extracting Signed Cookie using axum-extra & leptos-axum libs. 
SignedCookieJar needs Key to be in AppState. And thus, extractor need to be passed with AppState so that internally in SignedCookieJar::from_request_parts , using state cookie can be verified and decoded to be returned.

Thanks @gbj & @Beiri22 - I came across this discussion - [#1755](https://github.com/leptos-rs/leptos/issues/1755) 

Adding use case to the question [here](https://github.com/leptos-rs/leptos/issues/1755#issuecomment-1729035465) 

Please do guide me if documentation needs more details. 

```
 // AppState is some type that implements `Deserialize + Serialize`
 struct AppState {
     pub key: Key,
 }

 impl FromRef<AppState> for Key {
     fn from_ref(state: &AppState) -> Self {
         state.key.clone()
     }
 }
 #[server]
 pub async fn cookie_extract() -> Result<SignedCookieJar, ServerFnError> {
     use axum_extra::extract::cookie::SignedCookieJar;
     use cookie::secure::Key;
     use leptos_axum::*;
     let state: AppState = use_context::<AppState>()?;
     let jar: SignedCookieJar<Key> = extractor_with_state(state).await?;

     Ok(jar)
 }
```

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1943,6 +1943,52 @@ where
         .map_err(|e| ServerFnError::ServerError(format!("{e:?}")))
 }
 
+/// A helper to make it easier to use Axum extractors in server functions, with a
+/// simpler API than [`extract_with_state`].
+///
+/// It is generic over some type `T` that implements [`FromRequestParts`] and can
+/// therefore be used in an extractor. The compiler can often infer this type.
+///
+/// Any error that occurs during extraction is converted to a [`ServerFnError`].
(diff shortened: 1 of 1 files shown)


gbj:
Thanks. I had already done this in the 0.6 branch, which was just merged into main.

PR #2218, opened 2024-01-22
Adding a Beginner's Tip to HtmlElement<El>

sjud:
I kept tripping up on this and for some reason it was hard to wrap my head around so I added a "beginner's tip" to explain the transparent nature of HtmlElement<El>. I think tips like this in docs will go a long way in on boarding new community members! Thanks in advance!

### DIFF
--- a/leptos_dom/src/html.rs
+++ b/leptos_dom/src/html.rs
@@ -276,6 +276,12 @@ impl ElementDescriptor for Custom {
 cfg_if! {
   if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
     /// Represents an HTML element.
+    /// Beginner's tip: 
+    /// HtmlElement<El> implements Deref where El implements Deref.
+    /// When El has a corresponding web_sys::HtmlElement -> El will implement Deref for it.
+    /// For instance leptos::HtmlElement<Div> impl Deref for web_sys::HtmlDivElement
+    /// Because of [Deref Coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion) you can call applicable web_sys::HtmlElement methods on HtmlElement<El> as if it were that type.
+    /// In other words leptos::HtmlElement<leptos::html::Div> is a [Smart Pointer](https://doc.rust-lang.org/book/ch15-00-smart-pointers.html) to [web_sys::HtmlDivElement](https://docs.rs/web-sys/latest/web_sys/struct.HtmlDivElement.html)
     #[must_use = "You are creating an HtmlElement<_> but not using it. An unused view can \
     cause your view to be rendered as () unexpectedly, and it can \
     also cause issues with client-side hydration."]
--- a/leptos_dom/src/html.rs
+++ b/leptos_dom/src/html.rs
@@ -276,12 +276,14 @@ impl ElementDescriptor for Custom {
 cfg_if! {
   if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
     /// Represents an HTML element.
-    /// Beginner's tip: 
-    /// HtmlElement<El> implements Deref where El implements Deref.
-    /// When El has a corresponding web_sys::HtmlElement -> El will implement Deref for it.
-    /// For instance leptos::HtmlElement<Div> impl Deref for web_sys::HtmlDivElement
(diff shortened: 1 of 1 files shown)


leptos_dom/src/html.rs
```diff
@@ -276,6 +276,12 @@ impl ElementDescriptor for Custom {
 cfg_if! {
   if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
     /// Represents an HTML element.
+    /// Beginner's tip:
```
rakshith-ravi, line 279: There seems to be trailing whitespace here, which is why `cargo fmt` is failing. Would you mind running rustfmt on the code once?

gbj, line 279: Could you make this a header? 
```md
### Beginner's Tip
```

leptos_dom/src/html.rs
```diff
@@ -276,6 +276,12 @@ impl ElementDescriptor for Custom {
 cfg_if! {
   if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
     /// Represents an HTML element.
+    /// Beginner's tip: 
+    /// HtmlElement<El> implements Deref where El implements Deref.
+    /// When El has a corresponding web_sys::HtmlElement -> El will implement Deref for it.
```
gbj, line 281: Currently, all of these are listed as plain text. If you make them links like \[\`web_sys::HtmlElement\`\] then they will link to the appropriate docs. This could be done for `HtmlElement` and `Deref` at the least. Other types can be wrapped in backticks \`\` so they appear as code not plain text in the docs.

gbj (commented):
Thanks, I think this makes sense to add as a tip and I added a couple notes on formatting/linking.

leptos_dom/src/html.rs
```diff
@@ -276,6 +276,12 @@ impl ElementDescriptor for Custom {
...
+    /// For instance leptos::HtmlElement<Div> impl Deref for web_sys::HtmlDivElement
+    /// Because of [Deref Coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion) you can call applicable web_sys::HtmlElement methods on HtmlElement<El> as if it were that type.
+    /// In other words leptos::HtmlElement<leptos::html::Div> is a [Smart Pointer](https://doc.rust-lang.org/book/ch15-00-smart-pointers.html) to [web_sys::HtmlDivElement](https://docs.rs/web-sys/latest/web_sys/struct.HtmlDivElement.html)
```
blorbb, line 284: I don't think this counts as a smart pointer, even if it implements `Deref`. The book page says "smart pointers implement the `Deref` and `Drop` traits" - I don't think `leptos::HtmlElement` implements `Drop`.

leptos_dom/src/html.rs
```diff
@@ -276,6 +276,12 @@ impl ElementDescriptor for Custom {
 cfg_if! {
   if #[cfg(all(target_arch = "wasm32", feature = "web"))] {
     /// Represents an HTML element.
+    /// Beginner's tip: 
+    /// HtmlElement<El> implements Deref where El implements Deref.
+    /// When El has a corresponding web_sys::HtmlElement -> El will implement Deref for it.
+    /// For instance leptos::HtmlElement<Div> impl Deref for web_sys::HtmlDivElement
+    /// Because of [Deref Coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion) you can call applicable web_sys::HtmlElement methods on HtmlElement<El> as if it were that type.
```
blorbb, line 283: Maybe add something about needing to occasionally manually deref. Feel free to copy the below.

If both `HtmlElement` and one of its derefs have a method with the same name, the dot syntax will call the method on the inherent impl (i.e. `HtmlElement`). You may need to manually deref to access other methods, for example, [`(*el).style()`](https://docs.rs/web-sys/latest/web_sys/struct.HtmlElement.html#method.style) to get the [`CssStyleDeclaration`](https://docs.rs/web-sys/latest/web_sys/struct.CssStyleDeclaration.html) instead of calling [`leptos::HtmlElement::style`].

blorbb (commented):
sorry for the random intrusion, I ran into this confusion recently too! Added some comments about other things I've noticed.

sjud:
Hi Everyone,
I'll add everyone's changes and update asap! Thanks :)

sjud:
Awesome, thanks so much for the feedback everyone!

gbj:
Looks good -- thanks!

PR #2227, opened 2024-01-25
fix: importing `quote` fixes the `private macro use` error in the build (in CI)

sabify:


rakshith-ravi (approved):

gbj:
Thanks, much appreciated! I didn't even think to check something so simple, and didn't have any idea the error message could be pointing to that. This is a great help.

PR #2232, opened 2024-01-26
Update README.md framework comparisons

gbj:
I removed the earlier comparisons from the README, as I was getting the feeling they were out of date, and I don't like the feeling of "competition brain" between frameworks; these have been in there since 0.0.x, but Leptos is much more established now and I didn't feel the need to differentiate as strongly. @rakshith-ravi pointed out that these are still helpful so I'll add them back, updated/edited. Open to feedback on any details here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,25 @@ pub fn SimpleCounter(initial_value: i32) -> impl IntoView {
     }
 }
 
+// we also support a builder syntax rather than the JSX-like `view` macro
+#[component]
+pub fn SimpleCounterWithBuilder(initial_value: i32) -> impl IntoView {
+    use leptos::html::*;
+
+    let (value, set_value) = create_signal(initial_value);
+    let clear = move |_| set_value(0);
+    let decrement = move |_| set_value.update(|value| *value -= 1);
+    let increment = move |_| set_value.update(|value| *value += 1);
+
+    // the `view` macro above expands to this builder syntax
+    div().child((
+        button().on(ev::click, clear).child("Clear"),
+        button().on(ev::click, decrement).child("-1"),
+        span().child(("Value: ", value, "!")),
+        button().on(ev::click, increment).child("+1")
+    ))
+}
+
 // Easy to use with Trunk (trunkrs.dev) or with a simple wasm-bindgen setup
 pub fn main() {
     mount_to_body(|| view! {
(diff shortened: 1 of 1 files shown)


rakshith-ravi:
Doesn't necessarily need to be a comparison or competition of sorts, but rather, just wanted to point out how leptos is focused on performance and the core tenets of leptos. Not having a comparison with any other framework works too

SilasLock:
In my opinion, these kinds of inter-framework comparisons are super helpful for new users! Lots of people come into the Rust+wasm space unsure what the basic lay of the land is, need to decide which framework to use for their project, and their first question is "what's the difference between Yew/Dioxus/Leptos?" It's a valid question for people to ask, and I think it's good that the readme provides a good answer!

[Here](https://www.reddit.com/r/rust/comments/y5scij/comment/isodoyy/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) and [here](https://www.reddit.com/r/rust/comments/y5scij/comment/isnldq1/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) are two comments from the Rust subreddit where people learned about Leptos for the first time, and their first question was "how is this different from \<other framework\>" and though I'm having trouble finding them now, I've seen several YouTube comments on Leptos-related videos where people ask the exact same question. Newcomers to Rust frontend web frameworks will always exist for whom Leptos/Sycamore/Yew/Dioxus are novel technologies, so even though it's been a year and Leptos is a lot more of an well-known name than it used to be, it's still helpful to provide a summary of how it differs from other options.

Personally, I've never seen them as creating "competition brain" between frameworks. I think people come into this with more of a "please educate me on the differences between my options!" mindset than a "let's find out which framework is the best, and would totally destroy the others in a duel to the death!" mindset. But that's just my personal experience. :blush:

Looking over the new version in this pull request, I like it a little better than the old version. It removes some of the outdated information about Dioxus (e.g. the performance boosts from Dioxus 0.3) and emphasizes that Dioxus has an additional focus on non-web/desktop applications, which is an important comparison point missing from the old readme. So, it looks good to me!

gbj:
Thanks to both of you -- and I've restored that JSX reference, as well as adding the builder syntax so people know it exists!

PR #2235, opened 2024-01-27
chore: define `edtion = "2021"` in `rustfmt.toml`

chrisp60:
See [leptosfmt](https://github.com/bram209/leptosfmt?tab=readme-ov-file#using-with-rust-analyzer) for why this may be needed.

TLDR: when rustfmt is hooked by leptosfmt, you need to specify an edition explicitly.


benwis:
LGTM

PR #2238, opened 2024-01-28
fix: correctly track source in `create_local_resource` (closes #2237)

### LINKED ISSUE #2237: create_local_resource does not refetch when source signal changes on experimental-islands
**Describe the bug**
A create_local_resource inside an island does not re-run the fetcher when the source signal changes, in Leptos 6.0 beta versions (so far).

It was previously working on Leptos 5.5 (3f34d9bb23a180abd083b059d7d796994442682c)

In this reproduction, you can see that the signal when routed through the LocalResourceTest ServerFn, does not re-run and match the signal itself. It remains blank, as it only appears to run the fetcher on initial load.

**To Reproduce**
```rust
use leptos::*;

#[server(LocalResourceTest, "/api")]
pub async fn local_resource_test(input: String) -> Result<String, ServerFnError> {
    Ok(input)
}

#[component]
pub fn Test() -> impl IntoView {
    view! { <TestIsland/> }
}

#[island]
pub fn TestIsland() -> impl IntoView {
    let (input, set_input) = create_signal("".to_string());

    let local_resource_test = create_local_resource(move || input(), move |input| local_resource_test(input));

    view! {
        <input style="border: 1px solid black;" type="text" on:input=move |ev| { set_input(event_target_value(&ev)) }/>

        <div>
            <Transition fallback=move || ()>

                {move || match local_resource_test.get() {
                    Some(Ok(result)) => format!("Here is the response from the local resource: {}", result),
                    _ => "".to_string(),
                }}

            </Transition>
        </div>

        {move || format!("Here is the input signal, so you can see it does update and change and should trigger the re-fetch: {}", input())}
    }
}
```

gbj:


benwis:
LGTM :)

PR #2241, opened 2024-01-29
fix: error rather than panicking if unable to send response in Axum integration

gbj:
Because of the awkward setup of spawning into a pinned thread, the Axum integration tries to send its response through a channel (due to the weird Send/Sync workaround currently in place) and unwraps awaiting the receiver
https://github.com/leptos-rs/leptos/blob/0ff1e279a2774730ab01f28d2c794217de35a19d/integrations/axum/src/lib.rs#L332

Panicking here (or anywhere in these handlers) is a bad idea, and can cause difficult-to-reproduce issues.

I'm wondering if replacing all these with something like this might be a good idea.


benwis:
Sure, I don't think it hurts. I've yet to see someone hit these, but a better experience is good if the do

PR #2243, opened 2024-01-30
Added missing doc comment and changed the type for as_query() 

benwis:
as_query() is unimplementable since it returns Option<&str> and not everyone has a way to get a &str of the Headers, so I transformed it into a Cow<'_, str> like all the other methods


benwis:
I'm going to close this for now, since we came up with a workaround. Will reopen later if it doesn't pan out

PR #2244, opened 2024-01-30
example/counter_isomorphic Removed console warning.

martinfrances107:
This warning is seen in the browsers console window.

```
counter_isomorphic.js:1068 At src/counters.rs:138:17, you are reading a resource in `hydrate` mode outside a Suspense or Transition. This can cause hydration mismatch errors and loses out on a significant performance optimization. To fix this issue, you can either:
1. Wrap the place where you read the resource in a Suspense or Transition/ component, or
2. Switch to using create_local_resource(), which will wait to load the resource until the app is hydrated on the client side. (This will have worse performance in most cases.)
```

Two derived signals "value" and "error_msg" need to be wrapped in a Suspense block.

"value" falls back to just the initial text.
"error" uses the default fallback.

### DIFF
--- a/examples/counter_isomorphic/src/counters.rs
+++ b/examples/counter_isomorphic/src/counters.rs
@@ -132,15 +132,6 @@ pub fn Counter() -> impl IntoView {
         |_| get_server_count(),
     );
 
-    let value =
-        move || counter.get().map(|count| count.unwrap_or(0)).unwrap_or(0);
-    let error_msg = move || {
-        counter.get().and_then(|res| match res {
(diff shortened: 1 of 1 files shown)


martinfrances107:
In terms of review... 

I am constantly dipping in and out of the these examples... 

so I want them to be as close to a golden reference as possible 

if you want a different solution.. say involving Transition and create_local_resource 
just let me now and I will put in any extra work needed.

benwis:
Hi Martin! This change looks good, thanks for updating the example

PR #2249, opened 2024-02-01
fix: make directive `.into()` calls consistent

### LINKED ISSUE #2248: inconsistent directive behaviour
**Describe the bug**
The implementation of directives on the macro seems to have somewhat inconsistent behaviour.

The current implementation is this:
```rust
pub(crate) fn directive_call_from_attribute_node(
    attr: &KeyedAttribute,
    directive_name: &str,
) -> TokenStream {
    let handler = format_ident!("{directive_name}", span = attr.key.span());

    let param = if let Some(value) = attr.value() {
        quote! { #value.into() }
    } else {
        quote! { () }
    };

    quote! { .directive(#handler, #param) }
}
```

I'm not sure why `.into()` is called on `#value` - was this a deliberate decision to automatically call `.into()`? This might be a good API choice, so that many different inputs can be used, without the directive author needing to make the parameter `impl Into<MyParameter>`. This should also be documented, something I can add to the book once this is clarified.

However, this is inconsistent as `into` is not called on the no-value alternative, `()`. There may be cases where there is a sensible default if the user doesn't provide any parameter, but right now the directive author can't add a `impl From<()> for MyParameter` to accept it. 

*❗ I'm not sure if this is a breaking change to add `.into()` though.*

Another bit: the first branch should be `(#value).into()` so that `into` is called on the whole value passed in, not just added to the end. For example, this wouldn't work:
```rust
struct Param;
impl<T: Fn() -> i32> From<T> for Param {
    // ...
}

fn directive(el: HtmlElement<AnyElement>, p: Param) { /* ... */ }

// this doesn't work: expands to `|| 10.into()`
view! {
    <div use:directive=|| 10 />
}
// currently have to do this to get `{|| 10}.into()`
view! {
    <div use:directive={|| 10} />
}
```

I think fixing this won't be a breaking change.

I can make a PR to fix these issues, just wanted to check what the intended behaviour was and if this might be a breaking change.

**Leptos Dependencies**

```toml
leptos = { version = "0.6", features = ["csr"] }
```

**To Reproduce**
described above.

**Expected behavior**
also described above.

**Screenshots**
N/A

**Additional context**
N/A

blorbb:
fixes #2248 

added to the directive examples as well.

### DIFF
--- a/examples/directives/src/lib.rs
+++ b/examples/directives/src/lib.rs
@@ -1,5 +1,6 @@
 use leptos::{ev::click, html::AnyElement, *};
 
+// no extra parameter
 pub fn highlight(el: HtmlElement<AnyElement>) {
     let mut highlighted = false;
 
@@ -14,6 +15,7 @@ pub fn highlight(el: HtmlElement<AnyElement>) {
     });
 }
 
+// one extra parameter
 pub fn copy_to_clipboard(el: HtmlElement<AnyElement>, content: &str) {
     let content = content.to_string();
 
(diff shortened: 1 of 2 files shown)


blorbb:
Actually, there might be a way to call `Default` on the parameter if no argument is given to `use:`, instead of requiring `From<()>`.
However, this would mean that users might forget to provide an argument and it unknowingly falls back to default. Could be a problem if there is no sensible default for the directive function.
@gbj do you think this is a good idea, or should we just stick to requiring `From<()>`? Using From instead means that directive authors can't provide both a blanket impl and a 'default', but users could still explicitly use a default with `use:dir=Default::default()`
Using default might be a breaking change? I don't think calling `().into()` is

(Also, not sure why the CI failed above)

benwis:
Can you run cargo clippy and/or cargo fmt on your repo? The tests are failing because of that  @blorbb

blorbb:
> Can you run cargo clippy and/or cargo fmt on your repo? The tests are failing because of that @blorbb

ran both, the only warning I got is about auto traits being unstable from clippy.

benwis:
Nightly clippy is causing pain, no worries

gbj:
Thanks, this looks good to me for now. I agree that `Default` makes more sense than `().into()`. It would be a good breaking change to make in the future, as it seems like a more idiomatic way to handle the optional case (and is in line with what we do with optional component props).

gbj:
(@blorbb You have it marked as draft, I think because of that question — To me this seems ready to merge, unless you're waiting on anything else. Is it ready to go?)

PR #2253, opened 2024-02-02
examples/hackernews: Add a "Suspense" wrapper.

martinfrances107:
This warning appears in the browser's console log.

```
hackernews.js:933 At src/routes/stories.rs:39:17, you are reading a resource in `hydrate` mode outside a <Suspense/> or <Transition/>. This can cause hydration mismatch errors and loses out on a significant performance optimization. To fix this issue, you can either:
1. Wrap the place where you read the resource in a <Suspense/> or <Transition/> component, or
2. Switch to using create_local_resource(), which will wait to load the resource until the app is hydrated on the client side. (This will have worse performance in most cases.)
```

### DIFF
--- a/examples/hackernews/src/routes/stories.rs
+++ b/examples/hackernews/src/routes/stories.rs
@@ -62,16 +62,18 @@ pub fn Stories() -> impl IntoView {
                     }}
                 </span>
                 <span>"page " {page}</span>
-                <span class="page-link"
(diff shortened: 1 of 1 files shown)


martinfrances107:
PS ... I have manually tests this ...

the example still works as expected, and I can confirm the warning is removed.

benwis:
Than,s for noticing and fixing the error!

PR #2256, opened 2024-02-04
docs: fix broken link in leptos_router

nootr:
This commit simply fixes a typo which causes a broken link in the docs.

Thanks for this awesome framework!


gbj:
Thanks!

PR #2257, opened 2024-02-04
feat(axum): provide state to server fn context

chrisp60:
Note that this is a minimal implementation and will __not__ allow the
user to `expect_state` if they have external calls to rendering their
app (i.e. using `render_app_to_*` directly).

Additional work should be taken up to outline these changes through documentation.
Unfortunately, my personal work takes up most of my time right now.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1586,6 +1586,14 @@ where
     where
         IV: IntoView + 'static,
     {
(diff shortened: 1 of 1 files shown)


chrisp60:
oof did I screw something up or has CI just been fussy recently. I can take a deeper look if its needed

gbj:
It's not you, it's `clippy` testing `rkyv` and feeling sad.

This looks great -- thanks!

PR #2268, opened 2024-02-06
fix: add key/value pair from submit button when parsing form event

### LINKED ISSUE #2265: ActionForm does not send name-value pairs from type=submit inputs
(Leptos v0.6)

The following ActionForm doesn't send the `"accept": "true"` key-value pair when clicking "Accept" and JS/WASM is loaded. When JS/WASM is not loaded, yet, it works.

```rust
use leptos::*;
use leptos_router::*;

#[component]
    let handle_consent = create_server_action::<HandleConsent>();

    view! {
        <ActionForm action=handle_consent>
            <button type="submit" name="accept" value="true">Accept</button>
        </ActionForm>
    }
}

#[server(HandleConsent)]
async fn handle_consent(accept: Option<String>) -> Result<(), ServerFnError> {
    println!("{:?}", accept);
    Ok(())
}
```

(Leptos v0.6)

haslersn:
Fixes #2265

### DIFF
--- a/router/src/components/form.rs
+++ b/router/src/components/form.rs
@@ -593,21 +593,38 @@ where
     action_form
 }
 
-fn form_from_event(ev: &SubmitEvent) -> Option<HtmlFormElement> {
-    let submitter = ev.unchecked_ref::<SubmitEvent>().submitter();
-    match &submitter {
+fn form_data_from_event(
+    ev: &SubmitEvent,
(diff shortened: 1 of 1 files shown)


gbj:
Thanks for your work! This looks great. The CI issues are an unrelated issue and I have tested this manually and it works correctly. Thanks very much for the PR. :tada:

PR #2270, opened 2024-02-06
Create an axum-no-default feature for server_fn and sets leptos_axum to rely on it instead.

sjud:
This is required to run an axum leptos server in wasm post 0.6.0. Prior 0.6 server_fn didn't rely on axum and now that it does it needs a no-default version because the default axum is not wasm compatible because it brings in tokio. 

leptos_axum doesn't require server_fn to have axum default features so I set it to rely on the new no-default-feature version of server_fn. 

I don't think server_fn requires axum to have default features at all, so instead of adding a no-default feature flag we could just set no-default feature for axum to be the default feature set of server_fn.

This fixes

https://github.com/leptos-community/deployment-deno/pull/2

### DIFF
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -14,7 +14,7 @@ axum = { version = "0.7", default-features = false, features = [
 futures = "0.3"
 http-body-util = "0.1"
 leptos = { workspace = true, features = ["ssr"] }
-server_fn = { workspace = true, features = ["axum"] }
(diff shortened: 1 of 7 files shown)


gbj:
I was prepared to eat the "okay this is technically a breaking change, but would it really break anyone's Axum build by disabling features? surely not" but then I actually looked at the PR and wow, I think you actually managed to do this without a breaking change. 

You have my respect and appreciation.

CI issues are #2272, will run it locally just to make sure but this all looks fine to me.

Issue #1764, opened 2023-09-21
Typing double slashes in URL causes leptos router error


PianoPrinter:
**Describe the bug**
Typing double slashes in URL causes leptos router error: ``[Leptos Router] Invalid path //``

URL ex) ``mysite.com//``

**Leptos Dependencies**
```toml
leptos = { version = "0.5.0-rc2", features = ["csr"] }
leptos_router = { version = "0.5.0-rc2", features = ["csr"] }
```

**To Reproduce**
1. Create leptos project with above described dependencies.
2. Edit main.rs to this
```rust
use leptos::*;
use leptos_router::*;

fn main() {
    leptos::mount_to_body(|| view! { <App/> })
}

#[component]
fn App() -> impl IntoView {
    view! {
        <Router>
            <Routes>
                <Route path="/" view=|| { "hello" }/>
                <Route path="/test" view=|| { "another one" }/>
                <Route path="*" view=|| { "anything else" }/>
            </Routes>
        </Router>
    }
}
```
3. ``trunk serve --open``
4. type double slashes in URL then enter.
5. See error in console.

**Expected behavior**
URL with double slashes should also match to ``path="*"``

**Screenshots**
![image](https://github.com/leptos-rs/leptos/assets/43927752/3c95c05a-ce41-4634-806e-3246e153f2b7)

**Additional context**
similar case that i found on github:
https://github.com/remix-run/react-router/issues/8428

PianoPrinter:
i just realized this error also happens when you type multiple slashes in a row in URL, not only when double slashes.

mysite.com///
mysite.com////
mysite.com////...

gbj:
This was a pretty fun issue. I'm not sure the fix is perfect but it works for the test cases I have.

PR #2271, opened 2024-02-06
fix: handle cross-origin redirects in server function redirect hook

### LINKED ISSUE #2269: Server actions: redirect to different origin does not work in HTTPS scenario
**Describe the bug**

When accessing a leptos app over HTTPS (via a TLS-terminating reverse proxy), redirects returned by server actions do not work, when the location points to a different origin. Instead of following the redirect, the following error is printed to the JS console:

```
Uncaught Error: `unwrap_throw` failed ([...]/leptos_router-0.6.5/src/history/mod.rs:124:18)
```

This seems to have to do with same origin policy.

In my limited experience, this error only happens when a `<Router>` component is involved.

**Leptos Dependencies**

```toml
```

**To Reproduce**

1. Run the following leptos component:
    ```rust
    use leptos::*;
    use leptos_meta::*;
    use leptos_router::*;
    
    #[component]
        provide_meta_context();
    
        view! {
            <Stylesheet id="leptos" href="/pkg/leptos-playground.css"/>
            <Title text="Leptos Playground"/>
    
            <Router>
                <Routes>
                    <Route path="" view=MyForm/>
                </Routes>
            </Router>
        }
    }
    
    #[component]
    fn MyForm() -> impl IntoView {
        let redirect = create_server_action::<Redirect>();
    
        view! {
            <ActionForm action=redirect>
                <button type="submit">Redirect</button>
            </ActionForm>
        }
    }
    
    #[server(Redirect)]
    async fn redirect() -> Result<(), ServerFnError> {
        leptos_axum::redirect("https://example.com");
    
        Ok(())
    }
    ````
2. Set up a TLS-terminating reverse proxy
3. Access the app over HTTPS
4. Click "Redirect"

When WASM/JS is enabled and loaded, the server function is called via the `fetch()` API and the redirect is not followed. Instead the following error is printed in the JS console:

```
Uncaught Error: `unwrap_throw` failed ([...]/leptos_router-0.6.5/src/history/mod.rs:124:18)
```

Leptos tried to call the `pushState()` API with url `https://example.com` and that returned a `SecurityError` which got unwrapped using `unwrap_throw()`.

**Additional context**

The `<Router>` component seems to be necessary to reproduce the issue. Maybe, without it, leptos doesn't manage a Histroy object.

haslersn:
Fixes #2269

### DIFF
--- a/router/src/components/router.rs
+++ b/router/src/components/router.rs
@@ -1,7 +1,7 @@
 use crate::{
-    create_location, matching::resolve_path, scroll_to_el, use_location,
-    use_navigate, Branch, History, Location, LocationChange, RouteContext,
-    RouterIntegrationContext, State,
+    create_location, matching::resolve_path, resolve_redirect_url,
+    scroll_to_el, use_location, use_navigate, Branch, History, Location,
+    LocationChange, RouteContext, RouterIntegrationContext, State,
 };
 #[cfg(not(feature = "ssr"))]
 use crate::{unescape, Url};
@@ -24,6 +24,7 @@ use std::{
 use thiserror::Error;
 #[cfg(not(feature = "ssr"))]
 use wasm_bindgen::JsCast;
+use wasm_bindgen::UnwrapThrowExt;
 
 static GLOBAL_ROUTERS_COUNT: AtomicUsize = AtomicUsize::new(0);
 
@@ -56,15 +57,24 @@ pub fn Router(
     // set server function redirect hook
     let navigate = use_navigate();
     let navigate = SendWrapper::new(navigate);
-    let router_hook = Box::new(move |path: &str| {
-        let path = path.to_string();
-        // delay by a tick here, so that the Action updates *before* the redirect
-        request_animation_frame({
+    let router_hook = Box::new(move |loc: &str| {
+        let Some(url) = resolve_redirect_url(loc) else {
+            return; // resolve_redirect_url() already logs an error
+        };
+        let current_origin =
+            leptos_dom::helpers::location().origin().unwrap_throw();
+        if url.origin == current_origin {
             let navigate = navigate.clone();
-            move || {
-                navigate(&path, Default::default());
-            }
-        });
+            // delay by a tick here, so that the Action updates *before* the redirect
+            request_animation_frame(move || {
+                navigate(&url.pathname, Default::default());
+            });
+            // Use set_href() if the conditions for client-side navigation were not satisfied
+        } else if let Err(e) =
+            leptos_dom::helpers::location().set_href(&url.href())
+        {
+            leptos::logging::error!("Failed to redirect: {e:#?}");
+        }
     }) as RedirectHook;
     _ = server_fn::redirect::set_redirect_hook(router_hook);
 
--- a/router/src/hooks.rs
+++ b/router/src/hooks.rs
@@ -1,10 +1,10 @@
 use crate::{
     Location, NavigateOptions, Params, ParamsError, ParamsMap, RouteContext,
-    RouterContext,
+    RouterContext, Url,
 };
 use leptos::{
-    create_memo, request_animation_frame, signal_prelude::*, use_context, Memo,
-    Oco,
+    create_memo, request_animation_frame, signal_prelude::*, use_context,
+    window, Memo, Oco,
 };
 use std::{rc::Rc, str::FromStr};
 
@@ -215,4 +215,26 @@ pub fn use_navigate() -> impl Fn(&str, NavigateOptions) + Clone {
 pub(crate) fn use_is_back_navigation() -> ReadSignal<bool> {
     let router = use_router();
     router.inner.is_back.read_only()
+}
+
+/// Resolves a redirect location to an (absolute) URL.
+pub(crate) fn resolve_redirect_url(loc: &str) -> Option<Url> {
+    let origin = match window().location().origin() {
+        Ok(origin) => origin,
+        Err(e) => {
+            leptos::logging::error!("Failed to get origin: {:#?}", e);
+            return None;
+        }
+    };
+
+    // TODO: Use server function's URL as base instead.
+    let base = origin;
+
+    match Url::new_with_base(loc, &base) {
+        Ok(url) => Some(url),
+        Err(e) => {
+            leptos::logging::error!("Invalid redirect location: {}", e);
+            None
+        }
+    }
 }
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -692,12 +692,12 @@ fn provide_contexts(req: &HttpRequest, res_options: ResponseOptions) {
 }
 
 fn leptos_corrected_path(req: &HttpRequest) -> String {
-    let path = req.path();
+    let path = req.path().to_string();
     let query = req.query_string();
     if query.is_empty() {
-        "http://leptos".to_string() + path
+        path
     } else {
-        "http://leptos".to_string() + path + "?" + query
+        path + "?" + query
     }
 }
 #[tracing::instrument(level = "trace", fields(error), skip_all)]
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -659,12 +659,11 @@ where
                     // Need to get the path and query string of the Request
                     // For reasons that escape me, if the incoming URI protocol is https, it provides the absolute URI
                     // if http, it returns a relative path. Adding .path() seems to make it explicitly return the relative uri
-                    let path = req.uri().path_and_query().unwrap().as_str();
+                    let path = req.uri().path_and_query().unwrap().to_string();
 
-                    let full_path = format!("http://leptos.dev{path}");
                     let (_, req_parts) = generate_request_and_parts(req);
                     move || {
-                        provide_contexts(full_path, req_parts, default_res_options);
+                        provide_contexts(path, req_parts, default_res_options);
                         app_fn().into_view()
                     }
                 };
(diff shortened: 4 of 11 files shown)


router/src/components/router.rs
```diff
@@ -57,14 +57,29 @@ pub fn Router(
...
-        request_animation_frame({
-            let navigate = navigate.clone();
-            move || {
-                navigate(&path, Default::default());
+        match Url::try_from(path) {
+            Ok(url) => {
+                let current_origin = leptos_dom::helpers::location()
+                    .origin()
+                    .unwrap_or_default();
+                if url.origin == current_origin {
```
gbj, line 65: This check fails for local URLs like `/foobar` because in the browser, `Url::try_from` inserts a fake `http://leptos` base:

https://github.com/leptos-rs/leptos/blob/a903e19eb24371e59b6c36dfdddc2ca1940866b0/router/src/history/url.rs#L47-L57

This is simply because the `new URL()` constructor will fail with an incomplete URL
```js
new URL("/foobar");
// throws `Uncaught TypeError: URL constructor: /foobar is not a valid URL.`
```

I *think* it's safe to add a `url.origin = current_origin || url.origin == "http://leptos"`, maybe refactoring these two places to share a constant, or adding a helper method to `Url` like `is_app_origin`.

gbj (changes requested):
Thanks very much! This is the right way to do it, but see the note about client-side redirects to the same origin.

router/src/components/router.rs
haslersn: Why does this have a double slash here?
https://github.com/leptos-rs/leptos/blob/a903e19eb24371e59b6c36dfdddc2ca1940866b0/router/src/history/url.rs#L48

haslersn: What about the server? Would it also fail there (but doesn't, because that code isn't run on the server)?

The client-side code for `Url::try_from()` doesn't seem quite correct: I think we should apply the rules of https://datatracker.ietf.org/doc/html/rfc3986#section-4, i.e., treat the input as an URI-reference and, if it's a relative reference, resolve it against a base URI (probably the current window location). Eexamples in section 5.4.1.

gbj: The only place that 

> Why does this have a double slash here?
> 
> https://github.com/leptos-rs/leptos/blob/a903e19eb24371e59b6c36dfdddc2ca1940866b0/router/src/history/url.rs#L48

#1764 describes an issue in which using a route like `http://my-app.com///foo` causes an error, instead of matching the path `//foo`. Using something like `new URL("//foo", window.location.origin)` causes the browser to misinterpret this as the URL `"https://foo/"`. So in that case, we manually build the whole URL from `window.location.origin`. In other cases, we just use `new URL()` with a fake base.

If you fire up the browser console it might make more sense: observe the difference in behavior between these two:
```js
// this is fine
new URL("/foo", window.location.origin);
// this is not
new URL("//foo", window.location.origin);
```

Hence the check that's currently in place.

It would be possible to change this by always getting the `window.location.origin` and doing string concatenation, but note you'll need to check for a `/` at the beginning, as the following set also do not behavior properly
```js
// good
new URL(window.location.origin + "/foo")
// good
new URL(window.location.origin + "//foo")
// bad
new URL(window.location.origin + "foo")
```

> The client-side code for `Url::try_from()` doesn't seem quite correct: I think we should apply the rules of https://datatracker.ietf.org/doc/html/rfc3986#section-4, i.e., treat the input as an URI-reference and resolve it against a base URI (probably the current window location). Examples in section 5.4.1.

I might be misunderstanding you, but as far as I can tell neither the `url` crate nor the `new URL()` constructor in the browser do this kind of resolution; they simply require complete URLs, hence the workaround. I don't think it makes sense to implement the kind of resolution described in 5.4.1 in the context of our router.

haslersn: > Using something like `new URL("//foo", window.location.origin)` causes the browser to misinterpret this as the URL `"https://foo/"`.

Why "misinterpret"?  According to the RFC, `//foo` is a network-path reference and it's correct behavior to resolve it as `http://foo` or `https://foo` (depending on whether the current origin uses `http` or `https`).

Try the following with JS/WASM disabled. The browser correctly links / redirects to `http://example.com`.

```rust
use leptos::*;
use leptos_meta::*;
use leptos_router::*;

#[component]
    provide_meta_context();

    let redirect = create_server_action::<Redirect>();
    
    view! {
        <Stylesheet id="leptos" href="/pkg/leptos-playground.css"/>
        <Title text="Leptos Playground"/>

        <a href="//example.com">Click me</a>
        <ActionForm action=redirect>
            <button type="submit">Redirect</button>
        </ActionForm>
    }
}

#[server(Redirect)]
async fn redirect() -> Result<(), ServerFnError> {
    leptos_axum::redirect("//example.com");

    Ok(())
}
```

haslersn: > as far as I can tell neither the `url` crate nor the `new URL()` constructor in the browser do this kind of resolution

The `new URL()` constructor, when passing a base URI as 2nd argument, seems to do the resolution specified by https://datatracker.ietf.org/doc/html/rfc3986#section-4

haslersn:
I just updated this PR. It now consists of two separate commits:

1. feat: provide two different Url constructors: with and without base
    
    The URL constructor without base (`Url::try_from`) requires an absolute URL, so we need to enforce this for all callers. The only caller which called this with a relative path was `create_location()`, so I changed `create_location` to always prepend `http://leptos.dev` to the path. The origin in this case doesn't matter, because `create_location()` doesn't use the origin.
    
    However, there was an inconsistency between client and server: previosly, `create_location()` was called with a relative path on the client and an absolute URL (with origin `http://leptos.rs` or `http://leptos.dev`) on the server. I fixed this by always using a relative path on the server, too.

2. fix: handle cross-origin redirects in server function redirect hook
    
    In client-side navigation we now handle redirects returned from server functions the same way how the browser would handle them. This fixes cross-origin redirects.
    
    Note: this is a breaking change for relative locations such as `Location: foo`. Previously that was resolved to `<origin>/foo`. Now it's resolved as a URI-reference (RFC 3986) relative to the server function URL's (typically `<origin>/api/something`), resulting in `<origin>/api/foo`. In case you require the old behavior, make sure that relative locations start with a single slash.

server_fn/src/lib.rs
```diff
@@ -328,7 +329,23 @@ where
             // if redirected, call the redirect hook (if that's been set)
             if let Some(redirect_hook) = redirect_hook {
                 if (300..=399).contains(&status) || has_redirect_header {
-                    redirect_hook(&location);
+                    let origin = web_sys::window()
```
gbj, line 332: The server function client is not guaranteed to have access to browser APIs, because server functions can also be used in desktop applications via `reqwest`, so this logic need to live somewhere in `leptos_router`

gbj (commented):
**Specific comments:**
1. This currently panics if you try to load a page using Actix. The PR updates the logic for handling paths from `ServerIntegration` in the router, and updates the Axum integration, but not the Actix integration.
2. This would panic on a redirect in the server fn client if the client is not running in a web browser (see specific comment there).

**General comment:** I am hesitant to introduce these sorts of breaking changes in order to address a relatively small bug. Not all of the server integrations live in this repository any more (e.g., viz and spin integrations), and changing the behavior of the router in a way that introduces panics in code that still compiles would be a serious issue. (This is the case for the Actix integration, and would be for the others even if the Actix one were fixed in this PR.) 

I think your approach is probably *correct*, for the long term, but it would either need to be new minor version (i.e., from 0.6 to 0.7) to prevent these runtime issues, or it would need to accommodate both full URLs and paths only — maybe that's another possible route.

However, I do wonder whether a much more minimal approach would be sufficient to fix the bug, with a more-correct-but-breaking fix like this PR included in a future release.

For example, simply changing `router_hook` to the below seems to me to yield correct behavior for both local client-side redirects and other origins:
```rust
    let router_hook = Box::new(move |path: &str| {
        let path = path.to_string();
        let Ok(origin) = window().location().origin() else {
            return;
        };
        let Ok(to_url) = web_sys::Url::new_with_base(&path, &origin) else {
            logging::error!("could not parse url {path:?}");
            return;
        };
        let to_origin = to_url.origin();
        // delay by a tick here, so that the Action updates *before* the redirect
        if to_origin != origin {
            _ = window().location().set_href(&path);
        } else {
            request_animation_frame({
                let navigate = navigate.clone();
                move || {
                    navigate(&path, Default::default());
                }
            });
        }
    }) as RedirectHook;
```

haslersn:
I pushed a change to the 1st commit that makes it backwards-compatible by also allowing an URL instead of a path in `create_location()`. In case an URL is provided, a warning is logged.

However, for the 2nd commit, in order to move the logic into `leptos_router`, the redirect hook would need to know which server function was called (or at least the server function's path), because the redirect hook needs use the server function's path as a base. This requires changing the signature of the redirect hook – what do you think about this?

haslersn:
Now I also changed the 2nd commit, moving the resolving logic into `leptos_router`. Currently it uses the window's origin as a base for resolving the redirect location. This is still incorrect (as the server function's URL should be used as a base instead). But at least cross-origin redirects are now fixed.

@gbj please review again.

I left a TODO for using the correct base. This probably requires a breaking change in the signature of the redirect hook.

router/src/hooks.rs
```diff
@@ -216,3 +216,25 @@ pub(crate) fn use_is_back_navigation() -> ReadSignal<bool> {
     let router = use_router();
     router.inner.is_back.read_only()
 }
+
+/// Resolves a redirect location to an (absolute) URL.
+pub(crate) fn resolve_redirect_url(loc: &str) -> Option<Url> {
+    let origin = match window().location().origin() {
```
haslersn, line 222: Would it be ok to `unwrap()` or `unwrap_throw()` here? Is the `.origin()` call always expected to succeed?

gbj:
Doesn't the **General Comment** section of what I wrote above still apply? i.e., that changing the expected URL that's provided by the server integration would break the server integrations that live outside this repo (viz, spin)?

The **Suggested Solution** I posted above fixes issue #2269 in a non-breaking way. It seems this PR is trying to address that and the other issues you've identified at the same time, but in a breaking way.

I would propose opening separate issues for those separate issues, so that we can fix the specific issue of #2269 now and address those separately.

If you would prefer, we can keep this PR open, and I can fix #2269 separately.

haslersn:
I created a smaller PR #2329 to fix #2269, and I will create one or more separate PRs for the remaining issues as soon as #2329 is merged.

haslersn:
I created a smaller PR #2329 to fix #2269, and I will create one or more separate PRs for the remaining issues as soon as #2329 is merged.

PR #2280, opened 2024-02-08
pub export server is either from leptos_macro or leptos_spin_macro depending on if spin feature is enabled.

sjud:
This request add leptos_spin_macro and pub exports the server macro defined there if spin is a feature, it adds leptos_spin_macro as a dependency (based on benwis fork of it) but for this to work he needs to accept my PR there.


benwis:
I like the PR, but I don't want to introduce a dependency on my fork of leptos-spin, so I think we can wait for them to release before we do this

sjud:
> I like the PR, but I don't want to introduce a dependency on my fork of leptos-spin, so I think we can wait for them to release before we do this

Sounds good to me!

PR #2283, opened 2024-02-09
feat: use cdn url for static assets (#2281)

### LINKED ISSUE #2281: How to Use CDN Url for Static Assets When Deploying with SSR?
When deploying using SSR, I would like to specify an absolute path for all static client-side assets, including the initial client-side java-script file and wasm file. This will allow using a CDN for static assets. Currently, I have not found a way to alter the relative paths for static assets; a preceding slash `/` is always added. I'm using the [start-aws template](https://github.com/leptos-rs/start-aws). Is there a way to accomplish this?

zoomiti:
Closes #2281

### DIFF
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -55,7 +55,9 @@ pub fn html_parts_separated(
     options: &LeptosOptions,
     meta: Option<&MetaContext>,
 ) -> (String, &'static str) {
-    let pkg_path = &options.site_pkg_dir;
+    let pkg_path = option_env!("CDN_PKG_PATH")
+        .map(String::from)
+        .unwrap_or(format!("/{}", options.site_pkg_dir));
     let output_name = &options.output_name;
     let nonce = use_nonce();
     let nonce = nonce
@@ -107,8 +109,8 @@ pub fn html_parts_separated(
                     <meta charset="utf-8"/>
                     <meta name="viewport" content="width=device-width, initial-scale=1"/>
                     {head}
-                    <link rel="modulepreload" href="/{pkg_path}/{output_name}.js"{nonce}>
-                    <link rel="preload" href="/{pkg_path}/{wasm_output_name}.wasm" as="fetch" type="application/wasm" crossorigin=""{nonce}>
+                    <link rel="modulepreload" href="{pkg_path}/{output_name}.js"{nonce}>
+                    <link rel="preload" href="{pkg_path}/{wasm_output_name}.wasm" as="fetch" type="application/wasm" crossorigin=""{nonce}>
                     <script type="module"{nonce}>
                         function idle(c) {{
                             if ("requestIdleCallback" in window) {{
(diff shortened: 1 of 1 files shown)


zoomiti:
me when i forgor to run cargo fmt

integrations/utils/src/lib.rs
```diff
@@ -55,7 +55,9 @@ pub fn html_parts_separated(
     options: &LeptosOptions,
     meta: Option<&MetaContext>,
 ) -> (String, &'static str) {
-    let pkg_path = &options.site_pkg_dir;
+    let pkg_path = option_env!("CDN_PKG_PATH")
+        .map(String::from)
+        .unwrap_or(format!("/{}", options.site_pkg_dir));
```
gbj, line 60: Two smallish but important things here:
1. `option_env!()` returns `Option<&'static str>`, you can make these a `Cow<'static, str>` and avoid the allocation in the CDN case
2. `.unwrap_or_else(|| format!` is better here because `.unwrap_or` always constructs its argument

So in other words:
```rust
    let pkg_path = option_env!("CDN_PKG_PATH")
        .map(Cow::from)
        .unwrap_or_else(|| format!("/{}", options.site_pkg_dir).into());
```

gbj:
Thanks, this looks great!

@zoomiti I recommended a small change in the comment above. The CI issues here are unrelated and I need to fix things up on main.

@CodingFlow could you let us know if this solves it for your use case? The approach here would be something like
```sh
CDN_PKG_PATH=https://my-cool-cdn.com/my-leptos-app/pkg cargo leptos build
```
i.e., including the `CDN_PKG_PATH` at compile time.

integrations/utils/src/lib.rs
zoomiti: I always forget `Cow`. I stared at my solution at first knowing I missed something haha.
And then I figured since I'm already making a needless allocation, doing another unconditionally wouldn't matter. But you're right, if we can avoid it trivially, why not?

zoomiti:
Wow it looks like CI is so big the container ran out of diskspace, at least for `server_fn`
Not sure what happened for select_stable

CodingFlow:
> Thanks, this looks great!
> 
> @zoomiti I recommended a small change in the comment above. The CI issues here are unrelated and I need to fix things up on main.
> 
> @CodingFlow could you let us know if this solves it for your use case? The approach here would be something like
> 
> ```shell
> CDN_PKG_PATH=https://my-cool-cdn.com/my-leptos-app/pkg cargo leptos build
> ```
> 
> i.e., including the `CDN_PKG_PATH` at compile time.

Yes, that looks it should do it for me. 👍

gbj:
Thanks very much!

rakshith-ravi:
Okay. Small regression. What happens when my leptos app is not hosted at /? Let's say it's hosted at /ui?

The current code assumes / as the root regardless of what's given in the cargo.toml. Isn't this something we should fix?

zoomiti:
The current code doesn't change any routing. All it does is make it so that static assets can be hosted on a CDN instead of on disk.

gbj:
@rakshith-ravi I don't think this changes anything — before the PR, the integration assumed the pkg directory is relative to `/`, always. Now, the integration assumes the pkg directory is relative to `/`, unless you specify a build-time env variable `CDN_PKG_PATH`. 

I might be wrong, in which case if you open an issue with a reproduction of problem I'll look at it.

rakshith-ravi:
> before the PR, the integration assumed the pkg directory is relative to `/`, always

Ahh, got it. Then we shouldn't have an issue

PR #2284, opened 2024-02-09
fix: hydration ID mismatch with nested Suspense

gbj:
@nicoburniske This does seem to fix the nested Suspense issue in the reproduction case you shared with me (I think) but it also seems like kind of a stupid solution to me, so I'd appreciate it if you could help test it in your initial use case.


benwis:
@gbj What do you think about this?

gbj:
I think this one's just broken for other cases, will close for now.

PR #2290, opened 2024-02-10
Example: server-fns without cargo-leptos/ssr

crajcan:
I made this before I noticed the [todo_app_sqlite_csr](https://github.com/leptos-rs/leptos/tree/main/examples/todo_app_sqlite_csr) example had been added.

It is a bit different in that it shows how to add server_fns to an existing Trunk app without introducing cargo-leptos or the `ssr`/`hydrate` features.

If this is too redundant, feel free to close, nbd.


gbj:
Thanks for sharing this. I wonder if we should start collecting interesting examples of how to do things somewhere else, as the examples directory in the repo is starting to become a little overwhelming. Adding more examples also slowly adds to maintenance burden as I end up updating them long term.

To me, this seems really useful as a guide for how to add server fns to an existing Trunk project, but if you know you want CSR with server functions from the beginning the `todo_app_sqlite_csr` approach is slightly better (just with the live-reloading/coordinated rebuild, etc.,) so I'd probably recommend that for new projects.

I've thought of splitting `examples` into something like `examples/basic` and `examples/projects`, and grouping the more complex examples under the latter.

gbj:
@crajcan Sorry for my slow response here. Thanks for your contribution. We have created a new `projects` directory to house a wider variety of projects without the same maintenance burden as examples — Would you be interested in updating your PR to add this to `projects` instead?

crajcan:
> @crajcan Sorry for my slow response here. Thanks for your contribution. We have created a new `projects` directory to house a wider variety of projects without the same maintenance burden as examples — Would you be interested in updating your PR to add this to `projects` instead?

Hey no problem. I think I actually left you hanging. Yes, I would be happy to do that. I'll request your review when it's ready.

PR #2298, opened 2024-02-13
Error type detection only worked without type path

maccesch:
In most cases when you return `Result<..., ServerFnError<E>>` this worked but when you tried `Result<..., leptos::ServerFnError<E>>` then it didn't.


gbj:
Thanks! As the change suggests, clearly I intended this and then missed that reference :smile: Good catch.

maccesch:
Happened to me various times. Macro code is very finicky to write.

I think the 2 failing checks are not because of my code.

gbj:
Thanks!

PR #2305, opened 2024-02-14
meilisearch_searchbar example

sjud:
This example shows how to use a meilisearch server to provide a good search experience out of the box and with only a couple of basic parts.


sjud:
Nobody fix this for me, I can do it!!

sjud:
ok i give up

gbj:
(Silly question but does it make a difference if you rename `makefile.toml` to `Makefile.toml`?)

I just left a similar comment on #2290: I wonder if we should start collecting interesting examples of how to do things somewhere else, as the examples directory in the repo is starting to become a little overwhelming. Adding more examples also slowly adds to maintenance burden as I end up updating them long term.

I've thought of splitting `examples` into something like `examples/basic` and `examples/projects`, and grouping the more complex examples under the latter. Any thoughts?

sjud:
> (Silly question but does it make a difference if you rename `makefile.toml` to `Makefile.toml`?)
> 
> I just left a similar comment on #2290: I wonder if we should start collecting interesting examples of how to do things somewhere else, as the examples directory in the repo is starting to become a little overwhelming. Adding more examples also slowly adds to maintenance burden as I end up updating them long term.
> 
> I've thought of splitting `examples` into something like `examples/basic` and `examples/projects`, and grouping the more complex examples under the latter. Any thoughts?

I like that idea.
examples/basic (or core) could demonstrate core functionality of the library and be something that the maintainers or broader team is willing to keep up to date with the quick velocity of the project, it would function as an extension of the docs/book.

And example/projects (or extra) could be an extension of awesome Leptos where people show off integrations, quirky little things, neat usages of Leptos. There was recently a Leptos camera that was posted to showcase which I think would be a great fit. And then this directory would not necessarily be up to date but still be useful to show how certain things could be done. 

And awesome Leptos could be real world use cases, actual open source projects etc.

I think right now there's a gap between showing off  big projects and showing off core functionality. @diversable had floated a leptos magazine type of blog but having this extra examples directory could provide source material and a good launching point for a more expansive look into all the neat little usecases/integrations for Leptos.

Let me know what you think about my thoughts about your thoughts, thanks!! :)

gbj:
Superseded by #2500.

PR #2307, opened 2024-02-15
Added missing generics to convert into server_fn Error

maccesch:
Without this `IntoView` is not implemented for `Result<View, ServerFnError<String>>` for example.

This is related to https://discord.com/channels/1031524867910148188/1207174766289748048


maccesch:
Hm this gives conflicting trait implementations. So closing this for now. But this should be resolved somehow, right?

Issue #2309, opened 2024-02-15
session_auth_axum example doesn't build


drebbe-intrepid:
**Describe the bug**
session_auth_axum example doesn't build with nightly on windows

**Leptos Dependencies**

```toml
[dependencies]
anyhow = "1.0"
console_log = "1.0"
rand = { version = "0.8", features = ["min_const_gen"], optional = true }
console_error_panic_hook = "0.1"
futures = "0.3"
leptos = { path = "../../leptos", features = ["nightly"] }
leptos_meta = { path = "../../meta", features = ["nightly"] }
leptos_axum = { path = "../../integrations/axum", optional = true }
leptos_router = { path = "../../router", features = ["nightly"] }
log = "0.4"
simple_logger = "4.0"
serde = { version = "1.0", features = ["derive"] }
axum = { version = "0.7", optional = true, features = ["macros"] }
tower = { version = "0.4", optional = true }
tower-http = { version = "0.5", features = ["fs"], optional = true }
tokio = { version = "1", features = ["full"], optional = true }
http = { version = "1.0" }
sqlx = { version = "0.7.2", features = [
	"runtime-tokio-rustls",
	"sqlite",
], optional = true }
thiserror = "1.0"
wasm-bindgen = "0.2"
axum_session_auth = { version = "0.10", features = [
	"sqlite-rustls",
], optional = true }
axum_session = { version = "0.10", features = [
	"sqlite-rustls",
], optional = true }
bcrypt = { version = "0.15", optional = true }
async-trait = { version = "0.1", optional = true }
```

**To Reproduce**
1. git clone git@github.com:leptos-rs/leptos.git
2. cd leptos/examples/session_auth_axum
3. cargo leptos build
4. multiple errors

**Expected behavior**
session_auth_axum to build

**Additional context**
```powershell
> cargo leptos watch
info: syncing channel updates for 'nightly-2024-01-29-x86_64-pc-windows-msvc'
info: latest update on 2024-01-29, rust version 1.77.0-nightly (635124704 2024-01-27)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
[...]

PR #2310, opened 2024-02-15
session_auth_axum compiles and works again

drebbe-intrepid:
Fix for https://github.com/leptos-rs/leptos/issues/2309


gbj:
Looks like they yanked all previous versions of the crate. Not sure why this would be done, but thanks for the update.

PR #2311, opened 2024-02-15
fix(meta): provide meta context missing in example

EstebanBorai:
Otherwise user gets:

```
use_head() is being called without a MetaContext being provided. We'll automatically create and provide one, but if this is being called in a child route it may cause bugs. To be safe, you should provide_meta_context() somewhere in the root of the app.
```


gbj:
Thanks!

PR #2318, opened 2024-02-16
Fix Broken Doc links and Deprecate `FromUtf8Error` in `oco.rs`

### LINKED ISSUE #859: Broken Doc Links across Crates
There are many instances across the documentation where the `[member]` syntax is used to create a link to a member but, since the member in question is from a different crate, the link is not generated. Here's an example from the [leptos_router](https://docs.rs/leptos_router/latest/leptos_router/fn.Route.html) crate:

> This can be any function that takes a [Scope] and returns a type that implements [IntoView] ...

This also applies to the type annotations in the documentation automatically generated by the `component` macro for a component's properties. The macro issue is more difficult to resolve, but the issue with the handwritten doc comments is trivial.

zoomiti:
Potentially fixes #859

The relative paths for some doc links comes inspired from what tokio does for its workspace links. See [here](https://docs.rs/tokio-macros/1.6.0/src/tokio_macros/lib.rs.html#28) for an example.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -203,7 +203,7 @@ pub fn handle_server_fns() -> Route {
 /// context, allowing you to pass in info about the route or user from Actix, or other info.
 ///
 /// **NOTE**: If your server functions expect a context, make sure to provide it both in
-/// [`handle_server_fns_with_context`] **and** in [`leptos_routes_with_context`] (or whatever
+/// [`handle_server_fns_with_context`] **and** in [`LeptosRoutes::leptos_routes_with_context`] (or whatever
 /// rendering method you are using). During SSR, server functions are called by the rendering
 /// method, while subsequent calls from the client are handled by the server function handler.
 /// The same context needs to be provided to both handlers.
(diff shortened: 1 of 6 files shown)


benwis:
Can you run cargo fmt on your PR?

benwis:
Thanks for looking into this!

gbj:
I'm removing the deprecation with #2362. Adding deprecation warnings can be tricky, since adding them in a patch release can cause previously-passing CI to fail if it is set up to deny based on clippy warnings. This was happening in our own repo after this change; I would not have expected it to be a problem, but clippy was warning about the  use of it in `leptos_reactive` itself when run in `leptos_dom` etc. You can see the warning showing up in CI for recent PRs.

```
warning: use of deprecated enum `oco::FromUtf8Error`: This error was intended for a `TryFrom` implementation for `Oco`. `Oco` will likely move to its own crate after 0.7 so this type will be removed in the future
   --> leptos_reactive/src/oco.rs:450:10
    |
450 | pub enum FromUtf8Error {
    |          ^^^^^^^^^^^^^
```

Issue #2327, opened 2024-02-17
tailwind_csr example css is not loaded.


itsDrac:
**Describe the bug**
The tailwind_csr example works but the css file was not loading due to incorrect path to output.css

**Leptos Dependencies**

N/A (Works fine with existing dependency)

**To Reproduce**
1. Go to 'example/tailwind_csr'.
2. run the example.
3. verify the css is not working as expected.

**Expected behavior**
css file should load up and tailwind class should work as expected.

PR #2328, opened 2024-02-17
Fixed href path in tailwind_csr example

itsDrac:
Fix for #2327 
Updated `href` path in `Stylesheet` tag to `output.css`


gbj:
Thanks!

PR #2329, opened 2024-02-17
fix: handle cross-origin redirects in server function redirect hook

### LINKED ISSUE #2269: Server actions: redirect to different origin does not work in HTTPS scenario

haslersn:
Fixes #2269

In client-side navigation we now handle redirects returned from server functions by resolving the location against the current origin as a base. The base is only relevant if the location doesn't already include an origin. This fixes cross-origin redirects.

Note: in order to handle redirects in the same way as the browser would handle them, we need to use the server function's URL (typically `<origin>/api/something`) as a base. I leave this as a TODO for a future leptos version, because it probably requires changing the signature of the `server_fn` redirect hook.

In order to not be affected by a future breaking change, users should already start making sure that their redirect locations either include an origin or at least start with a single slash (e.g. `Location: /foo`).

### DIFF
--- a/router/src/components/form.rs
+++ b/router/src/components/form.rs
@@ -1,6 +1,6 @@
 use crate::{
-    hooks::has_router, use_navigate, use_resolved_path, NavigateOptions,
-    ToHref, Url,
+    hooks::has_router, resolve_redirect_url, use_navigate, use_resolved_path,
+    NavigateOptions, ToHref, Url,
 };
 use leptos::{
     html::form,
@@ -447,8 +447,10 @@ where
 {
     let has_router = has_router();
     if !has_router {
-        _ = server_fn::redirect::set_redirect_hook(|path: &str| {
-            _ = window().location().set_href(path);
+        _ = server_fn::redirect::set_redirect_hook(|loc: &str| {
+            if let Some(url) = resolve_redirect_url(loc) {
+                _ = window().location().set_href(&url.href());
+            }
         });
     }
     let action_url = if let Some(url) = action.url() {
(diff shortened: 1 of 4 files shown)


gbj:
Okay, this looks good — Thank you!

PR #2331, opened 2024-02-17
Make it so that cancelled timeouts don't leak

### LINKED ISSUE #2330: Cancelling a timeout or animation frame request leaks the associated closure
**Describe the bug**
`request_animation_frame_with_handle` and `set_timeout_with_handle` convert the Rust closure into a Javascript one with `Closure::once_into_js`.  This function cleans up the Rust side once it's called, but if it never is then it is [documented to leak](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/closure/struct.Closure.html#method.once_into_js).

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.6.5", features = ["csr"] }
```

**To Reproduce**
1. Create a timeout or a request for animation frame which closes over something whose Drop impl is observable e.g., by logging.
2. Cancel that timeout or animation frame.
3. Don't observe the drop

**Expected behavior**
Cancelled timeouts to drop their closures.

rjmac:
Instead of using `Closure::once_into_js`, this uses `into_js_value`, which uses weak references to clean up the closure when Javascript no longer has need of it.

It would be nice to make this (and the similar interval function) drop the callback promptly when cancelled, but I don't think that's possible while keeping the handles Copy.

Fixes #2330


gbj:
Nice — makes sense and thanks!

PR #2334, opened 2024-02-17
feat: add `impl_into` argument to `#[server]` proc_macro

videobitva:
This PR adds an optional `impl_into` named  argument for `server_fn` proc macro.
This argument defaults to `true` and enables (or disables) the impl of `From` trait for server function argument type.

In cases when the server function has more than one argument, the `impl_into` has no effect on the trait impl generation, as it has been already disabled by function arguments number check.

The related issue has been discussed here: https://discord.com/channels/1031524867910148188/1203001939089035295

TL;DR: orphan rules disable (in some cases) the implementation of `From` trait for conversion between the only server function argument and server function argument type. 

The minimal example of the issue lives here: https://github.com/videobitva/test_server_fn

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,22 +25,22 @@ members = [
 exclude = ["benchmarks", "examples"]
 
 [workspace.package]
-version = "0.6.1"
+version = "0.6.2"
 
 [workspace.dependencies]
-leptos = { path = "./leptos", version = "0.6.1" }
-leptos_dom = { path = "./leptos_dom", version = "0.6.1" }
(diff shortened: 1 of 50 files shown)


benwis:
You might need to recreate your PR against the most recent version, you're pulling in a lot of old/extra PRs

videobitva:
> You might need to recreate your PR against the most recent version, you're pulling in a lot of old/extra PRs

Ok, will do that then, thank you!

videobitva:
PR without old/extra PRs is now here: https://github.com/leptos-rs/leptos/pull/2335

PR #2335, opened 2024-02-17
feat: add `impl_from` argument to `#[server]` proc_macro

videobitva:
This PR adds an optional `impl_from` named  argument for `server_fn` proc macro.
This argument defaults to `true` and enables (or disables) the impl of `From` trait for server function argument type.

In cases when the server function has more than one argument, the `impl_from` has no effect on the trait impl generation, as it has been already disabled by function arguments number check.

The related issue has been discussed here: https://discord.com/channels/1031524867910148188/1203001939089035295

TL;DR: orphan rules disable (in some cases) the implementation of `From` trait for conversion between the only server function argument and server function argument type. 

The minimal example of the issue lives here: https://github.com/videobitva/test_server_fn

### DIFF
--- a/leptos_macro/src/lib.rs
+++ b/leptos_macro/src/lib.rs
@@ -883,9 +883,11 @@ pub fn slot(args: proc_macro::TokenStream, s: TokenStream) -> TokenStream {
 ///     - `"GetCbor"`: `GET` request with URL-encoded arguments and CBOR response
 /// - `req` and `res` specify the HTTP request and response types to be used on the server (these
 ///   should usually only be necessary if you are integrating with a server other than Actix/Axum)
-/// - `impl_into`: specifies whether to implement trait Into for server function's argument or not,
-///   given that there is only one argument; on making server action dispatch, this allows to call `.into()` on the
-///   server function argument, eliminating the need to fill server function’s type manually (defaults to `true`)
+/// - `impl_from`: specifies whether to implement trait `From` for server function's type or not.
+///   By default, if a server function only has one argument, the macro automatically implements the `From` trait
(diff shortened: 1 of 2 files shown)


server_fn_macro/src/lib.rs
```diff
@@ -119,6 +119,7 @@ pub fn server_macro_impl(
         res_ty,
         client,
         custom_wrapper,
+        impl_from: impl_into,
```
gbj, line 122: This line is what made me think it should be renamed: if what it does is `impl From<_>`, and we're using that language internally, it might make sense to do the same thing in the API.

videobitva: The funny thing is that initially it was named actually as `impl_from`, but later on I renamed it to be `impl_into` -- this line of code seems to be an artifact of that process. I've changed it back in the receing commit.

leptos_macro/src/lib.rs
```diff
@@ -883,6 +883,9 @@ pub fn slot(args: proc_macro::TokenStream, s: TokenStream) -> TokenStream {
 ///     - `"GetCbor"`: `GET` request with URL-encoded arguments and CBOR response
 /// - `req` and `res` specify the HTTP request and response types to be used on the server (these
 ///   should usually only be necessary if you are integrating with a server other than Actix/Axum)
+/// - `impl_into`: specifies whether to implement trait Into for server function's argument or not,
```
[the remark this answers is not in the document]
videobitva: I've renamed the argument back and merged the initial description with your variant, hope it is better now.

gbj:
Excellent, thanks!

PR #2340, opened 2024-02-20
feat(leptos-axum): propagate trace context to server functions

janu-cambrelen:
## What

This PR modifies the `handle_server_fns_inner` function within the Axum integration to enable trace context propagation to server functions.

## Why

Parent contexts — whether passed from the client or created via middleware — were not associated with downstream spans. This led to disjointed and incomplete traces. With this change, downstream spans will now be properly associated with their upstream context:

```rust
#[server]
#[cfg_attr(feature = "ssr", tracing::instrument)]
pub async fn some_server_fn() -> Result<(), ServerFnError> {
    use server_imports::*;
    
    // The generated `__some_server_fn` span will now be a child of the propagated parent span.
    // Furthermore, any child events or spans herein will also share the common trace ID.

    Ok(())
}
```

## How

By introducing two changes:

1. *Capturing Current Span*: Before spawning the task that contains/runs server functions, the current span is captured via `let current_span = tracing::Span::current()`.
2. *Entering Span Scope*: Inside the task, a guard is created using `current_span.enter()`. This ensures that the current span is set as the active span for the duration of the task.


benwis:
I'd love to have better tracing support. Does it work well on regular pages?

pythoneer:
@janu-cambrelen do you know if that would work with sentry(-tracing .. etc)? If yes, how would that look like, i guess you would `#[tracing::instrument]` the server functions, like in your example but i don't really get what the leptos/client side would look like to "start" a span in the browser and sync the trace ID.

Issue #2338, opened 2024-02-19
`<!DOCTYPE html>` tag is never rendered


thinety:
**Describe the bug**
`<!DOCTYPE html>` tag does not show up when rendering a view.

**Leptos Dependencies**
```toml
leptos = { version = "0.6", features = ["ssr", "experimental-islands"] }
```

**To Reproduce**
Code to reproduce the behavior:
```rust
use leptos::{ssr, component, view, IntoView};

#[component]
fn App() -> impl IntoView {
    view! {
        <!DOCTYPE html>
        <html>
        </html>
    }
}

fn main() {
    println!("{}", ssr::render_to_string(App));
}
```

**Expected behavior**
Expected to print `<!DOCTYPE html><html></html>`, but prints only `<html></html>`.

**Additional context**
When server side rendering, these are the relevant lines:

https://github.com/leptos-rs/leptos/blob/ffcf3c2952240598ab9f019a75c77c9145965670/leptos_macro/src/view/server_template.rs#L42
https://github.com/leptos-rs/leptos/blob/ffcf3c2952240598ab9f019a75c77c9145965670/leptos_macro/src/view/server_template.rs#L389

When rendering on the client side, the logic is about the same.

Since the `Node::Doctype(_)` pattern is being handled explicitly, I am not sure if this behavior is by design. It was a little surprising for me when I first encountered it.

gbj:
I can confirm this already works in the different implementation for the next version. I never intended this to be used as a general-purpose HTML templating library, so DOCTYPE just does nothing. The alternative for this versio would be to emit a compile error in the proc macro in the `Node::Doctype(_)` branch, to make it clearer. A PR adding that would be welcome.

PR #2355, opened 2024-02-22
Add compile error when using comment and doctype tags within the `view!` macro

thinety:
Discussed in https://github.com/leptos-rs/leptos/issues/2338.

Preview of the error messages:
image

The first commit contains just style changes (using `proc_macro_error::abort!` everywhere instead of just `abort!`, to be consistent with `proc_macro_error::emit_error!`), so it can be reverted if desired.


thinety:
I've changed the compile-time errors to warnings:
image
This only works in nightly though.

Up to you if you want to leave this change alone and just wait for the next major.

PR #2360, opened 2024-02-23
Add MSRV

paul-hansen:
Adds a Minimum Supported Rust Version using the [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field.

Current motivation for this is we are seeing people try to compile Leptos with older versions of Rust and it's failing with confusing error messages. With this PR it now gives a nice error message letting them know which version to upgrade to.

I simply chose Rust 1.75 because it's the oldest that compiles currently. Lmk if we want to use something newer.

You can test this by running this from the project root:
```
rustup install 1.74
rustup default 1.74
cargo build
```

Which using this PR will now output:

> error: package `leptos_macro v0.6.6 (/home/paul/Projects/Rust/leptos/leptos_macro)` cannot be built because it requires rustc 1.75 or newer, while the currently active rustc version is 1.74.1

I added it to the stable examples as well, the nightly examples already enforce a version in their rust-toolchain.toml files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ exclude = ["benchmarks", "examples"]
 
 [workspace.package]
 version = "0.6.6"
+rust-version = "1.75"
 
 [workspace.dependencies]
 leptos = { path = "./leptos", version = "0.6.5" }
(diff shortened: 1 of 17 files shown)


paul-hansen:
Reviewed CI failures and they all seem unrelated to me. Let me know if I missed anything though.

gbj:
Thanks!

PR #2368, opened 2024-02-25
Multiple Client Multiple Server cross communication using Nginx.

sjud:
This example shows how multiple clients can communicate with multiple servers while being shared over a single domain i.e localhost:80 using nginx as a reverse proxy.


gbj:
(See comment on #2305 re: what's the best place to put interesting example projects)

sebadob:
@sjud just as an idea, you could do this without nginx as well, when you use cloudflares [Pingora](https://github.com/cloudflare/pingora) which they officially released last week.  
This could be used directly inside the Leptos SSR code and proxy requests to your other apps. With this you have a less complex setup and you could stay Rust only as well.

gbj:
Superseded by #2500.

PR #2371, opened 2024-02-26
Add MessagePack codec

johnbchron:
This adds [MessagePack](https://msgpack.org/) codec support.

To quote MessagePack, "it's like JSON but fast and small". I was wanting to make a param payload that included a decently-sized byte array inline, and JSON is pretty inefficient at that, so I added MessagePack support.

The CI hasn't finished running on my machine but I decided to go ahead and submit it anyways because the change is so isolated.

### DIFF
--- a/server_fn/src/codec/msgpack.rs
+++ b/server_fn/src/codec/msgpack.rs
@@ -1,0 +1,74 @@
+use super::{Encoding, FromReq, FromRes, IntoReq, IntoRes};
+use crate::{
+    error::ServerFnError,
+    request::{ClientReq, Req},
+    response::{ClientRes, Res},
+};
+use bytes::Bytes;
+use http::Method;
+use serde::{de::DeserializeOwned, Serialize};
+
+/// A codec for MessagePack.
+pub struct MsgPack;
+
+impl Encoding for MsgPack {
+    const CONTENT_TYPE: &'static str = "application/msgpack";
+    const METHOD: Method = Method::POST;
+}
+
+impl<T, Request, Err> IntoReq<MsgPack, Request, Err> for T
+where
+    Request: ClientReq<Err>,
+    T: Serialize,
+{
+    fn into_req(
+        self,
+        path: &str,
+        accepts: &str,
+    ) -> Result<Request, ServerFnError<Err>> {
+        let data = rmp_serde::to_vec(&self)
+            .map_err(|e| ServerFnError::Serialization(e.to_string()))?;
+        Request::try_new_post_bytes(
+            path,
+            MsgPack::CONTENT_TYPE,
+            accepts,
+            Bytes::from(data),
+        )
+    }
+}
+
+impl<T, Request, Err> FromReq<MsgPack, Request, Err> for T
+where
+    Request: Req<Err> + Send,
+    T: DeserializeOwned,
+{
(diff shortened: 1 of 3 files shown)


server_fn/src/codec/msgpack.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl<T, Request, Err> FromReq<MsgPack, Request, Err> for T
+where
+    Request: Req<Err> + Send,
+    T: DeserializeOwned,
+{
+    async fn from_req(req: Request) -> Result<Self, ServerFnError<Err>> {
+        let string_data = req.try_into_string().await?;
```
gbj, line 46: Shouldn't this use [`try_into_bytes`](https://docs.rs/server_fn/latest/server_fn/request/trait.Req.html#tymethod.try_into_bytes) instead? I would think `try_into_string` fails if the MessagePack data is not valid UTF-8. I don't know the format but that seems unlikely.

server_fn/src/codec/msgpack.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl<T, Response, Err> FromRes<MsgPack, Response, Err> for T
+where
+    Response: ClientRes<Err> + Send,
+    T: DeserializeOwned,
+{
+    async fn from_res(res: Response) -> Result<Self, ServerFnError<Err>> {
+        let data = res.try_into_string().await?;
```
gbj, line 70: Same note re `try_into_bytes` as above.

gbj (commented):
Thanks, it's great to have added implementations like this. See notes above about `into_string` vs `into_bytes`.

server_fn/src/codec/msgpack.rs
johnbchron: You were correct. It is not valid UTF-8.

server_fn/src/codec/msgpack.rs
johnbchron: Same as above.

benwis:
I think there is a deeper question here, of how many codecs we want to have available in leptos itself. I don't think this is an issue for MessagePack or this commit, but it's worth thinking about moving forward

benwis:
Overall this looks good, thanks for contributing!

luckynumberke7in:
> I think there is a deeper question here, of how many codecs we want to have available in leptos itself. I don't think this is an issue for MessagePack or this commit, but it's worth thinking about moving forward

I haven't worked with this package before, so someone please feel free to correct me, but from skimming the package...honestly...should something like this be a flat replacement in most cases (1:1)?

These little improvements towards an optimal Leptos sound wonderful, especially for a nosql or graphql DB when possible.

benwis:
> > I think there is a deeper question here, of how many codecs we want to have available in leptos itself. I don't think this is an issue for MessagePack or this commit, but it's worth thinking about moving forward
> 
> I haven't worked with this package before, so someone please feel free to correct me, but from skimming the package...honestly...should something like this be a flat replacement in most cases (1:1)?
> 
> These little improvements towards an optimal Leptos sound wonderful, especially for a nosql or graphql DB when possible.

A flat replacement? This isn't really for db access, this is for the server fn encoding between server and client.

PR #2373, opened 2024-02-27
added hashes generated from cargo-leptos

maccesch:
This is accompanying PR for https://github.com/leptos-rs/cargo-leptos/pull/256


benwis:
Overall this looks pretty good, my question would be what would happen if you were not using cargo-leptos? We still want to maintain compatibility with that use case

maccesch:
I believe that shouldn't be a problem. The server code looks if the hash.txt exists and only if it does adds the hashes to the file names.

maccesch:
The failing check doesn't seem to be from my code

BrandonDyer64:
Any way this gets backported to leptos 6?

maccesch:
This was done for Leptos 0.6 👍🏼

PR #2378, opened 2024-02-28
Make the ScopedFuture abort on cleanup of the owner

TethysSvensson:
Currently the four `spawn` functions that use the `ScopedFuture` has some non-intuitive behavior, which this PR attempts to fix.

#### Persistence across re-evaluation of the owning effect

When you spawn a future with the current owner, the most intuitive behavior would be for it to live as long as e.g. signals created in the same closure.

Consider for instance the following code:

```rust
create_effect(move |_| {
    let value = some_signal.get();
    spawn_local_with_current_owner(async move {
        let log_stream = register_new_log_stream().await;
        loop {
            log_stream.send(format!("Value is {value:?}")).await;
            gloo::timers::future::TimeoutFuture::new(100).await
        }
    })
    .unwrap();
});
```

I would expect this code it to spawn a new future every time `some_signal` changes, which it does. However I would also expect it to stop the previous future which it does not. The result is that multiple futures will run concurrently.

#### Predictable clean-up

Currently long-running futures are not cleaned up at a predictable time.

Consider for instance the following code:

```rust
#[component]
fn LiveInfoPage() -> impl IntoView {
    let (state, set_state) = create_signal(State::default());

    spawn_local_with_current_owner(async move {
        let resource = initialize_expensive_resource().await;
        let connection = open_websocket().await;
        while let Some(msg) = connection.next_update().await {
            let state = resource.process(msg);
            set_state.set(state);
        }
    })
    .unwrap();

    "..."
}
```

Ideally the resources initialized by the future would be cleaned up as soon as the user navigates away from the `LiveInfoPage`, however that is not how `ScopedFuture` currently works.

Instead the resources only cleaned up once the `next_update()` future makes progress at which point the `ScopedFuture` will finally finish.

## Solution

Make sure the future is woken up and aborted whenever the cleanup effects run on the owner.

## Alternatives

Document the current behavior, and potentially add new functions with the behavior implemented in this PR.

### DIFF
--- a/leptos_reactive/src/runtime.rs
+++ b/leptos_reactive/src/runtime.rs
@@ -1550,7 +1550,7 @@ pub fn untrack_with_diagnostics<T>(f: impl FnOnce() -> T) -> T {
 pub struct ScopedFuture<Fut: Future> {
     owner: Owner,
     #[pin]
-    future: Fut,
+    future: futures::future::Abortable<Fut>,
 }
 
 /// Errors that can occur when trying to spawn a [`ScopedFuture`].
@@ -1573,7 +1573,8 @@ impl<Fut: Future + 'static> Future for ScopedFuture<Fut> {
 
         if let Ok(poll) = try_with_owner(*this.owner, || this.future.poll(cx)) {
             match poll {
-                Poll::Ready(res) => Poll::Ready(Some(res)),
+                Poll::Ready(Ok(res)) => Poll::Ready(Some(res)),
+                Poll::Ready(Err(futures::future::Aborted)) => Poll::Ready(None),
                 Poll::Pending => Poll::Pending,
             }
         } else {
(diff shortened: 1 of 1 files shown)


gbj:
Given the breaking changes and additional overhead, I'm going to tag this one as 0.7 to make sure it behaves as expected in the future.  Alternatively, feel free to duplicate + rename these functions, leaving the original ones as they are.

gbj:
To be clear I mean that it's a breaking change primarily because it changes the signature of public functions, and secondarily because of the behavior change. Something like `_abortable` for the function names maybe?

TethysSvensson:
Do you want me to write also update the docs for the non `abortable` versions?

gbj:
Sure, I think updating the docs for the non-abortable version would be great.

And re: a timeline for 0.7, I don't have a very firm answer but "within the next few months" seems like it's in the right zone. I'd say: if the abortable version is something you need right now, I'm happy to take the PR now, and can then decide whether to make this the default behavior in the next version.

TethysSvensson:
It *is* something I need right now, but it is also something I don't mind keeping around as a utility function in my own crate for a few months until v0.7 gets released.

I would prefer if v0.7 contained my versions, as I think the current implementation is a footgun. Alternatively I would also think it would be fine, if the current implement was available under a different name that, to make it less likely for somebody to get the wrong impression about their behavior.

gbj:
Just leaving an update:

0.7 is a significant enough rewrite that the functions this PR modifies in `leptos_reactive` are not used (that entire crate may no longer be in existence when actually released)

The idea of the PR (aborting Futures on cleanup) is probably worth including as an opt in, and possibly even by default.

It's worth saying that this can introduce its own confusions and will need to be documented clearly as to what it does/does not do: for example, if this is used with a `fetch` request aborting the Future will *not* abort the request, unless a browser `AbortSignal` is used.

gbj:
#3030 adds two functions in `leptos::task`: `spawn_local_scoped` (which preserves `Owner`, but without cancellation) and `spawn_local_scoped_with_cancellation` (which adds cancellation). The default behavior remains as currently, without the overhead of the abortable `Future`, with an easy opt-in, such that the example below behaves as expected, canceling the prior async loop when you spawn a new task.

```rust
let value = RwSignal::new(1);
Effect::new(move |_| {
    let value = value.get();
    spawn_local_scoped_with_cancellation(async move {
        loop {
            leptos::logging::log!("iterating on {value}");
            gloo_timers::future::TimeoutFuture::new(100).await;
        }
    });
});

view! { <button on:click=move |_| *value.write() += 1>{value}</button> }
```

PR #2382, opened 2024-02-29
fix: re-throw errors in delegated event listeners

Giovanni-Tably:
Hi there,

This just re-throws errors that happen in delegated event listeners, it prevents confusion from a panic not showing up anywhere (besides the control flow being interrupted).

### DIFF
--- a/leptos_dom/src/events.rs
+++ b/leptos_dom/src/events.rs
@@ -150,7 +150,10 @@ pub(crate) fn add_delegated_event_listener(
                         if !maybe_handler.is_undefined() {
                             let f = maybe_handler
                                 .unchecked_ref::<js_sys::Function>();
-                            let _ = f.call1(&node, &ev);
+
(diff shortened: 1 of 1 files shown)


leptos_dom/src/events.rs
```diff
@@ -150,7 +150,10 @@ pub(crate) fn add_delegated_event_listener(
                         if !maybe_handler.is_undefined() {
                             let f = maybe_handler
                                 .unchecked_ref::<js_sys::Function>();
-                            let _ = f.call1(&node, &ev);
+                            match f.call1(&node, &ev) {
```
gbj, line 153: style: I don't know if `clippy` has a lint for this or not, but if you have an empty match branch I'd think
```rust
if let Err(e) = f.call1(&node, &ev) {
```
would be better

gbj (commented):
Thanks, this makes sense and is a good change: see style note above.

PR #2383, opened 2024-02-29
fix: prevent setting the same event listener twice

Giovanni-Tably:
Delegated event listeners do not support adding more than one event listener of the same type. This can cause confusion if two listeners are added, as one is silently dropped.

Here I just assert that the value does not exist yet, which was what we decided to go with internally (it seems good practice to avoid double listeners, if possible).

Alternatives:
- Switch to a warning.
- Combine the two event listeners.

### DIFF
--- a/leptos_dom/src/events.rs
+++ b/leptos_dom/src/events.rs
@@ -70,6 +70,12 @@ pub fn add_event_listener<E>(
 
     let cb = Closure::wrap(cb as Box<dyn FnMut(E)>).into_js_value();
     let key = intern(&key);
+    debug_assert_eq!(
+        Ok(false),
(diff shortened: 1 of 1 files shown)


leptos_dom/src/events.rs
```diff
@@ -70,6 +70,12 @@ pub fn add_event_listener<E>(
 
     let cb = Closure::wrap(cb as Box<dyn FnMut(E)>).into_js_value();
     let key = intern(&key);
+    assert_eq!(
```
gbj, line 73: It think it's worth gating this with `#[cfg(debug_assertions)]` to keep release binaries as small as possible. What do you think?

(Thank you for your work on these PRs!)

PR #2390, opened 2024-03-01
Minor: examples/todomvc -   Rename Todos::new()  as  Todos::default().

martinfrances107:
This is just a minor point seen while copying code snippets from the examples directory

Here is a snippet from this  clippy rule [new_without_default](https://rust-lang.github.io/rust-clippy/v0.0.212/index.html)

struct Foo(Bar);

impl Foo {
    fn new() -> Self {
        Foo(Bar::new())
    }
}
Instead, use:

struct Foo(Bar);

impl Default for Foo {
    fn default() -> Self {
        Foo(Bar::new())
    }
}

both new() and default() are not needed - when just default() will suffice.

### DIFF
--- a/examples/todomvc/src/lib.rs
+++ b/examples/todomvc/src/lib.rs
@@ -7,9 +7,8 @@ pub struct Todos(pub Vec<Todo>);
 
 const STORAGE_KEY: &str = "todos-leptos";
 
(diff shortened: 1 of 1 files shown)


benwis:
LGTM thanks for the contribution!

PR #2391, opened 2024-03-01
fix(ci): "needless borrow" error and example never exiting

paul-hansen:
`options` is already a reference so it was passing a double reference.

Looks like this was inadvertently added in the last commit in PR https://github.com/leptos-rs/leptos/pull/2373 while other CI failures were happening so it was missed.


paul-hansen:
PR seems to not be picking up my latest change. Trying reopening it.

PR #2392, opened 2024-03-01
fix(ci): "needless borrow" error and example never exiting

paul-hansen:
`options` is already a reference so it was passing a double reference. Looks like this was inadvertently added in the last commit in PR https://github.com/leptos-rs/leptos/pull/2373 while other CI failures were happening so it was missed.

The error_boundary example has been leaving the server running and eventually failing due to timing out. This PR makes the start task run in parallel with the playwright task instead of waiting forever on the server start task to exit it uses a dumb two minute sleep timer to ensure the server has compiled and started. This could probably be improved but might risk being too complex for bash scripts so I left it for now.

### DIFF
--- a/examples/cargo-make/playwright-trunk-test.toml
+++ b/examples/cargo-make/playwright-trunk-test.toml
@@ -6,9 +6,17 @@ extend = [
 [tasks.integration-test]
 description = "Run integration test with automated start and stop of processes"
 env = { SPAWN_CLIENT_PROCESS = "1" }
-dependencies = ["start", "wait-one", "test-playwright", "stop"]
+run_task = { name = ["start", "wait-test-stop"], parallel = true }
 
-[tasks.wait-one]
+[tasks.wait-test-stop]
+private = true
+dependencies = ["wait-server", "test-playwright", "stop"]
+
+[tasks.wait-server]
 script = '''
-sleep 1
+for run in {1..6}; do
+    echo "Waiting to ensure server is started..."
+    sleep 10
+done
+echo "Times up, running tests"
 '''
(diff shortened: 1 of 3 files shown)


gbj:
Thanks for both of these commits: truly heroic to investigate and fix weird CI behavior :smile:

Will let CI run (hopefully for less than 6 hours :-p) and merge if it's good.

paul-hansen:
Thanks for Leptos!

Unfortunately it didn't work to kill the example :disappointed: 
Looks like that example is the only one to use [playwright-trunk-test.toml](https://github.com/leptos-rs/leptos/blob/main/examples/error_boundary/Makefile.toml#L3), I'll look at that some, if I don't get it figured out I'll pull that commit from the PR so we can at least merge the other fix.

paul-hansen:
New fix is working, updated the description in the first post with some details. Looks like the integration/axum CI is failing now but that shouldn't be related to this PR as I only changed the cargo-make files in the examples which isn't used there.

paul-hansen:
Interestingly one of the lines the axum integration is failing on isn't even in the branch for this PR: 
https://github.com/leptos-rs/leptos/actions/runs/8117600089/job/22190537155?pr=2392#step:17:4323
```
error[E0425]: cannot find function `expect_context` in this scope
 --> integrations/axum/src/lib.rs:94:16
  |
7 |     let opts = expect_context::<leptos_axum::ResponseOptions>();
  |                ^^^^^^^^^^^^^^ not found in this scope
  |
help: consider importing this function
  |
2 + use leptos::expect_context;
  |
  ```
  
  It was added in this commit:
https://github.com/leptos-rs/leptos/commit/f70ebc1289654141c50e6556e65e1389a2343f95

gbj:
Thanks!

PR #2394, opened 2024-03-01
Unnecessary trait bound `PartialEq` removed from `create_owning_memo`

### LINKED ISSUE #2388: Drop `PartialEq` requirement in `create_owning_memo`
`create_owning_memo` requires `T` to implement `PartialEq`. This is unnecessary because `f` returns a boolean that determines if the value has changed. It can also be confusing because it makes it unclear whether the equality is being used from `PartialEq` or from the boolean.

Drop the requirement for `PartialEq`.

Use of a wrapper type that has a dummy implementation for `PartialEq`.

**Additional context**
`create_owning_memo` was added in the new 0.6.7 release.

zroug:
Removes trait bound `PartialEq` from `create_owning_memo` which is not needed. Closes #2388.

### DIFF
--- a/leptos_reactive/src/memo.rs
+++ b/leptos_reactive/src/memo.rs
@@ -165,7 +165,7 @@ pub fn create_owning_memo<T>(
     f: impl Fn(Option<T>) -> (T, bool) + 'static,
 ) -> Memo<T>
 where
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

PR #2396, opened 2024-03-02
feat: allow #[prop(attrs)] on slots

Upbolt:
Saw somebody in the Discord server ask for the ability to do
```rs
#[slot]
struct Slot {
  #[prop(attrs)]
  attributes: Vec<(&'static str, Attribute)>
}
```
Which currently is not implemented, so I just extended the existing options and slapped in the impls and methods in `leptos_macro/slot.rs` and `leptos_macro/slot_helper.rs`

### DIFF
--- a/leptos_macro/src/slot.rs
+++ b/leptos_macro/src/slot.rs
@@ -65,14 +65,49 @@ impl ToTokens for Model {
             body,
         } = self;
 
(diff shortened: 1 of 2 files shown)


benwis:
Some real serendipity here, thanks for contributing!

PR #2402, opened 2024-03-04
remove printlns that run many times

ChristopherBiscardi:
On an experimental islands site I upgraded today, these printlns run >1000 times on a single page load. I believe they're just for development, but also happy to change to tracing if desired


gbj:
... Yeah that's just a "leftover while debugging" log, for sure :smile:

gbj:
(Unrelated CI flake due to how many combos of features are in the `server_fn` crate => running out of disk space.)

Thanks for the PR!

PR #2404, opened 2024-03-04
chore: Disallow prints to stdout

mondeja:
Prevents PRs like #2402

### DIFF
--- a/router/src/components/static_render.rs
+++ b/router/src/components/static_render.rs
@@ -270,6 +270,7 @@ where
         if let Some(data) = static_data.get(path.path()) {
             path.add_params(data);
         }
+        #[allow(clippy::print_stdout)]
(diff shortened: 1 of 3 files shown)


router/src/components/static_render.rs
```diff
@@ -270,6 +270,7 @@ where
         if let Some(data) = static_data.get(path.path()) {
             path.add_params(data);
         }
+        #[allow(clippy::print_stdout)]
         for path in path.into_paths() {
             println!("building static route: {}", path);
```
mondeja, line 275: I'm not sure if this is really needed.

gbj: Thanks -- let's just leave it for now

gbj (commented):
Thanks, this looks great!

PR #2407, opened 2024-03-05
Properties with into and default attribute calls into() in proc macro

### LINKED ISSUE #2386: Make #prop(into, default) a bit more ergonomic.
At the momement MaybeSignal and similar component properties can be constructed like this:
```
#prop(into, default = false.into()) disabled: MaybeSignal<bool>
#prop(into, default = "Some value".into()) disabled: MaybeSignal<String>
```

While this is working when components with many default props are created, the into() calls on the default creates a lot of noise, boilerplate. So I'd suggest to make the .into() call part of the proc-macro:
```
#prop(into, default = false) disabled: MaybeSignal<bool>
#prop(into, default = "Some value") disabled: MaybeSignal<String>
```

I think the extra into when it is not required would be compiled to a no-op.
Also this feature could be enabled only when both the input and default tags are present in the annotation.

gzp79:
resolves #2386 

Into() is added to the default_code when the prop is into and there's a default expression

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -844,8 +844,12 @@ impl TypedBuilderOpts {
 impl TypedBuilderOpts {
     fn to_serde_tokens(&self) -> TokenStream {
         let default = if let Some(v) = &self.default_with_value {
-            let v = v.to_token_stream().to_string();
-            quote! { default=#v, }
+            let mut v = v.to_token_stream();
+            if self.into {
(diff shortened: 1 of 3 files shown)


gbj:
Thanks for this contribution! Apart from all the noise of examples failing CI due to the need for a cargo-leptos update (I'm rerunning those right now) this also fails on the formatting check in `leptos_macro` -- if you just `cargo fmt` there and push it should resolve that issue. Looks good to me otherwise.

PR #2410, opened 2024-03-06
Make `For` component each prop work with signals

SleeplessOne1917:
Despite what the docs say, I have been unable to pass signals to the `For` component. To test this, try running the changed code for sso auth axum without the changes I made to the `For` component.

### DIFF
--- a/leptos/src/for_loop.rs
+++ b/leptos/src/for_loop.rs
@@ -1,5 +1,6 @@
 use leptos_dom::IntoView;
 use leptos_macro::component;
+use leptos_reactive::{Signal, SignalGet};
 use std::hash::Hash;
 
 /// Iterates over children and displays them, keyed by the `key` function given.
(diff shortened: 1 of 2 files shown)


leptos/src/for_loop.rs
```diff
@@ -24,19 +25,7 @@ use std::hash::Hash;
...
-///         each=move || counters.get()
-///         // a unique key for each item
-///         key=|counter| counter.id
-///         // renders each item to a view
-///         children=move |counter: Counter| {
-///           view! {
-///             <button>"Value: " {move || counter.count.get()}</button>
-///           }
-///         }
-///       />
```
SleeplessOne1917, line 36: This was repeated twice. I'm assuming that was unintentional. Let me know if you want to keep it.

gbj:
I do not understand your statement of the problem. What is the code you are expecting to compile that doesn't? This function signature has been the same since ~0.1 so I'd be very surprised if it is broken somehow, rather than this being a misunderstanding.

SleeplessOne1917:
> What is the code you are expecting to compile that doesn't?

Among other things, the example that I modified as part of this PR. If you try to compile that example by passing in the signal instead of a closure that calls 'get' on that signal using the code as it is in main, you will find that it doesn't compile.

SleeplessOne1917:
I was not aware of the nightly limitation. Closing this PR.

SleeplessOne1917:
Strangely, I ran into this issue in another project despite having a rust-toolchain.toml in the repo root that sets the channel to nightly.

gbj:
Using the nightly features requires two things:
1) having the nightly compiler 
2) enabling the `nightly` feature on the Leptos crate

PR #2413, opened 2024-03-08
fix some comments

battmdpkq:


integrations/axum/src/lib.rs
```diff
@@ -308,8 +308,8 @@ async fn handle_server_fns_inner(
             let (status, mut res_headers) =
                 (res_options_inner.status, res_options_inner.headers.clone());
 
-            // it it accepts text/html (i.e., is a plain form post) and doesn't already have a
-            // Location set, then redirect to to Referer
+            // it accepts text/html (i.e., is a plain form post) and doesn't already have a
```
gbj, line 311: should be "if it"

leptos_dom/src/components/each.rs
```diff
@@ -1065,7 +1065,7 @@ fn unpack_moves(diff: &Diff) -> (Vec<DiffOpMove>, Vec<DiffOpAdd>) {
 //             ]
 //         );
 
-//         // Now we're going to to the same as above, just with more items
+//         // Now we're going to the same as above, just with more items
```
gbj, line 1068: should be "going to do"

gbj (commented):
Thanks! I just left notes on two changes that were, I think, typos not duplications.

gbj:
LGTM, thanks very much!

Issue #2431, opened 2024-03-15
Spreading both attributes and event handlers


lpotthast:
Libraries like `react-aria` and now `Leptonic` try to ease creation of html elements with a lot of pre-attached logic by letting their users spread properties onto an an element.
With the alternative way, manipulating a `NodeRef`, composing multiple behaviors would be harder to implement (regarding precedence).

Leptos's spreading feature currently only supports attributes.  Which already works great btw!
Most hooks coming to Leptonic will require event listeners to be present on elements. Having a way to spread them as well would be much more ergonomic for users of such libraries.

**Describe the solution you'd like**
Instead of
```
let (attrs, handlers) = use_foo();
view! {
  <div {..attrs} on:click=handlers.on_click on:keydown=handlers.on_keydown [...] />
}
```
where every user of `use_foo` would have to now ALL event handlers that must be attached by looking at the type of `handlers` and would have to do follow-ups whenever this type changes, the following solution would be much easier and less error-prone to use.
```
let (attrs, handlers) = use_foo();
view! {
  <div {..attrs} {..handlers} />
}
```

Even more concise, when attributes and event handlers could live interleaved in a single structure. 
```
let bindings = use_foo();
view! {
  <div {..bindings} />
}
```

I chose `bindings` as a variable name, as both the concepts of `attributes` and `properties` already have a DOM-related conotation in the framework. A binding may attach/bind data through an attribute or logic through an event handler.

See PR: https://github.com/leptos-rs/leptos/pull/2432

lpotthast:
@gbj Made the PR ready. It supports both proposed solutions and should keep backward-compatibility. Would love to hear some feedback on this! Code locations, naming, the concept in general.
CI is still failing with some `error[E0658]: `#[diagnostic]` attribute name space is experimental` errors, wich I cannot connect to any of my changes. Any input on that?

PR #2432, opened 2024-03-15
Allow spreading of both attributes and event handlers

lpotthast:
The `{..foo}` block parsing in a `view!` now supports everything `impl IntoIterator<Item = B>`, where `B: Into<Binding>` with `Binding` being either an attribute of type `{ name: &'static str, value: Attribute }` or  an `EventHandlerFn` (typed, boxed Fn).

The old `Vec<(&'static str, Attribute)>` can still be used as `{..foo}`, as shown by the added `spread` example, to only spread attributes.
`Vec<EventHandlerFn>` can be used as `{..foo}`, to only spread event handlers.
`Vec<Binding>` can be used as `{..foo}`, to possibly spread both attributes and event handlers.
Multiple (`{..foo} {..bar}`) spreads can be performed on one element (possibly keeping attributes and handlers separate).

Closes: https://github.com/leptos-rs/leptos/issues/2431


leptos_dom/src/events/typed.rs
```diff
@@ -167,6 +167,86 @@ impl DOMEventResponder for crate::View {
     }
 }
 
+/// A statically typed event handler.
+pub enum EventHandlerFn {
```
gbj, line 171: I'm wondering if there's a way to implement this without a hardcoded list of event handlers: for example by storing the event name and a type-erased handler. The enum-based approach doesn't support custom events, and the big match statement seems like it could expensive in terms of binary size.

examples/spread/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// Demonstrates how attributes and event handlers can be spread onto elements.
+#[component]
+pub fn SpreadingExample() -> impl IntoView {
+    fn alert(msg: impl AsRef<str>) {
+        let _ = window().alert_with_message(msg.as_ref());
+    }
+
+    let attrs_only: Vec<(&'static str, Attribute)> = vec![
+        ("data-foo", Attribute::String(Oco::Borrowed("42"))),
```
gbj, line 11: Is there a less explicit way of constructing `Attribute`? This is pretty imposing for an example.

leptos_dom/src/html.rs
```diff
@@ -651,8 +677,14 @@ impl<El: ElementDescriptor + 'static> HtmlElement<El> {
         }
     }
 
-    /// Adds multiple attributes to the element
+    /// Adds multiple attributes to the element.
     #[track_caller]
+    #[deprecated(
```
gbj, line 682: I try to avoid adding `#[deprecated]` in patch releases because it can break CI/CD for people if they are set up to deny on warnings and their build pulls in a newer patch version than they have locally.

gbj (commented):
Thanks for your work! This is a good feature to have. 

I left a few comments on the implementation. I am open to being wrong about any of them!

It's likely all of this will change with the more-statically-typed rendering in 0.7 but it is great to continue developing the 0.6 approach as well.

Now that I've fixed the CI on main if you rebase again I think it should run properly for you.

examples/spread/src/lib.rs
Upbolt: .into_attribute() exists

lpotthast: Switched to `.into_attribute()` calls. Much more readable, and the explicitness is not needed here.

leptos_dom/src/html.rs
lpotthast: Removed it.

leptos_dom/src/events/typed.rs
lpotthast: Yeah, I don't particularly like this either.. I tried it a few times but couldn't find  a solution that works. The main purpose of this type is to be used in some sort of collection, so making it generic in any way should be avoided. But I do not know any trait that would be suitable used as a trait object here. Do you? Or how would you otherwise erase the function type? Maybe there is something around which I'm not aware of. Our `EventDescriptor` is only ever used in generic contexts and certainly not object safe. The `FromWasmAbi` trait also didn't led me to a solution. `gloo` (https://github.com/rustwasm/gloo/blob/master/crates/events/src/lib.rs) also has nothing of interest in this regard.

I think I would leave it at that for know. If you have any ideas on this, please let me know. Happy to change this!

gbj: What I meant was something like this:
```rust
pub struct EventHandlerFn {
    event_name: Oco<'static, str>,
    handler: Box<dyn FnMut(web_sys::Event)>,
}

impl EventHandlerFn {
    pub fn new<E: EventDescriptor>(ev: E, handler: impl FnMut(E::EventType)) -> Self {
        Self {
            event_name: ev.name(),
            handler: Box::new(move |ev: Event| {
                // convert `Event` into `E::EventType` using JsCast
            }),
        }
    }
}
```
Unfortunately based on the way I defined `EventDescriptor` at some point, I don't think we can actually convert an `E::EventType` into a `web_sys::Event`, even using `.unchecked_into()` — even though all of these types actually inherit from Event and that conversion should be fine.

For now then I think your solution is probably best!

gbj:
Okay, this looks good to me, now, having looked into that last comment and discovered the `enum` is probably the way to go! Thanks so much.

PR #2433, opened 2024-03-16
Fix: Environment variables do not overwrite Config.toml options

### LINKED ISSUE #2429: Environment variables do not overwrite Config.toml options
**Describe the bug**

When calling `get_configuration()` with `None`, environment variables like `LEPTOS_SITE_ADDR` are used correctly, but when the method is called with `Some("Cargo.toml")`, the environment variable gets ignored.

**Leptos Dependencies**

I am encountering this with the axum starter and leptos `0.6.9`.

```toml
```

**To Reproduce**

1. Create a new leptos project using the axum starter.
2. Run `cargo leptos build`.
3. Set the environment variable `LEPTOS_SITE_ADDR` to `0.0.0.0:1234`.
4. Run `target/debug/<project name>`.
5. You should see "listening on http://0.0.0.0:1234".
6. Edit the `site-addr` option in the `Cargo.toml` to something non default (e.g. `site-addr = "127.0.0.1:3001"`).
7. Edit the `main.rs` file and change `get_configuration(None)` to `get_configuration(Some("Cargo.toml"))`.
8. Rebuild and launch the server again (still with the environment variable set).
9. Now it shows "listening on http://127.0.0.1:3001"

**Expected behavior**

According to the documentation, configuration via environment variables should always take precedence.

GrumiumXD:
Closes #2429 

This is my first PR on this project and it turned out a bit bigger than expected. So I am not sure about some of the decisions I made.

- Removed `separator("_")` from the Environment source setup, because this is intended for nested structures.
- Added  `convert_case(Case::Kebab)` to the Environment source setup to match the casing serde expects.
- Removed `"[leptos-options]"` from toml string to let the settings be parsed as root level, otherwise the settings from the file and the ones set by the environment variables would be considered in different structures, therefore not affecting each other. Because of this `get_config_from_str()` now returns a `LeptosOptions` object instead of a `ConfFile`.

The test cases gave me some troubles as I had some inconsistent results. I assume (because tests are executed in parallel) that sometimes variables set in one test could interfere with a test that does not expect them (now that they could actually overwrite settings originated from a file source).
I added `temp-env` as a dev dependency to address my issues. Afterwards the tests were working consistently again and I added a new test to explicitly test the override functionality.

Because I changed the parsing from `ConfFile` to `LeptosOptions`, I think the `serde::Deserialize` trait could actually be removed from the `ConfFile` struct. Any opinions on that?

### DIFF
--- a/leptos_config/Cargo.toml
+++ b/leptos_config/Cargo.toml
@@ -19,3 +19,4 @@ typed-builder = "0.18"
 [dev-dependencies]
 tokio = { version = "1", features = ["rt", "macros"] }
 tempfile = "3"
+temp-env = { version = "0.3.6", features = ["async_closure"] }
--- a/leptos_config/Cargo.toml
+++ b/leptos_config/Cargo.toml
@@ -10,7 +10,7 @@ readme = "../README.md"
 rust-version.workspace = true
 
 [dependencies]
-config = { version = "0.14", default-features = false, features = ["toml"] }
+config = { version = "0.14", default-features = false, features = ["toml", "convert-case"] }
 regex = "1.7.0"
 serde = { version = "1.0.151", features = ["derive"] }
 thiserror = "1.0.38"
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -3,7 +3,7 @@
 pub mod errors;
 
 use crate::errors::LeptosConfigError;
-use config::{Config, File, FileFormat};
+use config::{Case, Config, File, FileFormat};
 use regex::Regex;
 use std::{
     convert::TryFrom, env::VarError, fs, net::SocketAddr, path::Path,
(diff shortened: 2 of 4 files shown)


benwis:
Nice work! I'd love to remove the Deserialize struct if that's possible.

GrumiumXD:
I was just talking about getting rid of the `Deserialize` trait for the `ConfFile` struct, but getting rid of the whole `ConfFile` struct and only working with `LeptosOptions` everywhere sounds even better, I guess.
I will see what I can do.

benwis:
> I was just talking about getting rid of the `Deserialize` trait for the `ConfFile` struct, but getting rid of the whole `ConfFile` struct and only working with `LeptosOptions` everywhere sounds even better, I guess. I will see what I can do.

To be honest that is my bad wording, We have both of those because of how the config crate works. At any rate, I think this looks good, and can merge and see if anything breaks

PR #2435, opened 2024-03-16
register server_fn first to allow for wildcard Route path

sify21:
It's normal to have a `NotFound` page with a wildcard path like this
```
<Routes>
    ...
    <Route path="*any" view=NotFound>
</Routes>
```
In `ssr` mode, most servers do a `first match win` approach, so we should register server functions before view routes, or else a wildcard route would block all api requests.

https://discord.com/channels/1031524867910148188/1218508054442545185


benwis:
LGTM, thanks for the PR!

PR #2444, opened 2024-03-18
chore(ci): upgrade actions to node 20

agilarity:
Resolve node warnings from the following actions:

- mbround18/install-jq@v1
- actions-rs/toolchain@v1
- jetli/trunk-action@v0.4.0
- actions/cache@v3


benwis:
Love that you're keeping on top of the CI

PR #2446, opened 2024-03-19
For the session_auth_axum example, move the passhash into a separate non-serializable struct.

rcythr:
The current example for `session_auth_axum` has a serializable `User` struct which contains the user's hashed password:  

https://github.com/leptos-rs/leptos/blob/fda4dba2370fdb4f5f81961bdeb7ca139d2e612d/examples/session_auth_axum/src/auth.rs#L5C1-L11C2

This change separates the password hash into a new, non-serializable struct to prevent it from being returned to the client. Existing auth logic is modified to use new versions of `User::get` and `User::get_from_username` which return `(User, UserPasshash)` tuples.

### DIFF
--- a/examples/session_auth_axum/src/auth.rs
+++ b/examples/session_auth_axum/src/auth.rs
@@ -6,9 +6,12 @@ use std::collections::HashSet;
 pub struct User {
     pub id: i64,
     pub username: String,
-    pub password: String,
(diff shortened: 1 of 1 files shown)


benwis:
LGTM, thanks for your contribution!

PR #2456, opened 2024-03-21
Persist parent span context within resource fetchers

boyswan:
This PR fixes a loss in parent span context when fetching resources in SSR. 

Most notably, calling `server_fn` in a resource during SSR currently results in a break in span context, and server_fn  span runs as a 'sibling' instead of entering as a child.
 
With this fix the `server_fn` span will correctly continue as a child of it's caller.

### DIFF
--- a/leptos_reactive/src/resource.rs
+++ b/leptos_reactive/src/resource.rs
@@ -1389,6 +1389,7 @@ where
                 }
             }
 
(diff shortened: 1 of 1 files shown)


benwis:
Thanks for continuing your work on this! Keep 'em coming!

PR #2461, opened 2024-03-23
Minor: Ran cargo clippy --fix

martinfrances107:
I Manually reviewed the changes. All look like reasonable nudges.

A summary :-

In one place removed a redundant call to .clone().

In two places, now using clone_from() which clippy says **MAY** be an optimisation.


benwis:
Thanks!

PR #2462, opened 2024-03-24
Implement IntoView for Rc<str>

ydirson:


ydirson:
Indeed as suggested on Discord, `Option<str>` is a much better choice than `Option<String>`, and already supported.

gbj:
Thanks for noticing this and your work to implement it! This implementation can be improved. Adding it to the `viewable_primitives` list will call `.to_string()` on the `Rc<str>`, allocating a new `String` with the same contents, and then converting that into an `Oco<'static, str>`. `Oco<_>` is a type we use internally that works like `Cow<_>`, but also has a reference-counted branch (so it can hold a `&'static str`, `Rc<str>`, or `String`) 

You can convert an `Rc<str>` directly into an `Oco<'_, str>` without the additional `.to_string()` allocation and copy. So adding something like this, in the same file, right below the `impl IntoView for Cow<'static, str>` would be ideal!
```rust
impl IntoView for Rc<str> {
    #[cfg_attr(
        any(debug_assertions, feature = "ssr"),
        instrument(level = "trace", name = "#text", skip_all)
    )]
    #[inline(always)]
    fn into_view(self) -> View {
        View::Text(Text::new(self.into()))
    }
}
```
(note this is identical code to the `Cow<_>` implementation -- since both implement `Into<Oco<_>>` they are the same)

gbj:
Thanks so much! :tada:

PR #2466, opened 2024-03-25
Allow CDN_PKG_PATH at runtime.

zakstucke:
This PR implements runtime support for CDN_PKG_PATH, preferring it when available.
Beforehand, only build time would be read.

## Why
In my case, my cdn path is categorised as a sensitive key, therefore only available at runtime later than the build stage.

### DIFF
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -56,9 +56,16 @@ pub fn html_parts_separated(
     options: &LeptosOptions,
     meta: Option<&MetaContext>,
 ) -> (String, &'static str) {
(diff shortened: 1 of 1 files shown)


benwis:
I would lean towards a solution that kept the env var check during compile time, and could be overriden by an env var at runtime. This way both users get what they want

zakstucke:
@benwis could you elaborate? I read what you're saying to be exactly how I implemented it.

Before: 
avail at build time ? use else default

After:
avail at run time ? use else before logic

benwis:
I think you're right, I made a mistake with that. Jet lagged 8 hours in London is not helping. Thanks for the PR

Issue #2414, opened 2024-03-08
Hard to Debug NodeRef Warning


soundprojects:
**Describe the bug**
When getting the warning:
`[Warning] You are setting a NodeRef that has already been filled. It’s possible this is intentional, but it’s also possible that you’re accidentally using the same NodeRef for multiple _ref attributes.`

There is no stack trace or way to debug where this warning originates, which NodeRef has already been filled and where to look for it

**Leptos Dependencies**

```toml
leptos = { version = "0.6.9", features = ["csr"] }
leptos_meta = { version = "0.6.9", features = ["csr"] }
leptos_router = { version = "0.6.9", features = ["csr"] }
```

**To Reproduce**
1. That's the issue, I don't know how to reproduce this

**Expected behavior**
A stack trace or more information of any kind would really help 

**Screenshots**

**Additional context**

gbj:
The solution here would be to do what we do for other signals and add a debug-mode-only `defined_at: &'static Location<'static>` to `NodeRef`, with `#[track_caller]` on the functions that allow you to create node refs, so that it can log "You are setting a NodeRef, defined at {defined_at}...` instead. It is not likely I will get to this in the near future but would welcome a PR from anyone.

martinfrances107:
I think that PR above is a good reponse to the problem.

PR #2467, opened 2024-03-25
Debug NodeRef Warning (#2414)

martinfrances107:
This is a response to #2414

In production #[repr(transparent)] is use to ensure a compact representation in memory. 
In developement the memory is slightly expanded.

Things I am thinking about 

A) I have introduced a maintence hazard... the documentation block for NodeRef is repeated twice 
So any future PRs will have to make sure things are kept in sync.
I would love to avoid that...

B) I may have been over eager.. ( adding  #[track_caller ] in too many places. ) 
impl<T: ElementDescriptor> Clone for NodeRef<T> {
    #[track_caller]
    fn clone(&self) -> Self {
        *self
    }

### DIFF
--- a/leptos_dom/src/node_ref.rs
+++ b/leptos_dom/src/node_ref.rs
@@ -189,10 +189,12 @@ impl<T: ElementDescriptor + 'static> NodeRef<T> {
         self.element.update(|current| {
             if current.is_some() {
                 crate::debug_warn!(
-                    "You are setting a NodeRef that has already been filled. \
+                    "You are setting the NodeRef defined at {}, which has \
+                     already been filled \
                      It’s possible this is intentional, but it’s also \
                      possible that you’re accidentally using the same NodeRef \
-                     for multiple _ref attributes."
+                     for multiple _ref attributes.",
+                    self.defined_at
                 );
             }
             *current = Some(node.clone());
--- a/leptos_dom/src/node_ref.rs
+++ b/leptos_dom/src/node_ref.rs
@@ -218,7 +218,6 @@ impl<T: ElementDescriptor + 'static> NodeRef<T> {
 }
 
 impl<T: ElementDescriptor> Clone for NodeRef<T> {
-    #[track_caller]
     fn clone(&self) -> Self {
         *self
     }
--- a/leptos_dom/src/node_ref.rs
+++ b/leptos_dom/src/node_ref.rs
@@ -2,38 +2,76 @@ use crate::{html::ElementDescriptor, HtmlElement};
 use leptos_reactive::{create_render_effect, signal_prelude::*};
 use std::cell::Cell;
 
-/// Contains a shared reference to a DOM node created while using the `view`
-/// macro to create your UI.
-///
-/// ```
-/// # use leptos::{*, logging::log};
-/// use leptos::html::Input;
-///
-/// #[component]
-/// pub fn MyComponent() -> impl IntoView {
-///     let input_ref = create_node_ref::<Input>();
-///
-///     let on_click = move |_| {
-///         let node =
-///             input_ref.get().expect("input_ref should be loaded by now");
-///         // `node` is strongly typed
-///         // it is dereferenced to an `HtmlInputElement` automatically
-///         log!("value is {:?}", node.value())
-///     };
-///
-///     view! {
-///       <div>
-///       // `node_ref` loads the input
-///       <input _ref=input_ref type="text"/>
-///       // the button consumes it
-///       <button on:click=on_click>"Click me"</button>
-///       </div>
-///     }
-/// }
-/// ```
-#[repr(transparent)]
-pub struct NodeRef<T: ElementDescriptor + 'static>(
-    RwSignal<Option<HtmlElement<T>>>,
+cfg_if::cfg_if!(
+  if #[cfg(debug_assertions)] {
+    /// Contains a shared reference to a DOM node created while using the `view`
+    /// macro to create your UI.
+    ///
+    /// ```
+    /// # use leptos::{*, logging::log};
+    /// use leptos::html::Input;
+    ///
+    /// #[component]
+    /// pub fn MyComponent() -> impl IntoView {
+    ///     let input_ref = create_node_ref::<Input>();
+    ///
+    ///     let on_click = move |_| {
+    ///         let node =
+    ///             input_ref.get().expect("input_ref should be loaded by now");
+    ///         // `node` is strongly typed
+    ///         // it is dereferenced to an `HtmlInputElement` automatically
+    ///         log!("value is {:?}", node.value())
+    ///     };
+    ///
+    ///     view! {
+    ///       <div>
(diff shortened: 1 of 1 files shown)


leptos_dom/src/node_ref.rs
```diff
@@ -4,38 +4,76 @@ use leptos_reactive::{
...
-///       // the button consumes it
-///       <button on:click=on_click>"Click me"</button>
-///       </div>
-///     }
-/// }
-/// ```
-#[repr(transparent)]
-pub struct NodeRef<T: ElementDescriptor + 'static>(
-    RwSignal<Option<HtmlElement<T>>>,
+cfg_if::cfg_if!(
```
gbj, line 7: The `cfg_if` block and duplicated code can be replaced with
```rust
#[cfg(debug_assertions)]
defined_at: &'static std::panic::Location<'static>,
```
in the struct definition and 
```rust
#[cfg(debug_assertions)]
defined_at,
```
when it's used.

leptos_dom/src/node_ref.rs
```diff
@@ -146,7 +188,7 @@ impl<T: ElementDescriptor + 'static> NodeRef<T> {
     where
         T: Clone,
     {
-        self.0.update(|current| {
+        self.element.update(|current| {
             if current.is_some() {
                 crate::debug_warn!(
                     "You are setting a NodeRef that has already been filled. \
```
gbj, line 194: You can update the warning here to something like `"You are setting the NodeRef defined at {}, which has already been filled...", self.defined_at` so it actually uses the new information.

leptos_dom/src/node_ref.rs
```diff
@@ -178,6 +220,7 @@ impl<T: ElementDescriptor + 'static> NodeRef<T> {
 }
 
 impl<T: ElementDescriptor> Clone for NodeRef<T> {
+    #[track_caller]
```
gbj, line 223: not needed

leptos_dom/src/node_ref.rs
```diff
@@ -186,6 +229,7 @@ impl<T: ElementDescriptor> Clone for NodeRef<T> {
 impl<T: ElementDescriptor + 'static> Copy for NodeRef<T> {}
 
 impl<T: ElementDescriptor + 'static> Default for NodeRef<T> {
+    #[track_caller]
```
gbj, line 232: this one should stay, I think

gbj (changes requested):
Thanks for working on this! I proposed a few small changes in the comments above. Much appreciated as always.

leptos_dom/src/node_ref.rs
martinfrances107: Thanks that look better .. pushed

leptos_dom/src/node_ref.rs
benwis: @martinfrances107 Any chance you can or have made this change so we can merge this?

leptos_dom/src/node_ref.rs
benwis: @martinfrances107 Same

martinfrances107:
Sorry for the late reply

There is a small nuance here.

This directive is dropped for a debug build
something I was hoping to avoid .

```
  #[repr(transparent)]
```

I have to do it by adding a

```
 #[allow(missing_docs)]
```

leptos_dom/src/node_ref.rs
martinfrances107: Sure, fixed.

martinfrances107:
I recently, rebased and force pushed - as I think it will  fix build errors

benwis:
To be honest I think the CI is just unhappy

leptos_dom/src/node_ref.rs
```diff
@@ -2,38 +2,48 @@ use crate::{html::ElementDescriptor, HtmlElement};
...
-///       </div>
-///     }
-/// }
-/// ```
-#[repr(transparent)]
-pub struct NodeRef<T: ElementDescriptor + 'static>(
-    RwSignal<Option<HtmlElement<T>>>,
+  cfg_if::cfg_if!(
+  if #[cfg(debug_assertions)] {
+    #[allow(missing_docs)]
```
gbj, line 7: Is there a reason to think docs.rs builds docs using release mode? My assumption would be that doing this just hides the docs.

leptos_dom/src/node_ref.rs
```diff
@@ -2,38 +2,48 @@ use crate::{html::ElementDescriptor, HtmlElement};
...
+    ///       <div>
+    ///       // `node_ref` loads the input
+    ///       <input _ref=input_ref type="text"/>
+    ///       // the button consumes it
+    ///       <button on:click=on_click>"Click me"</button>
+    ///       </div>
+    ///     }
+    /// }
+    /// ```
+    #[repr(transparent)]
```
gbj, line 42: It's not clear to me that there's any benefit to `#[repr(transparent)]`. If you think it's important, I'd propose that rather than a `cfg_if` block that removes the docs in debug mode, this should be `#[cfg_attr(not(debug_assertions), repr(transparent)]` if possible, and `#[cfg(debug_assertions)]` on the `defined_at` field.

leptos_dom/src/node_ref.rs
```diff
@@ -65,9 +75,13 @@ pub struct NodeRef<T: ElementDescriptor + 'static>(
 ///     }
 /// }
 /// ```
+#[track_caller]
 #[inline(always)]
 pub fn create_node_ref<T: ElementDescriptor + 'static>() -> NodeRef<T> {
-    NodeRef(create_rw_signal(None))
+    NodeRef {
+        defined_at: std::panic::Location::caller(),
```
gbj, line 82: Needs `#[cfg(debug_assertions)]` on the `defined_at` field, otherwise this won't compile in release I don't think.

leptos_dom/src/node_ref.rs
martinfrances107: > It's not clear to me that there's any benefit to #[repr(transparent)]

I just providing context - not expressing an opinion..
When I "git blame" this line I am directed to this issue - [831](https://github.com/leptos-rs/leptos/pull/831)

With this subheading 

> The goal of all these changes is to reduce WASM binary size with opt-level = "z" and reduce the number of allocations for tiny objects.

All very normal  -- I am going to try and keep it rather than throw it out in favor of better debug

leptos_dom/src/node_ref.rs
martinfrances107: Thanks... I will fix that in the next PR.

martinfrances107:
Ok this next version looks much less like a dogs breakfast

```rustlang
#[cfg_attr(not(debug_assertions), repr(transparent)]
```
They was the key to this, that I was not seeing ... thank you - thank you

gbj (approved):

gbj:
Thanks very much for the contribution, and for your patience through all the back and forth!

PR #2468, opened 2024-03-28
Remove panic for axum ResponseOptions

jollygreenlaser:
It's not clear what's causing `ResponseOptions` to get removed from context but it causes occasional panics. This removes the panics and starts providing some debug info to track the issue down further.


benwis:
I hope we can get to the bottom of this!

PR #2474, opened 2024-03-31
Add repository field to server_fn_macro

paolobarbolini:
Hi! While scraping crates.io I've found `server_fn_macro` to be missing the `repository` field. This fixes it.


gbj:
Thanks!

Issue #2481, opened 2024-04-01
Help converting examples from `nightly` to `stable`


gbj:
**Is your feature request related to a problem? Please describe.**

We've had many requests over time to switch from using the `nightly` feature and nightly compiler as the default for examples and templates to using stable by default, to make it easier to understand for newcomers and to avoid giving the impression that Leptos only works on nightly or is particularly experimental or unstable.

The use of `nightly` everywhere also causes annoying issues in CI from time to time.

It's been a low-priority item for a while, and it would be great to have community help to get there.

**Describe the solution you'd like**
Remove the `nightly` feature on all examples and adjust the syntax accordingly, typically just switching the function-call syntax for signals to use `.get()` or `.set()`.

Alt-iOS:
Just curious, wouldnt it be better to have both? Personally I always use nightly because I like not having the `.get()` or `.set()` syntax (and I can use stuff like the parallel frontend in the compiler) but I understand why its a good idea to show that it works fine in stable, so why not keep both?

But yeah,  I think I can take care of the hackernews examples and maybe some of the counter ones

gbj:
> Personally I always use nightly because I like not having the `.get()` or `.set()` syntax (and I can use stuff like the parallel frontend in the compiler) 

Me too, which is why it's been the default until now.

> but I understand why its a good idea to show that it works fine in stable, so why not keep both?

My main concern is actually CI-related. We have not had many problems until fairly recently, but the instability of nightly APIs means that occasionally, things break and need to be fixed in CI in ways that can be confusing for people trying to use the examples. 

One small case study that prompted me to open this: A few months ago, we found that `rkyv` was throwing clippy warnings on new nightly versions, which was breaking our CI during `cargo check` on `leptos_reactive` featurs that pulled in `rkyv`.  I probably should've investigated more deeply at the time (it turns out just disabling the `copy` feature on `rkyv` would've fixed it) but we just pinned to a nightly version that worked, instead.

Then some time in the last few weeks, an Axum patch update was released that was *only* compatible with more recent nightly compilers — but `rkyv` still wasn't. This meant we were in an uncanny valley where there was *no* nightly version of the compiler that could actually cause our CI to pass. I ended up spending a few hours of my day yesterday just dealing with the consequences of that combination and updating to a new nightly.

Ultimately this stuff is just a pain, and it will continue to be a pain as long as we have nightly examples that we're running CI against — hence why I'd rather replace them than just add more stable examples.
 
[...]

PR #2485, opened 2024-04-02
ci(examples/error-boundary): use stable syntax

agilarity:
WIP - Work in Progress

Supports #2481

## Verification

- `cargo +stable build`

## Notes

- I decided to fix to #2486 here.


agilarity:
Fix #2486 first.

PR #2487, opened 2024-04-02
Moving hackernews examples to stable

Alt-iOS:


Alt-iOS:
Hey so with the stable specification in the `rust-toolchain.toml` it was getting an error that said target `wasm32-unknown-unknown` not installed so maybe it was searching for the latest stable version and thats not with what the CI is config to use, thats why I took that out in the latest commit, hoping now it does pass

benwis:
I wouldn't worry about the CI, change it to what you'd expect and we'll fix the CI. Looks like you have some conflicts too

gbj:
Thanks for the contribution! It looks like it has crossed paths with other contributions, so I'm going to close it now -- sorry about that.

PR #2499, opened 2024-04-04
ex: counter_url_query; to stable

solweo:
Addressing: #2481 
> New code contains only _stable_ features

### Verification

`cargo +stable build`


benwis:
Thank you!

PR #2500, opened 2024-04-04
projects directory with 4 projects

sjud:
This adds a project directory to crate root based of discord discussion with gbj.
I think the idea here is to have a dumping ground for everybody's projects in one convenient location for easy user discovery. 
By having it be a separate directory we can exclude it from CI, and it won't have the same guarantees that the examples directory does. The project can aid in discovery of projects while distancing itself from the actual content, somewhat.

If people are storing lots of different Leptos projects here, and there aren't the same standards (which there probably shouldn't be to encourage lots of contributions), then individual contributors of those projects will have a lot of freedom to update these projects whenever they see fit which will hopefully translate into a lot of pull requests for people pushing code and taking responsibility for their own projects in the projects folder.

I'm not sure how fine-grained the merge approval process can get and if you can do it per-directory but I'd be happy to oversee the projects directory, maintain a list of which projects do what, and work with contributors to put their projects here and approve & merge pulls requests to the projects directory.

Thanks,
Sam


gbj:
Thanks so much.

It would be helpful if you could write something for a small `README.md` for the `projects` dir that just describes each of these projects. I will open a README file with some small notes, so at your convenience you can just make a PR to update that.

Thanks for your patience and for putting these all together!

sjud:
> Thanks so much.
> 
> It would be helpful if you could write something for a small `README.md` for the `projects` dir that just describes each of these projects. I will open a README file with some small notes, so at your convenience you can just make a PR to update that.
> 
> Thanks for your patience and for putting these all together!

Absolutely! Thank you. :)

Issue #2513, opened 2024-04-10
Make NavigateOptions configurable for create_query_signal


kryesh:
Hi, really enjoying using leptos so far.

**Is your feature request related to a problem? Please describe.**
When setting the value of a query signal, my app that isn't mounted on the root URL of the domain gets redirected to an incorrect url, for example, on page:
`domain.com/prefix/app/page?key=value`
When updating the signal for "key" the app redirects to an invalid uri:
`domain.com/prefix/app/prefix/app/page?key=new_value`
The desired uri in this case would be:
`domain.com/prefix/app/page?key=new_value`

**Describe the solution you'd like**
create_query_signal is hardcoded to use NavigateOptions::default(); 
https://github.com/leptos-rs/leptos/blob/454a4f4ccbf5de77a76ddd393cfafed20a67fbcd/router/src/hooks.rs#L80
I'd like to be able to customize NavigateOptions to have "resolve: false" as that is how I fixed the behavior for other redirects in the application.

**Describe alternatives you've considered**
Looked at use_query and use_query_map, but they weren't fit for purpose as Memo<> doesn't provide a method to mutate the inner ParamsMap

gbj:
PR welcome to add a `create_query_signal_with_options()` (or similar)

PR #2517, opened 2024-04-11
Add 'create_query_signal_with_options' to leptos_router

kryesh:
Addresses https://github.com/leptos-rs/leptos/issues/2513


gbj:
This looks good, thank you! Will let the relevant CI run.

PR #2518, opened 2024-04-12
fix: make TextProp's IntoView and IntoAttribute impls reactive

0e4ef622:
This makes passing a `TextProp` directly into an attribute behave as one would expect.


benwis:
Thanks for the fix!

PR #2520, opened 2024-04-12
docs: remove unnecessary type parameter and trait bound in component macro 'bad' example

ethanniser:
the removed type parameter is redundant to the 'bad' example, and is confusing with it being one of the 'valid' options


gbj:
Oops! Thanks!

Issue #2522, opened 2024-04-13
Allow multiple classes  in the class-tuple syntax


bicarlsen:
**Is your feature request related to a problem? Please describe.**
When using the class tuple syntax (i.e. `class=("my-class", true)`), multiple classes are not supported. e.g. `class=("class-1 class-2", true)`.

**Describe the solution you'd like**
Extend the class-tuple syntax to allow for multiple classes.

**Describe alternatives you've considered**
I currently move each of the classes into its own tuple. e.g. `class=("class-1", true) class=("class-2", true)`.

**Additional context**
The issue seems to come because [`DomTokenList.add_1(class_name)`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.DomTokenList.html#method.add_1) is [used to add the class](https://github.com/leptos-rs/leptos/blob/e29d31e686b75b62a6413e9b386dccba425b68f9/leptos_dom/src/macro_helpers/into_class.rs#L122), which [expects the classes to already be separated](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add#invalidcharactererror).

This could be remedied by first parsing the classes into a `Vec`, `split`ting on whitespace, then
1. Using [`DomTokenList.add()`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.DomTokenList.html#method.add) in place of `add_1`, or 
2. By iterating over the `Vec`, repeatedly calling `add_1`
I haven't played with the performance tradeoffs.

bicarlsen:
If this seems like something wroth tackling, I'm happy to issue a PR.

gbj:
I am open to providing some alternate API to simultaneously set or un-set multiple classes in response to the same reactive change, and possibly checking for whitespace in the class name in the view macro and splitting them at runtime.

I'm not open to splitting the string at runtime unless it's purely opt-in: if `class:foo=some_condition` requires checking whether `"foo"` contains whitespace or wrapping it in a JS `Array`, it's just pure overhead.

bicarlsen:
I'll take a look and see what I can put together. Agreed that it should be done in the `view` macro.

gbj:
Feature added by #2532. Nice work!

PR #2532, opened 2024-04-16
Adds ability to use multiple classes in view macro using array syntax.

bicarlsen:
Adds array syntax to the `fancy class name` syntax in the `view` macro for multiple classes to be controlled by the same `class` attribute.
e.g.
```rust
#[component]
fn App() -> impl IntoView {
  let (classes, set_classes) = create_signal(true);

  view! {
    <div
      class=(["class-1", "class-2"], classes)
    >
      <button
        on:click=move |_| set_classes(!classes)
      >
        Toggle
      </button>
    </div>
  }
}
```
Compared to the current style which requires creating a new `class` attribute for each class.
e.g.
```rust
#[component]
fn App() -> impl IntoView {
  let (classes, set_classes) = create_signal(true);

  view! {
    <div
      class=("class-1", classes)
      class=("class-2", classes)
    >
      <button
        on:click=move |_| set_classes(!classes)
      >
        Toggle
      </button>
    </div>
  }
}
``` 
This can be especially useful when using styling systems like Tailwind.

This was discussed in #2522.

### DIFF
--- a/leptos_macro/src/view/client_template.rs
+++ b/leptos_macro/src/view/client_template.rs
@@ -321,19 +321,14 @@ fn attr_to_tokens(
     // Classes
     else if let Some(name) = name.strip_prefix("class:") {
         let value = attribute_value(node);
-        match value {
-            syn::Expr::Lit(value) => {
-                todo!();
-            }
 
-            _ => expressions.push(quote_spanned! {
-                span=> ::leptos::leptos_dom::class_helper(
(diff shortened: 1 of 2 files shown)


bicarlsen:
Another route to go on this would be to use the current syntax, but split the class string at spaces.
e.g.
```rust
<div
  class=("class-1 class-2", true)
>
</div>
```

gbj:
I like the array syntax a lot, and it seems to work in my testing. Thanks for this!

Issue #2403, opened 2024-03-04
Spreading a dynamic list of attributes on a component


bram209:
**Describe the bug**
Not entirely sure if this is a bug or a feature request, but I would like to be able to spread a dynamic list of attributes to a component.

**Leptos Dependencies**

```toml
leptos = { version = "0.6", features = ["nightly"] }
leptos_axum = { version = "0.6", optional = true }
leptos_meta = { version = "0.6", features = ["nightly"] }
leptos_router = { version = "0.6", features = ["nightly"] }
```

**To Reproduce**
```rust
#[component]
fn Comp1() -> impl IntoView {
    let attrs = vec![("id", Attribute::String("1".into()))];
    view! {
        <>
            // this works (id is set to 1)
            <div {..attrs}>Hello</div>

            // this doesn't work (no id is set)
            <Comp2 {..attrs}/>

            // this does work
            <Comp2 attr:test=5/>
        </>
    }
}

#[component]
fn Comp2(#[prop(attrs)] attrs: Vec<(&'static str, Attribute)>) -> impl IntoView {
    let attrs_len = attrs.len();
    view! { <div {..attrs}>{attrs_len}</div> }
}

```
**Expected behavior**
I would expect that `Comp2` renders into `<div id="1">1</div>`, but instead it renders to `<div>0</div>`.

Upbolt:
i've got a working solution for this, but my fork also includes a way to spread things other than attributes which https://github.com/leptos-rs/leptos/pull/2432 seems to have beaten me to

so if/when it gets merged i'll adapt it to that

bram209:
> i've got a working solution for this, but my fork also includes a way to spread things other than attributes which #2432 seems to have beaten me to
> 
> so if/when it gets merged i'll adapt it to that

Very cool, thanks for your contributions : )

PR #2534, opened 2024-04-16
feat: spread component attrs

Upbolt:
Solves https://github.com/leptos-rs/leptos/issues/2403

In the following code `Bar` will not inherit attributes from `Foo` when it spreads its attributes
```rs
fn main() {
    let (count, set_count) = create_signal(0);

    mount_to_body(move || {
        view! {
            <Foo
                attr:hello=move || count.get().to_string()
            />

            <button on:click=move|_| { set_count.update(|count| *count += 1) }>"+ count"</button>
        }
    });
}

#[component]
fn Foo(#[prop(attrs)] attrs: Vec<(&'static str, Attribute)>) -> impl IntoView {
    view! {
        <Bar {..attrs} />
    }
}

#[component]
fn Bar(#[prop(attrs)] attrs: Vec<(&'static str, Attribute)>) -> impl IntoView {
    view! {
        <div {..attrs}>"hello world"</div>
    }
}
```

This PR allows for this functionality

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -365,33 +365,6 @@ impl ToTokens for Model {
             })
             .collect::<TokenStream>();
 
-        let dyn_binding_props = props
-            .iter()
-            .filter(
-                |Prop {
-                     prop_opts: PropOpt { attrs, .. },
-                     ..
-                 }| *attrs,
-            )
-            .filter_map(
-                |Prop {
-                     name,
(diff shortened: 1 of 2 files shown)


bram209:
This is great, personally I found it confusing that HTML components behaved differently from custom components and think this is a great addition DX wise 👍

benwis:
Thanks for the PR, can't say I've tried to do this before. Definitely going to try it though

PR #2553, opened 2024-04-24
Add an example for generating sitemaps

JoeyMckenzie:
Huge fan of Leptos and all the work everyone has done to make it the most fun web framework on the market. I recently rebuilt my website using Leptos and couldn't find an example (though I probably wasn't looking hard enought) for generating dynamic sitemaps. I thought it would be helpful for anyone looking to do the same on the websites they're building, especially for sites with lots of a pages. No harm done if the maintainers don't see any value in this example - this was purely for my use case.

The example is based on the Axum start template and uses sqlx for querying content, alongside the xml crate for generating the sitemap file.


sjud:
This is a good project for the Projects directory! https://github.com/leptos-rs/leptos/tree/main/projects

JoeyMckenzie:
@sjud Ah, good to know. Moved everything over to the project directory, thanks for calling that out!

gbj:
Thanks!

PR #2557, opened 2024-04-24
Use trunk built-in way of handling tailwind

SleeplessOne1917:
Update the tailwind CSR example to use [tailwind trunk asset](https://trunkrs.dev/assets/#tailwind).


gbj:
Thanks!

PR #2558, opened 2024-04-24
docs: add caveats for ProtectedRoute

gbj:


router/src/components/route.rs
```diff
@@ -106,6 +106,38 @@ where
...
+/// # use leptos::*; use leptos_router::*;
+/// # if false {
+/// let has_permission = move || true; // TODO!
+///
+/// view! {
+///  <Routes>
+///    // parent route
+///    <Route path="/" view=move || {
+///      view! {
+///        // only show the outlet when `is_loaded` is `true`, and hide it when it is `false`
```
reedwoodruff, line 128: I think that you meant:
> only show the outlet when ~~is_loaded~~ __has_permission__ is true"

gbj: Thanks! I changed the name to make more sense halfway through and missed this.

PR #2563, opened 2024-04-28
The counter example, with Dwarf debugging (breakpoints, single stepping in vscode and the browser)

Hecatron:
Hi,
I've recently been looking into if it's possible to single step / debug wasm rust within the browser or vscode
It turns out it is possible although there doesn't seem to be much in the way of examples or documentation on getting it to work.

So I've put together an example of the counter that also can do debugging.

  * Readme lists what's needed to get it to work.
  * index.html modified slightly.
  * I've included examples for the .vscode files for this one since it's main focus is on how to get debugging to work easily

So far I've only really tried this under the windows environment.


gbj:
Cool! I am wondering if this belongs in `examples` or in `projects` — I am trying to move more things into `projects` over time because the maintenance burden for examples is too high with the number we have in there right now, but it's great to be able to point to things saved somewhere.

Hecatron:
I've now moved this to projects, it's not perfect in that you can add breakpoints within chrome and visual studio code which is pretty cool, but so far I can't see what's inside variables with the example setup (so no variable inspection).

gbj:
Awesome — Thanks! The write-up in the README is really helpful.

PR #2570, opened 2024-05-02
Update leptos-spin-macro reference

itowlson:
The currently referenced `leptos-spin-macro` crate depends on `spin-sdk` v2, which can cause conflicts for users wanting to run Leptos on Spin using `spin-sdk` v3.

This PR updates the reference.  `leptos-spin-macro` 0.2 does not depend on the Spin SDK so should avoid any potential conflicts.


itowlson:
I am getting a whole lot of CI errors that don't seem related to this change, and am not sure how (or whether) to fix them.  Any advice welcome...!

benwis:
Thanks for the update!

PR #2577, opened 2024-05-06
Example using the bevy 3d game engine and leptos

Hecatron:
Hi,
This example is for using the bevy 3d engine and leptos in combination.
The controls inside the component move the cube up and down via events.

Currently there are events going from leptos -> bevy.
I think it should be possible to do the other way around as well, although I've not tried setting that up yet so it's a bit experimental.


gbj:
Thanks! Could you put this in the `projects` directory instead? I'm trying to keep most new examples there so that we have them for reference and to point people toward, but without me needing to maintain them long term.

Hecatron:
Ok this one now moved to the projects directory

gbj:
Thanks!

PR #2583, opened 2024-05-08
fix: StoredValue and Resource panic during cleanup

luxalpa:
If the value inside a `StoredValue` implements the drop trait, and that drop causes another read, update or dispose on another StoredValue, we'd get a cryptic panic with `already mutably borrowed: BorrowError`. 
I encountered this issue with a `StoredValue<Disposer>` with the disposer type returned by `as_child_of_current_owner()`. Because the Disposer cleans up a reactive sub-scope on Drop, it will try to mutably borrow `Runtime::stored_values` while it is still mutably borrowed by the function that drops it.

### DIFF
--- a/leptos_reactive/src/runtime.rs
+++ b/leptos_reactive/src/runtime.rs
@@ -294,10 +294,12 @@ impl Runtime {
                 drop(node);
             }
             ScopeProperty::Resource(id) => {
-                self.resources.borrow_mut().remove(id);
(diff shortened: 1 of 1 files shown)


gbj:
Oh, wow. This is one of those very awkward examples where the scoping of when the `BorrowMut` handle gets dropped makes a real difference. Nice catch.

PR #2589, opened 2024-05-11
fix: grammar typo for MultiActon doc comment

pitoniak32:
Noticed this small typo when digging into some of the examples :)

Thank you for all that you are doing!


gbj:
Thanks!

PR #2598, opened 2024-05-21
Allow slice! macro to index tuples

SleeplessOne1917:
Currently, the `slice!` macro breaks when trying to index a tuple. Example:

```rust
use leptos::*;

#[derive(Clone, PartialEq, Eq)]
struct PostView {
    pub post: Post,
    pub creator: String
}

#[derive(Clone, PartialEq, Eq)]
struct Post {
    pub body: String,
    pub id: PostId
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
struct PostId(pub i32);

let post = RwSignal::new(
    PostView {
        creator: "Bob".to_owned(),
        post: Post {
            body: "Hello World!!!".to_owned(),
            id: PostId(42)
        }
    }
);

let id = slice!(post.post.id.0); // Does not compile!
```

This change allows both identifiers and tuple indices to be used.

### DIFF
--- a/leptos_macro/src/slice.rs
+++ b/leptos_macro/src/slice.rs
@@ -11,7 +11,7 @@ use syn::{
 
 struct SliceMacroInput {
     root: syn::Ident,
-    path: Punctuated<syn::Ident, Token![.]>,
+    path: Punctuated<syn::Member, Token![.]>,
 }
 
 impl Parse for SliceMacroInput {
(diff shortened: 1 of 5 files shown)


gbj:
Thanks! The `leptos_macro/tests/ui` tests that you changed are unrelated to the changes in this PR, I think, and your changes appear to cause the tests in CI to fail. Could you revert those changes?

(There may be a version mismatch between the compiler you're running them against locally and the compiler version running in CI, which would account for the different error messages maybe)

SleeplessOne1917:
@gbj I fixed the issue you pointed out. No clue why the semver step and the hacker new js fetch step are both failing.

gbj:
It's okay they are failing on main and I just haven't had time recently to look into it. Thanks very much, this looks good.

PR #2601, opened 2024-05-27
fix: URL encoding issue (closes #2602)

### LINKED ISSUE #2602: Query parameters are not properly escaped
**Describe the bug**
Because leptos_router uses `js_sys::encode_uri` instead of `js_sys::encode_uri_component` under the hood, it does not escape characters like `%` and `+`. In #2601 I fixed the issue. Please note that the fix also revealed an issue with double decoding values, which causes errors with `%`/`%25` in the URL, which I also fixed in the same PR.

**Leptos Dependencies**
leptos = { version = "0.6.11", features = ["nightly"] }
leptos_meta = { version = "0.6.11", features = ["nightly"] }
leptos_actix = { version = "0.6.11", optional = true }
leptos_router = { version = "0.6.11", features = ["nightly"] }
leptos_dom = "0.6.11"
leptos_reactive = { version = "0.6.11" }
leptos_macro = { version = "0.6.11" }

**To Reproduce**
Run the following code, then try to enter `+` into input.

```rs
use leptos::*;
use leptos_meta::*;
use leptos_router::{
    params_map, use_navigate, use_query, NavigateOptions, Params, Route, Router, Routes, SsrMode,
};

#[component]
    provide_meta_context();

    view! {
        <Router>
            <Routes>
                <Route path="/" view=HomePage ssr=SsrMode::PartiallyBlocked />
            </Routes>
        </Router>
    }
}

#[derive(Clone, PartialEq, Params, Default, Debug)]
struct Query {
    q: String,
}

#[component]
pub fn HomePage() -> impl IntoView {
    let navigate = use_navigate();
    let q = use_query::<Query>();
    let value = move || {
        logging::log!("q: {:?}", q.get());
        q.get().unwrap_or_default().q
    };
    let on_input = move |event| {
        let v = event_target_value(&event);
        let params = params_map! {
            "q" => v,
        };

        navigate(
            &format!("/{}", params.to_query_string()),
            NavigateOptions {
                replace: true,
                ..Default::default()
            },
        );
    };

    view! {
        <h1>Home Page</h1>
        <input type="text" name="q" prop:value=value on:input=on_input />
    }
}
```

**Expected behavior**
Characters like `+` show up in the input because they were properly escaped into the query parameters

**Actual behavior**
Instead of a `+`, we get a space character

luxalpa:
`encode_uri` does not encode URL critical characters like `&` and `+`, for that `encode_uri_component` must be used.
This also revelaed some issues with too much unescaping in several cases. Fixes #2602

### DIFF
--- a/router/src/history/params.rs
+++ b/router/src/history/params.rs
@@ -30,8 +30,6 @@ impl ParamsMap {
     /// Inserts a value into the map.
     #[inline(always)]
     pub fn insert(&mut self, key: String, value: String) -> Option<String> {
-        use crate::history::url::unescape;
-        let value = unescape(&value);
         self.0.insert(key, value)
     }
 
--- a/router/src/history/url.rs
+++ b/router/src/history/url.rs
@@ -25,7 +25,7 @@ pub fn unescape(s: &str) -> String {
 
 #[cfg(not(feature = "ssr"))]
 pub fn unescape(s: &str) -> String {
-    js_sys::decode_uri(s).unwrap().into()
+    js_sys::decode_uri_component(s).unwrap().into()
 }
 
 #[cfg(feature = "ssr")]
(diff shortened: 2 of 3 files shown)


gbj:
Awesome! Thank you so much.

PR #2607, opened 2024-05-31
(draft) Leptos 0.7

gbj:
(Opening this so that I can start tracking test/docs/example coverage)

Things I'd like to achieve before a beta release
- [x] Clear all warnings (`todo!()` is okay)
- [ ] Complete docs
- [ ] All tests passing in core
- [x] Actix integration
- [x] All examples either a) ported, b) missing features listed in TODOs, or c) moved to `projects`

Incomplete list of feature TODOs
- [x] nonces
- [x] `.await` on resources: clone by default for `IntoFuture`, with alternate method to read them by reference
- [x] Setting ResponseOptions in integrations
- [x] local resources (resources that only run on the client, not server)
- [x] `<ProtectedRoute/>`
- [x] Render individual `Suspend(_)` without a `<Suspense/>`
- [ ] Nested router `set_is_routing`
- [x] Attribute spreading for attributes on `<Body/>` etc. instead of `Vec<AnyAttribute>`
- [x] Directives
- [ ] Equivalent to a user implementing `IntoView` on an arbitrary struct
- [ ] Figuring out the interaction between `Lazy` routes and hydration (what do you do while hydrating if the code for the route hasn't been lazy-loaded yet?)
- [x] `path!()` macro to make route definitions easier in router
- [ ] Reactive channel implementation for single-use signals
- [x] blocking resources
- [x] partially-blocked SSR
- [x] Thread-local arena that yields `Copy` but `!Send` for `!Send` data (like browser things)
- [x] global class support
- [ ] hot reloading (and therefore view markers)
- [ ] static routes
- [ ] trailing slash modes
- [ ] Router/Route base
- [x] custom elements/custom attributes
- [ ] fallible hydration => start from suspense boundary
- [x] add `GlobalAttributes` methods for `on____` HTML attributes

Forward-ported bug fixes (things that have been fixed on 0.6 but reimplemented in 0.7, and need to be checked)
- [ ] pending action dispatches (#2653)
- [ ] unescaping during navigation (#2651)
- [ ] ownership and disposal fixes and tests (#2639)
- [ ] ToChildren docs (#2643)
- [ ] #2621

Finalizing: 
- [ ] simplify tachys APIs by removing duplicated content (HTML length, checking on template duplication, etc.)


tachys/src/html/element/elements.rs
```diff
@@ -0,0 +1,488 @@
...
+        rndr: PhantomData,
+        attributes: (),
+        children: (),
+        #[cfg(debug_assertions)]
+        defined_at: std::panic::Location::caller(),
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Option_;
```
feathecutie, line 462: `Option_` does not implement `ElementWithChildren` but should, and also does not implement `<option>`'s typed attribute methods. This makes using `<option>` properly in 0.7.0-alpha impossible.

tachys/src/html/element/elements.rs
```diff
@@ -0,0 +1,488 @@
...
+
+impl ElementType for Option_ {
+    type Output = web_sys::HtmlOutputElement;
```
feathecutie, line 465: Also, `Option_` incorrectly uses `HtmlOutputElement` instead of `HtmlOptionElement`.

tachys/src/html/element/elements.rs
gbj: Thanks — I've fixed these two and will make a new `tachys` alpha release too.

NiklasEi:
Will there be follow-up tickets for tracking the open tasks from this PR? I am mainly interested in static routes, but some of the other open points certainly are big enough to also be tracked on their own.

gbj:
@NiklasEi Yes, it is just way easier to manage a single `main` branch once people start making PRs, and it was time! Everything else will be completed. I will probably just create a single tracking item that copy-pastes the list above.

PR #2613, opened 2024-06-03
feat: Add Compression to Hacker News w/ Islands Example

Th3Whit3Wolf:
One of the things I love about leptos is how easy it is to ship a small frontend. 

Leptos provides islands that reduce the amount of total HTML generated and it's build tool minifies CSS with `lightningcss`, soon will be minifying javascript with `swc`, and it can even precompress these static assets. That last of which I think may be the most important because web assembly compress fairly well and it allows the user to ship their entire app as 1 static binary (with musl). 

Since precompression was added as a feature in `cargo-leptos` I haven't seen it mentioned a whole lot so I wanted to help people see how to use it. I have also seen a couple question about setting up compression in general so I added that too.

I've updated the `hackernews_islands_axum` example to serve all static assets precompressed with `cargo-leptos` and to dynamically compress the HTML.

The type of compression sent will depend on what the client supports.

The resulting network usage

| Asset | No Compression | Compression | Difference |
| :---- | :------------- | :---------- | :--------- |
| HTML  | 16.25 kB       | 4.80 kB     | - 70.46%   |    
| CSS   | 3.84 kB        | 1.17 kB     | - 69.53%   |        
| JS    | 16.86 kB       | 4.63 kB     | - 72.53%   |         
| WASM  | 200.15 kB      | 69.93 kB    | - 65.06%   |       
| ICON  | 15.54 kB       | 3.29 kB     | - 78.83%   |
|       |                |             |            |
| **TOTAL** | **252.64 kB** | **83.82 kB** | **- 66.82%** |

### DIFF
--- a/examples/cargo-make/cargo-leptos-compress.toml
+++ b/examples/cargo-make/cargo-leptos-compress.toml
@@ -1,5 +1,13 @@
+extend = [
+    { path = "./lint.toml" }
+]
+
+[tasks.make-target-site-dir]
+command = "mkdir"
+args = ["-p", "target/site"]
+
 [tasks.install-cargo-leptos]
-install_crate = { crate_name = "cargo-leptos", binary = "cargo-leptos", test_arg = "--help" }
+install_crate = { crate_name = "cargo-leptos", install_command = "cargo install --git https://github.com/leptos-rs/cargo-leptos --locked cargo-leptos",binary = "cargo-leptos", test_arg = "--help" }
 
 [tasks.cargo-leptos-e2e]
 command = "cargo"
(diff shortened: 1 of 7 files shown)


Th3Whit3Wolf:
Fix lints for CI

Th3Whit3Wolf:
Looks like the issue is that when building it is trying to access `target/site` which doesn't exist until `cargo leptos build` is run. I have updated the makefile to account for this.

Th3Whit3Wolf:
I also add `hash_file` to the `Cargo.toml`. Since the compressed file are served with a `Cache-Control` policy `hash_file` is necessary to ensure that client receives the correct file when updates are made to an asset. I don't expect the example files to change much but for users copying & pasting this example to base their own app off of this will be helpful.

gbj:
Cool -- This looks great, thanks!

Issue #2618, opened 2024-06-09
`method 'dyn_bindings' is private` when spreading attrs on components 


Upbolt:
components in outside crates using https://github.com/leptos-rs/leptos/pull/2534 throw a `method 'dyn_bindings' is private` compile error

a fix pr should be pending shortly

PR #2619, opened 2024-06-09
fix: extract dyn_bindings impl into DynBindings trait

Upbolt:
solves #2618


gbj:
Thanks. Could you please run `cargo fmt` and update the PR?

PR #2621, opened 2024-06-09
fix: try_with should not panic on disposed resources (closes #2620)

### LINKED ISSUE #2620: `try_with` on disposed resource panics instead of returning `None`
**Describe the bug**
According to [documentation](https://docs.rs/leptos/latest/leptos/struct.Resource.html#method.try_with), calling `try_with` should return `None` if signal is no longer valid. Instead it calls `panic!`.

It calls `runtime.resource`:
https://github.com/leptos-rs/leptos/blob/2ef27cb0bb485b7ad33e0e3d5f144718504ad757/leptos_reactive/src/resource.rs#L801
which panics:
https://github.com/leptos-rs/leptos/blob/2ef27cb0bb485b7ad33e0e3d5f144718504ad757/leptos_reactive/src/runtime.rs#L1279-L1284

**Leptos Dependencies**

```toml
leptos = { version = "0.6.12", features = ["csr"] }
leptos_router = { version = "0.6.12"}
```

**Expected behavior**
`try_with` to return `None` for disposed resources.

**Additional context**
Attempted to implement resource caching as described [here](https://github.com/leptos-rs/leptos/discussions/1162#discussioncomment-6126800).

Navigating away from the page causes resources in the cache to be disposed,
after returning to the page, the cached (and now disposed) resources are unusable, with no way to test if cached resource needs to be replaced by new one.

The only user of `runtime.try_resource()` without `unwrap` is `try_update`, which cannot be used for testing whether resource is disposed, calling `try_update` breaks loading for non-disposed resources.

otopetrik:
With this change, the following code:
```
...
Some(existing_resource) => {
    let signal_alive = existing_resource.try_with(|r| {});
    match signal_alive {
        Some(()) => *existing_resource,
        None => {
            let r = self.create_resource(&id);
            self.resource_cache.insert(id, r);
            r
        },
    }
}
...
```
can be used to check whether the cached resource is still usable (in an implementation of resource cache as described in https://github.com/leptos-rs/leptos/discussions/1162).

### DIFF
--- a/leptos_reactive/src/resource.rs
+++ b/leptos_reactive/src/resource.rs
@@ -798,9 +798,11 @@ where
     fn try_with<O>(&self, f: impl FnOnce(&Option<T>) -> O) -> Option<O> {
         let location = std::panic::Location::caller();
         with_runtime(|runtime| {
-            runtime.resource(self.id, |resource: &ResourceState<S, T>| {
-                resource.with_maybe(f, location, self.id)
-            })
+            runtime
+                .try_resource(self.id, |resource: &ResourceState<S, T>| {
+                    resource.with_maybe(f, location, self.id)
+                })
+                .flatten()
         })
         .ok()
         .flatten()


gbj:
Thanks. Could you please run `cargo fmt` on the commit?

Issue #1284, opened 2023-07-05
Allow for a streaming value with #server.


MinaMatta98:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Currently, #server macro does not allow for a return of a stream type, due to the requirement to implement Serialize.

Take the following server_fn as an example:

```rust

#[server(GetImage, "/api", "Url")]
pub async fn get_image(
    cx: Scope,
    path: String,
) -> Result<
    Option<
        impl futures_util::Stream<Item = Result<Vec<u8>, std::io::Error>>
            + Unpin
            + Serialize
            + Deserialize<'static>,
    >,
    ServerFnError,
> {
    use tokio::io::AsyncReadExt;

    let mut path = path;
    path.remove(0);
    let path = std::env::current_dir().unwrap().join(path);

    if let Ok(mut file) = tokio::fs::File::open(path).await {
        let mut buffer = Vec::new();
        file.read_to_end(&mut buffer).await?;

        let stream = futures_util::stream::once(async move { Ok(buffer) });
        tokio::pin!(stream);

        Ok(Some(stream))
    } else {
        Ok(None)
    }
}

```
A solution would be to stream the output of the return value.

Let's say that I had a rather large Vec<_> to return, a stream return value would be very useful in this case.

gbj:
I'm open to implementing streaming server functions, but I've heard several different proposals from people imagining that they mean several different things, which makes me cautious. For example, some people assume that a server function that returns a `Stream` should be a websocket connection or server-sent events. In this case, it's clearly a streaming HTTP response with a clear beginning and end.

This kind of use case can currently be done by creating an API route in your server framework of choice that returns a streaming response. It could then be consumed with `reqwest` from the client using something like [`Response::bytes_stream()`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes_stream)

Integrating `Stream` into something like `<Suspense/>`, HTTP streaming, etc. is a more daunting task.

MinaMatta98:
Hello, 

Forgive the ambiguity. The request was for a stream to be consumed within a Suspense tag.

You don't even have to automate the consumption. You could implement a server fn that returns an impl Stream and allow the user to handle the sink internally within suspense.

gbj:
> The request was for a stream to be consumed within a Suspense tag.
> 
> You don't even have to automate the consumption. You could implement a server fn that returns an impl Stream and allow the user to handle the sink internally within suspense.

[...]

PR #2623, opened 2024-06-10
Server function streaming with serializable types

ealmloff:
This PR handles serialization and deserialization for custom types with streaming server functions (https://github.com/leptos-rs/leptos/issues/1284)

It adds a new `StreamingJson` encoding which serializes each item in the stream with `serde_json`. The implementation is a fairly straight forward combination of the streaming bytes and json codecs with two exceptions:
1) The content type is the default streaming type instead of JSON because if the chunks are read as merged chunks they cannot be parsed as valid json
2) When using a custom stream type with `StreamingJson`, this PR requires the bound `Stream<Item=T>` in order to make the `T` type used in the trait implementation. We don't use the bound, but it makes the `T` type constrained. Alternatively, I could add an unused generic to `StreamingJson` or `FromReq`, but I would imagine both of those would effect user code more

### DIFF
--- a/server_fn/src/codec/json.rs
+++ b/server_fn/src/codec/json.rs
@@ -1,12 +1,15 @@
-use super::{Encoding, FromReq, FromRes};
+use super::{Encoding, FromReq, FromRes, Streaming};
 use crate::{
-    error::ServerFnError,
+    error::{NoCustomError, ServerFnError},
     request::{ClientReq, Req},
     response::{ClientRes, Res},
     IntoReq, IntoRes,
 };
+use bytes::Bytes;
+use futures::{Stream, StreamExt};
 use http::Method;
 use serde::{de::DeserializeOwned, Serialize};
+use std::pin::Pin;
 /// Pass arguments and receive responses as JSON in the body of a `POST` request.
 pub struct Json;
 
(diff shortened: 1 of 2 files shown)


gbj:
This looks good to me. Thanks very much for the addition!

boblehest:
I don't think this is correct; it decodes each chunk read from the stream using `serde_json::from_slice`, but doesn't seem to do any framing to keep track of message boundaries. E.g. sending `1` immediately followed by `2` will usually be read as the integer `12`, and sending `1.0` immediately followed by `2.0` will usually fail when it tries to decode it as `1.02.0`.

ealmloff:
> I don't think this is correct; it decodes each chunk read from the stream using `serde_json::from_slice`, but doesn't seem to do any framing to keep track of message boundaries. E.g. sending `1` immediately followed by `2` will usually be read as the integer `12`, and sending `1.0` immediately followed by `2.0` will usually fail when it tries to decode it as `1.02.0`.

`StreamingJson` uses the http chunked encoding. HTTP streaming is chunked with a built in delimiter instead of a single uninterrupted stream of binary data. More info [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked)

boblehest:
> > I don't think this is correct; it decodes each chunk read from the stream using `serde_json::from_slice`, but doesn't seem to do any framing to keep track of message boundaries. E.g. sending `1` immediately followed by `2` will usually be read as the integer `12`, and sending `1.0` immediately followed by `2.0` will usually fail when it tries to decode it as `1.02.0`.
> 
> `StreamingJson` uses the http chunked encoding. HTTP streaming is chunked with a built in delimiter instead of a single uninterrupted stream of binary data. More info [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked)

Thank you for the explanation. Which part of the code is responsible for adding the `Transfer-Encoding` header? I tried using `StreamingJson` in my app but it fails in the way I described in my previous comment -- I might get around to providing a test case later this week.

ealmloff:
While the response itself is chunked correctly, the browser API that receives the response re-chunks the request into larger sections. When running natively with `reqwest`, this encoding works fine, but for browser compatibility I think we will need to switch to SSE

PR #2630, opened 2024-06-14
Simplify the example for using tailwindcss with leptos.

### LINKED ISSUE #109: Bug with components with element siblings in hydrate mode
This came up in Primeagen's stream but I couldn't reproduce; I think it may have been because it was in hydrate mode. 

Something like 

view! { cx,
  <div><ComponentA/><div><ComponentB/></div></div>
}

TheRealSalmon:
https://github.com/leptos-rs/book/issues/109

I felt that for a tutorial, the example was a little too complex. Here I've made it much more minimal and added more information to the README for using tailwindcss.


TheRealSalmon:
I was wrong about this whole thing. This example is already using the best solution, the direct integration of trunk with tailwindcss. I misunderstood.

PR #2639, opened 2024-06-20
fix: ensure everything is disposed of consistently

Giovanni-Tably:
Hi there!

This one popped up for us as a `BorrowMut` error on the `on_cleanups` secondary map, here's what I think was going on:
- `dispose_node` was removing nodeA from the reactive graph, but not running the cleanups.
- The slot in the `slotmap` was reused by nodeB, and nodeB used `on_cleanup`.
- The old cleanups were still there, so, when inserting the new cleanup, the old ones were dropped since the slot is the same but has a higher version.
- The old cleanups contained `Drop` impls that triggered some work, which were triggered [inside `slotmap`](https://github.com/orlp/slotmap/blob/c905b6ced490551476cb7c37778eb8128bdea7ba/src/sparse_secondary.rs#L291).
- If any of the triggered work used `on_cleanup`, we recurse and panic.

With this patch I tried to simplify and make it more obvious what cleaning up a node and disposing of it means, hopefully my understanding of what's going on here is correct.

The commits are meant to be easy to audit individually, and `refactor` commits are meant to match the original behaviour exactly.

And as always, thank you for `leptos`, it has been great!

### DIFF
--- a/leptos_reactive/src/node.rs
+++ b/leptos_reactive/src/node.rs
@@ -13,10 +13,7 @@ pub struct Disposer(pub(crate) NodeId);
 impl Drop for Disposer {
     fn drop(&mut self) {
         let id = self.0;
-        _ = with_runtime(|runtime| {
-            runtime.cleanup_node(id);
-            runtime.dispose_node(id);
-        });
+        _ = with_runtime(|runtime| runtime.dispose_node(id));
     }
 }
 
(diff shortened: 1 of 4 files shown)


gbj:
It would be very helpful to have a reproducible example of the issue(s) this is addressing, either here or in a linked issue.

gbj:
Thanks very much! The tests are great and will be helpful to make sure this correct behavior carries over into the next version, too.

PR #2643, opened 2024-06-21
docs: Add docs for `ToChildren`

spencewenski:
As discussed in https://github.com/leptos-rs/leptos/discussions/2640, the `ToChildren` trait is useful to consumers who want to use the builder syntax. However, because it is currently annotated with `#[docs(hidden)]`, it's not visible in docs and also not included in Jetbrains's auto-complete.

Add a doc comment for the `ToChildren` trait, including doc tests that demonstrate how to use the trait and how it compares to directly creating children.

### DIFF
--- a/leptos/src/children.rs
+++ b/leptos/src/children.rs
@@ -16,8 +16,62 @@ pub type ChildrenFnMut = Box<dyn FnMut() -> Fragment>;
 // This is to still support components that accept `Box<dyn Fn() -> Fragment>` as a children.
 type BoxedChildrenFn = Box<dyn Fn() -> Fragment>;
 
-#[doc(hidden)]
+/// This trait can be used when constructing a component that takes children without needing
+/// to know exactly what children type the component expects. This is used internally by the
+/// `view!` macro implementation, and can also be used explicitly when using the builder syntax.
+///
+/// # Examples
+///
+/// ## Without ToChildren
+///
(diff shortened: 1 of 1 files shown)


leptos/src/children.rs
```diff
@@ -16,8 +16,62 @@ pub type ChildrenFnMut = Box<dyn FnMut() -> Fragment>;
...
+///             .children(Box::new(|| p().child("Foo").into())),
+///         ShowProps::builder().children(Rc::new(|| p().child("Foo").into())),
+///     )
+/// }
+/// ```
+///
+/// ## With ToChildren
+///
+/// With [ToChildren], consumers don't need to know exactly which type a component uses for
+/// its children. This also removes the need to call `into()` on items in the children closure.
```
spencewenski, line 51: > This also removes the need to call `into()` on items in the children closure.

I thought this was true because I wasn't getting an error in my IDE inside this example block, but it seems this is not true, and I do get an error if I try this outside of an example block.

spencewenski:
Hey @gbj , I believe I resolved the issue that was causing the [checks to fail](https://github.com/leptos-rs/leptos/actions/runs/9620892003/job/26551599383). Can you re-run the workflow and review when you get a chance? Thanks!

benwis:
LGTM. Thanks for the PR

PR #2648, opened 2024-06-27
book_ru: SUMMARY.md

kakserpom:


benwis:
Thanks for clarifying these, it's so hard to know with languages we don't speak

Issue #2652, opened 2024-06-29
Inconsistent Action::pending() state


domwst:
TLDR;
When several calls of Action are pending, pending() is set to false on the first finished task.

While reading through the sources, I've found out this oddly looking line: https://github.com/leptos-rs/leptos/blob/c53fc67d382616f84b6d28b663a3d24fb56debf9/leptos_server/src/action.rs#L395
Which decreases pending count instead of increasing it. CSR example below showcases this bug:

```rs
#[component]
pub fn App() -> impl IntoView {
    let action = create_action(|v: &i32| {
        let v = *v;
        async move {
            gloo_timers::future::sleep(std::time::Duration::from_secs(4)).await;
            v + 1
        }
    });

    spawn_local(async move {
        action.dispatch(10);
        gloo_timers::future::sleep(std::time::Duration::from_secs(2)).await;
        action.dispatch(41);
    });

    view! {
        <div class="container">
            <h1>"Computation"</h1>
            <span>"Computation result: " {action.value()}</span>
            <span>
            { move ||
                if action.pending()() { "Loading..." } else { "" }
            }
            </span>
        </div>
    }
}
```

After the `action.value()` becomes 11, the "Loading..." text disappears and only after 2 seconds the `action.value()` becomes 42.

I've prepared [PR](https://github.com/leptos-rs/leptos/pull/2653), that fixes this.

**Leptos Dependencies**
N/A

**To Reproduce**
See above

**Expected behavior**
`action.pending()` to represent pending state of an `Action`

**Screenshots**
N/A

**Additional context**
N/A

PR #2653, opened 2024-06-29
Fix of #2652

domwst:


gbj:
Yes, this line is clearly wrong (and you can see the `.saturating_sub()` a few lines below.) Thanks for the PR!

PR #2655, opened 2024-06-29
Destructuring `let` (0.7)

Jinxit:
This PR adds a new `let()` syntax to components, while also keeping `let:` alive for backwards compatibility.

Example usage:
```rust
view! {
    <For
        each=move || data.get()
        key=|n| *n
        // stores the item in each row in a variable named `data`
        let(data)
    >
        <p>{data}</p>
    </For>
}
```

This also allows for destructuring of tuples and structs, as it's just directly embedded in the children function's arguments.
```rust
view! {
    <For
        each=move || data.get()
        key=|n| *n
        // stores the item in each row in a variable named `data`
        let((tup1, tup2)) // note the extra parentheses for destructuring
    >
        <p>{tup1} {tup2}</p>
    </For>
}
```

I haven't gone through all the examples and changed the usage as I wasn't sure what you wanted to present as the default. Would you like me to do that?

### DIFF
--- a/leptos_hot_reload/Cargo.toml
+++ b/leptos_hot_reload/Cargo.toml
@@ -20,7 +20,7 @@ syn = { version = "2", features = [
 	"printing",
 ] }
 quote = "1"
-rstml = "0.11.0"
+rstml = "0.11.2"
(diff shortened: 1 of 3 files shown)


benwis:
Lovely, thanks for the PR

PR #2658, opened 2024-06-30
fix: impl `Copy` for `Callback`

brofrain:
I suppose it's an oversight rather than an intentional change compared to v0.6?


gbj:
Thanks!

PR #2664, opened 2024-07-02
feat: Ensure services in axum `#[middleware()]` macro Implement Clone

thlsrms:
This change requires that services utilized within the `#[middleware()]` macro for server functions to implement `Clone`. This ensures that the service being wrapped by the middleware layer can also be cloned, useful for moving a clone of the service into a boxed future to be polled next:

```rust
fn call(&mut self, req: Request) -> Self::Future {
    let not_ready_inner = self.inner.clone();
    let ready_inner = std::mem::replace(&mut self.inner, not_ready_inner);

    let future = Box::pin(async move {
        /*
            Async operations
        */
        let inner = ServiceBuilder::new()
            .boxed_clone()
            .map_response(IntoResponse::into_response)
            .service(ready_inner);
        let next = Next { inner };
    };
    
    ResponseFuture {
        inner: future
    }
}
```
```rust
fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
    let inner = self.inner.clone();
    let authorization = Box::pin(async move {
        /* Query the DB and authorize this user */
    });

    ResponseFuture {
        authorization,
        inner,
    }
}
```

### DIFF
--- a/examples/server_fns_axum/src/middleware.rs
+++ b/examples/server_fns_axum/src/middleware.rs
@@ -18,6 +18,7 @@ impl<S> Layer<S> for LoggingLayer {
     }
 }
 
+#[derive(Clone)]
 pub struct LoggingService<T> {
     inner: T,
 }
--- a/server_fn/src/middleware/mod.rs
+++ b/server_fn/src/middleware/mod.rs
@@ -17,13 +17,48 @@ impl<Req, Res> BoxedService<Req, Res> {
     }
 }
 
+#[cfg(feature = "axum-no-default")]
+impl<Req, Res> Clone for BoxedService<Req, Res> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone_box())
+    }
+}
+
 /// A service converts an HTTP request into a response.
+#[cfg(not(feature = "axum-no-default"))]
 pub trait Service<Request, Response> {
     /// Converts a request into a response.
     fn run(
         &mut self,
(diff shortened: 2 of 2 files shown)


thlsrms:
> I just want to flag the difference between "it's useful to _allow_ services to be cloned" and "therefore we should (make a breaking change to) _require_ services to be `Clone`".
> 
That's fair, and badly worded on my part.

> I'm not particularly familiar with the Tower ecosystem beyond what I needed to know of the APIs to implement this a few months ago, so I am not sure how common/uncommon cloning services is.
> 
Cloning is necessary when transferring the inner service's ownership to the `Future`. This is needed when moving the request into an async block, which is often done to access or modify it asynchronously before passing it to the next service `call`, or when moving the inner service itself into the async block. Because of `&mut self` we end up with no other options.

That is until tower, hopefully some day, get blessed by the async trait.


> I wonder if it's possible to achieve the goal of allowing services to be cloned without the breaking change to require them to be cloneable.

I'm not sure if this is possible.
For a moment I thought that [`tower::ServiceBuider`](https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#method.boxed_clone) would be able to do that with `boxed_clone` but it requires the service to already be `Clone`

gbj:
I guess the alternative is to have `BoxedService` actually be an `Arc<Mutex<_>>` rather than a `Box<_>`.

Issue #2663, opened 2024-07-02
Static site sub-routes missing interactivity/not playing well with hydration


Tehnix:
**Describe the bug**
I've been exploring SSG with Leptos a bit, trying to document my steps along the way. I've got everything _mostly_ working, except for reloading a statically generated page.

One thing that's broken though seems to be interactivity on pages when you navigate directly to them or reload while on them.

My suspicion is that it has something to do with hydration and a mismatch between the element IDs that it's looking for to reestablish interactivity. Not too sure though 😅 

**Leptos Dependencies**

I've done some slight adjustments to the standard [start-axum](https://github.com/leptos-rs/start-axum) template, which you can find here https://github.com/Tehnix/comparison-ssg-rust-js/tree/main/leptos-example.

**To Reproduce**

1. Clone the example repo `git clone git@github.com:Tehnix/comparison-ssg-rust-js.git`
2. Go into the Leptos project `cd leptos-example`
3. Build the Leptos project `cargo leptos build --release`
4. Run the binary to generate the static routes `./target/release/leptos-example`
5. You should now have files in `./target/site`
6. Serve the files via e.g. `(cd target/site && bunx simplehttpserver)`
7. Visit a sub-page directly in the browser http://localhost:8000/another-page.html
8. Interactivity is broken because hydration can't find the elements/components

**Expected behavior**

I would expect sub-routes to hydrate correctly and support interactivity, such as using signals etc.

When navigating to the root route at http://localhost:8000 everything works, and the sub-routes also work when navigating to them directly from the links on the root page. Things break when we reload on any of the sub-routes where hydration seems to fail (at least that's my guess as to what causes the lack of interactivity).

**Screenshots**

Loading a sub-route directly, e.g. http://localhost:8000/another-page.html, failing hydration and not calling the WASM/JS:

<img width="953" alt="Screenshot 2024-07-02 at 16 54 37" src="https://github.com/leptos-rs/leptos/assets/1189998/d8000cb8-012c-4097-813a-8ce4a58d72d0">

Loading a sub-route by loading the root page at http://localhost:8000 and then navigating to it, with everything working (note the `328` indicating it can fetch the `window.innerHeight` value):

<img width="953" alt="Screenshot 2024-07-02 at 16 57 11" src="https://github.com/leptos-rs/leptos/assets/1189998/9e6ec19c-7ab1-4ae3-ad63-b4d2da1cc88e">

**Additional context**

Tehnix:
[...]

PR #2667, opened 2024-07-04
fix: Handle trailing slash in routes when building StaticRoute's

Tehnix:
In https://github.com/leptos-rs/leptos/issues/2663 I ran into an issue with hydration for sub-routes. While I was initially a bit off, after @gbj's comment I was able to fix the route I was generating but then ran into another issue with `build_static_routes` not supporting paths ending in a trailing slash (e.g. `another-page/`).

It would instead think these paths had no trailing slash, and when combined with the `TrailingSlash::Exact` setting, you would see output like this

```
building static route: /another-page
Errors: [
    NotFound,
]
building static route:
```

This PR does a couple of things:

- Changes `spawn_task` to return a handle that you can await on
  - This allows `build_static_routes` to return when its done generating the files
  - It then enables the user to do any file manipulation they need in their `main` function after calling `build_static_routes` (previously any attempt to do this would break since the files were generated async and the `.await` on `build_static_routes` wouldn't actually wait for the files to be generated)
- Adds `trailing_slash` to `StaticPath` to track `/` route endings
- Retains the trailing slash when turning the `StaticPath` into a `ResolvedStaticPath`, making sure the correct route gets built
- Changes the file output to be `<route>/index.html` for routes ending with a trailing slash

I've also extended the output a bit, which makes it easier for users to follow along in what route corresponds to what file, written after the current `building static route`:

```
...
building static route: /sub-page/
writing static route to: target/site/sub-page/index.html
building static route: /another-page/
writing static route to: target/site/another-page/index.html
building static route:
writing static route to: target/site/index.html
```

I'm rather new to Rust and especially new to the Leptos codebase, so if there's any improvements I can make I'd love to hear!

I've tested it locally, and it completely fixes https://github.com/leptos-rs/leptos/issues/2663 :)

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -221,18 +221,21 @@ pub async fn handle_server_fns(req: Request<Body>) -> impl IntoResponse {
 /// to panic so we define a macro to conditionally compile the correct code.
 macro_rules! spawn_task {
     ($block:expr) => {
-        cfg_if::cfg_if! {
-            if #[cfg(feature = "wasm")] {
-                spawn_local($block);
-            } else if #[cfg(feature = "default")] {
-                let pool_handle = get_leptos_pool();
-                pool_handle.spawn_pinned(move || { $block });
-            } else {
-                eprintln!("It appears you have set 'default-features = false' on 'leptos_axum', \
-                but are not using the 'wasm' feature. Either remove 'default-features = false' or, \
-                if you are running in a JS-hosted WASM server environment, add the 'wasm' feature.");
-                spawn_local($block);
+        {
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "wasm")] {
+                    let handle = spawn_local($block);
+                } else if #[cfg(feature = "default")] {
+                    let pool_handle = get_leptos_pool();
+                    let handle = pool_handle.spawn_pinned(move || { $block });
+                } else {
+                    eprintln!("It appears you have set 'default-features = false' on 'leptos_axum', \
+                    but are not using the 'wasm' feature. Either remove 'default-features = false' or, \
+                    if you are running in a JS-hosted WASM server environment, add the 'wasm' feature.");
+                    let handle = spawn_local($block);
+                }
             }
+            handle
         }
     };
 }
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1254,7 +1254,7 @@ where
     generate_route_list_with_exclusions_and_ssg(app_fn, excluded_routes).0
 }
 
-/// Builds and outputs each route to the target/site directory. This is useful for static site generation.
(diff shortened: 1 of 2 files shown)


integrations/axum/src/lib.rs
```diff
@@ -1275,6 +1278,7 @@ pub async fn build_static_routes<IV>(
         .await
         .expect("could not build static routes")
     });
+    let _ = handle.await;
```
gbj, line 1281: `leptos_reactive::spawn_local` (like the underlying [`wasm_bindgen_futures::spawn_local`](https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.spawn_local.html)) returns `()`, not a handle that can be awaited, so you can see the following error in CI when the `wasm` feature is enabled:
```
 error[E0277]: `()` is not a future
    --> integrations/axum/src/lib.rs:1281:20
     |
1281 |     let _ = handle.await;
     |                   -^^^^^
     |                   ||
     |                   |`()` is not a future
     |                   help: remove the `.await`
     |
     = help: the trait `futures::Future` is not implemented for `()`, which is required by `(): std::future::IntoFuture`
     = note: () must be a future or must implement `IntoFuture` to be awaited
     = note: required for `()` to implement `std::future::IntoFuture`
```

gbj (changes requested):
This mostly looks good to me but the changes to spawning cause it not to compile with the `wasm` feature.

integrations/axum/src/lib.rs
Tehnix: Hadn't seen that CI was finally running 🥳 

Hmm, while I've got rid of all the errors, I did end up with a bit of an iffy solution to it, essentially:

```
    #[allow(unused_variables, clippy::let_unit_value)]
    let handle = spawn_task!(async move {
        leptos_router::build_static_routes(
            &options,
            app_fn,
            &routes,
            &static_data_map,
        )
        .await
        .expect("could not build static routes")
    });
    #[cfg(all(not(feature = "wasm"), feature = "default"))]
    let _ = handle.await;
```

I'd be curious if you have any suggestions to alternatives? For now I've been trying to be minimally invasive, but that might be the wrong approach here 🤔

gbj:
I have since merged 0.7 into the main branch. 0.7 does not yet have static routing implemented. I will make sure the case this addresses is working once that is implemented.

Tehnix:
> I have since merged 0.7 into the main branch. 0.7 does not yet have static routing implemented. I will make sure the case this addresses is working once that is implemented.

Awesome, very excited for 0.7! 🥳 Let me know if you end up with wanting help on it (at the very least I can help test).

@gbj should I close this PR? There's still https://github.com/leptos-rs/leptos/issues/2663 to track the actual issue itself. (feel free to close it if you feel that's appropriate)

gbj:
This is handled by #2875 so should work as intended in the new version.

PR #2670, opened 2024-07-07
leptos 0.7: remove FromRef LeptosOptions in `LeptosRoutes` axum integration

b4-io:
Hi im trying out leptos for SSR template rendering and saw in the new version LeptosOptions was no longer needed for routes generation.

Maybe this line should be removed o maybe it was left there on purpose 😄 and this should just be closed


b4-io:
Hi @gbj could you apply this changes to the 0.7 branch so i don't need to use my fork in my project? :).

Thank you for the great work in this version

PR #2674, opened 2024-07-09
chore: temprorarily disable `hackernews-js-fetch` example

sabify:
The `axum-js-fetch` crate has not been successful in updating in the long run and prevents the `hackernews-js-fetch` example and CI from succeeding. Disabling it as part of the workflow helps manage PRs and merges.


gbj:
This is something that works on main and that I intend to update and have not yet updated. The reason the CI is red is that the PR isn't ready yet!

PR #2677, opened 2024-07-11
feat: added ARIA attributes

luoxiaozero:
Added some missing ARIA attributes, refer to https://www.w3.org/TR/wai-aria/#state_prop_def


gbj:
Thanks!

PR #2678, opened 2024-07-11
fix: `hackernews_js_fetch` example for `leptos_0.7`

sabify:


examples/hackernews_js_fetch/src/lib.rs
```diff
@@ -1,40 +1,67 @@
...
     provide_meta_context();
+    let (is_routing, set_is_routing) = signal(false);
+
     view! {
-        <Link rel="shortcut icon" type_="image/ico" href="/public/favicon.ico"/>
-        <Stylesheet id="leptos" href="/public/style.css"/>
-        <Meta name="description" content="Leptos implementation of a HackerNews demo."/>
-        <Router>
-            <Nav />
+        <head>
```
gbj, line 36: This extra `<head>` tag should be removed -- you create the `<head>` in the `shell` above.

gbj (changes requested):
Thanks. If you run it you'll notice the JS/WASM don't load. This is because `.leptos_routes()` needs to be updated so that it serves the `shell()`, not just the body.

```rust
                    let options = leptos_options.clone();
                    move || shell(options.clone())
                })
```

See also comment re duplicated `<head>`

examples/hackernews_js_fetch/src/lib.rs
sabify: Could you please confirm that it is working now?

gbj:
Thanks! This is great. Much appreciated.

PR #2682, opened 2024-07-12
add impl IntoStyle for Style

### LINKED ISSUE #2681: Implement Trait IntoStyle for Style
I want to express my appreciation for the Leptos project. The framework has significantly streamlined my web development workflow, and I am genuinely impressed with its capabilities and design. The documentation have been incredibly helpful, making it a pleasure to work with Leptos.

Currently, there is no straightforward way to store and pass `Style`\`s between components and apply them transparently in Leptos.

```rust
// Ok
leptos::html::div().style("foo", "bar"); 

// Compilation error, IntoStyle is not implemented for Style:
// the trait bound `leptos::leptos_dom::Style: leptos::IntoStyle` is not satisfied
leptos::html::div().style("foo", leptos::leptos_dom::Style::Value("bar".into()));
```

I propose adding an implementation of IntoStyle for Style, as follows:

```rust
impl IntoStyle for Style {
    fn into_style(self) -> Style {
        self
    }

    fn into_style_boxed(self: Box<Self>) -> Style {
        *self
    }
}
```

The only current method for passing set of styles involves merging them into a string via `as_value_string`, which overwrites existing styles and results in a loss of reactivity. I would like to pass a style using a container that is better suited for storing a set of styles than concatenated raw strings:
```rust
pub struct StyleSet(pub Vec<(Oco<'static, str>, Style)>);
```

An alternative approach is to add a method similar to ToHtmlElement::bindings():

```rust
impl<El: ElementDescriptor + 'static> HtmlElement<El> {
  /// Adds multiple styles (values or callbacks) to the element css styles.
  #[track_caller]
  pub fn with_styles<B: IntoStyle>>(
      mut self,
      styles: impl std::iter::IntoIterator<Item = B>,
  ) -> Self {
      for style in styles {
          self = self.style(style.into_style());
      }
      self
  }
}
```

Without implementing IntoStyle for Style, it is impossible to create any custom style container effectively.

alfatm:
This will close #2681

### DIFF
--- a/leptos_dom/src/macro_helpers/into_style.rs
+++ b/leptos_dom/src/macro_helpers/into_style.rs
@@ -46,6 +46,16 @@ pub trait IntoStyle {
     fn into_style_boxed(self: Box<Self>) -> Style;
 }
 
+impl IntoStyle for Style {
+    fn into_style(self) -> Style {
+        self
+    }
+
+    fn into_style_boxed(self: Box<Self>) -> Style {
+        *self
+    }
+}
+
 impl IntoStyle for &'static str {
     #[inline(always)]
     fn into_style(self) -> Style {


gbj:
Thanks!

PR #2685, opened 2024-07-13
allow using enums on Router StaticSegment

mahmoud-eltahawy:
i just wanted to suggest the idea of enums in router path


mahmoud-eltahawy:
the added example is a copy of router example but i changed routes to use enum instead

router/src/matching/horizontal/static_segment.rs
```diff
@@ -15,10 +15,20 @@ impl PossibleRouteMatch for () {
     fn generate_path(&self, _path: &mut Vec<PathSegment>) {}
 }
 
+pub trait Routable: Debug + Clone {
+    fn into_str(&self) -> &'static str;
```
gbj, line 19: I'd recommend `as_path()` as a name for this method: `as_` in line with the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) for a lightweight `&`-to-`&` conversion, and `as_path()` to give a little more context than `as_str()`.

router/src/matching/horizontal/static_segment.rs
```diff
@@ -15,10 +15,20 @@ impl PossibleRouteMatch for () {
     fn generate_path(&self, _path: &mut Vec<PathSegment>) {}
 }
 
+pub trait Routable: Debug + Clone {
```
gbj, line 18: For single-method traits, it's best to name it the same as the trait itself -- So given my proposal of `as_path`, I'd suggest `AsPath` for the trait name.

The `Debug + Clone` trait bounds are not necessary -- as long as they are implemented on your enum they do not need to be bounds on the trait.

gbj (changes requested):
Thanks! This looks good.

I would rather not include a separate example, as multiplying examples for specific features just increases the maintenance burden over time. Not sure if you have a good idea for documenting/publicizing this? Maybe a comment in the `router` example? Or it can be added to the future version of the book when it is updated for 0.7

gbj:
Looks good, thanks!

PR #2687, opened 2024-07-13
Support dynamic attribuutes in components

### LINKED ISSUE #2594: `HtmlElement::attrs` and `DynAttrs::dyn_attrs` Accept `&'static str` Instead of `Oco<'static, str>`
**Describe the bug**
The `leptos_dom::html::HtmlElement::attrs` method does not accept `Vec<(Oco<'static, str>, Attribute)>`.

However, it will accept `Vec<(&'static str, Attribute)>` and will convert `name` to `Oco<'static, str>` in the `HtmlElement::attr` method.

The same issue occurs with `leptos::DynAttrs::dyn_attrs`, which will only accept `Vec<(&'static str, Attribute)>`.

**Leptos Dependencies**

```toml
leptos = { version = "0.6", features = ["csr", "nightly"] }
```

**To Reproduce**

```rust
use leptos::*;

#[component]
pub fn MyComponent(
    #[prop(attrs, optional)] attrs: Vec<(Oco<'static, str>, Attribute)>,
) -> impl IntoView {
    html::span().attrs(attrs)
}
```

**Expected behavior**

The method should accept `Vec<(Oco<'static, str>, Attribute)>` as a type for dynamic attributes.

**Additional context**

I am working on Storybook integration for Leptos and parsing props for components from `JsValue`. Currently, it is impossible to use regular arguments, so I have to do something like this:

```rust
use leptos::*;
use wasm_bindgen::prelude::*;

// we have to create wrapper type for attributes
#[derive(Default)]
pub struct AttributeSpread(pub Vec<(Oco<'static, str>, Attribute)>);

impl From<Vec<(&'static str, leptos::Attribute)>> for AttributeSpread {
    fn from(value: Vec<(&'static str, leptos::Attribute)>) -> Self {
        Self(
            value
                .into_iter()
                .map(|(name, value)| (Oco::Borrowed(name), value))
                .collect(),
        )
    }
}

#[component]
pub fn MyComponent(#[prop(attrs, optional)] attrs: AttributeSpread) -> impl IntoView {
    let mut span = html::span();

    // have to add each attr manually
    for (name, value) in attrs.0 {
        span = span.attr(name, value);
    }

    span
}
```

alfatm:
# Overview
I propose an enhancement to allow dynamically generated sets of attributes to be passed to components. Currently, this is limited to predefined attributes because the Binding struct requires field names to be `&'static str`. This proposal aims to address this limitation by enabling components to accept dynamically generated attributes. To support dynamic strings, I suggest updating the Binding struct to use Oco<'static, str> for the name field.

```rust
pub enum Binding {
    /// A statically named attribute.
    Attribute {
        /// Name of the attribute.
        name: Oco<'static, str>,
        /// Value of the attribute, possibly reactive.
        value: Attribute,
    },
    /// A statically typed event handler.
    EventHandler(EventHandlerFn),
}
```

# Technical Details
Leptos components are implemented as structs with a builder macro. This means that the declared `attributes` field in the component is a struct field and it store a passed `Binding`s (i.e. `String`'s), However, it is not possible to store an `Oco::Owned(String)` inside `&'static str`. Therefore, the current attributes field, defined as `Vec<(&'static str, Attribute)>`, should be modified to `Vec<(Oco<'static, str>, Attribute)>` in every component across all projects. This modification will allow the storage of dynamic names from Binding within components. Updating the attribute field signature is essential and it is a breaking change.

```rust
#[component(transparent)]
pub fn FooBar(
	...

    attributes: Vec<(&'static str, Attribute)>,
) -> impl IntoView {
	...
}
```

replaced to 

```rust
#[component(transparent)]
pub fn FooBar(
	...

    attributes: Vec<(Oco<'static, str>, Attribute)>,
) -> impl IntoView {
	...
}
```

and will be expanded to 

```rust
pub struct FooBarProps {
	...

    pub attributes: Vec<(Oco<'static, str>, Attribute)>,
}
```

# Backward Compatibility
Supporting both new and previous syntax while retaining the old component definitions is feasible. However, this might cause potential panics when non-static strings are passed to components that only accept static string attributes. This can lead hidden errors and user frustration. Therefore, it is better to update all components than to introduce a design flaw.

# Usage
The proposed change will allow for the following:

```rust
#[component(transparent)]
pub fn MyComponent(name: String, value: String) -> impl IntoView {
    let attrs: Vec<(Oco<'static, str>, Attribute)> = vec![(name.into(), value.into_attribute())];
    view! {
        <div {..attrs}>foo bar</div>
    }
}

```

This enhancement provides the flexibility to dynamically generate attribute sets, making components more versatile and adaptable to varying use cases.

# More context

FooBar component will be expanded into this:

```rust
pub struct FooBarProps {
	...

    pub attributes: Vec<(Oco<'static, str>, Attribute)>,
}

...

impl leptos::DynAttrs for FooBarProps {
    fn dyn_attrs(
        mut self,
        v: Vec<(::leptos::Oco<'static, str>, ::leptos::Attribute)>,
    ) -> Self {
        self.attributes = v.into();
        self
    }
}

impl ::leptos::DynBindings for FooBarProps {
    fn dyn_bindings<B: Into<Binding>>(
        mut self,
        bindings: impl std::iter::IntoIterator<Item = B>,
    ) -> Self {
        for binding in bindings.into_iter() {
            let binding: Binding = binding.into();
            match binding {
                Binding::Attribute {
                    name,
                    value,
                } => self.attributes.push((name.into(), value)),
                _ => {}
            }
        }
        self
    }
}

...
```

I believe this enhancement will significantly improve the flexibility and functionality of attribute handling in components. 

This will close #2594

### DIFF
--- a/examples/spread/src/lib.rs
+++ b/examples/spread/src/lib.rs
@@ -8,7 +8,7 @@ pub fn SpreadingExample() -> impl IntoView {
     }
 
     let attrs_static_str: Vec<(&'static str, Attribute)> =
-        vec![("data-foo".into(), "42".into_attribute())];
+        vec![("data-foo", "42".into_attribute())];
 
     let attrs_string: Vec<(String, Attribute)> =
         vec![("data-foo".into(), "42".into_attribute())];
--- a/examples/spread/src/lib.rs
+++ b/examples/spread/src/lib.rs
@@ -7,7 +7,10 @@ pub fn SpreadingExample() -> impl IntoView {
         let _ = window().alert_with_message(msg.as_ref());
     }
 
-    let attrs_str: Vec<(&'static str, Attribute)> =
+    let attrs_static_str: Vec<(&'static str, Attribute)> =
+        vec![("data-foo".into(), "42".into_attribute())];
+
+    let attrs_string: Vec<(String, Attribute)> =
         vec![("data-foo".into(), "42".into_attribute())];
 
     let attrs_only: Vec<(Oco<'static, str>, Attribute)> =
@@ -35,8 +38,12 @@ pub fn SpreadingExample() -> impl IntoView {
         }))];
 
     view! {
-        <div {..attrs_str}>
-            "<div {..attrs_str} />"
+        <div {..attrs_static_str}>
+            "<div {..attrs_static_str} />"
+        </div>
+
+        <div {..attrs_string}>
+            "<div {..attrs_string} />"
         </div>
 
         <div {..attrs_only}>
@@ -55,24 +62,10 @@ pub fn SpreadingExample() -> impl IntoView {
             "<div {..partial_attrs} {..partial_event_handlers} />"
         </div>
 
-        <SpreadingAttrsComponent {..attrs_only} />
-
         // Overwriting an event handler, here on:click, will result in a panic in debug builds. In release builds, the initial handler is kept.
         // If spreading is used, prefer manually merging event handlers in the binding list instead.
         //<div {..mixed} on:click=|_e| { alert("I will never be seen..."); }>
         //    "with overwritten click handler"
         //</div>
-    }
-}
-
-#[component]
-pub fn SpreadingAttrsComponent(
-    #[prop(into)]
-    #[prop(optional)]
-    #[prop(attrs)]
-    attrs: Vec<(Oco<'static, str>, Attribute)>,
-) -> impl IntoView {
-    view! {
-        <div {..attrs}>attrs</div>
     }
 }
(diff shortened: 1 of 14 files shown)


gbj:
See [my comment on the linked issue](https://github.com/leptos-rs/leptos/issues/2594#issuecomment-2116042066):

> I don't think it's possible to change these function signatures in a way that isn't technically a breaking change, although I may be wrong. I do know that they will change significantly in the next release as part of the wider renderer rewrite, so I will mark this issue as 0.7 to make sure there's a decent solution then.
> 
> 0.7 does support spreading arbitrary attributes of any kind onto any component, which seems to me like it should help.

0.7 supports spreading arbitrary attributes onto components and elements anyway, so the `attributes` field and `Vec<_>` are no longer needed; I don't plan to introduce a separate release with breaking changes before 0.7.

PR #2694, opened 2024-07-17
feature/router_macro

boyswan:
This PR adds the `path!` router proc_macro.

As well as constructing an n-tuple of `StaticSegment | ParamSegment | StaticSegment`, the macro errors on a handful of invalid cases:

- deny consecutive_slashes  
example: `//`

- deny non_trailing_wildcard 
example:`/home/*any/end`

- deny invalid segment 
example: `/foo/^/"`
characters that are not `ascii_lowercase`, `asci_digit` or `-` are deemed invalid

### DIFF
--- a/router_macro/src/lib.rs
+++ b/router_macro/src/lib.rs
@@ -1,11 +1,7 @@
 use proc_macro::{TokenStream, TokenTree};
+use proc_macro2::Span;
+use proc_macro_error::abort;
 use quote::{quote, ToTokens};
-use std::borrow::Cow;
-use syn::{
-    parse::{Parse, ParseStream},
-    parse_macro_input,
-    token::Token,
-};
 
 #[proc_macro_error::proc_macro_error]
 #[proc_macro]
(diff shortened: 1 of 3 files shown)


router_macro/src/lib.rs
```diff
@@ -45,32 +41,102 @@ impl SegmentParser {
...
+            }
+        }
+    }
+}
+
+impl Segment {
+    fn is_valid(segment: &str) -> bool {
+        segment
+            .chars()
+            .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '-')
```
gbj, line 85: I'm wondering if this is slightly too restrictive — for example, I'd think `_` should be a valid character for static segments, at least, and possibly other [rarer but spec-compliant characters](https://stackoverflow.com/questions/4669692/valid-characters-for-directory-part-of-a-url-for-short-links).

gbj (commented):
Thanks, this looks great overall! See one comment about permitted characters.

router_macro/src/lib.rs
boyswan: Great point!

I've added the following permitted characters:
```rust
const RFC3986_UNRESERVED: [char; 4] = ['-', '.', '_', '~'];
const RFC3986_PCHAR_OTHER: [char; 1] = ['@'];
```

And also added uppercase ascii:
```rust
c.is_ascii_digit()
    || c.is_ascii_lowercase()
    || c.is_ascii_uppercase()
    || RFC3986_UNRESERVED.contains(&c)
    || RFC3986_PCHAR_OTHER.contains(&c)
```

Technically `RFC3986_PCHAR_OTHER` also permits `:`, however I've removed it due to potential clashes with param logic.

gbj: Thanks! This seems good to me. If someone really `:` in a static route they can always use the tuple syntax... This is just a convenience.

Issue #2043, opened 2023-11-18
Add ability to create an enumerated `<For />` component


jquesada2016:
**Is your feature request related to a problem? Please describe.**
There's currently no sound way of getting notified when a `<For />` child gets moved around from it's previous index.

**Describe the solution you'd like**

Since this feature adds a non-negligible amount of overhead to the `<For />` component, it doesn't make too much sense baking it in....especially since it would break the existing API. I propose adding a trait and helper state to make bookkeeping work.

**Describe alternatives you've considered**

API design is the place where this can mostly vary.

**Additional context**
Context is actually a great alternative place to put this API in....

PR #2695, opened 2024-07-18
feat: adds ForEnumerate component

luoxiaozero:
Implement the functionality described in #2043 issue.

Please feel free to rename the component name.


leptos/src/for_loop.rs
```diff
@@ -76,6 +81,83 @@ where
...
+///   count: RwSignal<i32>
+/// }
+///
+/// #[component]
+/// fn Counters() -> impl IntoView {
+///   let (counters, set_counters) = create_signal::<Vec<Counter>>(vec![]);
+///
+///   view! {
+///     <div>
+///       <For
```
gbj, line 106: Is this meant to be `ForEnumerate`?

leptos/src/for_loop.rs
```diff
@@ -76,6 +81,83 @@ where
...
+    each: IF,
+    /// A key function that will be applied to each item.
+    key: KF,
+    /// A function that takes the index and the item, and returns the view that will be displayed for each item.
+    children: EF,
+) -> impl IntoView
+where
+    IF: Fn() -> I + Send + 'static,
+    I: IntoIterator<Item = T> + Send + 'static,
+    EF: Fn(MaybeSignal<usize>, T) -> N + Send + Clone + 'static,
```
gbj, line 135: Why does it take `MaybeSignal`, when we know that it is always a signal? Why not just have it take a `ReadSignal`?

tachys/src/view/keyed/keyed_enumerate.rs
```diff
@@ -0,0 +1,445 @@
...
+    items: I,
+    key_fn: KF,
+    view_fn: VF,
+) -> KeyedEnumerate<T, I, K, KF, VF, V, Rndr>
+where
+    I: IntoIterator<Item = T>,
+    K: Eq + Hash + 'static,
+    KF: Fn(&T) -> K,
+    V: Render<Rndr>,
+    VF: Fn(usize, T) -> (ArcWriteSignal<usize>, V),
```
gbj, line 30: This will probably fail to compile unless the `reactive_graph` feature is enabled on this crate. One of the goals is to abstract away the reactive system, and `tachys` as a whole doesn't depend on `reactive_graph`. This either needs to be feature-gated, or it needs to abstract over the concept of "set a signal" to something like "call this function if the index has changed"

gbj (commented):
Thanks for working on this!

I wrote down a few specific comments.

Generally: This ends up with a lot of duplicated code. I am wondering if there is a way to refactor so that the two implementations can be merged again. I will probably try to do that in the future, unless you want to, now.

Overall, it looks great and thank you!

luoxiaozero:
How about changing the keyed function to this?

```rust
pub fn keyed<VF, VFS, V, T>(view_fn: VF)
where
    VF: Fn(usize, T) -> (VFS, V),
    VFS: Fn(usize),
```
This will enable the logic of `ForEnumerate` and `For` to be reused and also move the related logic of `reactive_graph` into the lepots library.

luoxiaozero:
I refactored it as described in my comments above.

gbj:
I have now merged the 0.7 branch into main, to help avoid the conflicts that have come from juggling main, 0.7, PRs to main, and PRs to 0.7. Could you please open a new PR against main?

PR #2699, opened 2024-07-20
<For/> empty items

tetuaoro:
### Description
This PR introduces a new `ifempty` prop to the `<For>` component, allowing you to render a specific view when the array of items is empty. This enhancement ensures better handling of empty states in your views without any breaking changes because the prop is optional and `default`.

### Example
Here’s an example of how to use the new ifempty prop in your component :

```rust
use leptos::*;

let (phones, _) = create_signal(Vec::<u32>::new());

let fifempty = || view! {
    <tr>
        <td colspan="2">"you don't have a girlfriend, dude!"</td>
    </tr>
};

view! {
    <table>
        <tbody>
            <For each=move || phones.get() key=|n| *n let:phone ifempty=fifempty>
                <tr>
                    <td>"Phone number"</td>
                    <td>{phone}</td>
                </tr>
            </For>
        </tbody>
    </table>
}
```

### Changes
- added `ifempty` prop to the <For> component
- when the array is empty, the view specified by ifempty is rendered.

### Benefits
- improved handling of empty states in lists
- more flexible and customizable views for empty data scenarios
- no breaking changes, as the `ifempty` prop is optional.


gbj:
Thanks! I like the idea, but I think this implementation is broken, for a few reasons:
1) It runs the `each` function an additional time, to convert into an iterator, to check if it's empty -- for something like a `Vec<_>`, this could be expensive, and it's a cost paid whether the user is user the fallback function or not!
2) Unless I'm mistaken, this only checks the initial state of the list, non-reactively; i.e., if you have 0 items, add 10 items, then clear it back down to 0 items, you'll get a) fallback, b) ten items, c) blank. I would think a) and c), when they both have 0 items, should show the same thing.

The logic for this will ultimately need to live in the inner implementation (in 0.1-0.6, that's `leptos_dom::Each`, in 0.7 it's the `Keyed` implementation in `tachys`).

tetuaoro:
Got it, thanks for the feedback! I'll rewrite the implementation and update the PR accordingly. Thanks again!

gbj:
I have since merged 0.7 into the main branch, and the history has gotten messed up here. If you'd like, you could open a new PR against main.

PR #2703, opened 2024-07-20
feat: reintroduce `queue_microtask`

CorvusPrudens:
After putting some effort into migrating [leptos-use](https://github.com/Synphonyte/leptos-use) to 0.7, I noticed there were a few items missing. One of these is [queue_microtask](https://docs.rs/leptos/latest/leptos/fn.queue_microtask.html).

Since this is a web API, I though it would be most appropriate to place it with the other helpers in `leptos_dom`, rather than in `reactive_graph` which is closer to its original location.

However, `leptos_dom` does not have access to the `hydrate` and `csr` features that the original code [uses to run the task outright outside of the browser](https://docs.rs/leptos_reactive/0.6.12/src/leptos_reactive/spawn_microtask.rs.html#9). I removed this functionality, **which would be a breaking change for anyone who depends on the code to run in all contexts**.

If there is a more appropriate place to put this code so that its behavior can be made identical, please let me know!


gbj:
Thanks! The breaking change is fine -- that it wasn't present already is because the whole library has moved over to async reactivity, which uses microtasks (implicitly, via wasm-bindgen-futures). Running it immediately in non-browser and as a microtask in the browser is weird behavior anyway, so this Should Be Fine.

PR #2711, opened 2024-07-24
fix(examples): `action-from-error-handling`, `counters`, `counters_isomorphic`, `stores`

sabify:


gbj:
@sabify Thanks! It looks like several of these are cases I actually want to fix: specifically the ones where you had to manually specify the types, I will fix by adding a `_local` method so that the `::new` or `::derive` correctly infers it's supposed to be using `SyncStorage`. I also force-pushed due to a conflict with main so the PR is out of whack now, my apologies.

Issue #2726, opened 2024-07-27
Bump MSRV to 1.8, so we can drop lazy_static


martinfrances107:
To me this  is minor housekeeping, but I wanted to generate an issue, so this may  inform those  who are not focused on the detailed flow of PR's. But  want changes to Architecture (  or the to tool/infrastructure ).

I want to reduce the attack surface by droping the crate "lazy_static" as the functionality has now been stabilized in std  ( as of last weeks release of rust 1.8) 

**This issue has the awkward aspect that we must bump the MSRV to rust 1.8.** 

I plan to generate two PR for this isuse.. one for changes to the library only and one for changes to the examples.

For reference  see the  "Stabilized API" section of the release notes here 
[ release notes](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-180-2016-04-14)

rakshith-ravi:
Thanks for the PR @martinfrances107! If I may add my two cents here - Would it make sense to add this for 0.8? Might give people some time to upgrade

rakshith-ravi:
Closing this since #2767 is merged. Feel free to reopen it in case I'm missing something

PR #2727, opened 2024-07-27
#2726 (Part1) Drop crate lazy_static, prefer std::sync::LazyLock

martinfrances107:
**MSRV change**

rust-version is now 1.80

This is just the smallest change possible the framework.. 

smallest patch easiest  to review. I want to do this as the implications of the MSRV change need carefull consideration.

There are lots of  bulky changes to the examples/project subdirectories which will need careful testing in themselves .. but that is part 2.

### DIFF
--- a/.github/workflows/ci-semver.yml
+++ b/.github/workflows/ci-semver.yml
@@ -15,7 +15,7 @@ jobs:
   test:
     needs: [get-leptos-changed]
     if: needs.get-leptos-changed.outputs.leptos_changed == 'true'
-    name: Run semver check (nightly-2024-04-14)
+    name: Run semver check (nightly-2024-07-26)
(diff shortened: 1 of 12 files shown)


martinfrances107:
Looking at the errors ( indentation errors).

```
error: doc list item without indentation
   --> leptos_reactive/src/signal_wrappers_read.rs:892:5
    |
892 | ///   without reactively tracking it.
```
These errors are  being fixed here .. https://github.com/leptos-rs/leptos/pull/2728

So guess this PR should be delayed until this lands

benwis:
@martinfrances107 If you want to patch this in the 0.7 branch too, that'd be appreciated :)

gbj:
Closing this as 0.7 is now the main branch.

PR #2729, opened 2024-07-27
Fix doc comment for `spawn_local` function

TheBestTvarynka:
Hi,
I'm using Leptos for my projects and it's great! I can say it's a wonderful framework.
I noticed a small mistake in the `spawn_local` function doc comment today. The `.read` method is deprecated:
image
I've changed it to the `.get` method.


benwis:
I believe this error message is actually incorrect now. We've created .read() and .write() for general Signals, but on Resources we used to have .read() before we switched to .get(). I think the right answer is to delete that warning

TheBestTvarynka:
@benwis, got it. Then I clone this PR
Thank you for the explanation and your time

PR #2731, opened 2024-07-27
Allowing effect functions without parameters for DX boost

maccesch:
It has long bothered me (much more than is reasonable) that you always had to provide a param to the effect functions while you almost never need it. This is in no small part due to the fact that on my keyboard layout typing the combination `|_|` is really awkward.

Now it's possible to do

```rust
Effect::new(move || { 
    // do sth
});
```

as well as

```rust
Effect::new(move |prev| {
    // do sth

    value
});
```

### DIFF
--- a/reactive_graph/src/effect.rs
+++ b/reactive_graph/src/effect.rs
@@ -2,7 +2,10 @@
 
 #[allow(clippy::module_inception)]
 mod effect;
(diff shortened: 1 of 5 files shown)


maccesch:
Also I removed the compile warnings when running the reactive_graph tests

PR #2732, opened 2024-07-28
Implemented Effect::watch

maccesch:
Needed so Leptos-Use can continue to port to 0.7

https://github.com/Synphonyte/leptos-use/issues/140


gbj:
Awesome! Thanks so much.

PR #2737, opened 2024-07-28
fix: add `leptos_router_macro::path` to public api [`0.7`]

chrisp60:
Seems like this is currently missing from the public API on v0.7.

I chose to re-export it through `leptos_router`, I am assuming it doesn't
have use outside of that crate. Please let me know if it is preferred to be
through the base `leptos` crate instead.

This macro was mentioned in the release docs for 0.7-beta


benwis:
LGTM!

PR #2744, opened 2024-07-30
fix: a more robust `aria-current="page"` algorithm for `A`

metatoaster:
With this change, anchors with href ending with a single trailing slash (e.g. `<A href="/item/">`) will now get the `aria-current="page"` marker set when navigating under its subpaths such as `/item/1`.  The fix being pushed here is a bit tricky. Enumerate the comparisons and only allow the path be empty beyond the first _two_ items, as it assumes the zeroth character for both `loc` and `path` are going to be a `'/'` (which will result in a comparison between two empty `str` as the zeroth pair), and then the first pair cannot be assumed to be valid as the `'/'` path should only match if the location is also `'/'`, being the root.  Only comparison 2 and beyond can have the path fragment be empty as that signifies a potential trailing slash after some fragment so unconditionally allow that to match.

The prior version I tested will bake the assumption of `'/'` being the first char in and just split whatever is after and do the comparison (only assume empty path fragment to be matching from comparison 1 and beyond).

```rust
std::iter::zip(loc[1..].split('/'), path[1..].split('/'))
    .enumerate()
    .all(|(c, (loc_p, path_p))| loc_p == path_p || path_p == "" && c > 0)
```

The following is alternative fix is also naive (but it lead to the one for this push request); it maintains the more simple `.all(...)` check while stripping out just a single trailing `'/'` char, and it also special cases the root path. It also is unfortunately more verbose, but no enumeration is needed in this case.

```rust
let path_split = if path != "/" && path.chars().last() == Some('/') { 
    path[0..path.len() - 1].split('/')
} else { 
    path.split('/')
};
std::iter::zip(loc.split('/'), path_split)
    .all(|(loc_p, path_p)| loc_p == path_p)
```

Ideally there should be some testing for this part, maybe this part should be a standalone function and that can at least be tested, but I am rather unsure about how to best organize this into the module.

### DIFF
--- a/router/src/link.rs
+++ b/router/src/link.rs
@@ -105,8 +105,7 @@ where
                         if exact {
                             loc == path
                         } else {
-                            std::iter::zip(loc.split('/'), path.split('/'))
-                                .all(|(loc_p, path_p)| loc_p == path_p)
+                            is_active_for(path, loc)
                         }
                     })
                 })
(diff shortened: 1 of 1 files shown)


metatoaster:
I have since then refactor that particular check to its own function and added a test module with a number of cases (some are exhaustive combinations) and actually found that multi-level matching never worked with this algorithm.  For instance, given `href="/tag/special"`, when location is at `/tag`, it's flagged as active, but clearly this is wrong because `/tag` lies above `/tag/special` (and `/tag/anythingelse`).  I'll think about this after I sleep on it and come up with a better solution.

metatoaster:
Well, I posted what mostly works now, except for the commented out `assert!`s as the ambiguity with the interpretations of trailing slashes rear their ugly heads as usual.  The edge case can be argued either way, but for the most common cases, i.e. where the non-slash version is just an alias to the slash version, is typically the expected pattern and so it should flag the combination as active. This leaves the less common interpretation of `/item/` being a resource with proper subitems while `/item` just simply browsing the flat `item` while still at `/`, as the user hasn't "initiate the descent" into the item tree (e.g. a certain filesystem tried to implement a directory can be read as a file, and this could have been one of the interpretation).

gbj:
I have now merged the 0.7 branch into main, to help avoid the conflicts that have come from juggling main, 0.7, PRs to main, and PRs to 0.7. Could you please open a new PR against main?

PR #2745, opened 2024-07-30
 #2726 Bump MSRV to 1.8, so we can drop lazy_static (leptos 0.7 ) 

martinfrances107:
As state in the acompaning issue #2726 the drive of this issue is to reduce the attach surface 
by removing a dependecy.

It been suggested here 

https://github.com/leptos-rs/leptos/pull/2727#issuecomment-2258746196

That I generate a PR for 0.7 

In terms of review , the follwing change is needed to pick up a nighlty build .. consistent with the new rust version 1.8

- rust-toolchain: nightly-2024-04-14
+ rust-toolchain: nightly-2024-07-26

the jump to the new branch, means I had to reimplement the idea behind the patch .. not just copy and paste. .. so there manybe some unexpected build errors.


martinfrances107:
When I visually inspect that last auto matic merge .. I see errors 

The best next step is the close .. and reupload a clean PR

Sorry for the noise.

PR #2751, opened 2024-07-31
chore: update gloo-net and reqwest to http 1.0 (closes #2688) (leptos 0.6)

### LINKED ISSUE #2688: server_fn depedency on gloo_net 0.5 causing duplicate inclusion of http library with different versions.
Hey all, this isn't exactly a bug, but I was trying to reduce the number of duplicate packages in my project with different versions in an attempt to cut down on build times and binary sizes. I've noticed that `server_fns` depends on `gloo_net` v0.5, which in turn depends on http v0.2 when the rest of the ecosystem has moved onto http v1. We should bump gloo_net to a more recent version eg. 0.9 to help reduce the inclusion of http v0.2.

sabify:


rakshith-ravi (approved):

gbj:
Thanks!

Issue #2196, opened 2024-01-18
0.7 attribute values


blorbb:
**Is your feature request related to a problem? Please describe.**
How attribute values work is slightly inconsistent and rust-analyzer support isn't the best

**Describe the solution you'd like**
This was discussed briefly on discord:

**1. Require braces for attribute values** except for str/num/bool literals.

Right now you can do `attr=arbitrary.expression() + thing` - but this requires that the expression after the `=` is valid - if you write something like `arbitrary.`, the macro won't expand and you don't get any autocomplete suggestions after the `.`.

This change would require any non-literal expressions to be wrapped in braces, like `attr={arbitrary.expression() + thing}`. This makes it easier to parse (just take everything inside the `{}`), providing r-a support even in incomplete expressions.

**2. More consistent value shorthands**

On HTML attributes, the attribute `checked` is equivalent to `checked=true`, while on leptos components, it is `checked={checked}`.
Make this consistent by making `checked` always equivalent to `checked=true`, and `{checked}` equivalent to `checked={checked}`. 

Both of these changes are also what other frameworks like Svelte does.

**Describe alternatives you've considered**

staying with the current macro.

**Additional context**

My own view macro [leptos-mview](https://github.com/blorbb/leptos-mview) implements these requirements, and does have significantly better r-a support. Feel free to steal parts of my implementation for this change.

jquesada2016:
I like this!

JonahLund:
Yes, I would love a stricter/more consistent view macro aswell

benwis:
I could see this in 0.7 if someone wants to PR it and we can get a migration aid.

blorbb:
ill have a go at implementing this

rakshith-ravi:
@blorbb just cleaning up some older issues, and wanted to check - has this been merged into main? Can this issue be closed?

blorbb:
@rakshith-ravi the first point has, I'm not sure what the status of the second point is in 0.7

gbj:
Point 1 is resolved, I disagree with point 2.

`{checked}` probably would have been a better syntax for components in the first place, but I think making breaking syntax changes is a losing proposition (you will always annoy at least as many people as you please) unless there is a particularly compelling reason (i.e., there's something that can't be achieved with syntax A and can only be achieved with syntax B).

PR #2753, opened 2024-08-01
rstml 0.12 and enforce braces in attribute values

blorbb:
This is an implementation of the first half of #2196 

Invalid values were added in rstml 0.12.0 (related: https://github.com/rs-tml/rstml/issues/54) which allows incomplete expressions to be expanded. Alongside this implementation, rust-analyzer support is significantly improved:

image

Even with an invalid expression, rust-analyzer can provide autocomplete suggestions.

Note: rstml 0.12 adds a generic to a few structs (`NodeElement`, `Node`). I'm not actually sure what this generic is for, I just added what the compiler told me to add. Testing the changes to `leptos_hot_reload` may be good.

Minor issue: for some reason, adding `#[allow(unused_brace)]` directly at the expression expansion would break rust-analyzer and it wouldn't provide suggestions. `#[allow(unused_braces)]` has been added to the entire `view!` expansion, which allows all unused braces, so some legitimate warnings might disappear.

### DIFF
--- a/benchmarks/src/ssr.rs
+++ b/benchmarks/src/ssr.rs
@@ -11,20 +11,24 @@ fn leptos_ssr_bench(b: &mut Bencher) {
 				let (value, set_value) = create_signal(initial);
 				view! {
 					<div>
-						<button on:click=move |_| set_value.update(|value| *value -= 1)>"-1"</button>
+						<button on:click={move |_| {
+							set_value.update(|value| *value -= 1)
+						}}>"-1"</button>
 						<span>"Value: " {move || value().to_string()} "!"</span>
-						<button on:click=move |_| set_value.update(|value| *value += 1)>"+1"</button>
+						<button on:click={move |_| {
+							set_value.update(|value| *value += 1)
+						}}>"+1"</button>
 					</div>
 				}
 			}
 
-			let rendered = view! { 
+			let rendered = view! {
 				<main>
 					<h1>"Welcome to our benchmark page."</h1>
 					<p>"Here's some introductory text."</p>
-					<Counter initial=1/>
-					<Counter initial=2/>
-					<Counter initial=3/>
+					<Counter initial=1 />
+					<Counter initial=2 />
+					<Counter initial=3 />
 				</main>
 			}.into_view().render_to_string();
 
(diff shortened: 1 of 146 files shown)


benwis:
I'm excited, looks like you have a couple conflicts. Once those are resolved, is it ready for testing?

blorbb:
Oh this PR was mergeable before and had 1 commit, I think @gbj your force push broke some things. Is it possible to revert that? Otherwise I can just fix it up later, looks like I can just accept everything on the upstream side and it should work fine.

blorbb:
nvm fixed it up (thanks to some friends who know git way better than me), should be ready now @benwis

gbj:
Sorry @blorbb I suspect it is at least in part on my own limited git skills — when I rebase `leptos_0.7` onto `main` and have conflicts, after resolving I tend to force-push but of course that screws up PRs that are onto `leptos_0.7` themselves. I'm trying to finish things off in the next few days so it can merge into main and we can be in a more normal dev cycle at least.

This looks good to me in general, and the compile error seems helpful at least. Maybe worth adding a note about how to migrate automatically with leptosfmt? 

I assume that all of the examples and many doctests will be broken when this is merged, because they aren't currently using braces for attribute values — Does that sound accurate? 

If so one of us should
1) run `leptosfmt` on all the examples as a starting point (and restore any deleted comments...)
2) run `leptosfmt` to add braces

blorbb:
Oh yeah I just fixed up the braces rust told me about, totally forgot about examples! Does leptosfmt work on examples and docs?

Adding a comment explaining how to migrate might be quite a lot of noise, could you perhaps create a migration guide on GitHub/leptos.dev so that the note can link to a website(+heading) instead of a full explanation? It doesn't seem like there's a one-line command to configure leptosfmt to use the "AlwaysUnlessLit" brace style.

blorbb:
I ran leptosfmt on the examples and tried to fix up all the comments. Also manually added braces to the docs (many doctests are still broken due to other reasons, I haven't fixed other issues in the commit).

I went through most of the diff but I'm not 100% sure that I fixed every comment changed by leptosfmt, you may want to look through it again to double-check

gbj:
I have now merged the 0.7 branch into main, to help avoid the conflicts that have come from juggling main, 0.7, PRs to main, and PRs to 0.7. Could you please open a new PR against main? This will also allow CI to run, which will catch missing cases.

PR #2761, opened 2024-08-04
Provide deprecated `watch` function

CorvusPrudens:
This PR provides a deprecated `watch` function, similar to `create_effect`. This should ease migration.

In order for `watch` to maintain the same signature as 0.6, `Effect` must be at least `Clone`. I provided both `Clone` _and_ `Copy`. Since the underlying type is always `Copy`, I think it makes sense to allow easy ergonomic access to `Effect::stop`, even if this may result in `stop` being called more than once. As noted [in this message](https://discord.com/channels/1031524867910148188/1263897766929764363/1269441434222592000), multiple calls to `stop` aren't a correctness problem; they're simply redundant.

I also provided a single test to ensure it runs properly. Since `watch` is a very light wrapper around `Effect::watch`, this can be argued to be redundant.


gbj:
Thanks!

PR #2770, opened 2024-08-05
fix: a more robust `aria-current="page"` algorithm for `A`

metatoaster:
This started with the fix to setting of `aria-current="page"` for `href` that ends with a single trailing slash, such as `<A href="/item/">`, during navigation to it, and further testing found additional problems with the initial algorithm which was introduced in #1656 (which was a much appreciated starting point).  Work was done to refactor the `std::iter::zip` call into its own function and extended it to support the initial desired use-case, along with multi-level matching, and includes a decent number of test cases with a fairly exhaustive set of href/location combination (which lead to the discovery of the issue with multi-level matching).  There are some edge-case tests also, though given the nature of that less common interpretation I have not include the ability to allow the configuration of that for the helper function.  I imagine if this is needed a future change can just be a boolean flag to determine the match condition when the final `href_f` fragment is `Some("")` - further details on how this all works are left as comments in the relevant sections of this pull request.

This pull request was re-submitted as per discussion at #2744.

Edit: I might actually go ahead and implement this as a flag with this pull request if needed, since shortly after I wrote that I saw the path on how to actually do this (and came up with the name to the flag).  In any case this additional change will be a separate commit so if it is required to be part of a separate pull request we can have this done easily.

### DIFF
--- a/router/src/link.rs
+++ b/router/src/link.rs
@@ -158,7 +158,7 @@ fn is_active_for(
     // with the loc fragment on an emtpy href fragment for non root related parts.
     std::iter::zip(location.split('/'), href_f.by_ref())
         .enumerate()
-        .all(|(c, (loc_p, href_p))| loc_p == href_p || href_p == "" && c > 1)
+        .all(|(c, (loc_p, href_p))| loc_p == href_p || href_p.is_empty() && c > 1)
         && match href_f.next() {
             // when no href fragments remain, location is definitely somewhere nested inside href
             None => true,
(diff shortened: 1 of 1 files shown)


metatoaster:
Went ahead and did it, but turns out there can be another scenario where only the immediate parent should be highlighted, and now there's way more possible combination which will complicate things if pursued further, so I am taking a break from this problem with these two changesets.  Please let me know if I should split the second change into a separate pull request to be considered in another time, otherwise I am fine with them being merged as is.

metatoaster:
Both commits should pass the `cargo fmt` step now.

gbj:
Thanks! Failing clippy (see its suggestion, and I'd always suggest using clippy in your editor with rust-analyzer) but the test cases look good to me so presumably if it's passing these I will merge.

PR #2772, opened 2024-08-05
feat: `custom_view` macro to allow implementing view types on arbitrary data

gbj:
In 0.1-0.6, `IntoView` is basically `Into<View>`: `View` is an actual type that is a big enum, and all other types are erased. (With binary size and performance drawbacks). In 0.7, `IntoView` is a trait, consisted of composing some other traits from the renderer `(Render + RenderHtml + AddAnyAttr)` that you should never really need to deal with yourself.

This makes the pattern of implementing `IntoView` for a custom struct quite cumbersome though, with a bunch of boilerplate.

It turns out it's almost impossible to do this via the trait system, due to Rust's relatively conversative rules. So here's my (implemented) proposal:

```rust
struct Foo<T>(T);

#[custom_view]
impl<T> CustomView for Foo<T>
where
    T: ToString + Send + 'static,
{
    // this should usually just be AnyView<Rndr>
    // for simple types, you may be easily able to specify the actual type
    type View = String;

    fn into_view(self) -> Self::View {
        self.0.to_string().to_ascii_uppercase()
    }
}
```
where `CustomView` is not an actual trait but a macro convention: this expands to implementing the `Render`, `RenderHtml`, and `AddAnyAttr` traits appropriately so you can actually use it.

I'm very open to feedback on this one. It's definitely more complex than doing the same thing in 0.1-0.6, but I don't think that complexity is avoidable: previously, we basically pay the price *everywhere* for being able to do this simply.

### DIFF
--- a/leptos_macro/src/custom_view.rs
+++ b/leptos_macro/src/custom_view.rs
@@ -1,0 +1,166 @@
+use proc_macro::TokenStream;
+use quote::{quote, ToTokens};
+use syn::{parse::Parse, parse_macro_input, ImplItem, ItemImpl};
+
+pub fn custom_view_impl(tokens: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(tokens as CustomViewMacroInput);
+    input.into_token_stream().into()
+}
+
+#[derive(Debug)]
+struct CustomViewMacroInput {
+    impl_block: ItemImpl,
+}
+
+impl Parse for CustomViewMacroInput {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let impl_block = input.parse()?;
+        Ok(CustomViewMacroInput { impl_block })
+    }
+}
+
+impl ToTokens for CustomViewMacroInput {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let ItemImpl {
+            impl_token,
+            generics,
+            self_ty,
+            items,
+            ..
+        } = &self.impl_block;
+        let impl_span = &impl_token;
+        let view_ty = items
+            .iter()
+            .find_map(|item| match item {
+                ImplItem::Type(ty) => (ty.ident == "View").then_some(&ty.ty),
+                _ => None,
+            })
+            .unwrap_or_else(|| {
+                proc_macro_error::abort!(
+                    impl_span,
+                    "You must include `type View = ...;` to specify the type. \
+                     In most cases, this will be `type View = AnyView<Rndr>;"
+                )
+            });
+
+        let view_fn = items
+            .iter()
+            .find_map(|item| match item {
+                ImplItem::Fn(f) => {
+                    (f.sig.ident == "into_view").then_some(&f.block)
+                }
+                _ => None,
+            })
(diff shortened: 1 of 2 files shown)


dustypomerleau:
Could a derive macro address everything burdensome in the trait bounds, but still allow custom implementations of `IntoView`?: 

```rust
pub trait Viewable: RenderHtml<Dom> + Render<Dom> + AddAnyAttr {}

pub trait IntoView
where
    Self: Viewable + Sized + Send,
{
    fn into_view(self) -> View<Self>;
}

#[derive(Viewable)]
pub struct Foo<T>(T);

impl IntoView for Foo<T> {
    type View = String;

    fn into_view(self) -> Self::View {
        //...
    }
}
```

If that isn't possible, maybe custom structs could provide conversion to an existing concrete type that already meets the trait bounds? For example, it looks like you've already put a lot of effort into implementations of `Render<Rndr>` and `RenderHtml<Rndr>` for `HtmlElement`. I can envision a path to `IntoView` that transforms the custom struct into an `HtmlElement` (maybe by implementing a one-method `IntoHtmlElement` trait), and lets Leptos pick up from there.

gbj:
I think I've tried every possible trait-based solution here, and generally rustc sees some conflict. 

> maybe custom structs could provide conversion to an existing concrete type that already meets the trait bounds

This is exactly the point here: allowing custom structs to provide a way of converting them to a concrete type that meets the trait bounds.

Implementing `IntoView` for a concrete type isn't the issue. The problem is that we cannot do a blanket `impl<T> Render for T where T: IntoView` — the compiler will not allow it.

Maybe I'm misunderstanding what you're suggesting above.

dustypomerleau:
Regarding trait-based solutions, I trust your actual experience more than my theoretical approach - sounds like a blanket impl is off the table.

What I was suggesting with the latter comment was that you could expose the fields of `HtmlElement` in such a way that people could easily implement `From<MyType> for HtmlElement`, and basically just sidestep any implementation of `IntoView` on the custom struct by calling `into()` whenever they want to use it in a view.

Baptistemontan:
I think what most people do is to have a way to compose `IntoView` elements, so I think somethink like 

```rust
pub trait Viewable: RenderHtml<Dom> + Render<Dom> + AddAnyAttr {
    fn into_view(self) -> View<Self>;
}

pub trait IntoView {
    type View: Viewable;

    fn into_view(self) -> Self::View;
}


impl<T: Viewable> IntoView for T {
    type View = Self;
    fn into_view(self) -> Self::View {
        self
    }
}

struct Foo(u8, String);

impl IntoView for Foo {
    type View = (&'static str, u8, &'static str, String);
    fn into_view(self) -> Self::View {
        ("id:", self.0, "message:", self.1)
    }
}

enum Bar {
    A(u8)
    B(String)
}

impl IntoView for Bar {
    type View = Either<u8, String>
    fn into_view(self) -> Self::View {
        match self {
            Bar::A(id) => Either::Left(id),
            Bar::B(message) => Either::Right(message)
       }
}
```

This will add one layer for a lot of implementations, for tuples and `Either` for examples, but would be very easy to implement on the user side

Issue #2766, opened 2024-08-04
`<img />` has missing attributes


CorvusPrudens:
I noticed that `<img />` is missing the `loading` attribute when porting over some of my code. Looking at the set of all attributes in the macro, it looks like it's actually missing several:

https://github.com/leptos-rs/leptos/blob/1f4c410f7818c9f81317cfacadff7a1ef28f6ced/tachys/src/html/element/elements.rs#L253

[Here's the MDN docs for reference.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes)

Should all of the missing attributes be added?

gbj:
Yes! PRs adding any and all missing attributes are welcome, it should be trivial to add them to the `elements.rs` file unless they have not even been defined as attributes yet ([here](https://github.com/leptos-rs/leptos/blob/main/tachys/src/html/attribute/key.rs)), which this one is
https://github.com/leptos-rs/leptos/blob/1f4c410f7818c9f81317cfacadff7a1ef28f6ced/tachys/src/html/attribute/key.rs#L317-L318

PR #2774, opened 2024-08-05
Complete `img` and `video` attributes set

CorvusPrudens:
Addresses #2766.

This rounds out all attributes for `img` and `video` as defined [here for `img`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes) and [here for `video`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attributes).


gbj:
Thanks!

PR #2749, opened 2024-07-31
 Allowing effect functions without parameters for DX boost 

maccesch:
Same as #2731 but with conflicts resolved

PR #2776, opened 2024-08-05
 Allow effect functions without parameters for DX boost

maccesch:
Third time's the charm 😀

This is the same as #2749 and #2731.


gbj:
This is great! Thanks so much, good improvement.

PR #2779, opened 2024-08-05
Cleanup some examples

mondeja:
- Remove some redundant imports (they were giving me too much output).
- Fix `timer` example. The counters were not being shown because the `leptos/csr` feature was missing.
- Remove `csr` feature from `action_form_error_handling` as needs a server function to work.
- Remove uneeded `csr` feature from `todomvc` example.


gbj:
Thanks! I have not read through all of your changes, but wanted to point out a misunderstanding of `todo_app_sqlite_csr`. If you run it (using `cargo leptos serve` as the README says) you will see that it does in fact work; it is an example of how to integrate a client-side-rendered UI with an Axum server that uses server functions (running on that Axum server) but does not server-side render the UI. You can see the HTML shell it uses for CSR here: https://github.com/leptos-rs/leptos/blob/1f4c410f7818c9f81317cfacadff7a1ef28f6ced/examples/todo_app_sqlite_csr/src/fallback.rs#L26-L37

If you restore that one I'll be happy to review the other changes. And thank you for the clean-up work!

mondeja:
Ah, my bad. Re-added it!

gbj:
Thanks — all looks good and much appreciated!

PR #2780, opened 2024-08-05
Allow changing SSR meta tags outside components

mondeja:
This is just a really naive proposal and I'm not sure about the implications of the change. Any help will be appreciated.

My problem is that I need to change `<html>` attributes on SSR before being rendered to the DOM. I know that I can use the `<Html/>` component, something like:

```rust
view! {
    <Html lang="es" dir="auto"/>
}
```

But that forces me to be inside a `view!` or `-> impl IntoView`. I don't have access to the view.

With the change included in this PR, I can do this:

```rust
provide_meta_context();

let meta = leptos::prelude::expect_context::<leptos_meta::ServerMetaContext>();
_ = meta.html.send(format!(" lang=\"es\" dir=\"auto\""));
```

Which sucks but works.

There is an alternative? Is the updating of meta tags forbidden outside the view of a component for some reason or the API has not been built ever? Thanks in advance.

### DIFF
--- a/meta/src/lib.rs
+++ b/meta/src/lib.rs
@@ -167,14 +167,14 @@ impl Default for MetaContext {
 #[derive(Clone, Debug)]
 pub struct ServerMetaContext {
     /// Metadata associated with the `<title>` element.
-    pub(crate) title: TitleContext,
+    pub title: TitleContext,
     /// Attributes for the `<html>` element.
-    pub(crate) html: Sender<String>,
+    pub html: Sender<String>,
     /// Attributes for the `<body>` element.
-    pub(crate) body: Sender<String>,
+    pub body: Sender<String>,
     /// Arbitrary elements to be added to the `<head>` as HTML.
     #[allow(unused)] // used in SSR
-    pub(crate) elements: Sender<String>,
+    pub elements: Sender<String>,
 }
 
 /// Allows you to access `<head>` content that was inserted via [`ServerMetaContext`].


gbj:
Could you clarify: In what sense does using the `Html` component require you to be "inside a `view!` or `-> impl IntoView`"? Or do you just mean that you are doing it in some code that will never be rendered? Can you not just update the `<html>` tag in the shell (which is why this is exposed more directly in the 0.7 approach)?

I am very hesitant to make fields pub in this way because it means that *any* change to the struct is a breaking change for the library.

mondeja:
> Or do you just mean that you are doing it in some code that will never be rendered?

That's it. The crate `leptos-fluent` only exposes a macro that expands before any `view!`. Like:

```rust
#[component]
fn App() -> impl IntoView {
    leptos_fluent! {{
        // ... all the config
    }};

    view! {
        ...
    }
}
```

So I can't access the view and I don't want to split the configuration of i18n between components and macro expansion. Currently is exposing a `SsrHtmlTag` component to solve this problem, but I want to get rid of it:

```rust
#[component(transparent)]
pub fn SsrHtmlTag() -> impl IntoView {
    let lang = expect_i18n().language.get_untracked();
    view! { <Html lang=lang.id.to_string() dir=lang.dir.as_str()/> }
}
```

> I am very hesitant to make fields pub in this way because it means that any change to the struct is a breaking change for the library.

Is just a naive example. What about a builder pattern for `ServerMetaContext`?

```rust
leptos::prelude::expect_context::<leptos_meta::ServerMetaContext>()
    .with_html_attrs(...)  // whatever name
```

gbj:
Manipulating the `ServerMetaContext` directly will not work with client-side routing though, for example.

I would suggest that rather than an `SsrHtmlTag` component you have something like a `FluentProvider` component, which takes `children` and uses the `Html` component, then evaluates the children. This is a pretty typical pattern in component-based frameworks, and will ensure that modifying the `<html>` tag works as expected on any platform, rather than only using SSR.

mondeja:
> Manipulating the `ServerMetaContext` directly will not work with client-side routing though, for example.

I don't understand this statement. The examples that worked for me were using the same that Leptos internally does. What is the difference inside Leptos to make them work in client side routing?

> I would suggest that rather than an SsrHtmlTag component you have something like a FluentProvider component, which takes children and uses the Html component, then evaluates the children.

That was my first attempt when crafting the library, create a macro that expand to a `Provider` component because knew that there are things that only work inside a `view!`, but it became a nightmare quick. Mainly because I need a macro to only expand the minimum code possible, that's the point.

Something that works would break the `Provider`/`Component`-children syntax.

```rust
view! {
    {my_macro! {{
        children: {<Foo/>},
        my_compile_time_prop: true,
    }}}
}

// expanding to something like

view! {
    {
        // ... a lot of code
        <Provider value=i18n>
            <Foo/>
        </Provider>
    }
}
```

Looks really really weird. How can I do it better?

Ideally it would be something like:

```rust
view! {
    <FluentProvider! my_compile_time_prop=true>
        <Foo/>
    </FluentProvider>
}
```

Component-like macros would be awesome. Without macros and with Leptos being a component-based framework, IMO, I don't see the point of using Rust in the first place.

I suppose that will end expanding to a `Provider` if a `children` parameter is passed and documenting that the ugly syntax is the only thing that we can do.

gbj:
> The examples that worked for me were using the same that Leptos internally does. What is the difference inside Leptos to make them work in client side routing?

The approach that you suggested (getting access to `ServerMetaContext` and then injecting HTML attribute text directly into the `html` field) works during server-side rendering. It would do nothing for pages rendered in the browser (for example, if the user is using client-side rendering, or if the user is using SSR + hydration but this is a subsequent page navigation). The framework in general (talking now about 0.7 as that is what this PR is against) distinguishes between different rendering paths with different methods: `build()` builds something in the browser, `hydrate()` hydrates something in the browser that was rendered on the server, `to_html_with_buf()` renders something to HTML on the server, etc. 

For the `<Html/>` component, for example, the `build()` method adds attributes to the `<html>` element that exists in the current document:
https://github.com/leptos-rs/leptos/blob/5deedf721ecfcc4a780c0005fe3b187f03519756/meta/src/html.rs#L67-L75

Note that this doesn't use `ServerMetaContext` at all, or the `html` field — which is only used during server rendering.

> We can create an API that take those cases into account.

Right, and this is exactly what the `<Html/>` component does — handles the different cases. If you want to duplicate that it's fine of course, and maybe it does make sense to expose some of the `ServerMetaContext` fields via some API. The problem is just that doing this would make it easy to (accidentally) create incorrect implementations (i.e., implementations that only work during SSR)

> That was my first attempt when crafting the library, create a macro that expand to a `Provider` component, because knew that [there are things that only work inside a `view!`](https://github.com/lpotthast/leptonic/pull/28/commits/41f66aa4d1cc9aced9d83146208c64aa9a967cb7) as `provide_context` doesn't fit well enough with multiple instances of the same type

It's not quite true that this only works inside the `view!` macro. The logic of `Provider` is itself very simple: It just creates a new `Owner` which is the child of the current `Owner`, provides the context, and then does something (in this case, renders the children)

https://github.com/leptos-rs/leptos/blob/5deedf721ecfcc4a780c0005fe3b187f03519756/leptos/src/provider.rs#L40-L48

That said, I assume the macro is the only thing that provides its context type? You can just `provide_context()` once.

> Something that works would break the `Provider`/`Component`-children syntax.
> 
> ```rust
> view! {
>     {my_macro! {{
>         children: {<Foo/>},
>         my_compile_time_prop: true,
>     }}}
> }
> ```
>
> ...
>
> Looks really really weird. How can I do it better?

The `view` macro doesn't do anything here. This could just be
```rust
my_macro! {{
  children: Foo,
  my_compile_time_prop: true
}}
```
(I don't know if there's a reason you require the double braces here. My assumption would be that the following feels more natural:
```rust
my_macro! {
  children: Foo,
  my_compile_time_prop: true
}
```

> Thanks. IMO, I would remove the `provide_context` in a future release as is confusing.

Maybe I'm misunderstanding, but the only caveat with `provide_context` is that if you provide a context of the same type multiple times, they will be shadowed -- just like if you name a variable the same thing multiple times. `Provider` exists for that use case. My assumption is that you are only providing the i18n context once so you don't need it, but maybe I'm wrong.

mondeja:
> Right, and this is exactly what the <Html/> component does — handles the different cases. If you want to duplicate that it's fine of course, and maybe it does make sense to expose some of the ServerMetaContext fields via some API. The problem is just that doing this would make it easy to (accidentally) create incorrect implementations (i.e., implementations that only work during SSR)

That can be easily solved only adding the API when SSR with `#[cfg(feature = "ssr")]`.

> (I don't know if there's a reason you require the double braces here. My assumption would be that the following feels more natural

As I remember doesn't compiles because is not a Rust valid construct.

> Maybe I'm misunderstanding, but the only caveat with `provide_context` is that if you provide a context of the same type multiple times, they will be shadowed -- just like if you name a variable the same thing multiple times. `Provider` exists for that use case. My assumption is that you are only providing the i18n context once so you don't need it, but maybe I'm wrong.

That context shadowing I think that is indeed creating incorrect implementations. Perhaps a warning on debug mode could mitigate the problem.

PR #2795, opened 2024-08-07
docs: bring warning on shadowing context correctly from 0.6

agraboso:
Brings the section titled "Warning: Shadowing Context Correctly" in the documentation for `provide_context` back from the 0.6 docs, updating `use leptos::*` to `use leptos::prelude::*` in the examples, and adding a `use leptos::context::Provider` in the appropriate example.

### DIFF
--- a/reactive_graph/src/owner/context.rs
+++ b/reactive_graph/src/owner/context.rs
@@ -128,6 +128,118 @@ impl Owner {
 /// });
 /// # });
 /// ```
+///
+/// ## Warning: Shadowing Context Correctly
+///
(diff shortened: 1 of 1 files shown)


agraboso:
That makes sense. Maybe the best place for it is chapter 8 of the book, in the section titled [Option #2: Passing Signals through Context](https://book.leptos.dev/15_global_state.html#option-2-passing-signals-through-context) (with some modifications to adapt it to the narrative there).

rakshith-ravi:
@agraboso @gbj Hate to be "that guy" but wanted to revive this dead PR to see if we can figure out how to move forward?

Issue #2789, opened 2024-08-07
`leptos_axum::redirect` fails if there is a delay issuing it inside a server function


metatoaster:
**Describe the bug**

Given the following server function:

```rust
#[server] 
pub async fn goto_target() -> Result<String, ServerFnError> {
    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
    leptos_axum::redirect("/target");
    Ok("/target".to_string())
}
```

The sleep is there to emulate a fetch from some database; in my use case it determines whether or not to redirect the request to a different route set up directly on the underlying `axum::Router`.

Using the above in the following component:

```rust
#[component]
pub fn DemoPage() -> impl IntoView {
    let target = Resource::new(|| (), |_| async move { goto_target().await });
    
    view! {
        <Suspense fallback=move || view! { <p>"Loading..."</p> }>
            <span>"Redirecting to "{target}</span> 
        </Suspense>
    }   
}   
```

The above is setup inside a `<Routes>` at the root level with a `<Route path=StaticSegment("demo") view=DemoPage />`.

Issuing the following curl command will not result in the expected redirect. Note the `200 OK` and the lack of `location` header.

```shell
$ curl -H "Accept: text/html" -v http://127.0.0.1:4000/demo
*   Trying 127.0.0.1:4000...
* Connected to 127.0.0.1 (127.0.0.1) port 4000 (#0)
> GET /demo HTTP/1.1
> Host: 127.0.0.1:4000
> User-Agent: curl/8.0.1
> Accept: text/html
> 
< HTTP/1.1 200 OK
< content-type: text/html; charset=utf-8
< transfer-encoding: chunked
< date: Wed, 07 Aug 2024 04:56:47 GMT
... rest of output snipped ...
```

Commenting out the `sleep` will result in the expected behavior, i.e. the `302 Found` is set along with the `location` header:

```shell
$ curl -H "Accept: text/html" -v http://127.0.0.1:4000/demo
*   Trying 127.0.0.1:4000...
* Connected to 127.0.0.1 (127.0.0.1) port 4000 (#0)
> GET /demo HTTP/1.1
> Host: 127.0.0.1:4000
> User-Agent: curl/8.0.1
> Accept: text/html
> 
< HTTP/1.1 302 Found
< location: /target
< content-type: text/html; charset=utf-8
< transfer-encoding: chunked
< date: Wed, 07 Aug 2024 05:13:22 GMT
```

[...]

PR #2796, opened 2024-08-08
doc: more details on how redirect works (#2789)

metatoaster:
Include the note about blocking resource for both the actix and axum, and expanded the wording on the axum version.

Note that I did add a `FIXME` note to `leptos_axum::redirect` because I don't know how should this be properly addressed in the documentation (or the code for that matter).

I am open to suggestion to better wording, etc.

### DIFF
--- a/examples/ssr_modes_axum/src/app.rs
+++ b/examples/ssr_modes_axum/src/app.rs
@@ -1,5 +1,5 @@
 use lazy_static::lazy_static;
-use leptos::prelude::*;
+use leptos::{prelude::*, spawn::spawn_local};
 use leptos_meta::MetaTags;
 use leptos_meta::*;
 use leptos_router::{
@@ -20,7 +20,7 @@ pub fn shell(options: LeptosOptions) -> impl IntoView {
             <head>
                 <meta charset="utf-8"/>
                 <meta name="viewport" content="width=device-width, initial-scale=1"/>
-                <AutoReload options=options.clone() />
+                <AutoReload options=options.clone()/>
                 <HydrationScripts options/>
                 <MetaTags/>
             </head>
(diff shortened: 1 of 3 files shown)


gbj:
Thanks! I split it into two sections, one describing usage and one with more depth about the implementation (what you wrote). Hopefully this will point others in the right direction if they are using curl etc.

metatoaster:
Thanks for that. Looking into the `serverfnredirect` header I found myself inside `server_fn/src/redirect.rs`, and it provides these hooks and I see that by default it is set for the various form components (i.e. `ActionForm` and `MultiActionForm`).  This means if the component that calls the function that will do the redirect will also need to set this hook up much like those components did?  I will note that those sections are prefixed with

```rust
// if redirect hook has not yet been set (by a router)`
```

It may be useful to also include a reference from the redirect functions too to lead to the explanation on how this bit actually works, as what ended up happening when I first came across this was that I more or less re-invented the wheel by setting the 302 status code directly immediately after the server function call (after checking whether that function returned an `Ok(Err(...))` to indicate that a redirect is needed.  Granted what I did works but it would be useful to have something more canonical/idiomatic for this package that uses the provided pieces.  All that being said, with these pieces I've found I may take yet another stab at wrapping up this problem.

gbj:
> Thanks for that. Looking into the `serverfnredirect` header I found myself inside `server_fn/src/redirect.rs`, and it provides these hooks and I see that by default it is set for the various form components (i.e. `ActionForm` and `MultiActionForm`).

Correct: This is so that they still work outside the `Router` (for example, if you use them with islands and there is no client-side routing)

> This means if the component that calls the function that will do the redirect will also need to set this hook up much like those components did? I will note that those sections are prefixed with

The user should not need to set these manually, unless they are not using the router. I notice that `leptos_router` does not currently call `set_redirect_hook`, which it should – that is just a bug, not something that users should need to think about.

metatoaster:
I see, and digging around further I found [this TODO note in `router/src/components.rs`](https://github.com/leptos-rs/leptos/blob/27dbadb7d2/router/src/components.rs#L90).  I hope the relevant hooks will be in place and fully set up for the final 0.7 release.

gbj:
See #2815.

PR #2809, opened 2024-08-10
Callback changes

gbj:


gbj:
Hm, this turns out to be difficult. We can have two of three:
1. `Callback<_>` can implement `From<Fn(In) -> Out>`
2. `Callback` can be used directly as an event handler like `on:`, `.on()`
3.  Event handlers can automatically infer the event type correctly

Basically if you change `.on()` so that it takes `EventCallback<E>`, it can no longer infer `ev` in `|ev| { ... }`, because that `ev` is no longer necessarily `FnMut(E)`: `EventCallback` could be implemented for other types.

blorbb:
If I understand point 2. correctly, we won't be able to do something like `on:click={callback}`, instead we need `on:click={|ev| callback.run(ev)}`? If so, I think this would be the best one to leave out.

When I've used `Callback`, it's mostly been for structs to avoid a bunch of generics. In components, since impl trait is supported, I would usually have `callback: impl Fn(MouseEvent)` as the argument instead of `#[prop(into)] callback: Callback<MouseEvent, ()>`.

A lot of my event listeners also take no arguments anyways since I usually don't need `ev` to be passed in, so that would already require `on:click={|_| callback.run(())}`.

Imo 1 and 3 are most important to have since 1 is required for `#[prop(into)]` to work and annotating every single event listener would be a huge pain. Having 2. seems to have the least benefit.

gbj:
Thanks, this is helpful feedback and matches my own intuitions.

PR #2826, opened 2024-08-13
feat: add `memo!` macro (closes #2708)

### LINKED ISSUE #2708: `memo!` macro similar to `slice!` macro
## Is your feature request related to a problem? Please describe.
As described in the Leptos book, sometimes it is useful to memoize signals to avoid unnecessary rerenders. For signals that can be read and written to, there exists a macro, `slice!`, that allows cases where a specific property that is being read and written to directly can be expressed more consisely. e.g.

```rust
#[derive(Clone, Copy)]
struct State {
    pub foo: i32,
    pub bar: String
}

#[component]
fn MyComponent() -> impl IntoView {
    let state = RwSignal::new(State {
        foo: 42,
        bar: String::from("Bar")
    });

    // verbose version
    let (foo, set_foo) = create_slice(state, |state| state.foo, |state, n| state.foo = n);

    // conscise version
    let (foo, set_foo) = slice!(state.foo);

    // Render stuff with signals...
}
```

However, if I want a read-only memo of `foo`, my only option is something like this:

```rust
let foo = Memo::new(move |_| with!(|state| state.foo));
```

## Describe the solution you'd like
I believe a similar macro to `slice!`, `memo!`, can be created for such cases. The above memo would be able to be re-expressed like this:

```rust
let foo = memo!(state.foo);
```

sabify:
Closes #2708

### DIFF
--- a/leptos_macro/tests/memo/red.stderr
+++ b/leptos_macro/tests/memo/red.stderr
@@ -1,0 +1,31 @@
+error: unexpected end of input, expected identifier
+  --> tests/memo/red.rs:19:13
+   |
+19 |     let _ = memo!();
+   |             ^^^^^^^
+   |
+   = note: this error originates in the macro `memo` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: expected `.`
+  --> tests/memo/red.rs:21:13
+   |
+21 |     let _ = memo!(outer_signal);
+   |             ^^^^^^^^^^^^^^^^^^^
+   |
+   = note: this error originates in the macro `memo` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: unexpected end of input, expected identifier or integer
+  --> tests/memo/red.rs:23:13
+   |
+23 |     let _ = memo!(outer_signal.);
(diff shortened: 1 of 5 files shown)


leptos_macro/tests/memo/red.stderr
benwis, line 1: You probably should remove this file, right?

sabify: I believe it is needed as part of `leptos_macro` test suite like what we have for slice.

benwis:
Could be, I'm not familiar with that part of it

sabify:
The main branch needs `cargo fmt` over merging 3f751551a84625f393a87e86fe73ff6a60d0f97f to make CI work.

benwis:
Indeed, I've merged it

mondeja:
> The main branch needs `cargo fmt` over merging [3f75155](https://github.com/leptos-rs/leptos/commit/3f751551a84625f393a87e86fe73ff6a60d0f97f) to make CI work.

Ups sorry, fixed by #2831

benwis:
Lovely!

PR #2827, opened 2024-08-13
fix: update to wasm-bindgen 0.2.93

sabify:
This solves the CI issue with the new wasm-bindgen version.


benwis (commented):
Nice, that'll fix those

PR #2835, opened 2024-08-14
fix: `#[component]` unknown attributes propagation

mondeja:
Currently, when you mark a component as deprecated, the meta attribute is propagated to the dummy model, when it should be propagated to the "real" model.

To check the behaviour, you can just mark as `#[deprecated]` the `SimpleCounter` component of the `counter` example and run `trunk serve`:

```rust
#[deprecated="Hello"]
#[component]
pub fn SimpleCounter( ... )
```

Before:

```
warning: use of deprecated function `__SimpleCounter`: Hello
 --> src/lib.rs:8:8
  |
8 | pub fn SimpleCounter(
  |        ^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default
```

After:

```
warning: use of deprecated function `counter::SimpleCounter`: Hello
 --> src/main.rs:1:14
  |
1 | use counter::SimpleCounter;
  |              ^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default

warning: use of deprecated function `counter::SimpleCounter`: Hello
 --> src/main.rs:8:18
  |
8 |         view! { <SimpleCounter initial_value=0 step=1/> }
  |                  ^^^^^^^^^^^^^

warning: `counter` (bin "counter") generated 2 warnings
```

Initially I started implementing for `#[deprecated]` attributes but I realize that can be extended to every attribute that is not `#[doc =]`. Not totally sure about the implications but it seems safe to me.

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -19,6 +19,7 @@ use syn::{
 pub struct Model {
     is_island: bool,
     docs: Docs,
+    unknown_attrs: UnknownAttrs,
     vis: Visibility,
     name: Ident,
     props: Vec<Prop>,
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

PR #2289, opened 2024-02-10
Fix error reporting in view macro

Ar4ys:
**rust-analyzer:** [2024-01-29](https://github.com/rust-analyzer/rust-analyzer/releases/2024-01-29) (v0.3.1823)

This PR is aimed to improve DX of using leptos by fixing error reporting in `view!` macro, so that errors inside it are reported over the token, that caused an error, instead of the whole `view!` macro call site.

> [!NOTE]
> This PR is tied to [bugfix](https://github.com/rust-lang/rust-analyzer/issues/16407) in rust-analyzer (that is already merged), so "minimum expected rust-analyzer version" is `v0.3.1823`. Also, in order to get better rust-analyzer feedback - you need to add `"rust-analyzer.cargo.allFeatures": true` to your VSCode config (or similar flag for other IDEs). Should I add these two points to `leptos-rs/book/getting_started/leptos_dx.md`?

Here is a simple example:

Before | image
--------|------
After | image

<details>
<summary><h2>List of all errors that are covered by this PR</h2></summary>

### `unreachable_code`/`missing_field`
Disable "unreachable_code" lint when there is an error `view!` macro

Report "missing field" and other `.build()` errors on component's name

**Regression:** "Go to Definition" triggered on `Component` name in `view!` macro now suggests `build` method definition in addition to `Component` itself. Previously only `Component` was reported

Before | image
--------|------
After | image

### `component_with_props_without_macro`
Component with props but without `#[component]` macro is used in `view!` macro

Before | image
--------|------
After | image

### `generics`
When incorrect generics are passed or inferred in component the error is reported over the whole `view!` macro.

**Regression:** "Go to Definition" triggered on `Component` name in `view!` macro now suggests `leptos::component_props_builder` definition in addition to `Component` itself.

Before | image
--------|------
After | image

### `prop_into`
Error reporting with `#[prop(into)]`

Before | image
--------|------
After | image

### `events`
Error for "on:" events should be reported on it's name

Before | image
--------|------
After | image

### `directive`
Missing or incorrect type of the param when using directive

Before | image
--------|------
After | image

### `block_in_fragment`
If block inside fragment returns something, that is not "renderable"

Before | image
--------|------
After | image

### `slots_prop_into`
Error reporting with `#[prop(into)]` in slots

Before | image
--------|------
After | image

### `wrong_slot_name`
Wrong slot name

Before | image
--------|------
After | image

### `undefined_slot`
Using slot that is not defined in a component

Before | image
--------|------
After | image

[...]

PR #2836, opened 2024-08-14
fix: Rust-Analyzer hover information / redundant spans

chrisp60:
These changes hope to improve (I wish I could say fix, but I'm not entirely sure yet) the recent issues that the `view!` and `#[component]` macros had with providing incorrect / extra information through Rust-Analyzer.

In many instances, the procedural macros are spanning tokens with their own span redundantly, while also including lint attributes and unrelated tokens within their spanned interpolation.

I am **far** from an expert on procedural macros and token spans, while this seems to fix the issues on my editor, I just don't know enough to say if it actually fixes everything. From what I can tell, no diagnostic information is lost from these changes.

Additionally, I have no idea how to write a test for this. I am open to doing additional work to research and write tests if needed.

### DIFF
--- a/leptos_macro/src/view/component_builder.rs
+++ b/leptos_macro/src/view/component_builder.rs
@@ -78,10 +78,8 @@ pub(crate) fn component_to_tokens(
                 })
                 .unwrap_or_else(|| quote! { #name });
 
-            let value = quote_spanned!(value.span()=> { #value });
-
-            quote_spanned! {attr.span()=>
-                .#name(#[allow(unused_braces)] #value)
+            quote! {
+                .#name(#[allow(unused_braces)] { #value })
             }
         });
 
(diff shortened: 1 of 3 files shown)


benwis:
Is this at all related to this PR: https://github.com/leptos-rs/leptos/pull/2835 ?

chrisp60:
> Is this at all related to this PR: #2835 ?

I believe entirely unrelated. My changes are around the builder pattern that is constructed in the view macro, theirs is about the `#[component]` definition.

chrisp60:
There are pictures available on #2840 that mostly show the improvements. Obviously the code is different but the results are visually similar

gbj:
Thanks very much! Yes, as I understand it #2289 introduced some of these, but rust-analyzer later made changes that caused overlapping spans to break. Thank you for working to clean up the issue.

PR #2840, opened 2024-08-15
[0.6] fix: Rust-Analyzer hover information / redundant spans

chrisp60:
These changes hope to achieve the same result as #2836 but for the current stable version.

Everything stated in that PR is also applicable here. This works on my machine but other people may need to chime in and say if anything doesn't look correct

All images taken using `rust-analyzer@2024-08-12`

### Before
image

### After
Notice `on:click`
image

### Before
image

### After
image

Correct hover information is provided for each individual section of `on:event` handlers as well.

### DIFF
--- a/leptos_macro/src/view/client_builder.rs
+++ b/leptos_macro/src/view/client_builder.rs
@@ -332,16 +332,17 @@ pub(crate) fn element_to_tokens(
             quote! {}
         };
         let ide_helper_close_tag = ide_helper_close_tag.into_iter();
-        let result = quote_spanned! {node.span()=> {
-            #(#ide_helper_close_tag)*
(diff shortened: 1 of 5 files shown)


chrisp60:
I am learning about the proc macro snapshots. I will try to push a commit soon once I understand how to update them

chrisp60:
Here is an example of another issue I would like to fix.

image

one of the snapshot tests seems to be failing, but I can't seem to make sense of it right now. I will take some more time to look at this. Let me know if you would prefer I swap this to Draft status until then

gbj:
No worries about the tests, I am not worried about macro snapshot tests on the 0.6 branch at this point...

PR #2843, opened 2024-08-16
chore: set `tracing` as feature for all crates

mondeja:
This makes `tracing` optional.


gbj (commented):
Thanks, this all looks good to me! I appreciate it, especially restoring the tracing component props piece.

Issue #2771, opened 2024-08-05
0.7 todos


gbj:
I have merged the `leptos_0.7` branch from #2607 because it is much easier just to juggle main + PRs than main, a branch, PRs to main, and PRs to the branch!

Good first issues/PRs super welcome to fill in missing things:
- [x] `Resource::map` and `Resource::and_then` (use 0.6 as baseline, should be very similar)
- [x] Add in missing deprecated functions for things like `create_blocking_resource` and so on (or at least help compile a list of renamed functions that we can add deprecations for)

Outstanding feature TODOs, in order of priority
- [x] Equivalent to a user implementing `IntoView` on an arbitrary struct (`IntoRender`)
- [x] hot reloading (and therefore view markers)
- [x] `Resource::refetch()`
- [x] static routes
- [x] `Resource::once` (with a relatively more efficient implementation)
- [x] `Await` component (use `Resource::once` above)
- [x] `AnimatedShow`
- [x] Router/Route `base`
- [x] Nested router `set_is_routing`
- [x] `SsrMode::PartiallyBlocked`
- [ ] Figuring out the interaction between `Lazy` routes and hydration (what do you do while hydrating if the code for the route hasn't been lazy-loaded yet?)
- [ ] Reactive channel implementation for single-use signals
- [x] trailing slash modes
- [ ] fallible hydration => start from suspense boundary
- [x] `AnimatedRoutes`/`AnimatedOutlet`

Forward-ported bug fixes (things that have been fixed on 0.6 but reimplemented in 0.7, and need to be checked)
- [x] pending action dispatches (#2653)
- [x] unescaping during navigation (#2651)
- [x] ownership and disposal fixes and tests (#2639)
- [x] ToChildren docs (#2643)
- [x] #2621

Finalizing: 
- [x] simplify tachys APIs by removing duplicated content (HTML length, checking on template duplication, etc.) 
- [ ] reading through docs to make sure they're consistent now

When ready for release
- [x] updating book
- [x] updating sandboxes
- [x] making sure all docsrs docs build and are complete
- [ ] migrating templates to use the new versions of the templates

dmgolembiowski:
> - [ ] Figuring out the interaction between Lazy routes and hydration (what do you do while hydrating if the code for the route hasn't been lazy-loaded yet?)

[...]

PR #2865, opened 2024-08-20
feat: add Resouce::map and Resouce::and_then

mkane0814:
I am not sure if this will work for closing out one of the todos in #2771, but I figured I would take a stab at it at least. Let me know if I need to change anything.


benwis (approved):
Looks good to me, I'll let @gbj take a look since he wrote all the code

gbj:
Thanks!

PR #2871, opened 2024-08-22
(draft) More work on stores

gbj:
This adds support for working with enums with stores, specifically:
1. additional support for `Option<_>` store fields, via monadic `.unwrap()` and `.map()` that convert from "a store with an `Option<T>` field" to "an `Option<_>` of a `T` store field"
2. adding support for enum store fields (beginning with simple enum "does this match" fields, and progressing to projecting reactivity through to unnamed or named enum fields)

Note: I discovered that using `<For/>` with the store field iter and deleting a row causes a panic, so I'll fix that too but the `stores` example is broken when removing a row for the time being


reactive_stores/src/option.rs
```diff
@@ -0,0 +1,45 @@
...
+    <S as Read>::Value: Deref<Target = Option<T>>,
+{
+    type Output = T;
+
+    fn unwrap(self) -> Subfield<Self, Option<Self::Output>, Self::Output> {
+        Subfield::new(
+            self,
+            0.into(),
+            |t| t.as_ref().unwrap(),
+            |t| t.as_mut().unwrap(),
```
fiadliel, line 31: I got a panic here on line 31 more than once (I think on page reload), but not reliably reproducible, unwrapping a `None` value.

fiadliel: The problem actually was, I called `.set` on the subfield value inside the closure, and this broke after that. Works if I call `.set(Some(...))` on the value outside the closure.

gbj: Do you have a quick example? I think I understand what you mean but am not totally sure.

fiadliel: Here it is - the subfield coming out of `map` is stale unless the parent field is triggered.

Click `Button`, and then `Button 2`. Bug or PEBKAC?

```rust
#[derive(Debug, Clone, reactive_stores_macro::Store)]
pub struct User {
    pub name: Option<Name>,
}

#[derive(Debug, Clone, reactive_stores_macro::Store)]
pub struct Name {
    pub first_name: Option<&'static str>,
}

#[component]
fn App() -> impl IntoView {
    let store = Store::new(User {
        name: Some(Name {
            first_name: Some("John Doe"),
        }),
    });
    view! {
        {move || {
            store
                .name()
                .map({
                    move |name| {
                        view! {
                            <button on:click=move |_| {
                                name.set(Name { first_name: None });
                            }>Button</button>
                            <br />
                            {move || {
                                name.first_name()
                                    .map({
                                        move |name| {
                                            view! {
                                                <button on:click=move |_| {
                                                    name.set("James Doe")
                                                }>Button 2</button>
                                            }
                                        }
                                    })
                            }}
                        }
                    }
                })
        }}
    }
}
```

gbj: Ah ok interesting. This is a great test case for whether accessing a child field should also track the parent, which was the change [here](https://github.com/leptos-rs/leptos/pull/2871/commits/fb9347923138741ff51a946b668f0adcfef615cf).

The reasoning to *not* track them was that it over-re-runs some things: for example, changing from `Some("John Doe")` to `Some("James Doe")` would cause the `.map()`. 

But I think this is an obvious reason that change was wrong.

fiadliel:
One minor quibble - I'm not sure `map` is the best name for this functionality (though if you like it, so be it).

There's a lot of existing examples of types that use `.map` (not even mentioning other languages):

 - `Option<A>` takes a `Fn(A) -> B` and returns an `Option<B>`
 - `Result<A, E>` takes a `Fn(A) -> B` and returns a `Result<B, E>`
 - `Ref<'b, A>` takes a `FnOnce(&A) -> &B` and returns a `Ref<'b, B>`
 - `Poll<A>` takes an `FnOnce(A) -> B` and returns a `Poll<B>`

But this pattern isn't the case for `StoreField` here. The signature is more like that `StoreField<Option<A>>` takes a function `FnOnce(Option<StoreField<A>>) -> B` and returns an `Option<B>`, which is quite different. It's like it unwraps one layer of the effect. I'm not sure if there's a name for this kind of pattern already, so whatever name makes sense. What about `.map_option` if you want to stick with having `map` in the name?

reactive_stores/src/option.rs
gbj: Ok, I actually read your example too quickly: What's *really* going on here is a simpler issue, namely that reading from a subfield was not appropriately tracking its parents. What this meant was that mutating `first_name` by mutating `name` was not notifying the effect that read from `first_name`, which is clearly wrong.

97282f4 is the one-line fix, and thanks again for pointing this out.

gbj:
> One minor quibble - I'm not sure `map` is the best name for this functionality (though if you like it, so be it).
> ...
> Option<A> takes a Fn(A) -> B and returns an Option<B> 
> ...
> The signature is more like that `StoreField<Option<A>>` takes a function `FnOnce(Option<StoreField<A>>) -> B` and returns an `Option<B>`, which is quite different.

Here's how I'd present it: `StoreField<Option<A>> takes a Fn(A) -> B and (reactively) returns an Option<StoreField<B>>`.

I agree that this is *different* from the notion of `StoreField<A> takes a Fn(A) -> and returns a StoreField<B>` 
 
The point here is to be able to continue chaining method calls.

I'd explain the naming choice less at the type level and more at the level of usage, with the pattern of accessing subfields intended to be as close to the pattern of using a regular Option as possible
```rust
        let user = User {
            name: Some(Name {
                first_name: Some("Greg".into()),
            }),
        };
        let first_name1 = user.name.map(|n| n.first_name);

        let user = Store::new(User {
            name: Some(Name {
                first_name: Some("Greg".into()),
            }),
        });
        let first_name2 = user.name().map(|n| n.first_name().get());

        assert_eq!(first_name1, first_name2);
```

I guess there's something to be said for actually returning a `StoreField<Option<U>>` here instead, which would be more like a "map" for other types.

fiadliel:
Yeah - it's the difference between a more rigorous definition of `map` (which could not implement this functionality, i.e. the output of a true map/functor can't be both a getter and a setter), and a more casual definition. As I said, up to you.
Otherwise, I'm happy with the signature as is, I don't see any need to change the output type of the function.

brofrain:
Hi @gbj,
I was giving `reactive_stores` a try recently, and I've encountered a few difficulties with them, which I'd like to share. I hope it isn't the wrong place. I also might be just misusing the functionality of course.

The first little thing is that stores implement `.read()` / `.with()` and their untracked variants, but they lack `.write()` / `.update()` methods that would allow mutating them as a whole. In my use case, I want to implement a few reusable methods for a store that would modify its contents:

```rust
#[derive(Store)]
pub struct State {
    items: Vec<Item>,
    rng: StdRng,
}

impl State {
    pub fn shuffle_items(&mut self) {
        self.items.shuffle(&mut self.rng);
    }
}
```

My current workaround is:

```rust
state.writer().unwrap().shuffle();
state.items().trigger(); // since `state.writer().unwrap()` doesn't trigger updates
```

It definitely does feel like I'm hacking something 😄 

On second thoughts, mutating a store "as a whole" may not be the best idea performance-wise as it's not "fine-grained" at all.
I wonder how universal stores can eventually become. Nested signals are quite verbose and even more cumbersome when a struct is shared between a front-end and a server.

For example, I can use something like this for the server-side logic:

```rust
struct User {
    name: String,
    age: usize
}

impl User {
    pub fn increment_age(&mut self) {
        self.age += 1;
    }
}
```

But then, when I want to use it in a granular way on the client, I'm forced to rewrite the struct with signals inside:

```rust
struct UserReactive {
    name: RwSignal<String>,
    age: RwSignal<usize>
}

impl UserReactive {
    pub fn increment_age(&mut self) {
        *self.age.write() += 1;
    }
}
```

This is something I'd personally love to solve using stores.

In the first code snippet, there is also `rng` field, which I would like to keep private and prevent `.rng()` getter from being generated for it. I wonder if such flexibility is planned to be implemented or I'm just using this completely wrong.

Another problem I'm facing is using `Field` and `Store` types with more generic components accepting `MaybeSignal`, `Signal` or `RwSignal` props.
For example, `Memo` implements both `Into<Signal>` and `Into<MaybeSignal>`, but `Field` cannot be used in such a way. Since it's mutable with `.write()` and similar methods, it would be also nice to define a flexible prop accepting both `Field` and `RwSignal`. I don't know how to handle that currently though.

Thanks in advance!

gbj:
>  I hope it isn't the wrong place. 

I don't think either of the things you raise here are related to this PR -- separate issues will be better so I can keep track of them separately, rather than four or five separate requests in comments on a PR that doesn't address any of them.

> The first little thing is that stores implement `.read()` / `.with()` and their untracked variants, but they lack `.write()` / `.update()` methods that would allow mutating them as a whole. 

This is a pretty simple omission, I have opened #2936.

> On second thoughts, mutating a store "as a whole" may not be the best idea performance-wise as it's not "fine-grained" at all. I wonder how universal stores can eventually become. Nested signals are quite verbose and even more cumbersome when a struct is shared between a front-end and a server.
> 
> ...
> 
> But then, when I want to use it in a granular way on the client, I'm forced to rewrite the struct with signals inside:
> ... 
> This is something I'd personally love to solve using stores.

I have plans for a "patching" approach here, which allows you to provide a new version of the entire struct, and will run equality checks for each field before updating it. This work is just not complete yet. 

> In the first code snippet, there is also `rng` field, which I would like to keep private and prevent `.rng()` getter from being generated for it. I wonder if such flexibility is planned to be implemented or I'm just using this completely wrong.

That would prevent you from reactively accessing the `rng` field. Is that your intent?

I believe currently the `pub`ness for each field is passed down to its methods (or it easily could be), so if it is a private field then `rng()` will be a private method.

> Another problem I'm facing is using `Field` and `Store` types with more generic components accepting `MaybeSignal`, `Signal` or `RwSignal` props. For example, `Memo` implements both `Into<Signal>` and `Into<MaybeSignal>`, but `Field` cannot be used in such a way. 

Again, implementing `Into<Signal>` and `Into<MaybeSignal>` on `Field` is trivial but just not done. Feel free to make a PR to add an implement.

> Since it's mutable with `.write()` and similar methods, it would be also nice to define a flexible prop accepting both `Field` and `RwSignal`. I don't know how to handle that currently though.

You can take `impl Writeable` or `impl Update` or `impl Set` as a prop, no?

brofrain:
Thanks for the explanation

I'm not yet fully familiar with fine-grained reactivity and API concepts here, so I often doubt whether something is a missing feature or intentionally not meant for specific usages.
I've opened separate issues for my things (https://github.com/leptos-rs/leptos/issues/2947, https://github.com/leptos-rs/leptos/issues/2946) and so far failed to implement them rather miserably. But I will ask related questions there not to pollute this thread further.

> I have plans for a "patching" approach here, which allows you to provide a new version of the entire struct, and will run equality checks for each field before updating it. This work is just not complete yet.

That sounds great. Looking forward to it!

> You can take impl Writeable or impl Update or impl Set as a prop, no?

Thanks, that works 🙂 
I was expecting the traits to have complex values for example `Read` has.
So `MaybeSignal`/`Signal` for "reading any signal type" and `impl Writeable` for "writing any signal type". Got it!

PR #2873, opened 2024-08-23
DX: Variadic `From` for `Callback`implemented variadic from for callbacks

maccesch:
This allows the following:

```rust
let cb0: Callback<()> = (|| {}).into();  // No more parameter necessary
let cb1: Callback<(i32,)> = (|x| {}).into();
let cb2: Callback<(i32, String)> = (|x, s| {}).into();
let cb3: Callback<(i32, String, f32)> = (|x, s, f| {}).into();
```

Please note the lack of tuple parentheses in the closure paremeters. 
This makes it nicer to provide callbacks as params to components.

The only downside is that inside a component, when you run a callback with only one parameter, it requires a tuple. The same goes for specifying the type of a callback with one parameter as you can see above (`cb1: Callback<(i32,)>`).

```rust
cb0.run(());                             // same as before
cb1.run((42,));                          // this requires a tuple now
cb2.run((42, "bla".to_string()));        // same as before
cb3.run((42, "bla".to_string(), 1.0));   // same as before
```

IMO this tradeoff is worth it. It shifts the potential weirdness into a component while making it more straight-forward to use. This is especially nice for crate authors.

This is the same as #2837 because I messed up git.

### DIFF
--- a/leptos/src/callback.rs
+++ b/leptos/src/callback.rs
@@ -10,13 +10,13 @@
 //!
 //! #[component]
 //! fn MyComponent(
-//!     #[prop(into)] render_number: Callback<i32, String>,
+//!     #[prop(into)] render_number: Callback<(i32,), String>,
 //! ) -> impl IntoView {
 //!     view! {
 //!         <div>
-//!             {render_number.run(1)}
+//!             {render_number.run((1,))}
 //!             // callbacks can be called multiple times
-//!             {render_number.run(42)}
+//!             {render_number.run((42,))}
 //!         </div>
 //!     }
 //! }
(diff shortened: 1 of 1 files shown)


benwis (approved):
I feel like inserting the "Look at what they have to do to achieve a fraction of our power meme". Overall it looks good and I like the easier DX. We may or may not want to go higher than 12 here, we're definitely having a similar discussion on a different route variadic

maccesch:
Clippy by default complains already when functions have more than 7 parameters. If you need more you should use structs usually. 

However let me know and I'll add the implementations for up to 100 😂

gbj:
Works for me! Thanks for this.

PR #2874, opened 2024-08-23
updated codee version

maccesch:
This shouldn't affect anything in Leptos at all. Just updateing so leptos-use has the same version


benwis (approved):
Thanks for keeping up to date!

PR #2876, opened 2024-08-24
Allow for documentation and other attributes to fields in server fn

### LINKED ISSUE #2844: Server function param custom name
[Cloudflare Turnstile](https://developers.cloudflare.com/turnstile/) (CAPTCHA) injects a form `<input>` that has the name `cf-turnstile-response` which isn't a valid rust ident.
When using the `<ActionForm/>` (important for progressive degradation) I would need to do this, which obviously isn't possible:
```rust
#[server]
async fn signup(username: String, email: String, cf-turnstile-response: String) -> Result<_, _> {}
```

A serialization rename attribute for server function params.
```rust
#[server]
async fn signup(
    username: String,
    email: String,
    #[server(rename = "cf-turnstile-response")]
    cf_turnstile_response: String,
) -> Result<_, _> {}
```

I'm realizing now that there is a `data-response-field-name` attribute that will override the form param name on the client side, so I'm unblocked on this. However, I still think this would be a useful feature for those times when it isn't possible.

rakshith-ravi:
Fixes #2844

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -8,7 +8,7 @@
 
 use convert_case::{Case, Converter};
 use proc_macro2::{Literal, Span, TokenStream as TokenStream2};
-use quote::{quote, quote_spanned, ToTokens};
+use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::{
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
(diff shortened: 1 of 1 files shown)


rakshith-ravi:
Yet to test this out properly, so take this PR with a grain of salt

rakshith-ravi:
Yup. Tested it out and it compiles

gbj:
Thanks! This looks good overall but I think it can actually me slightly more permissive. 

For example:
```rust
#[server]
pub async fn add_todo(
    /// The actual content of the todo.
    #[server(default)]
    #[server(rename = "someOtherName")]
    title: String,
    #[allow(unused)] foo: usize,
) -> Result<(), ServerFnError> {
```
Will reject the `#[allow(unused)]`, which is fine in a normal fn definition.

Basically I think 
1) the error condition should be moved up into the `if attr.path().is_ident("server") {` branch: if it is `#[server]` and it's not `default` or `rename`, then it should return an error
2) then the `else` branch can simply pass through *all* other attributes, not just `doc`

Does that make sense?

(The rename looks fine to me and I think you can remove the comment about it!)

Thanks very much for this.

rakshith-ravi:
That makes sense, but wouldn't we want to control the list of attributes that are passed to the struct? My concern is that an end user might not know what happens behind the scenes in the server fn macro, so they might not know what attributes are allowed and what are not, thereby causing confusion when they try to use a macro but it ends up failing

rakshith-ravi:
Got it. Will compile a list of commonly used attributes and get them added

gbj:
lgtm, thank you!

PR #2877, opened 2024-08-25
chore: update suspense_test example to use Resource::map

mkane0814:
Let me know if I need to update the wording in the explanation comment.


examples/suspense_tests/src/app.rs
```diff
@@ -319,7 +325,7 @@ fn LocalResource() -> impl IntoView {
                 "Loading 1..."
             }>
                 {move || {
-                    one_second.get().map(|_| view! { <p id="loaded-1">"One Second: Loaded 1!"</p> })
+                    one_second/*.get()*/.map(|_| view! { <p id="loaded-1">"One Second: Loaded 1!"</p> })
```
rakshith-ravi, line 328: Perhaps it would make more sense to remove the comments?

examples/suspense_tests/src/app.rs
```diff
@@ -135,6 +135,12 @@ fn SecondaryNav() -> impl IntoView {
     }
 }
 
+/*
```
rakshith-ravi, line 138: We should set this up either as a component documentation or as a module documentation perhaps

mkane0814: I'm not sure about this honestly. The distinction I am trying to make with this comment is implementation specific and not specific to the module/component. I could be convinced otherwise, but that's how I feel at the moment.

gbj: I'm not sure what the comment is supposed to mean, to be honest: 

> However, you should use the provided Resource::map method to avoid any potential pitfalls.

What are the potential pitfalls? `.map()` is really just syntax sugar.

I'd suggest either clarifying or removing, so that users don't ask "Wait, what are the pitfalls of `.get().map()`?

mkane0814:
Yeah, I agree. I don't think the comments are necessary. I will update the PR.

gbj:
Looks good, thanks!

PR #2878, opened 2024-08-25
Include the axum_js_ssr demo.

metatoaster:
As per an earlier discussion with @benwis, I cleaned up the JavaScript integration demo and now submitting that for inclusion as one of the official examples.


gbj:
Thanks! I have not reviewed it yet; FYI the reason the CI is failing is that there is not a `Makefile.toml` that tells CI how to run, as is found in the other examples. (Just copy one of those)

metatoaster:
Thanks for the tip, I had just pulled and squashed what I've done from my original repo and I had left out that file for some reason. Anyway I am happy with this now - I am fine with the merge going ahead with the final touches done via the force pushes. If any of the explanations I've given for the examples isn't quite how it actually works please do let me know, as I am still fairly new to all of this.

gbj:
I've just walked through it and this all seems reasonable to me. It's particularly helpful that you included "ok, here's the working method for each approach." This will be useful to share with people in the future... Much appreciated.

PR #2879, opened 2024-08-25
Add support for postcard codec in server_fn

niklass-l:
According to the ser/de benchmarks on https://github.com/djkoloski/rust_serialization_benchmark, postcard should be almost as fast as rkyv, while being compatible with serde, which is very convenient.
Postcard also has a significantly smaller serialized size compared to rkyv.

I've implemented the codec in server_fn and it seems to be working well.

### DIFF
--- a/examples/server_fns_axum/Cargo.toml
+++ b/examples/server_fns_axum/Cargo.toml
@@ -16,6 +16,7 @@ server_fn = { path = "../../server_fn", features = [
   "serde-lite",
   "rkyv",
   "multipart",
+  "postcard",
 ] }
 log = "0.4.22"
 simple_logger = "5.0"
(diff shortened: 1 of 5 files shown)


gbj:
I think the CI is dying here because of the combinatoric growth of adding features when we test combinations of features -- could you add a few excluded feature combos for postcard starting around [here](https://github.com/leptos-rs/leptos/pull/2879/files#diff-0e08e6740eee1441699c4026db1a9971033352034a4a33ff13a3f83ad7f84693L138), as they are for the others? 

Otherwise this looks totally reasonable to me. I assume you've tested it with some examples and it works at runtime?

benwis:
Nice Work!

PR #2883, opened 2024-08-26
Added attributes to <source> element

purung:
I noticed some missing attributes on the `<source>` element, so I just added the ones listed on the [reference page on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#srcset).


gbj:
Thanks!

PR #2886, opened 2024-08-27
tests: fix `Effect` doctests not being executed

mondeja:
Calls to `tokio::task::LocalSet::new().run_until()` are not awaited. To reproduce it, put `assert!(false)` inside the tests, they still pass.


reactive_graph/src/computed/selector.rs
```diff
@@ -55,6 +55,7 @@ use std::{
 ///
 /// # any_spawner::Executor::tick().await;
 /// assert_eq!(is_selected.selected(5), false);
+/// # // TODO: this test is really failing, put `# }).await;` in the next line
```
mondeja, line 58: The call to `is_selected.v.read()` returns `Ok(None)`, so it panics with "called `Option::unwrap()` on a `None` value" at `selected()` method.

gbj:
IDK why this one integration test is failing now, I will try to take a look some time in the next few days.

PR #2889, opened 2024-08-28
implement Copy and Clone for HtmlElement without needing Rndr to be Clone/Copy

Baptistemontan:
Deriving `Clone` or `Copy` on `HtmlElement` is kind of useless because pretty much everytime you use HtmlElement the `Rndr` is `Dom`, which is not `Clone` or `Copy`. Because it is behind a `PhantomData` implementing `Clone` and `Copy` by hand makes it possible to use.


gbj:
Right, thanks. This is a good idea in general. I've also gone in and added Copy/Clone/various other trait impls for Dom, which is a unit struct anyway.

PR #2894, opened 2024-08-29
Add hash files support to 0.7

mrvillage:
This adds a `<HashedStylesheet />` component for inserting the `cargo-leptos` CSS hashed style sheet, and modifies `<HydrationScripts />` to insert the JS and CSS hashes into the file names.


benwis:
Looks solid to me, I kinda like the special stylesheet component, seems clearer than a prop on Stylesheet

PR #2900, opened 2024-09-01
feat: add either macro

geovie:
While porting some code from 0.6 to 0.7 beta I had a few occurences of this pattern:

```rust
match error {
    Some(error) => {
        view! {
            // view with custom component A
        }
     }
     None => {
         view! {
              // view with custom component B
          }
      }
}
```
which in 0.7.0 returns an error due to the type mismatch.

I know it can be resolved by adding `.into_any()` to erase the types, but sometimes you want to keep the types so I figured why not add a macro to help with the creation of the Either types.

For now only up to `EitherOf6` is implemented, adding more is trivial.

I've also changed some usages in the leptos crate, but they are not necessary

### DIFF
--- a/either_of/src/lib.rs
+++ b/either_of/src/lib.rs
@@ -133,3 +133,104 @@ tuples!(EitherOf13 + EitherOf13Future + EitherOf13FutureProj => A, B, C, D, E, F
 tuples!(EitherOf14 + EitherOf14Future + EitherOf14FutureProj => A, B, C, D, E, F, G, H, I, J, K, L, M, N);
 tuples!(EitherOf15 + EitherOf15Future + EitherOf15FutureProj => A, B, C, D, E, F, G, H, I, J, K, L, M, N, O);
 tuples!(EitherOf16 + EitherOf16Future + EitherOf16FutureProj => A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P);
+
+/// Matches over the first expression and returns an either ([`Either`], [`EitherOf3`], ... [`EitherOf6`])
+/// composed of the values returned by the match arms.
+///
+/// The pattern syntax is exactly the same as found in a match arm.
+///
+/// # Examples
+///
+/// ```
+/// # use either_of::*;
+/// let either2 = either!(Some("hello"),
+///     Some(s) => s.len(),
+///     None => 0.0,
+/// );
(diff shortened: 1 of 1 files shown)


benwis:
@geovie Can you show an example of usage? I suppose this is what you'd use instead of writing out the existing Either?

geovie:
Yes this can be used instead of manually using EitherOf* and keeping track of which EitherOf*::A/B/C... was already used.

A simple usage example is in the PR (https://github.com/leptos-rs/leptos/pull/2900/files#diff-6edbb281543fc38e436f246c07e6d225c8fddfc49b6cf2e05dfbac26f9e5b5ddL140) 

```rust
    if let Some(node_ref) = node_ref {
        Either::Left(action_form.node_ref(node_ref))
    } else {
        Either::Right(action_form)
    }
```

```rust
    either!(node_ref,
        Some(node_ref) => action_form.node_ref(node_ref),
        None => action_form,
    )
```

Altough I have to admit in the normal `Either::Left/Right` case there is not much difference but when the `EitherOf*` gets to a higher number the macro is more readable.  
e.g. if have an enum with different variants or a nested Result/Option:

```rust
let result: Result<Option<i32>> = Ok(Some(1));
either!(result,
        Ok(Some(id)) => view! { <User id /> },
        Ok(None) => view! { <NotFound /> },
        Err(err) => view! { <ErrorScreen err /> },
)
```

which turns into

```rust
map result {
        Ok(Some(id)) => EitherOf3::A(view! { <User id /> }),
        Ok(None) => EitherOf3::B(view! { <NotFound /> }),
        Err(err) => EitherOf3::C(view! { <ErrorScreen err /> }),
)
```

Again not a big difference 😅 but now lets imagine you need to add another arm to the map. With the macro you simply add a new arm and are done, with the manual `EitherOf*` you need to replace all the `EitherOf3` with `EitherOf4`.


Overall I just think it's a nice improvement of the either_of crate.

geovie:
Done

benwis (approved):

geovie:
FYI: It seems there was no new release of [either_of](https://crates.io/crates/either_of) so this macro is not yet available in the re-export of either_of for leptos 0.7

gbj:
Thanks for noting that -- published as 0.1.1 now.

Issue #2902, opened 2024-09-01
[0.7] Investigate effect of adding type erasure in router on build times 


gbj:
0.7 uses static typing everywhere. It seems like the increase in compile time is pretty consistent across the board, and while most projects have 20-30% smaller binaries I've had some reports of larger binaries.

There have also been some reports of recursion depth overflows in the compiler due to the depth of nested types.

I wonder whether the compile-time issue can be improved significantly by erasing types in the router and in some other specific control-flow locations, without too significantly adding to binary size.

PR #2905, opened 2024-09-01
test: does erasing types of components improve compile times?

gbj:
Related to #2902. There is a genuine trade-off between compile time and WASM binary size: there seems to be some exponential compile time growth (and occasional recursion limits!) as we build a larger and larger statically-typed view tree, but it does allow the compiler to eliminate unused code much more effectively.

**`hackernews`**
`cargo build --timings --features=ssr`: 19.67s (main) => 6.65s (branch)
WASM size: 544kb (main) => 583kb (branch)

**`todo_app_sqlite_axum`**
`cargo build --timings --features=ssr`: 5.74s (main) => 2.62s (branch)
WASM size: 535kb (main) => 566kb (branch)

### DIFF
--- a/examples/router/src/lib.rs
+++ b/examples/router/src/lib.rs
@@ -64,7 +64,7 @@ pub fn RouterExample() -> impl IntoView {
 
 // You can define other routes in their own component.
 // Routes implement the MatchNestedRoutes
-#[component]
+#[component(transparent)]
 pub fn ContactRoutes() -> impl MatchNestedRoutes + Clone {
     view! {
         <ParentRoute path=path!("") view=ContactList>
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -92,7 +92,6 @@ trace-component-props = [
   "leptos_dom/trace-component-props"
 ]
 delegation = ["tachys/delegation"]
-erase_types = ["leptos_macro/erase_types"]
 
 [package.metadata.cargo-all-features]
 denylist = [
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -92,6 +92,7 @@ trace-component-props = [
   "leptos_dom/trace-component-props"
 ]
 delegation = ["tachys/delegation"]
+no_erase = ["leptos_macro/no_erase"]
 
 [package.metadata.cargo-all-features]
 denylist = [
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -92,7 +92,7 @@ trace-component-props = [
   "leptos_dom/trace-component-props"
 ]
 delegation = ["tachys/delegation"]
-no_erase = ["leptos_macro/no_erase"]
+erase_types = ["leptos_macro/erase_types"]
 
 [package.metadata.cargo-all-features]
 denylist = [
--- a/leptos/src/await_.rs
+++ b/leptos/src/await_.rs
@@ -87,7 +87,7 @@ where
     T: Send + Sync + Serialize + DeserializeOwned + 'static,
     Fut: std::future::Future<Output = T> + Send + 'static,
     Chil: FnOnce(&T) -> V + Send + 'static,
-    V: IntoView,
+    V: IntoView + 'static,
 {
     let res = ArcOnceResource::<T>::new_with_options(future, blocking);
     let ready = res.ready();
--- a/leptos/src/provider.rs
+++ b/leptos/src/provider.rs
@@ -35,7 +35,7 @@ pub fn Provider<T, Chil>(
 ) -> impl IntoView
 where
     T: Send + Sync + 'static,
-    Chil: IntoView,
+    Chil: IntoView + 'static,
 {
     let owner = Owner::current()
         .expect("no current reactive Owner found")
--- a/leptos/tests/ssr.rs
+++ b/leptos/tests/ssr.rs
@@ -1,5 +1,6 @@
 use leptos::html::HtmlElement;
 
+#[cfg(feature = "ssr")]
 #[test]
 fn simple_ssr_test() {
     use leptos::prelude::*;
@@ -20,6 +21,7 @@ fn simple_ssr_test() {
     );
 }
 
+#[cfg(feature = "ssr")]
 #[test]
 fn ssr_test_with_components() {
     use leptos::prelude::*;
@@ -51,6 +53,7 @@ fn ssr_test_with_components() {
     );
 }
 
+#[cfg(feature = "ssr")]
 #[test]
 fn ssr_test_with_snake_case_components() {
     use leptos::prelude::*;
@@ -81,6 +84,7 @@ fn ssr_test_with_snake_case_components() {
     );
 }
 
+#[cfg(feature = "ssr")]
 #[test]
 fn test_classes() {
     use leptos::prelude::*;
@@ -98,6 +102,7 @@ fn test_classes() {
     assert_eq!(rendered.to_html(), "<div class=\"my big  red car\"></div>");
 }
 
+#[cfg(feature = "ssr")]
 #[test]
 fn ssr_with_styles() {
     use leptos::prelude::*;
@@ -119,6 +124,7 @@ fn ssr_with_styles() {
     );
 }
 
+#[cfg(feature = "ssr")]
 #[test]
 fn ssr_option() {
     use leptos::prelude::*;
--- a/leptos/tests/ssr.rs
+++ b/leptos/tests/ssr.rs
@@ -1,3 +1,4 @@
+#[cfg(feature = "ssr")]
 use leptos::html::HtmlElement;
 
 #[cfg(feature = "ssr")]
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -60,3 +60,6 @@ skip_feature_sets = [
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--generate-link-to-definition"]
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(erase_components)'] }
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -48,7 +48,7 @@ experimental-islands = []
 trace-component-props = []
 actix = ["server_fn_macro/actix"]
 axum = ["server_fn_macro/axum"]
-no_erase = []
+erase_types = []
 
 [package.metadata.cargo-all-features]
 denylist = ["nightly", "tracing", "trace-component-props"]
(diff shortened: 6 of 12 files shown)


gbj:
It's worth asking whether this should be opt-in or opt-out. (If you look at the commit history here, you can see I waffled.)

My thinking here is this: Ideally, the framework should be able to operate in "no erased types" mode all the time. It produces the best end result for your users, in both binary size (load times, bandwidth use) and runtime performance. It is purely a limitation of compiler performance that this becomes infeasible for some larger apps, with larger UI trees and therefore combinatorically-increasing type trees.

I would rather the default be the best outcome, and be able to tell people "oh, and it might help your dev-mode compile times to opt into the `erase-types` feature, but just FYI here are the limitations of that" rather than having people say "Wait why is my binary huge?" and have to say "oh did you add the `dont_erase_types` flag?"

benwis:
I suppose it’s the question of where’s the pain point. I expect developers to notice long compile times when they first try Leptos, and not binary size/runtime perf. Then when they go to release or deploy you say oh you need the release flag and can optimize size by adding this. I think people expect some hoops at that point.

Making this opt in, i.e slow to compile by default, reinforces a preconception I've  heard a bunch that rust on the web has slow iteration times due to long compile times. I imagine more people would simply have their preconception validated and quit if they’re trying it out than go through that friction of asking the question.

I understand the fear of packages escaping with poor performance/bundle size into the wild, but it would be quite easy to add a Warning/Check to `cargo leptos build --release` saying "Hey, you should enable the don't_erase_types for release to get all these benefits."

zakstucke:
Chiming in here. 

**TLDR:**
**This branch still does wonders for compile time compared to current main. I'm confident >=90% of people would choose erasure in debug, no erasure in release and default should be variable between release and debug.**

It's hard to deny that a 50%ish improvement in compile times will ever _**not**_ be wanted in dev. 
Dev compile times are just so important from an actual productivity standpoint, but also the rust stigma @benwis mentioned. 

It's also easy to see the opposite is true in release, an e.g. `3->6` minute compile time increase won't matter much to most people infrequently building, at the gain of bundle size and performance, **_this is where the magic @gbj has brought with the static tree shines_**.

**How could this be done?**
I don't think you can change the default feature set between debug and release, see https://github.com/rust-lang/cargo/issues/1197. 
So this could be done with 2 features, `release_erased` and `debug_no_erased`, so erasure is `cfg()` flagged based on `release/debug` unless the corresponding opt out feature is enabled. There might be other ways, but I think this would work off the top of my head.

I've done some concrete comparisons to remove any doubt  between main `2bf1f46` and this branch rebased on main for a real project.

Notes:
- `cargo clean` is used between each type and attempt, except incremental+clippy tests were run after an initial build, and had a small code edit between changes.
- These are all in debug mode, where the compile time pressure comes.
- The argument is particularly strong for ssr incremental and linting, both receiving

| Variant/Attempt no                     | 1      | 2      | 3      |
|----------------------------------------|--------|--------|--------|
| 1. Main: ssr no-incremental            | 2m22s  | 2m22s  | 2m19s  |
| 1. Erased: ssr no-incremental          | 1m24s  | 1m22s  | 1m22s  |
| 2. Main: hydrate/wasm no-incremental   | 1m18s  | 1m17s  | 1m18s  |
| 2. Erased: hydrate/wasm no-incremental | 59.15s | 1m00s  | 59.37s |
| 3. Main: incremental ssr               | 27.12s | 25.34s | 24.01s |
| 3. Erased: incremental ssr             | 12.85s | 12.17s | 12.08s |
| 4. Main: incremental wasm              | 11.66s | 11.55s | 11.41s |
| 4. Erased: incremental wasm            | 7.14s  | 6.80s  | 6.83s  |
| 5. Main: cargo clippy --all-features   | 5.56s  | 5.61s  | 5.66s  |
| 5. Erased: cargo clippy --all-features | 5.33s  | 5.36s  | 5.29s  |

I hope this helps, and thanks again for your work gbj, truly impressive.

gbj:
Thanks for the numbers @zakstucke 

Ben and I were discussing yesterday and I have a proposal: using a custom `--cfg` flag rather than a feature. `cargo-leptos` can be adapted to use this `cfg` by default when building for debug, and not to use it for release. (Users can control it via `RUSTFLAGS` in any case.)

Benefits of using `--cfg`:
- It does not require the user to opt out for release
- It does not need to be un-set by all ecosystem libraries (if we use a default feature it will need to be deactivated everywhere, lest it be included via Cargo feature unification by being included somewhere)

I've rebased the PR and updated accordingly.

zakstucke:
@gbj sounds great! 

I think the crux is `cargo leptos`/`leptos` depending on how it's implemented should be optimised for performance when "releasing" and optimised for compile times when "developing". 

Your flag solution and having it built into cargo leptos sounds like it covers more edge cases than my features suggestion.

zakstucke:
Hey @gbj, unfortunately this branch now brings back the dreaded infinite compile -> RAM overflow, this has happened post my benchmarks.

I've tested the current commit history of this PR, the breaking commit is: 
[support AddAnyAttr on AnyView](https://github.com/leptos-rs/leptos/pull/2905/commits/b4ab9f1d3d5d02cecfeaad517d8f135f7305e455)

zakstucke:
Interestingly, when I reverse the changes in that commit, I assumed it would be caused by the `todo!()` replacement on line `317`, but it's not,  still happened when moved back to `todo!()`. 

What actually causes it is the `add_any_attr` being added to the `AnyView` struct, when I remove that, it compiles fine again.

Exact changes to get compiling here:
https://github.com/zakstucke/leptos/commit/1179cab41d2472a0c9a9eb08a5bff90a7ac15d50


Hopefully that makes some sense to you...

zakstucke:
It seems also just replacing the contents of the `add_any_attr` closure with a `todo!()` also works, so it seems to be some compiler infinite recursion with `into_any()` chains or similar.

gbj:
@zakstucke Thanks! Looks like this particular change must have been the issue, then, more than the others.

That's quite tricky as this PR becomes virtually unusable with `todo!()` there, because it will break spreading props onto any component unless I'm mistaken.

zakstucke:
Unusable yes, but only because of this problem with `AnyView`, AFAIK this isn't breaking because of large types, but genuine recursion in the compiler due to the current implementation. (issue still exists when erased types turned off)

zakstucke:
But technically a site does work by just removing the closure entirely, and just returning `self` from the `AddAnyAttr` implementation, just missing functionality.

gbj:
@zakstucke Is it possible that the whole issue was that I foolishly named them the same thing? I don't think that's right because I think I'm correctly accessing it as the struct field function and not the method, but I guess maybe it is. Try [d799ede](https://github.com/leptos-rs/leptos/pull/2905/commits/d799edef3822e0c0702b5865a86e4cef935a2423) I guess. I don't think it will fix it but who knows.

zakstucke:
No, I'm quite confident there's a logic issue here, take a look at the `add_any_attr` closure contents, and the fact that it calls  `into_any()` on the generic output of the trait, which means `into_any()` recurses and produces a new `add_any_attr` closure etc etc

zakstucke:
For example, when I was testing I tried reworking things to reuse the same `AnyView` instead of calling the `into_any()` trait to produce another one. This compiles fine, but then fails at runtime downcasting because the `T` does actually always change during the `add_any_attr` closure.

zakstucke:
@gbj To put my thinking more formally:

Where `T = ()`

- `into_any()` called.
- `add_any_attr` closure is created for `T = ()`, `T` is concretely needed for the downcast and the `AddAnyAttr::Output`.
- `T::add_any_attr(attr)` -> `T2 = (attr,)`
- `T2::into_any()` repeats above infinitely at compile team, leading to `T3 = (attr, attr)`, `T4 = (attr, attr, attr)`, ... etc.

I'm not too familiar with compiler stuff, so I may be way off, but from my testing and how the logic looks, this seems to me what's happening.

gbj:
@zakstucke Thanks — This is a little tricky because I don't have an example in front of me that actually causes compiler issues. Do you have an example, or does it only occur in a sufficiently large/complex app that it's not really reproducible in small?

While investigating, I did find a possible source of the looping — Namely, that calling `.into_any()` on an `AnyView` creates an additional level of nesting, rather than simply returning that `AnyView`. It should really never be necessary for this to happen, so in [79ade93](https://github.com/leptos-rs/leptos/pull/2905/commits/79ade93f3806d4f48fcb400cb785a6d4091a74b9) I just check if the `T` that is being converted into `AnyView` is already `AnyView` and if so, simply return it rather than adding an additional layer of boxing. (This will probably also help with binary sizes I guess, in cases where we inadvertently end up with `.into_any().into_any()` chains).

I have no idea whether this will help with the particular issue, but it seems better to me in any case.

zakstucke:
Unfortunately your change doesn't fix it. 

I've been trying to repro: after deleting half the project the app starts compiling in 5m... still using >10GB of RAM whilst compiling though. (I wasn't deleting components, just the entire body of components, aka all the `view!{}` trees)

So seems like it might be a size issue,  shame, my recursive idea would've been much nicer 😢

zakstucke:
Does `AnyView` need to be still relying on `T`? Could it be reworked back into an enum?

zakstucke:
I've also realised I can actually get to the end of compilation with the untouched project too, it uses close to 50GB of RAM to get there (I must've been running too much other stuff last time it caused my PC to crash), compiles then get to this:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 9m 42s
     Running `target/debug/app`
dyld[90873]: dyld cache '(null)' not loaded: syscall to map cache into shared region failed
dyld[90873]: Library not loaded: /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
  Referenced from: <5B562BFE-8FC3-3113-AF60-B514DFEAB547> /Users/zak/z/code/app/rust/target/debug/app
  Reason: tried: '/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation' (no such file), '/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation' (no such file, no dyld cache)
[1]    90873 abort      cargo run --features ssr
```

zakstucke:
image

Note the size of the files, is it that `T` becomes a runtime cost/string for every single node, whereas normally they're just a compile time cost maybe.

zakstucke:
@gbj see discord dm

gbj:
@zakstucke I've just reverted the `add_any_attr` change to do nothing but return the same `AnyView` (`self`) rather than panicking, so hopefully that will return this to helpful territory. I'm glad you caught it.

I've realized that -- when someone runs into the case in which they *need* prop spreading but now it isn't supported because their component has been erased -- they can just mark that component `#[component(transparent)]` anyway, and it will not be erased even with component erasure turned on.

If you could confirm for me that the latest commit here is back to normal, I think it's in fine shape to be merged.

PR #2934, opened 2024-09-06
leptos 0.7: Switch to `proc-macro-error2` to address unmaintained security advisory.

azriel91:
`proc-macro-error` is unmaintained according to https://github.com/rustsec/advisory-db/pull/2057, and one of my CI checks failed on this.


gbj:
Thanks!

PR #2943, opened 2024-09-07
fix: remove dead `router::router` module leftover from `0.6`

chrisp60:
I believe this should be straightforward and just removes code that is no longer used / was exploratory to begin with


chrisp60:
unsure if these CI failures are related. I can take a look at them though

gbj:
Thanks!zd

PR #2945, opened 2024-09-07
More experimentation with compile times

gbj:


gbj:
Not useful in typical-sized projects, unfortunately.

PR #2949, opened 2024-09-08
doc: `router::{Wildcard, Static, Param}Segment`

chrisp60:
Adds documentation to the router path segments. Open to feedback if this is the correct way to structure the examples.

I am not super familiar with 0.7 yet, so please let me know if I am missing the mark anywhere


gbj:
Thanks!

Issue #2947, opened 2024-09-07
[0.7] Implement `Into<Signal<T>>` & `Into<MaybeSignal<T>>` for stores and their fields


brofrain:
**Is your feature request related to a problem? Please describe.**
`MaybeSignal` is a convenient wrapper for both dynamic and static values. All the primitive signals implement `Into<MaybeSignal<T>>` except for the newly added stores.

**Describe the solution you'd like**
All of the store-related types (`Field`, `AtIndex`, `Subfield` and `Store`) should be convertible to `MaybeSignal` using `Into` trait, so for example they can be used with `#[prop(into)] value: MaybeSignal<T>` syntax.

brofrain:
@gbj 
From my observations, converting a `Memo` / `RwSignal` / ... to `Signal` / `MaybeSignal` usually comes down to creating an inner `SignalTypes` wrapping the original signal. With `reactive_stores` a new approach may be needed, because `reactive_graph` is a dependency of `reactive_stores`, so `SignalTypes` cannot be expanded by a new variant.
Could you give me some hints on how to tackle this properly?

gbj:
It can just use `Signal::derive()` and `MaybeSignal::derive()`.

Something like this:
```rust
impl<T> From<Field<T>> for Signal<T> where T: Clone {
  fn from(value: Field<T>) -> Signal<T> {
    Signal::derive(move || value.get())
  } 
}
```

brofrain:
Closing this as I don't believe it's relevant anymore. I wasn't aware this issue was still open for so long 😄 

For visitors seeking a generic approach to signals: you can simply use generics with `Get`, `Write` etc. traits. For example:

```rust
#[component]
pub fn GenericInput(
    value: impl Get<Value = String> + Set<Value = String>,
) -> impl IntoView {
    // ...
}

#[derive(Debug, Store)]
struct SomeStore {
    value: String,
}

#[component]
pub fn SomeComponent() -> impl IntoView {
    let value = RwSignal::new("".to_string());
    let store = Store::new(SomeStore { value: "".to_string() });
    view! { 
        <GenericInput value />
        <GenericInput value=store.value() />
    }
}
```

PR #2951, opened 2024-09-08
feat(#2947): impl into Signal/MaybeSignal for store-related types

brofrain:


brofrain:
I've implemented #2951 here, but I have a few minor concerns:
- I wasn't able to implement the conversions for `AtIndex`:
image
They seem to conflict with https://github.com/leptos-rs/leptos/blob/main/reactive_graph/src/wrappers.rs#L489
- I have implemented the traits for `Then`, but I don't know what `Then` is and if it actually needs them 😄 
- Store values or their fields must implement `Clone` to be convertible to basic signals. This also introduces some cloning overhead when using them as such. I wonder if that's relevant. It doesn't seem to matter for primitive values, but I can't predict if use cases with more heavy data structures may be a thing 🤔

PR #2960, opened 2024-09-10
remove unused dependencies leptos_axum and leptos_router

vsuryamurthy:
In the process of preparing for 0.7 (atleast for leptos_axum), some dependencies were added but unnecessary as axum re-exports them.

### DIFF
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -16,8 +16,6 @@ axum = { version = "0.7.5", default-features = false, features = [
 ] }
 dashmap = "6"
 futures = "0.3.30"
-http = "1.1"
-http-body-util = "0.1.2"
 leptos = { workspace = true, features = ["nonce", "ssr"] }
 server_fn = { workspace = true, features = ["axum-no-default"] }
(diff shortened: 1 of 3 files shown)


gbj:
Thanks for the PR. The CI was complaining about several things but I took too long to get to it, so I just fixed the `cargo fmt` issues and will let it run again.

gbj:
Just to add some clarification for the CI error, the issue here is that `cargo check --no-default-features` on `leptos_axum`.

vsuryamurthy:
There seems to be two more checks that fail. Should I just fix it although it doesnt seem to be this particular PR that is causing the fail?

gbj:
Unless I'm misunderstanding, the two checks that I see failing are `integrations/axum` failing `cargo check --no-default-features` (because of the changes in the PR) and `hackernews_js_fetch` (failing because it uses `leptos_axum` without default features).

vsuryamurthy:
Okay my bad, I see where it went wrong.

benwis:
Thanks for working on this. It looks solid to reuse the axum exports here

PR #2963, opened 2024-09-11
Added `#[track_caller]`  to `to_any_source`

Baptistemontan:
Currently when `unwrap_signal!` is called inside `ToAnySource::to_any_source` the message is:
```
thread 'tokio-runtime-worker' panicked at C:\Users\bapti\.cargo\registry\src\index.crates.io-6f17d22bba15001f\reactive_graph-0.1.0-beta5\src\signal\subscriber_traits.rs:108:29:
At C:\Users\bapti\.cargo\registry\src\index.crates.io-6f17d22bba15001f\reactive_graph-0.1.0-beta5\src\signal\subscriber_traits.rs:108:29, you tried to access a reactive value which was defined at C:\Users\bapti\Documents\Code\Rust\hex-chess\leptos_i18n\leptos_i18n\src\context.rs:121:25, but it has already been disposed.
```
the "At {location}, you tried ..." location is not very helpfull, adding `#[track_caller]` to the function provide a better hint to where the bug is.

### DIFF
--- a/reactive_graph/src/signal/subscriber_traits.rs
+++ b/reactive_graph/src/signal/subscriber_traits.rs
@@ -97,6 +97,7 @@ impl<T: AsSubscriberSet + DefinedAt> ToAnySource for T
 where
     T::Output: Borrow<Arc<RwLock<SubscriberSet>>>,
 {
+    #[track_caller]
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

PR #2964, opened 2024-09-12
fix: do not panic in automatic Track implementation if source is disposed

gbj:


maccesch:
Thanks! Looks good. Will try it out tomorrow

PR #2966, opened 2024-09-12
make ParamsMap have multiple values per key

### LINKED ISSUE #2882: Multi-value query-string parameters discard all but last
**Describe the bug**
Passing a query string like `?param=a&param=b` redirects the page (or rewrites the location) to `?param=b`. This seems partially intentional since `ParamsMap` maps `String` to `String`, but as you can see in the example, serializing a form can itself produce multiple values.

**Leptos Dependencies**

```toml
leptos = { version = "0.6" }
leptos_meta = { version = "0.6" }
leptos_router = { version = "0.6" }
```

**To Reproduce**
1. Clone shivjm/leptos-multi-value-params-mwe.
2. Run <kbd>cargo leptos serve</kbd> (or <kbd>watch</kbd>).
3. Open localhost:3000.
4. Check the boxes.
5. Click on Submit.
6. Note that the URL in the address bar and the query string provided by Leptos are correct but the `ParamsMap` is not.
7. Either refresh the page or navigate to the same URL.
8. Note that the URL in the address bar has dropped one value for the parameter.

**Expected behavior**
Multi-value parameters should be collected into a `Vec` or equivalent structure. I imagine this would have to be a breaking change since `ParamsMap` currently maps `String` to `String`.

**Screenshots**

kczimm:
Attempts to fix #2882.

### DIFF
--- a/router/src/hooks.rs
+++ b/router/src/hooks.rs
@@ -101,7 +101,9 @@ where
     let get = Memo::new({
         let key = key.clone_inplace();
         move |_| {
-            query_map.with(|map| map.get_str(&key).and_then(|s| s.parse().ok()))
+            query_map.with(|map| {
+                map.get_str(&key).and_then(|value| value.parse().ok())
+            })
         }
     });
 
--- a/router/src/hooks.rs
+++ b/router/src/hooks.rs
@@ -101,8 +101,13 @@ where
     let get = Memo::new({
         let key = key.clone_inplace();
         move |_| {
-            query_map
-                .with(|map| map.get(&key).and_then(|value| value.parse().ok()))
+            query_map.with(|map| {
+                map.get(&key).and_then(|values| {
+                    // QUESTION: in a world where we have multiple values,
+                    // which do we want here?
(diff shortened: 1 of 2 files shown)


gbj:
Thanks very much for the PR! I added a couple commits with a few changes:
1) I want to keep the default `.get()` behavior so I added a `.get_all()` that returns all options, and a `.get()` that behaves as previously
2) The test is panicking on tests without the `ssr` feature due to using a browser API, so I just gated it
3) Just a nit, I swapped the `position()` followed by an index (in `.insert()`)  with `.find()` as it gives us direct access to the found object.
4) Note that #2966 was not actually fixed, here; specifically, because the `FromIterator` was *not* using the correct insert behavior, so it was still only catching the first case when we were actually generating a `ParamsMap` from URLs.

I've tested the updated version and it should be good to go. Thanks again for your work!

kczimm:
Thanks for the changes! I appreciate your work on the project and I hope I can help with other issues to get the next major release out sooner.

PR #2967, opened 2024-09-12
Fix an error with optionnal props for islands

Baptistemontan:
Currently with the "hydrate" feature this code
```rust
#[island]
pub fn Test(#[prop(optional)] _test: Option<bool>) -> impl IntoView {}
```
produce this error:

```
 --> src\app.rs:6:31
  |
6 | pub fn Test(#[prop(optional)] _test: Option<bool>) -> impl IntoView {}
  |                               ^^^^^
  |                               |
  |                               expected `bool`, found `Option<bool>`
  |                               arguments to this method are incorrect
  |
  = note: expected type `bool`
             found enum `std::option::Option<bool>`
note: method defined here
 --> src\app.rs:6:31
  |
6 | pub fn Test(#[prop(optional)] _test: Option<bool>) -> impl IntoView {}
  |                               ^^^^^-------------
help: consider using `Option::expect` to unwrap the `std::option::Option<bool>` value, panicking if the value is an `Option::None`
  |
6 | pub fn Test(#[prop(optional)] _test.expect("REASON"): Option<bool>) -> impl IntoView {}
```

This PR fix the usage of optionnal props with islands

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -416,41 +416,61 @@ impl ToTokens for Model {
             let island_props = if is_island_with_children
                 || is_island_with_other_props
             {
-                let (destructure, prop_builders) = if is_island_with_other_props
-                {
-                    let prop_names = props
-                        .iter()
-                        .filter_map(|prop| {
-                            if prop.name.ident == "children" {
-                                None
-                            } else {
-                                let name = &prop.name.ident;
-                                Some(quote! { #name, })
-                            }
-                        })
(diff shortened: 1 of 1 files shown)


Baptistemontan:
Same issue exist in 0.7 too

benwis:
Interesting, not sure I've tried this before. Thanks for the PR!

benwis:
@Baptistemontan you'll probably want to reproduce this against main, as that's the branch that will be used for 0.7 and future updates

PR #2969, opened 2024-09-13
fix: allow component to use span prop when tracing feature is enabled

luoxiaozero:
76d081ca-1eea-447b-90ec-11930c0eebfb


gbj:
Thanks!

Issue #2946, opened 2024-09-07
[0.7] Store field getters should inherit `pub` modifier from their respective fields


brofrain:
**Is your feature request related to a problem? Please describe.**
Currently, getters for store fields are always public, which prevents hiding them outside their modules.

**Describe the solution you'd like**
Whether a getter is `pub` or not should be copied from the original field

gbj:
Ah right ok, so I just opened this up to see — The reason that this is not so straightforward is that currently the implementation uses an extension trait so that it can generate implementations for foreign types
```rust
trait TodosStoreFields<AnyStoreField>
where
    AnyStoreField: reactive_stores::StoreField<Value = Todos>,
{
    fn user(self) -> reactive_stores::Subfield<AnyStoreField, Todos, String>;
    fn todos(self) -> reactive_stores::Subfield<AnyStoreField, Todos, Vec<Todo>>;
}

impl<AnyStoreField> TodosStoreFields<AnyStoreField> for AnyStoreField
where
    AnyStoreField: reactive_stores::StoreField<Value = Todos>,
{
    fn user(self) -> reactive_stores::Subfield<AnyStoreField, Todos, String> {
        reactive_stores::Subfield::new(
            self,
            0usize.into(),
            |prev| &prev.user,
            |prev| &mut prev.user,
        )
    }
    fn todos(self) -> reactive_stores::Subfield<AnyStoreField, Todos, Vec<Todo>> {
        reactive_stores::Subfield::new(
            self,
            1usize.into(),
            |prev| &prev.todos,
            |prev| &mut prev.todos,
        )
    }
}
```

Rust does not allow us to directly implement methods on, say, `Store<Todos>` — it needs to be done either with a local newtype or with an extension trait. The extension trait seems far preferable here.

But the problem is that trait methods share the visibility of the trait, and cannot be set individually. So this is not really possible to do with the current design.

brofrain:
Hmm
Would it be possible to generate a separate trait for each field?
So something like this:

```rust
trait TodosStoreUserField<AnyStoreField>
where
    AnyStoreField: reactive_stores::StoreField<Value = Todos>,
{
    fn user(self) -> reactive_stores::Subfield<AnyStoreField, Todos, String>;
}

impl<AnyStoreField> TodosStoreUserField<AnyStoreField> for AnyStoreField
where
    AnyStoreField: reactive_stores::StoreField<Value = Todos>,
{
[...]

PR #2975, opened 2024-09-15
feat(#2946): add #[store(skip)] for #[derive(Store)]

brofrain:
The feature has already been implemented in https://github.com/leptos-rs/leptos/pull/2958 for `more-stores` branch, but it got lost. Assuming it wasn't intentional, I've restored it here.


gbj:
Thanks! Probably my mistake.

PR #2977, opened 2024-09-16
Two-way data binding

maccesch:
This allows to do the following:

```rust
// You can use `RwSignal`s
let is_awesome = RwSignal::new(true);
let sth = RwSignal::new("one".to_string());

// And you can use split signals
let (text, set_text) = signal("Hello world".to_string());

view! {
    // Use `bind:checked` and a `bool` signal for a checkbox
    <input type="checkbox" bind:checked=is_awesome />

    // Use `bind:group` and `String` for radio inputs
    <input type="radio" value="one" bind:group=sth />
    <input type="radio" value="two" bind:group=sth />
    <input type="radio" value="trhee" bind:group=sth />

    // Use `bind:value` and `String` for everything else
    <input type="text" bind:value=(text, set_text) />
    <textarea bind:value=(text, set_text) />
}
```

Without template:

```rust
checkbox.bind(Checked, is_awesome);

radio_element.bind(Group, sth);

...

textarea.bind(Value, (text, set_text))
```

## Inspiration

This is closely modelled after Svelte's binding (Qwik seems to use a similar scheme). It is the same as how Vue does it but it's syntax doesn't fit Leptos' idiom. Angular's even less so. React and SolidJS don't have two-way data binding. Finally, Sycamore takes the same approach as this but the docs don't mention radio inputs.

## Limitations

Right now this only works for HTML Elements. Components are not supported. I suspect it won't be that useful for components, because you can design them to take a `RwSignal` as prop which kind of serves the same purpose.

Also, right now, only `bool`s are supported for checkboxes and `String`s for everything else. A neat future addition could be to use `codee` codecs to allow numbers in number inputs, enums in radio groups and so on.


benwis:
@maccesch This is fascinating, ergonomically I think it's a step forward for sure. If you'd like to fix the conflicts I'll get it merged in

maccesch:
@benwis Great, thanks! Done ✅

gbj:
Thanks! Give me a couple days to look through it and make sure I understand whats going on, and to test out how we handle some of the more dangerous edge cases of two-way binding.

maccesch:
@gbj Sounds good. Excited to see what you find!

jquesada2016:
Didn't we already have support for this via specifying the event that would trigger the two-way data binding? Like `input` for value, and the user could specify a custom event for other attributes?

gbj:
Thanks so much! My changes to remove generic rendering for compile-time/linking reasons (#3015) broke this pretty much entirely, so I went in and cleaned that up, and added it to the `todomvc` example for testing purposes.

I think my concerns about two-way binding are probably just overly cautious. This is just a *much* nicer way of doing something we already support (and encourage!) doing. I do think the separated reader-writer pattern is just for encoding unidirectional data flow in the type system as much as possible, so supporting the separated signal pairs here is (IMO) really useful and not just a workaround for our default pattern.

(For the curious: The concern with two-way binding is always that if you are passing a bunch of what we would call `RwSignal` all over the place, and then binding them to inputs, input in one part of the app can cause mutations and reruns in a totally different part of the app, in a way that's quite hard to trace. I'd still recommend passing component props with the least mutation necessary, i.e., defaulting to `ReadSignal` or `Signal` or `impl Fn() -> T` or whatever for props. But for a controlled input you need both the read and the write anyway, so this is just a much nicer syntax sugar.) 

**A note for anyone doing SSR**: This has no effect on the server. (That's a prop, in a sense). It is the equivalent of `prop:value=foo on:____=move |ev| foo.set(/* ... */)`. If you need a value rendered *in the HTML*, use `value=` along with this as previously.

gbj:
(PS I'll go ahead and merge this assuming the CI passes, or I'll fix the CI if not. Or, if I fall asleep, Ben should feel free to merge :-) )

benwis:
Thanks @maccesch !

PR #2981, opened 2024-09-16
fixed lifetimes for `PartialPathMatch`

Baptistemontan:
`PartialPathMatch` has only one use, be used with `PossibleRouteMatch::test`, where everything is used around the `'a` lifetime, but `remaining` and `matched` being private you need to use the accessors, but they are `fn(&self) -> &str`, so the lifetime is not tied to `'a` but tied to the struct iself. This makes `*Segment::test` unsable in a public API as the result can't be used.

This PR only change the lifetime of the returned str of the accessors to match the `'a` lifetime.


gbj:
Thanks!

PR #2989, opened 2024-09-18
Optimize inert HTML elements

gbj:
One of the main architectural shifts between Leptos 0.1-0.6 and 0.7 is the shift from a type-erased `View` enum to a fully-typed view tree. This enables the compiler to generate much more efficient code, because it knows (for example) that `move || something.get().to_string()` can *only ever* generate `String` text nodes, and never any other kind of view: so it does not need to emit code to handle swapping one branch of the old `View` enum for another. This has led to significant binary-size reductions on the order of 20-30% in many example applications.

On the other hand, creating deeply-nested generic types seems to have catastrophic effects on compile times. In particular, representing a big tree of static HTML with many nested components as a tree of `HtmlElement<El, Attributes, Children, Renderer>` structs creates massive slowdowns in rustc as that tree becomes larger. 0.7 uses the Rust *language* in a perfectly reasonable way to compose behavior, but in a way that the Rust *compiler* cannot actually handle very well in terms of its own performance.

This PR optimizes a specific case: Within the `view` macro, it replaces children that consist entirely of plain, static HTML with no dynamic attributes or dynamic children with a single `InertElement` type, compiling that segment of the tree into a static string.

This has several benefits:
- decrease compile times slightly in the general, relatively-dynamic case by optimizing the inert (non-dynamic) parts of the page
- decrease compile times massively for applications with large static portions of HTML in the view macro, because the proc-macro code to generate the static HTML is significantly faster than the trait-based approach for these cases
- increases server rendering performance marginally by pushing one big static string rather than many small static strings, which on average will reduce reallocations slightly
- reduces WASM binary sizes by some amount by shipping inert portions of client-rendered apps as a single static str and a pair of `createElement`/`innerHTML` calls
- increases hydration speed, because inert elements can simply be skipped over rather than walked through entirely

I can't see any downside to this PR (granted that anything it breaks can and should be fixed.) I'd appreciate feedback from people with real-sized applications who can help test compile time improvements.

### DIFF
--- a/examples/js-framework-benchmark/src/lib.rs
+++ b/examples/js-framework-benchmark/src/lib.rs
@@ -162,22 +162,24 @@ pub fn App() -> impl IntoView {
             <table class="table table-hover table-striped test-data">
                 <tbody>
                     <For
-                        each={move || data.get()}
-                        key={|row| row.id}
+                        each=move || data.get()
+                        key=|row| row.id
                         children=move |row: RowData| {
                             let row_id = row.id;
                             let label = row.label;
                             let is_selected = is_selected.clone();
-                            ViewTemplate::new(view! {
-                                <tr class:danger={move || is_selected.selected(Some(row_id))}>
-                                    <td class="col-md-1">{row_id.to_string()}</td>
-                                    <td class="col-md-4"><a on:click=move |_| set_selected.set(Some(row_id))>{move || label.get()}</a></td>
-                                    <td class="col-md-1"><a on:click=move |_| remove(row_id)><span class="glyphicon glyphicon-remove" aria-hidden="true"></span></a></td>
-                                    <td class="col-md-6"/>
-                                </tr>
-                            })
+                            template! {
+                                < tr class : danger = { move || is_selected.selected(Some(row_id)) }
+                                > < td class = "col-md-1" > { row_id.to_string() } </ td > < td
+                                class = "col-md-4" >< a on : click = move | _ | set_selected
+                                .set(Some(row_id)) > { move || label.get() } </ a ></ td > < td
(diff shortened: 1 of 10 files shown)


benwis:
I’m pretty sure you meant decreasing compile times in your write up, excited!

gbj:
lmao 
> increase compile times **massively**

rip

gbj:
Oops the CSR implementation is slightly broken because it, e.g., creates a `<button>` and then sets its innerHTML to include `<button>`, so it nests every element. Will fix obviously.

VincentBerthier:
Debug mode:
```toml
[profile.dev]
incremental = true
opt-level = 0
codegen-units = 256
debug = 1 # Reduce debug info
lto = false

[profile.dev.package."*"]
opt-level = 3 # optimize dependencies aggressively
debug = 1 # we don’t need much debug info for dependencies

```

`tokio_unstable` and `web_sys_unstable_apis` are `cfg` rustflags.
For `x86_64-unknown-linux-gnu`, I’m using `mold` and `cranelift`.

Not sure the from-scratch are relevant, but since I have them…

With 0.7.0-beta5 (patched to commit 642d6fc72b4ba03e93b384176a631dd692b9ebd8):
- From scratch: Front 188s, Server 359s
- Incremental (so just the app): Front 117s, Server 130s

With 0.7.0-beta5 (patched to this branch):
- From scratch: Front 185s, server 346s
- Incremental (so just the app): Front 77s, Server 125s

I’m attaching the full reports, results to take with a pinch of salt considering that it’s only one run obviously.

[timings.tar.gz](https://github.com/user-attachments/files/17046438/timings.tar.gz)

PR #3016, opened 2024-09-24
Fix custom root when not using islands.

zakstucke:
Follow on from #2925 to fix the hydration when not using islands and a custom root.


ChristopherBiscardi:
looks right to me 👍🏼

benwis:
Same, Thanks for the PR

PR #3027, opened 2024-09-26
Read and write for StoredValue

zakstucke:
`try_read_value()` + `read_value()` + `try_write_value()` + `write_value()` for `StoredValue`

I just derived from the `ReadUntracked` and and `Writeable` traits, which I think works.


gbj:
Beautiful! Thank you.

Issue #2976, opened 2024-09-15
`Signal<T, S = SyncStorage>` doesn't implement `Track`


maccesch:

gbj:
Note that it's not possible to implement `Track` as distinct from the behavior of `Get` for `Signal`, because derived signals can't be tracked without being called. Implementing `Track` here would cause (IMO) surprising behavior, because while calling `.track()` on a `ReadSignal` or `Memo` or other similar node simply tracks that node, calling `.track()` on a `Signal` potentially invokes that inner function, which may be relatively expensive.

A derived `Signal` that needs to be `.track()`ed is better represented as a `Memo`. 

If it's just a question of API design and you want to be able to take `#[prop(into)] Signal<_>` or something and then track it, you can use `with(|_| ())`, which will not clone the values of a `ReadSignal` or `Memo` that is converted into a signal, but will invoke a derived signal. (Of course this is basically what a `Track` implementation would do, so maybe I'm overthinking it.)

Open to being wrong.

zakstucke:
Chiming in, not having track seems to prevent `read()`/`write()` being possible on signals, I'm finding these so ergonomic it would be unfortunate.

gbj:
> prevent `read()`/`write()` being possible on signals

Again, this is because it *isn't* possible, and the trait system is trying to avoid lying to you, arguably at the expense of ergonomics. In fact, `.read()`/`.write()` will likely *never* be possible. `.with()` fakes it by actually calling `.get()` and cloning, but for derived signals there *is* not underlying read or write lock being returned.

I want to distinguish really carefully between "possible on signals" and "possible on the `Signal` wrapper type." `Signal` is specifically designed to allow boxing up derived signals, in addition to `ReadSignal`/`RwSignal`/`Memo` (all of which *do* support `.track()`, `.read()`, and `.write()` in the case of `RwSignal`). 

If you're in a situation where using `.read()` specifically is *necessary* and not just convenient, then you shouldn't accept `Signal`.

zakstucke:
Whoops sorry definitely didn't mean to include `write()`, I was just hoping for `read()` on the signal type. 

Understood on the way Signal fakes `.with()`, but given it would be doing that anyway, couldn't we just... do the same thing for read by putting this cloned value in the read guard? As far as I understand it this is an internal implementation detail the user doesn't need to know about anyway. I definitely may be missing something in which case sorry.

[...]

Issue #1072, opened 2023-05-20
Could `MaybeSignal` be made `Copy` to simplify generalization `T` vs `Signal<T>`?


bluenote10:
I'm struggling a lot with generalizing the `T` vs `Signal<T>` decision for component props. I'd assume as a library author we should strive for `MaybeSignal<T>`, because it offers users the maximum flexibility.

What makes working with `MaybeSignal<T>` relatively awkward is that it isn't really consistent with `Signal<T>` in terms of `Copy` / `Clone`. Let's imagine an existing component that takes some `my_prop: Signal<MyData>` and `MyData` is costly to clone (or isn't clonable at all). As long as the component uses `Signal<MyData>` it can get away with using move closures everywhere, because `Signal` is copy and everything works fine. But trying to refactoring the component to use `my_prop: MaybeSignal<MyData>` feels almost impossible in some cases, because now `my_prop` is neither `Copy` and we also don't want to (or cannot) manually clone `my_prop` before each invocation of a move closure. In such cases I don't really find a good way how to proceed. All options I see look pretty bad:
- If I stick to just `Signal<MyData>` it will cause pain for users who only have a plain `MyData` (do I force them into creating dummy signals just to satisfy the prop type?)
- I could start wrapping things perhaps into `Signal<Rc<MyData>>` to enable clonability (or make it cheap). But that is painful on user side as well, because they have to wrap everything on their side. And the refactoring of the component is also not straightforward, because lack of `Copy` requires lots of explicit `.clone()` calls (one for each move closure).

I'm wondering if it would be possible to use the same trick as in `StoredValue` for the `T` "side" of `MaybeSignal`, so that it could become `Copy` as well (only reading side)? I'd assume that this should make it much easier to refactor from `Signal` to `MaybeSignal`.

gbj:
Hm... So the "clone into closure" pattern is a very normal, albeit unfortunate, one in Rust UI libraries. It's something that Leptos tries to make easier on users by making signals `Copy`, but this is a matter of convenience—there's nothing *wrong* with
```rust
let on_click = {
  let data = my_data.clone();
  move |_| {
    do_something_with(data);
  }
};
```
`MaybeSignal<T>` *could* be backed by `StoredValue<T>`. However, there's an overhead to creating those stored values, and if `T: Copy` they don't even add any convenience. If `T: Clone` they add some convenience for you as a library author, at the expense of some runtime cost for users.

[...]

PR #3031, opened 2024-09-27
Read/ReadUntracked/Track for Signal/MaybeSignal/MaybeProp

zakstucke:
Solves #2976. 

In my opinion, given `with()` is allowed and exists, which also has the external user assumption of "I'm just reading the value", but internally it actually does a clone, in 1/3 cases, also applies to `read()` and `track()`. All three cases the user would assume no clone occurs, hence there's no benefit to not having `read()` and `track()` as ergonomic utilities and a standard interface across other signal types.

When a library needs to use `track()` etc, rather than forcing the author to use `Memo` in their external interface, this allows them to use some `Signal` wrapper type, and leaves the performance choice to the external user. 

The `track()` functions are manually implemented rather than using the trait, using the trait would mean `Read` was automatically implemented, but would do a double clone unnecessarily in the derived signal case, hence the manual `Read` implementation instead.

The `WithUntracked` impls are removed, because they're automatically implemented by the new `ReadUntracked` impls

You'll see there's a TODO on the `Read/ReadUntracked` implementations for `MaybeSignal`, I think the Clone could be avoided theoretically for the static case, but interested to hear your thoughts.


zakstucke:
Sorry messed up my rebase and git auto closed the PR haha

gbj (commented):
~~cargo fmt wants this, I think?~~ I am literally unable to use the GitHub interface, that was supposed to be a commit

zakstucke:
Haha thanks, don't merge yet, now I know you're happy with the idea, I want to get rid of the clone in MaybeSignal static case, plus find a way to use the Track trait instead of manual impls (maintaining the single clone I'm wanting by currently not using it)

zakstucke:
@gbj i've removed my todos and you can merge if you're happy (i didn't implement either)

**MaybeSignal/MaybeProp static case clone**
- I tried adding references to all the traits, but it makes the api so complex, probably not the right idea. 
- Could theoretically `Arc<>` or `StoredValue<>` it, but that would be breaking considering the types are public enums, I'm not sure whether you'd be happy with that.

**manual track impl**
Couldn't simply modify the track trait to work, the manual impls are fine imo

zakstucke:
@gbj thinking about this today, would you be open to breaking `MaybeSignal`/`MaybeProp` with `Static(T)` -> `Static(StoredValue<T>)`?

From personal experience, `MaybeSignal`/`MaybeProp` not being `Copy` is not very ergonomic, they're the types that end up being used all over component libraries. 

This benefit could warrant the breaking change and get 2 birds with one stone here: ergonomics of `Copy` and no static clone for `read`/`track` in this PR.

gbj:
There's a conversation about this in #1072, which I ultimately closed with the reasoning, "if you want a `Copy` handle to `MaybeSignal<T>` then you should just use `Signal<T>`, because `Signal<T>` implements `From<T>` anyway."

However, getting `T` out of a `Signal<T>` does require cloning it. 

That said, the only real benefit to `MaybeSignal` is if you are separately handling the static and the dynamic halves of the enum, in which case `.read()` doesn't really make sense anyway.

I am wondering if actually deprecating `MaybeSignal` and creating a `Static` branch of `Signal<_>`,  which creates a `StoredValue<T>` and therefore can return a read guard, might actually make the most sense here?

zakstucke:
This sounds like a decent idea actually! I'll play around with this. 

Out of interest has anyone benchmarked the cost of `T -> StoredValue<T>`?

zakstucke:
@gbj I implemented the static variant of `Signal`.

I think this should actually make `Signal` much more efficient: T's were previously auto morphing into `Signal`'s using `derive()`, meaning clone everywhere on usage rather than no-ops, now `From<T>` doesn't require `Clone` and has clone free reading.

`StoredValue` didn't have a backing `ArcStoredValue` variant, I'm assuming this wasn't done for convenience, but was needed for this so I implemented it. It meant it made sense to extract the `$_value` methods into their own traits too.

**I'm not sure on the method naming for the nonreactive/`Static` `Signal` variant. I've preliminarily gone with `nonreactive()`/`nonreactive_local()`, because static is protected, let me know if something's better!**

In general now `From<T>` doesn't require `Clone`, assuming `T -> StoredValue<T>` is more or less neglible when taking everything else into account, recommending `Signal` for component libs makes sense and as you say maybe even deprecating `MaybeSignal`.

zakstucke:
Changed `Signal::nonreactive()` -> `Signal::stored()` (and `stored_local`)

reactive_graph/src/owner/arc_stored_value.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::{
+    fmt::{Debug, Formatter},
+    hash::Hash,
+    panic::Location,
+    sync::{Arc, RwLock},
+};
+
+/// A reference-counted getter for any value non-reactively.
+///
+/// This is a reference-counted signal, which is `Clone` but not `Copy`.
```
gbj, line 14: nit: Here and elsewhere (in the new StoredValue traits), these docs refer to these as "signals" but they're really not; I'd probably just replace "signal" with "value".

(Presumably just copy-pasted from the other signal docs)

reactive_graph/src/traits.rs
```diff
@@ -639,3 +639,199 @@ pub fn panic_getting_disposed_signal(
...
+    /// or `None` if the signal has already been disposed.
+    #[track_caller]
+    fn try_update_value<U>(
+        &self,
+        fun: impl FnOnce(&mut Self::Value) -> U,
+    ) -> Option<U>;
+
+    /// Updates the value by applying a function, non-reactively, returning the value returned by that function.
+    ///
+    /// # Panics
```
gbj, line 771: Current behavior in both 0.6 and 0.7 is that `update_value` 1) doesn't return `U` and therefore 2) is guaranteed never to panic -- if the value has already been disposed it simply does nothing. I think we should stick to this pattern.

reactive_graph/src/traits.rs
```diff
@@ -639,3 +639,199 @@ pub fn panic_getting_disposed_signal(
...
+    /// passed in. Otherwise, returns `None`.
+    #[track_caller]
+    fn try_set_value(&self, value: Self::Value) -> Option<Self::Value>;
+
+    /// Updates the value by replacing it, non-reactively.
+    ///
+    /// # Panics
+    /// Panics if you try to access a signal that has been disposed.
+    #[track_caller]
+    fn set_value(&self, value: Self::Value) {
```
gbj, line 814: `set_value()` definitely doesn't need to panic, here, as it's not even returning a value.

gbj (commented):
Thanks so much. This is looking really good!

I left a couple minor comments. (In fact looking at the current docs on panicking for 0.7 I think they're just currently wrong for `update_value()` and `set_value`, neither of these needs to panic if they don't return values.

I really appreciate you taking the time, not only to make the PR but to understand things deeply enough to be able to make it!

reactive_graph/src/traits.rs
zakstucke: Yep fair enough, I think I'd copied from StoredValue so the API had changed, but for the set one I was thinking about it as if each disposed error is a downstream user error hence they should see it straight away. But yeah makes sense to avoid an error when possible. Updated

reactive_graph/src/traits.rs
zakstucke: Same reason as `update_value`. Updated

zakstucke:
I think I'm happy with this, I might try a few things afterwards in a separate PR, this is getting too big: 
- sorting out the `Track` trait
- adding `Read`/`Get`... traits for `Option<Signal>`
- deprecating `MaybeSignal` and `MaybeProp` (`MaybeProp` can be deprecated if the traits are added to `Option<Signal>`

gbj:
Thanks! Yes, looks good to me for now.

PR #3040, opened 2024-09-28
Double-ended-iterator-for-stores

mahdi739:
This PR implements DoubleEndedIterator for two structs: StoreFieldKeyedIter and StoreFieldIter


gbj:
The code looks good, and I tested these both out and they seem to work well -- Very nice, thanks!

PR #3044, opened 2024-09-29
Replace proc-macro-error with proc-macro-error2

SleeplessOne1917:
I ran cargo-audit on a project of mine that uses Leptos and got a warning that Leptos is using an unmaintained library: `proc-macro-error`. This PR replaces it with `proc-macro-error2`, a drop-in replacement. Interestingly, `leptos-macro` is already using the newer library.


benwis:
Cool, gotta love random changes like this. Thanks for the PR

PR #3046, opened 2024-09-29
remove unnecessary `PartialEq` trait bounds

seanaye:
This MR removes certain `PartialEq` bounds which are not required when creating a new Memo


benwis:
> This MR removes certain `PartialEq` bounds which are not required when creating a new Memo

It may not be required during creation, but do we have another way to check if a Memo needs to update?

gbj:
@benwis The two functions that this makes changes to are functions that specifically take a comparison function, rather than defaulting to using `PartialEq`.

PR #3047, opened 2024-09-29
Implement `IntoView` for `Arc<str>`

ydirson:
0.7 will be much harder to work with Rc, so this adds the Arc support in 0.6 to make it possible to transition away from Rc before taking the already-large step to 0.7.


gbj:
Thanks!

Issue #2086, opened 2023-11-28
module restructuring proposal


blorbb:
The module structure of `leptos` is almost completely flat, making it hard to find certain items or add documentation about groups of items. I discussed this briefly in https://github.com/leptos-rs/leptos/issues/2007#issuecomment-1807411843

I suggest we restructure the modules such that there are clearer sections (e.g. reactive, dom, async, components etc.. get their own module), and switch to using a prelude for common imports (`use leptos::prelude::*;` over `use leptos::*;`)

Restructuring the modules is obviously a major breaking change, and would probably also cause a lot of merge conflicts with existing PRs too. Maybe this change could be done right before 0.6, when there are probably fewer ongoing PRs and all the changes can be made at once. For the majority of users, the only changes should be changing `use leptos::*;` to `use leptos::prelude::*;`, and maybe a few more imports for more uncommon items not included in the new prelude.

Below is my idea for a potential restructure; please add your own suggestions/feedback too! hopefully the half made-up syntax makes sense. I didn't include every single module / crate, only the most commonly used ones

```rust
crate leptos_reactive {
    // move `Callback` to `leptos` instead? it's not used anywhere internally,
    // just a helper type, not really reactive either.

    // make these `pub mod` instead of re-exporting at crate root
    pub mod hydration { ... }
    // renamed from `runtime`
    pub mod runtime_primitives { ... }
    pub mod macros { ... }

    pub mod reactive {
        // traits are accessible in `crate::reactive::*` or `crate::reactive::traits::*`
        pub mod traits { SignalGet, SignalDispose, SignalWith, IntoSignal, etc... }
        pub use traits::*;

        pub mod primitives { 
            // move these files from old modules over here
            crate::{diagnostics, trigger, node} 
        }

        signals, Memo, Effect, watch, slices, selector, MaybeSignal, MaybeProp, TextProp
        // should `batch`, `on_cleanup`, `untrack`, `untrack_with_diagnostics`
        // be moved from `crate::runtime` to here?
    }

    // better name? anything that is stored in the runtime but not reactive
    pub mod stores {
        context, StoredValue
    }

    // does this need to be its own module, or is `crate::Oco` fine?
    pub mod oco { Oco }

    // rename this? i don't like the look of the trailing _
    pub mod async_ {
        // not sure where `queue_microtask` should go, not exactly async but 
        // it's mainly used in Resources
        Resource, SuspenseContext, spawn_local, queue_microtask
    }

[...]

PR #3050, opened 2024-10-01
Module restructuring and docs cleanup

gbj:
See discussion in #2086.


reactive_graph/src/traits.rs
```diff
@@ -61,6 +61,7 @@ use std::{
     panic::Location,
 };
 
+#[doc(hidden)]
 /// Provides a sensible panic message for accessing disposed signals.
 #[macro_export]
 macro_rules! unwrap_signal {
```
blorbb, line 67: If you want to make a declarative macro internal within the crate (which seems to be intended), you can use:
```rs
// No macro_export
macro_rules! unwrap_signal {...}
pub(crate) use unwrap_signal;
```

gbj: Thanks, TIL! Searching though I find that I do actually use it in `reactive_stores` for exactly the same purpose... so I will probably just hide it for now.

PR #3054, opened 2024-10-01
Migrate rkyv 0.8.x

### LINKED ISSUE #3006: Upgrade Lepros rkyv integration to support rkyv 8
Hello, currently Leptos server fns suport rkyv 7. Rkyv 8 has been recently released. Please consider upgrading to 8. 

Word of caution: v8 is not backwards compatible with 7, it will end up being a breaking change; this is a good opportunity to move to rkyv 8 with Leptos 0.7.

### LINKED ISSUE #3055: [0.7] Example `server_fns_axum` crashes
**Describe the bug**

Example `server_fns_axum` crashes

**To Reproduce**

Checkout latest `main`.

```sh
cd /examples/server_fns_axum
cargo make ci
cargo make start
```
Open browser http://127.0.0.1:3000 and check Console:

```
panicked at leptos/tachys/src/view/strings.rs:94:14:
couldn't cast text node from node


server_fns_axum.js:440:9
```

sectore:
~~Note: #3056 needs to be fixed to run and test `rkyv` with `server_fns_axum` example.~~

~~Fix #3055~~

Closes #3006


gbj:
Thanks for doing this work! I see that the PR moves the implementation of rkyv traits out from the server macro into userland, and away from sending the arguments directly into another wrapper struct. Does this mean that only a single argument can be used here, and that it needs to be a struct on which the user has implemented the rkyv traits? To me [these changes](https://github.com/leptos-rs/leptos/pull/3054/files#diff-c1c50fc156e2d3732b38b2ada07f0de007a0a790b00b893ad004ddf1d82c987fR314-R333) in the example are kind of unfortunate, as they cause the abstraction of the server fn encoding to leak out into the user's code.

Could you add some more context for this change? Is the previous form not possible to implement with rkyv 8?

sectore:
> I see that the PR moves the implementation of rkyv traits out from the server macro into userland, and away from sending the arguments directly into another wrapper struct. Does this mean that only a single argument can be used here, and that it needs to be a struct on which the user has implemented the rkyv traits? 

Thanks for pointing it out. a58e785 reverted changes of `examples/server_fns_axum/src/app.rs`, so everything looks and works as before.

benwis:
Thanks for the Pr!

PR #3063, opened 2024-10-05
Makes the `wasm32-wasip1/2` target a first-class citizen for Leptos's Server-Side

raskyld:
> **DISCLAIMER**: This is a draft, I am not a seasoned Rust developer so this PR need some attention to ensure the code is idiomatic and clean.
>
> This is a side-project, not backed by my company so I need time to make it production ready, thanks for your understanding!

## Abstract

This PR is a first step towards making Leptos capable of targeting `wasm32-wasip1` (and `wasm32-wasip2` when it is released) for the server-side.

ATM, I managed to [`wasmtime serve`](https://docs.wasmtime.dev/cli-options.html#serve) the component produced when running `cargo-leptos build` on [this codebase](https://github.com/raskyld/leptos-wasmcloud).

I started working on this PR with the end-goal of making it easy to host the server-side on a [wasmCloud lattice](https://wasmcloud.com/docs/concepts/lattice).

While wasmCloud is a **CNCF Sandbox project**, I would understand if you don't want to have anything related to this project *in-tree*. For this reason, this PR is focused on running on **purely standard WASI**. The produced component have 0 dependencies on any vendor. Which is why I think this PR could be merged *in-tree*.

For my following work, specifically focused on wasmCloud, I would do that *out-of-tree*. (Expect if maintainers have interest supporting wasmCloud as a first-class citizen aswell! 🚀)

## Features

* Add a `generic` feature to `server_fn` to make it portable across a wide range of platform.
* Add a single-threaded version of `any_spawner`'s custom executor feature (implemented by @stefnotch! 🎉) allowing the futures to run in the context of a WebAssembly runtime.
* Add an *ad-hoc* async runtime executor for the `wasm32` targets.
* Add an *integration crate* `leptos_wasi` allowing Leptos users targeting the `wasm32-wasip1` platform to make their app deployable on any WASI runtime (for now, I only tested [`wasmtime`](https://wasmtime.dev))
* Support for both synchronous response and response streaming so most `SsrMode` are compatible
* Users can choose how to serve static files, in my example implementation, I used a simple `wasi:filesystem` link provided
  by the `--dir target/site/pkg` flag of `wasmtime serve`.
* **Shortcut Mechanism**: The component don't have any "initialisation" or "warm-up" phase, all is done on the fly and there is a *shortcut* mechanism that will skip all UI Rendering and/or Reactivity setup if the user is simply requesting a static file or calling a Server Function.

## Limitations

* For now, `SsrMode::Static` is not supported, I don't think it should be too hard to do, but for now it's not on my roadmap.
* **Optimized for cold-start** : This crate makes the assumption that components are initiated for each incoming request, that is, the lifetime of the component is bound to the one of the request. For this reason, we do not do any memoization (like caching routes).

## Road-map

* [x] Ask reviewer how we could improve the changes made to `any_spawner`, I really don't like my current changes.
* [x] Add a video to show-case the end result.. Most people won't read the PR if they don't see first what it does 🙈 (I will participate to the [wasmCloud Community Call](https://wasmcloud.com/community/) and will post the replay there since I will present the MR.
* [x] Test Server Functions
* [x] Implement an async runtime.
* [x] Use the `wasi` crate for the bindings.
* [x] Wait for the repo `leptos-wasi` in the organisation
* [x] Migrate to integration to this repo
* [x] Improve the library's error handling.


benwis:
@raskyld This looks interesting. Adding generics to server fns to make them more versatile could be handy. I need to spend some time looking at what you did there.

I think you want to take a look at leptos-spin and leptos-spin-macro. We're keeping most new integrations out of the main crates to reduce the time needed to release a new feature, but I'm happy to create a leptos-wasi and leptos-wasi-macro in the leptos org if you'd like. 

As far as the Leptos specific changes, it seems like you've broken some of the examples, so you'll want to review those

raskyld:
Hello @benwis thanks a lot for your answer.

Let me check that! 
Also I kind of figured out your would off-tree some parts so that's fine for me!

I should move the whole `integration/wasi` in a standalone crate so right?
I don't have any macros yet, that's the advantage of using the new `generic` server_fn.

I will likely need an example / starter crate as well, ported from https://github.com/raskyld/leptos-wasmcloud

any_spawner/src/lib.rs
raskyld, line 1: @gbj since you are the author of this crate, could you guide me on how I implement a single-threaded implementation for running the server side in Wasm?

The current implementation works but I don't really like it.

I think I have no choice to expose a handle to `LocalPool::run()` since, unlike tokio and glib, there is no "ambiant" executor and the caller is responsible for starting the pool.

raskyld:
Potentially, waiting for the release of the next Rust release that should come with a promotion of the `wasm32-wasip2` target's Tier. It would bring a lot of nice things regarding the interop of Rust std but also I may not need to write a custom Executor.. c.f. https://github.com/tokio-rs/tokio/issues/4827

benwis:
> Hello @benwis thanks a lot for your answer.
> 
> Let me check that! Also I kind of figured out your would off-tree some parts so that's fine for me!
> 
> I should move the whole `integration/wasi` in a standalone crate so right? I don't have any macros yet, that's the advantage of using the new `generic` server_fn.
> 
> I will likely need an example / starter crate as well, ported from https://github.com/raskyld/leptos-wasmcloud

Yeah, I'd make this a whole seperate crate. You could choose to wrap it up in a single integration crate like leptos-wasmcloud, or if you want to support other wasm function implementations you can do leptos-wasi

any_spawner/src/lib.rs
raskyld: @dicej pointed me out the executor used by spin is part of the spin SDK crate, so I should definitely go and have a look!

I will also check the linked PR since that would allow me to implement that out of tree aswell! Thanks!

raskyld:
We definitely should go for `leptos-wasi`, the goal is to make the whole ecosystem benefit from that crate, individual runtimes and/or orchestrators can then implement their own logic separately.

raskyld:
^^ That may be my bad:

https://github.com/leptos-rs/leptos/pull/3063/files#diff-b0051fc93f2245d2fdd95cfff6790392036d7e84b3639181e26675ce8987a050R56-R59

I wanted to make my clippy more annoying since I count on that to grow my Rust knowledge a bit but It looks like it also applied to tests 🤔

brooksmtownsend:
Chiming in from the wasmCloud side, very in support of a `leptos-wasi` crate that uses wasm32-wasip2 😄 If there's any work to make it a "first class citizen" in wasmCloud from the standard crate we can take that on, I'd rather have Leptos work out of the box with any component implementation

integrations/wasi/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[allow(warnings)]
+pub mod bindings {
+    wit_bindgen::generate!({
+        path: "wit",
+        pub_export_macro: true,
+        world: "http",
+        generate_all,
+    });
+}
```
brooksmtownsend, line 69: One thing to consider which may simplify the changes in this PR is directly bringing in the [wasi](https://crates.io/crates/wasi/0.13.1+wasi-0.2.0) crate as a dependency (pinned to `=0.13.1+wasi-0.2.0` may be a good idea) instead of using wit_bindgen directly. The types from the `wasi` crate also have additional trait implementations, like using `std::io::Read/Write` on the stream types 😃 Here's a commit where I updated a draft to use wasi from wit-bindgen https://github.com/brooksmtownsend/reqwest/commit/1a98138b322a944b5ffb496272b0ba6cae783ef2#diff-e3a9842c622287180da65275fb59a80bf4f665098591996b5796f62ff811c41a

I'm not sure if you gave this a shot before, so let me know if this caused issues.

server_fn_macro/Cargo.toml
```diff
@@ -21,6 +21,8 @@ nightly = []
 ssr = []
 actix = []
 axum = []
+wasi = ["generic"]
```
brooksmtownsend, line 24: Once `wasm32-wasip2` is a stable target in Rust (coming in 1.82 afaik) the use of feature flags could be simplified, using the target directive instead https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-wasip2.html#conditionally-compiling-code

integrations/wasi/src/lib.rs
raskyld: That's really cool! And it also solves the issue of what types the user need to use etc to interop with the crate! Will definitely do that :D Thanks Brooks! 💪

server_fn_macro/Cargo.toml
raskyld: True! I initially went for the feature to kind of follow the other integrations but since we have a target I may just use that!

raskyld:
https://github.com/leptos-rs/leptos/pull/3091 has been merged, I will rework the PR on that point, my new approach will be to bundle a custom WASI async runtime exposed to Leptos as a `CustomExecutor`, I wanted to use a crate for that, but nothing seems "standard" atm so I will just default to an *ad-hoc* implementation.

This may be reconsidered once we add first-party support to `tokio` for **all** the `pollable`s, see https://github.com/tokio-rs/mio/pull/1836 for an exemple of early work (i.e., adding socket support to `mio`).

server_fn_macro/Cargo.toml
brooksmtownsend: edit: This is live now in Rust 1.82!

raskyld:
Implemented an async runtime which supports waiting for `Pollable` and waking specific `Futures`. 🎉 

Tested a dumb "sleep 3s" c.f. https://github.com/raskyld/leptos-wasmcloud/commit/a683304fd752945f14ffde9d40fd939acb546a21#diff-3f1994a48e642525e7a77d762690a04f386ec355bef4a97826ff0b4f54ef2239R61

This makes `Suspense` work correctly: the server return the HTML shell with the `fallback` and then streams the counter value 3 seconds later ⏱️

raskyld:
What is missing at this point is:

* See how we can enable *logging* through `wasi:logging`, I am not sure wasmtime support exposing this world though. This would help with debugging the integration crate (maybe @brooksmtownsend have some hints on the state of the art for logging in a component? 🙏)
* Improve the Error Handling, atm it's using a super generic Error type and I don't really like that. but that's more optimisation.
* See with @benwis for the repository so I can migrate the `leptos_wasi` from this PR.

brooksmtownsend:
> have some hints on the state of the art for logging in a component? 🙏

Yeah, wasmtime doesn't support `wasi:logging/logging@0.1.0-draft` since that interface hasn't been through the proposal process far enough to be integrated into wasmtime. I would recommend, if running directly in wasmtime, just using stdout/stderr for logging since that's supported via `wasi:cli`

raskyld:
Alright, I think the PR is ready-ish for a first review, but considering there is a lot of changes, we should likely start the migration to a dedicated repository.

As I test and play with this crate, I will make it incrementally more production ready (add test suites etc..).

benwis:
@raskyld I've created the leptos_wasi repo under the leptos org and added you as a maintainer. Lmk if that doesn't work and we can get started!

raskyld:
Yay! Thanks 🙏 

I will take the integration crate to a PR on this new repo after work :D 🎉

raskyld:
FYI, migration is done! I will present the PRs at the next wasmCloud Community Meeting (next Wednesday).
If you have an introduction speech somewhere so I can present Leptos more broadly, don't hesitate to tell me 🤝 !

raskyld:
I will rebase and run `make format` everywhere, but it's weird previous commit only had 1 failure 🤔 I need to check on that

raskyld:
Hopefully that should be it, I ran `fmt` target on crates I touched 🤞

raskyld:
Meh, I must be using cargo make in a wrong way or my rebase messed it, I will double check later today 😓

raskyld:
Here is the little demo I made at wasmCloud Community Call, excuse my terrible french accent that went wild 🤣 

https://www.youtube.com/watch?v=SAAyukYJGeY&t=25s

raskyld:
Oh god I've just got it!
`cargo make` wasn't using my nightly toolchain for some reason 🙃 
Now all should be gud!

raskyld:
Fixed the clippy error for `server_fn`, I am not sure why the other one fail though, all I see is CPU/Memory error from the GH Runner 🤔

raskyld:
Ah! Now I have the backtrace!
I forgot to make the features mutually exclusive, my bad, fixing it.. 🔧

benwis:
@raskyld Bet that feels good!

benwis (approved):
LGTM! Thanks for all the hard work here!

PR #3076, opened 2024-10-08
Use the `Track` trait for the `Signal` wrapper.

zakstucke:
#3031 continuation, allow overriding the auto implemented `Read::try_read` implementation from the `ReadUntracked` trait. 

Doing so in this way means the normal `Track` trait can be used, and the double-clone fix propagates automatically to `MaybeSignal`, `MaybeProp` and any user created trait derivations involving the `Signal` wrapper.

### DIFF
--- a/reactive_graph/src/serde.rs
+++ b/reactive_graph/src/serde.rs
@@ -75,7 +75,7 @@ impl<T: Serialize + 'static, St: Storage<T>> Serialize for ArcMemo<T, St> {
 
 impl<T, St> Serialize for MaybeSignal<T, St>
 where
-    T: Send + Sync + Serialize,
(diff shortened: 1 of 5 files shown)


zakstucke:
Okay, sadly this doesn't actually solve it automatically for `MaybeSignal`/`MaybeProp`/downstream impls, because you have to implement `custom_try_read` at each derivation layer. There's no non-insanely hacky way of fixing it I've found, short of not auto implementing `Read`.

Still, it goes back to using the `Track` trait, and provides a much clearer path to propagating `custom_try_read` for users if they wanted to impl something and avoid an extra clone.

gbj:
Thanks! This looks good to me.

PR #3096, opened 2024-10-10
Use full path to Result in Params derive

Arjentix:
Hello!

Just encountered this bug and decided to quickly fix it. Currently it fails if you define your own `Result` alias without second generic.

### Questions

- Should I also create a backport PR for branch with stable release?
- Should I bump version somewhere?


gbj:
Thanks for the PR! "No need" to both your questions.

PR #3098, opened 2024-10-10
Option<T> read-like traits & deprecate MaybeSignal/MaybeProp

zakstucke:
Got `Option<T>` working to automatically implement read-like traits for any `T` implementing read-like traits, so `MaybeProp` but generic for all `RwSignal`, `Memo`, `Signal`... etc. 

image
^😍 

I think my changes across the last few PR's now meet the requirements to deprecate `MaybeSignal` and `MaybeProp` but that's up to you. 

This PR depends on #3076, so this shouldn't be merged until after that's finished with and merged. 

The only actual change in this PR is the new `trait_options.rs` file, everything else is changes from the unmerged #3076 or deprecation stuff. **Edit**: and some more `From<>` impls for `Signal`.

I could implement `Write`-like traits too, wouldn't be hard, each op would just be a noop when `None`, but maybe this is too weird.

### DIFF
--- a/examples/errors_axum/src/error_template.rs
+++ b/examples/errors_axum/src/error_template.rs
@@ -7,7 +7,7 @@ use leptos_axum::ResponseOptions;
 // Feel free to do more complicated things here than just displaying them.
 #[component]
 pub fn ErrorTemplate(
-    #[prop(into)] errors: MaybeSignal<Errors>,
+    #[prop(into)] errors: Signal<Errors>,
 ) -> impl IntoView {
     // Get Errors from Signal
     // Downcast lets us take a type that implements `std::error::Error`
--- a/examples/errors_axum/src/error_template.rs
+++ b/examples/errors_axum/src/error_template.rs
@@ -6,9 +6,7 @@ use leptos_axum::ResponseOptions;
 // A basic function to display errors served by the error boundaries.
 // Feel free to do more complicated things here than just displaying them.
 #[component]
-pub fn ErrorTemplate(
-    #[prop(into)] errors: Signal<Errors>,
-) -> impl IntoView {
+pub fn ErrorTemplate(#[prop(into)] errors: Signal<Errors>) -> impl IntoView {
     // Get Errors from Signal
     // Downcast lets us take a type that implements `std::error::Error`
     let errors = Memo::new(move |_| {
--- a/examples/slots/src/lib.rs
+++ b/examples/slots/src/lib.rs
@@ -10,7 +10,7 @@ struct Then {
 // the type with Option<...> and marking the option as #[prop(optional)].
 #[slot]
 struct ElseIf {
-    cond: MaybeSignal<bool>,
+    cond: Signal<bool>,
     children: ChildrenFn,
 }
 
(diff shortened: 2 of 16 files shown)


gbj:
So, this is objectively nice, and a good idea to implement the traits on `Option<T>`.

I'd note that it's *different* from `MaybeProp`, which is more like `Option<Signal<Option<T>>>` but flattens them into a single `Option<T>` when you read it, whereas using the method in this PR yields an `Option<Option<T>>`.

I'd like to review in a little more depth. It also seems there is some overlapping code between this and the other PR, which led to a merge issue after I merged the other one, and also some duplicate code to review... Would you mind fixing the merge issues? Then I'll take a look again.

zakstucke:
Fixed conflicts!

You're right I was forgetting `MaybeProp` equated to `Option<Signal<Option<T>>>`, I've made some updates to provide feature parity with `MaybeProp`:
- More `From<>` impls
- A `OptProp` (plz rename haha) type alias for `Option<Signal<Option<T>>>`
- A `flatten()` trait helper method for `Option<&Option<T>>`

Aside from the non `StoredValue` static `T` case, after #3105 there should be complete feature parity with just using basic signals but the added benefit of:
- Native setting to dom nodes e.g. see #3104
- Much improved `From<>` type inference
- Less burden inside leptos to manage a all of the `impls` needed for `MaybeSignal`/`MaybeProp` to be as dynamic as `Signal`

image

gbj:
I'm fine with deprecating `MaybeSignal` given that the changes to `Signal`.

I don't think it's a good idea at the moment to deprecate `MaybeProp`, given that the flattening alternative is also available but requires additional steps.

I don't love the idea of continuing to multiply the set of possible options by adding something like `OptProp` into the mix. I think this is a perfectly reasonable type alias for someone to create if they find themselves using a lot of `Option<Signal<Option<T>>>` but I'm not sure having it in the core library adds much value, and I don't understand why the macro would rewrite the type instead of just using the type alias.

zakstucke:
The reason I internalised `OptProp` is the component macro can't currently handle type aliases of `Option<_>`  hence I have to replace it inside the proc macro.

But I definitely see your point, I'll try and do it the proper way: improve the proc macro so it can handle type aliases and not need to internalise `OptProp`.

zakstucke:
@gbj after playing around with it some more, the type alias has to be internalised like it is now and I did it like that for a reason 🥲 .

rust-type-builder, the view macro and component macro all rely on seeing the `Option` ident and there doesn't seem to be a way to get that from the type alias at proc time because semantic analysis hasn't been done yet.

So it's either no alias, or alias in leptos and not downstream, or you see something I'm missing. Let me know what you think, I would prefer to have it, but i'd understand the veto! Maybe `PropAlias` is a better name.

zakstucke:
@gbj if we're not going to deprecate `MaybeProp` and not include the type alias, are you open to me making it Copy? It'll be breaking I think as requires slightly different bounds going from `MaybeSignal` to `Signal`.

zakstucke:
I went ahead and did it, I actually don't think this is a real breaking change, just requires some different bounds, but nothing major extra than before.

If CI passes I'd be happy if this was merged as it is now.

gbj:
Okay, thanks. I'm happy with this at this point. I'm going to merge this, and then separately un-deprecate `MaybeProp` rather than complicating a large PR further by trying to change it!

PR #3104, opened 2024-10-12
impl IntoClass for Option<impl IntoClass>

zakstucke:
To make `view!{<div class={Option<impl IntoClass>} />}` work and an extra `Signal` impl.

### DIFF
--- a/reactive_graph/src/wrappers.rs
+++ b/reactive_graph/src/wrappers.rs
@@ -830,6 +830,20 @@ pub mod read {
         }
     }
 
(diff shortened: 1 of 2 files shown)


benwis:
Interesting, thanks for the PR. Doing anything interesting with your classes?

zakstucke:
@benwis thanks, this is just to get something like this working:

```rust
#[component]
pub fn Foo(#[prop(optional)] class: Option<Signal<String>>) -> impl IntoView {
    view! { <div class={class} /> }
}
```

PR #3105, opened 2024-10-12
New `nostrip:` prop prefix to pass `Option<T>` directly when `prop(optional)`

zakstucke:
Currently you can't provide `Option<T>` to props with `prop(optional)` because they auto-wrap, which can be very annoying when passing props through intermediary components. There isn't a way around it without leaving the `view!{}` macro or using `optional_no_strip` which has it's own tradeoffs and applies globally. 

A new prop prefix, `nostrip:` solves this really nicely. This is made possible by [rust-type-builder](https://github.com/idanarye/rust-typed-builder/blob/master/CHANGELOG.md) recently adding a config for fallback builder methods to not strip optional.

If you have a better, shorter name, or a way for this to be automatic rather than something you manually specify, let me know!

image

### DIFF
--- a/leptos_macro/src/view/component_builder.rs
+++ b/leptos_macro/src/view/component_builder.rs
@@ -274,6 +274,8 @@ pub(crate) fn component_to_tokens(
     let mut component = quote! {
         {
             #[allow(unreachable_code)]
+            #[allow(unused_mut)]
+            #[allow(clippy::let_and_return)]
             ::leptos::component::component_view(
                 #[allow(clippy::needless_borrows_for_generic_args)]
                 &#name,
--- a/leptos_macro/src/view/component_builder.rs
+++ b/leptos_macro/src/view/component_builder.rs
@@ -1,4 +1,6 @@
-use super::{fragment_to_tokens, TagType};
+use super::{
+    fragment_to_tokens, utils::is_nostrip_optional_and_update_key, TagType,
+};
 use crate::view::{attribute_absolute, utils::filter_prefixed_attrs};
 use proc_macro2::{Ident, TokenStream, TokenTree};
 use quote::{format_ident, quote, quote_spanned};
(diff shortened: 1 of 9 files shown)


zakstucke:
Converted to draft, I think I can improve this to a more generic `opt:` that allows `Option<T>` for any component prop that expects `T`. 

With new idea could also apply to html els and work with e.g. `<div opt:node_ref=Option($) />`

zakstucke:
No not worth it, it get's way more complicated than this, and the extra cases it solves have easy alternatives already, it's just for this case that `nostrip:` is really useful.

zakstucke:
I've switched to the alternate version requiring nothing new from `rust-type-builder` but I switched for a different reason and haven't extended the functionality: the fallback optional methods wouldn't call `.into()` on T, but is possible with the new method. Only downside is `nostrip:` will treat all props as `prop(optional, into)` whether they have `into` or not, but I think this is a fine edge case.

This should be ready now.

leptos_macro/src/view/component_builder.rs
```diff
@@ -267,11 +277,15 @@ pub(crate) fn component_to_tokens(
             ::leptos::component::component_view(
                 #[allow(clippy::needless_borrows_for_generic_args)]
                 &#name,
-                ::leptos::component::component_props_builder(&#name #generics)
-                    #(#props)*
-                    #(#slots)*
-                    #children
-                    .build()
+                {
+                    let mut props = ::leptos::component::component_props_builder(&#name #generics)
```
gbj, line 281: This pattern has the potential to generate `unused_mut` and `clippy::let_and_return` warnings. In my current version, I don't see clippy applying these to the macro version but if I expand the macro code, it does apply those warnings, so it might start applying them in the future. It would be good to add these to the list of allowed lints above.

zakstucke: Added!

PR #3115, opened 2024-10-15
feat: add two-way data binding support for stores

osmano807:
This add support for directly using `Store` fields with the new two-way `bind:` for elements.
Needed only to `impl IntoSplitSignal` for `Subfield`, also added a impl for `KeyedSubfield` but don't know the utility yet.
Tested with `Subfield`

```rust
use reactive_stores::{Field, Patch, Store};

#[derive(Debug, Store, Patch, Serialize, Deserialize, Clone)]
struct User {
    name: String,
    email: String,
}

let user = Store::new(User {
  name: "John Doe".to_string(),
  email: "john.doe@example.com".to_string(),
});

view! {
  <input type="text" bind:value=user.name() />
  <span>{move || user.name().get()}</span>
}
```

I had to add `reactive_stores` as a feature of `tachys` in the leptos Cargo.toml, not sure if that's a good approach, as leptos crate doesn't depend on the stores right now.

### DIFF
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -28,7 +28,7 @@ paste = "1.0"
 rand = { version = "0.8.5", optional = true }
 reactive_graph = { workspace = true, features = ["serde"] }
 rustc-hash = "2.0"
-tachys = { workspace = true, features = ["reactive_graph", "oco"] }
+tachys = { workspace = true, features = ["reactive_graph", "reactive_stores", "oco"] }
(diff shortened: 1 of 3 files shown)


gbj:
This looks great, thanks! I'm fine with adding the feature to tachys.

gbj:
Ah looks like the extra feature combo causes CI to die on `tachys` due to running out of disk space, I can fix that separately though.

PR #3118, opened 2024-10-15
Allow reading resources in effect outside suspense

zakstucke:


gbj:
Looks good, thanks. In theory this is some additional overhead that will never be used in release mode, but it is only on user effects not on rendering effects so I'm fine with it.

PR #3119, opened 2024-10-15
`impl IntoStyle` for `Option<impl IntoStyle` & `Oco`

zakstucke:


tachys/src/html/style.rs
```diff
@@ -170,6 +170,80 @@ pub trait IntoStyle: Send {
...
+            (el.clone(), None)
+        }
+    }
+
+    fn rebuild(self, state: &mut Self::State) {
+        let el = &state.0;
+        let prev_state = &mut state.1;
+        let maybe_next_t_state = match (prev_state, self) {
+            (Some(_prev_t_state), None) => {
+                Rndr::remove_attribute(el, "style");
```
gbj, line 209: I don't think this logic quite works, because `IntoStyle` doesn't do the same thing for every type. For example, the implementation for the tuples (which is what the `style:color=foo` kind of syntax becomes) updates individual pieces of the `CSSStyleDeclaration`, rather than setting and unsetting the `style`. This `Option<T>` logic definitely works for the string-like types but not the tuple-like types.

zakstucke: @gbj  If you're meaning here this `Option<impl IntoClass>` applies to e.g. `style:foo` it doesn't that's implemented as `impl<F, S> IntoStyle for (&'static str, F)` so would be `(&'static str, Option<F>)` for the optional case, something else not matched by this impl.

If you're meaning the case where e.g. `style=foo.then_some("color: red") style:background="transparent"` in the same tag, then yes there's a problem here as e.g. setting the `foo` to `None` would remove the `background` style too, but: 
1. I don't know if this is actually supported, both `style=""` and `style:attr` in the same tag.
2. I just tried solving by just changing the impl to using `T::default()` and calling the `T` trait methods directly, instead of the `None` handling and removing the attribute when `None` in this impl,  but this issue would still persist, so this would be a bug elsewhere in the renderer anyway if both versions in same tag is a supported patter.

For 2. you can see this happens here by just toggling the b prop from a parent component from true to false: 
```rust
#[component]
pub fn Sub(b: bool) -> impl IntoView {
    view! {
        <p style={ if b {"color: red" } else { "" }} style:background="transparent">"Sub"</p>
    }
}
```

gbj: Here's an example of what I mean:
```rust
#[component]
    let (value, set_value) = signal(0);

    div()
        .style("font-weight: bold")
        .style(move || (value.get() % 2 == 0).then_some(("color", "red")))
        .on(ev::click, move |_| *set_value.write() += 1)
        .child("Click me")
}
```
When you click this, the `rebuild()` definition for the inner `Option<(&'static str, &'static str)>` will un-set the entire `style` attribute, not just the `color` property.

This can't be created using the view macro, but can be created pretty easily using the builder.

gbj: No; this is not the same problem. The problem you describe in #3141 is that updating the entire `style` attribute overwrites individual styles, which is true. The problem I am describing is that updating individual styles can overwrite the rest of the `style` attribute, which should not be true.

gbj: Although to be fair; the way I wrote the example is a little ambiguous as to whether you'd think it should update the whole style, or just that one style. I don't know. IMO this introduces a bug but maybe it would be so unusual for someone to actually create it that it's worth the trade off.

zakstucke: I've provided a fix which seems to work for your example above, it starts going a bit further than my internal knowledge though.

zakstucke: Effectively just adding the style attr name to the states of implementations using it, then implementing a reset of the css property.

tachys/src/html/style.rs
```diff
@@ -327,23 +429,27 @@ impl IntoStyle for (Arc<str>, Arc<str>) {
...
-        (style, value)
+        (style, name, value)
     }
 
     fn rebuild(self, state: &mut Self::State) {
         let (name, value) = self;
-        let (style, prev) = state;
-        if value != *prev {
+        // state.1 was the previous name, theoretically the css name could be changed:
+        if name != state.1 {
```
gbj, line 445: If this is the case, then we should also `*prev_name = name`, or it will stick with the old name if this updates again.

tachys/src/html/style.rs
```diff
@@ -380,23 +496,27 @@ impl<'a> IntoStyle for (&'a str, &'a str) {
...
+            <(&'a str, &'a str) as IntoStyle>::reset(state);
+        }
+        let (style, _prev_name, prev_value) = state;
+        if value != *prev_value {
```
gbj, line 516: See above: same thing about setting `prev_name`  to the new value.

gbj:
Thanks so much for your work on this. Adding a `reset` is a clever way of doing this and definitely the right approach. (The same thing probably needs to happen for classes but I will take a look at that separately)

I left just two small comments, and then I think this is ready to go.

zakstucke:
Should be fixed. Yeah this'll apply to the `IntoClass` too which will suffer from the same issue currently.

gbj:
Thanks very much for all your work!

zakstucke:
@gbj thanks! Just fyi, we're still missing the optional variant of `impl<F, S> IntoStyle for (&'static str, F)` for the `style:foo=Option<T>` case, it seems those might have to be manual, but a bit annoying the duplication, so I wasn't sure. Probably the same for `IntoClass` too.

PR #3123, opened 2024-10-16
fix: remove unused `Params` attribute `params`

chrisp60:
See #1966 for original PR on older version.

This was held off as a breaking change. The breaking change is now acceptable prior to 0.7 release


gbj:
Thanks!

PR #3133, opened 2024-10-19
feat: allow axum to serve precompressed files

### LINKED ISSUE #3130: `get_static_file` handler doesn't serve compressed assets
The `get_static_file` handler, now helpfully moved into Leptos proper, doesn't serve pre-compressed files even if they're available. This doesn't compose super nicely with the `-P` option in `cargo-leptos`. [A few months ago I adjusted the `0.6` axum template to provide this behavior by default](https://github.com/leptos-rs/start-axum/pull/38).

The main issues is the `ServeDir` construction, which omits the relevant flags for pre-compressed files:

https://github.com/leptos-rs/leptos/blob/ee66f6c3957d79bcf214cdf275daca605b93c740/integrations/axum/src/lib.rs#L2037

However, you do need to pass along `accept-encoding` headers for `ServeDir` to actually serve the compressed assets.

All that to say -- unless there's some weird blocker, I think it should be pretty easy to get this in there! I figured I'd write this up before I forget again.

CorvusPrudens:
This PR mirrors leptos-rs/start-axum#38 for the now-incorporated file serving for the axum integration.

As noted in leptos-rs/start-axum#41, `ServeDir` can produce odd behavior when all headers are passed along, so this PR passes only `accept-encoding`.

Closes #3130.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1993,7 +1993,7 @@ where
     move |uri: Uri, State(options): State<S>, req: Request<Body>| {
         Box::pin(async move {
             let options = LeptosOptions::from_ref(&options);
-            let res = get_static_file(uri, &options.site_root);
+            let res = get_static_file(uri, &options.site_root, req.headers());
             let res = res.await.unwrap();
 
             if res.status() == StatusCode::OK {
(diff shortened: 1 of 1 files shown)


gbj:
Awesome, thank you!

PR #3138, opened 2024-10-21
fix: unregister server functions whose paths are in excluded routes (closes #2735)

### LINKED ISSUE #2735: .leptos_routes() server fn routing should respect excluded routes
Right now it's impossible to manually implement a server fn route with leptos_routes() because it does not consider excluded routes, there should be a mechanism to do that, either through`generate_route_list_with_exclusions` or some other arg

gbj:


yuhime:
@gbj instead of doing the unregister you should make sure that they are excluded from the automatic path generation, as it is now they are not usable once excluded because they are removed from`REGISTERED_SERVER_FUNCTIONS` this leads to the fact that handle_server_fns does not work

benwis:
> @gbj instead of doing the unregister you should make sure that they are excluded from the automatic path generation, as it is now they are not usable once excluded because they are removed from`REGISTERED_SERVER_FUNCTIONS` this leads to the fact that handle_server_fns does not work

Yeah, this is already an open issue, I haven't gotten to it. You can manually register them to make it work in the meantime. Feel free to PR it if you'd like, and let's move conversation there. #3150

Issue #2890, opened 2024-08-28
Conditional tailwind classes are not automatically included


maxbergmark:
When using Tailwind CSS with the `view!` macro, there are some quirks related to conditional Tailwind classes. Within the macro, using the `class:class-name=move || condition` syntax won't trigger Tailwind to include the class. But adding the class name as a comment within any view macro will cause the class to be included, and will render the component as expected.

## Non-working example
```rust
use leptos::*;

fn main() {
    mount_to_body(|| {
        view! { <p
            class:text-orange-600=true
        >"This text should be orange"</p> }
    })
}
```

## Working example
```rust
use leptos::*;

fn main() {
    mount_to_body(|| {
        view! { <p
            // text-orange-600
            class:text-orange-600=true
        >"This text should be orange"</p> }
    })
}
```

**Leptos Dependencies**

```toml
leptos = { version = "0.6.14", features = ["csr", "nightly"] }
```

**To Reproduce**
1. Create CSR application following the [Getting Started](https://book.leptos.dev/getting_started/index.html) guide
2. Add Tailwind CSS using the [Tailwind CSR example](https://github.com/leptos-rs/leptos/tree/leptos_0.6/examples/tailwind_csr)
3. Paste the non-working example above into `main.rs`
4. `trunk serve --open`
5. See error

**Expected behavior**
I would expect the conditional Tailwind classes to get compiled too, without having to add them separately as a comment.

**Screenshots**
<div align="center" style="display: flex;">
<img width="45%" alt="image" src="https://github.com/user-attachments/assets/d6025302-87ec-46b2-a0e8-d9bd5dd5ed12">
<img  width="45%" alt="image" src="https://github.com/user-attachments/assets/4be8f25f-4a68-47d7-9143-6033d7690a70">
</div>

gbj:
I'm not sure this is an actionable issue, from the framework's perspective: i.e., there's no commit to the codebase of Leptos that could solve this problem for you.

I think it's something that can either be fixed by adjusting the Tailwind config (setting up Tailwind to check for `class:` prefixes too? I am not familiar enough with Tailwind to know the answer here) or, alternately, by using the "complicated class" syntax (`class=("text-orange-600", move || true)`) which should detect it because now it's a separate string.

If it *is* possible to do it by adjusting Tailwind config then I guess a PR to update the example Tailwind config would be welcome!

maxbergmark:
Good suggestions! I'll try them out, and see what the outcome is.

[...]

PR #3143, opened 2024-10-22
improve tailwind config to also catch dynamic classes

jheuel:
The previous tailwind config did not catch classes that are set dynamically, e.g.
```
use leptos::*;

fn main() {
    mount_to_body(|| {
        view! { <p
            class:text-orange-600=true
        >"This text should be orange"</p> }
    })
}
```
would not add `text-orange-600`.

The tailwind config now uses the `content.transform` property to [transform](https://tailwindcss.com/docs/content-configuration#transforming-source-files) the leptos code enough for tailwind to extract all classes. Furthermore, there is now a button in all tailwind examples with a dynamic class.

Fixes https://github.com/leptos-rs/leptos/issues/2890

### DIFF
--- a/examples/tailwind_actix/end2end/tests/example.spec.ts
+++ b/examples/tailwind_actix/end2end/tests/example.spec.ts
@@ -1,7 +1,7 @@
 import { test, expect } from "@playwright/test";
 
-test("should see the welcome message", async ({ page }) => {
+test("homepage has title 'Leptos + Tailwindcss'", async ({ page }) => {
   await page.goto("http://localhost:3000/");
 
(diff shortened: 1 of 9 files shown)


AkinAguda:
Would be nice to also catch `class=("bg-red", some_bool)` and `class=(["bg-red", "bg-blue"], some_bool)`

jheuel:
Apologies, when I first opened the pull request I forgot to change the end2end tests. Should be fine now.

gbj:
Awesome! Thanks very much.

maxbergmark:
@jheuel I'm a bit late to the party, but thanks for fixing this issue that I opened a while ago!

PR #3145, opened 2024-10-22
Expose innards of SerializedDataId and SsrSharedContext to allow creating custom hydration contexts

zakstucke:
My usecase was needing to test behaviour of resources when hydrating, these changes allowed me to create a mock hydrate context downstream and use on non-wasm targets.

### DIFF
--- a/hydration_context/src/ssr.rs
+++ b/hydration_context/src/ssr.rs
@@ -62,7 +62,9 @@ impl SsrSharedContext {
     /// Consume the data buffers, awaiting all async resources,
     /// returning both sync and async buffers.
     /// Useful to implement custom hydration contexts.
-    /// This can only be called once.
+    ///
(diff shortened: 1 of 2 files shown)


hydration_context/src/ssr.rs
```diff
@@ -58,6 +58,25 @@ impl SsrSharedContext {
             ..Default::default()
         }
     }
+
+    /// Consume the data buffers, awaiting all async resources,
+    /// returning both sync and async buffers.
+    /// Useful to implement custom hydration contexts.
+    /// This can only be called once.
```
gbj, line 65: Probably worth clarifying this comment; it takes `&self`, so it certainly can be called more than once! Maybe explain what happens if you call it more than once, instead?

zakstucke: 😆 that's true. Updated!

gbj:
Thanks!

PR #3159, opened 2024-10-24
batch navigation updates

Evelyn-H:
simple change here to make history updates atomic. 

this is especially useful when doing navigation based on state. e.g.:
```
let hash_memo = use_location().hash;
let state_signal = use_location().state;

let location_signal = create_memo(move |_| {
    let hash = hash_memo.get();
    let state = state_signal.get()

    let path = todo!(); // some processing
    (path)
});
```

making the nav updates atomic ensures that this memo doesn't get updated twice for a single navigation event


gbj:
Thanks!

PR #3174, opened 2024-10-29
Add fixed size array rendering

LouisDISPA:
Hello, I was using fixed size array and the render method is not implemented on them.
I find it weird that it's implemented on Vec but not arrays. So I wrote a quick implementation.
Is there a reason it's not implemented ?

This allow to pass the array directly to the view method without converting it into a vec.

```rust
// before
view! { <div> {array.to_vec()} </div> }

// after
view! { <div> {array} </div> }

```

It's mostly a copy of the Vec implementation with minor adjustments :

- I removed the marker since it's used to insert new items on size change

```rust
pub struct VecState<T>
where
    T: Mountable,
{
    states: Vec<T>,
    // Vecs keep a placeholder because they have the potential to add additional items,
    // after their own items but before the next neighbor. It is much easier to add an
    // item before a known placeholder than to add it after the last known item, so we
    // just leave a placeholder here unlike zero-or-one iterators (Option, Result, etc.)
    marker: crate::renderer::types::Placeholder,
}
```

- I used the `<[]>::map` function most of the time.
The only exception is for the `future::join_all` that takes and return an iterator.

- For the `Render::rebuild` method I don't know if there is a smart way of handling it. So I just wrote a naive implementation.

### DIFF
--- a/tachys/src/view/iterators.rs
+++ b/tachys/src/view/iterators.rs
@@ -336,3 +336,149 @@ where
         VecState { states, marker }
     }
 }
+
+impl<T, const N: usize> Render for [T; N]
+where
+    T: Render,
+{
+    type State = ArrayState<T::State, N>;
+
+    fn build(self) -> Self::State {
+        Self::State {
+            states: self.map(T::build),
+        }
+    }
+
+    fn rebuild(self, state: &mut Self::State) {
+        let Self::State { states } = state;
+        let old = states;
+        // this is an unkeyed diff
+        self.into_iter()
+            .zip(old.iter_mut())
+            .for_each(|(new, old)| T::rebuild(new, old));
+    }
+}
+
+/// Retained view state for a `Vec<_>`.
+pub struct ArrayState<T, const N: usize>
+where
+    T: Mountable,
+{
+    states: [T; N],
+}
+
+impl<T, const N: usize> Mountable for ArrayState<T, N>
+where
+    T: Mountable,
+{
+    fn unmount(&mut self) {
+        self.states.iter_mut().for_each(Mountable::unmount);
+    }
+
+    fn mount(
+        &mut self,
+        parent: &crate::renderer::types::Element,
+        marker: Option<&crate::renderer::types::Node>,
+    ) {
+        for state in self.states.iter_mut() {
+            state.mount(parent, marker);
+        }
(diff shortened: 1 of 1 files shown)


tachys/src/view/iterators.rs
```diff
@@ -336,3 +336,149 @@ where
...
+        futures::future::join_all(self.into_iter().map(T::resolve))
+            .await
+            .into_iter()
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|_| unreachable!())
+    }
+
+    fn html_len(&self) -> usize {
+        self.iter().map(RenderHtml::html_len).sum::<usize>() + 2
```
LouisDISPA, line 439: Here I removed 1 from the constant compared to the Vec implementation. I did this guessing that the placeholder was counted in. But I have no idea what the constant is for.

gbj: The 3 was the length of the comment node `<!>` :-) So this can just be 0, but it's also just an estimate/heuristic

tachys/src/view/iterators.rs
```diff
@@ -336,3 +336,149 @@ where
...
+        self,
+        buf: &mut String,
+        position: &mut Position,
+        escape: bool,
+        mark_branches: bool,
+    ) {
+        for child in self.into_iter() {
+            child.to_html_with_buf(buf, position, escape, mark_branches);
+        }
+        buf.push_str("<!>");
```
gbj, line 452: Inserting the placeholder here will cause an issue, as it's not used during hydration (so children after this will hydrate incorrectly because there's an extra node

tachys/src/view/iterators.rs
```diff
@@ -336,3 +336,149 @@ where
...
+    {
+        for child in self.into_iter() {
+            child.to_html_async_with_buf::<OUT_OF_ORDER>(
+                buf,
+                position,
+                escape,
+                mark_branches,
+            );
+        }
+        buf.push_sync("<!>");
```
gbj, line 472: (See above)

gbj:
Thanks so much! This looks great over all, I added a couple comments about the placeholders for information, but I just added a quick commit to fix it, so I don't think there's any action needed on your part.

Issue #3175, opened 2024-10-30
excluded paths: server functions not found


yuhime:
**Server function not found after insert it on excluded paths**
now the server functions are successfully excluded but also removed from `REGISTERED_SERVER_FUNCTIONS` this leads to the fact that handle_server_fns no longer works over the excluded functions, is it possible to exclude them only from automatic path generation?

**To Reproduce**
```rust
    let routes = generate_route_list_with_exclusions(
        {
            let options = leptos_options.clone();
            move || shell(options.clone())
        },
        Some(vec!["/api/my_server_functions".into()]),
    );
    
    Router::new().leptos_routes(&leptos_options, routes, {
            let options = leptos_options.clone();
            move || shell(options.clone())
        }) .route(
            "/api/my_server_functions",
            post(handle_server_fn)
        )
```
```rust
Could not find a server function at the route /api/my_server_functions.

It's likely that either
 1. The API prefix you specify in the `#[server]` macro doesn't match the prefix at which your server function handler is mounted, or
2. You are on a platform that doesn't support automatic server function registration and you need to call ServerFn::register_explicit() on the server function type, somewhere in your `main` function.
```

**Expected behavior**
handle_server_fns should be able to find the function so that it can be handled with axum together with the route and the layer

gbj:
Duplicate of #3150.

gbj:
Fixed by #3176

PR #3176, opened 2024-10-31
fix: exclude excluded server fn paths instead of unregistering them (closes #3150, #3175)

### LINKED ISSUE #3150: generate_routes_with_exclusions() unregisters the server fn
Generally we want to exclude a route here if we want to manually create a route in the server router for it. This is useful for middlewares and other things that depend on the ordering of things in the router, and thus we don't want LeptosRoutes to auto generate a route for it.

Currently we're also unregistering the server fn, which confuses people and is not useful. We just want to exclude it from route generation

gbj:
@benwis I had a slightly different idea on how to implement this, without any type signature changes, by simply creating a separate "excluded route" category in the generated route list, and then checking against that. Could you test it out for me?

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -41,10 +41,10 @@ use once_cell::sync::Lazy;
 use parking_lot::RwLock;
 use send_wrapper::SendWrapper;
 use server_fn::{
-    actix::unregister_server_fns, redirect::REDIRECT_HEADER,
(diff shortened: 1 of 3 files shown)


benwis:
Sure, will do!

gbj:
I did a bit of testing and it works, so will merge!

benwis:
Great, it worked for me as well!

PR #3182, opened 2024-11-01
fix: ensure we check memos the first time a dependency uses them, even if the dependency always runs on its first run (closes #3181)

### LINKED ISSUE #3181: Resource subscribed to a Memo has inconsistent fetching behaviour depending on whether it's hydrated or not
1. Start with the axum-0.7 starter
2. You will need to add this line to `Cargo.toml`:
    ```toml
    getrandom = { version = "0.2", features = ["js"] }
    ```
3. [Use this `app.rs`](https://github.com/user-attachments/files/17600485/app.rs.txt)

To test the non-hydrated case:
1. Navigate to `localhost:3000`.
2. Click the "counter" link.
3. Click the counter button. It should behave as expected, and only one `resource fetching with param_dep=Some(())` message will be displayed in the console.

To test the hydrated case:
1. Navigate to `localhost:3000/count`
2. Click the counter button. Note that the number remains "0" and a duplicate `resource fetching with param_dep=Some(())` has been logged to the console.
3. After this, the button will work as expected.

The resource depends on a Memo which depends on a Params struct. Though the URL changes (deliberately) every time the button is clicked, the Memo does not, and thus I would assume the Resource would not re-fetch.

gbj:


gbj:
@metatoaster I'm trying to figure out the failing tests on `suspense_tests` on this branch. When I test manually I don't see the issue, but I'm not sure I understand the tests you added well enough to understand what's going wrong here. When you have a bit of time, could you help me try to figure out what's going on?

metatoaster:
Some of the tests I've added basically did document the _existing_ behavior, not the ideal behavior, which I did [comment in the test itself](https://github.com/leptos-rs/leptos/blob/3a8508df6c90f7c16b42f31152209eb29c6aeeb0/examples/suspense_tests/e2e/features/check_instrumented_suspense_resource.feature#L131-L166) - in this test I was documenting the existing behavior where on departure from a component, the resource it needed was retrieved despite how it was on the verge of being disposed when hydrated, and while under CSR this wasn't an issue.  Now both versions show the same expected behavior where this extra unnecessary fetch is no longer necessary, and that's a good thing.  Moreover, this fix did correct other extra fetches which means those other tests need to pare down the counts, and again this is a very good thing.

However, the concerning part is the base test case is picking up what appears to be the SSR counters from the CSR navigation, and its only for that particular test, which is weird, this is what I am more concerned about. I will get to the bottom of this, but in the mean time I am going to fix those tests.

gbj:
Okay, thanks for the context! I think I was reading the numbers backwards in the output, this is great :tada: 

This was merged along with your test change in 8252655.

Issue #2856, opened 2024-08-18
Allow to edit server meta context on SSR


mondeja:
**Is your feature request related to a problem? Please describe.**

This problem affects the two internationalization projects that currently exist in Leptos (leptos_i18n and leptos-fluent). Nowadays we are using the next dirty hack to set the `lang` (and in the case of leptos-fluent, the `dir` also) attribute(s) on SSR with Leptos v0.6:

```rust
leptos_meta::Html(
    leptos_meta::HtmlProps {
         lang: ...,
         dir: ...,
         class: ...,
         attributes: ...,
    }
);
```

Of course, the above code is a simplification. On leptos-fluent, I'm parsing the result of `leptos_meta::use_head().html.as_string()` to keep the previous values of `lang`, `dir` and `class`, altough `attributes` is not possible because is a private field. These values are possibly previously set by a previous invocation of an `<Html/>` component in the tree.

**Describe the solution you'd like**

I want that `use_head()` would return `ServerMetaContext` instead of `MetaContext` on SSR. These two types would need a refactor because are only exposing certain things when not on SSR, which IMO is really strange. And then, probably build an API to allow something like:

```rust
#[cfg(features = "ssr")]
{
    let context = use_head();
    let mut attrs = context.html().attributes();
    // ... manipulate the attributes
    context.html.set_attributes(attrs);
}
```

**Describe alternatives you've considered**

There is no alternatives currently.

**Additional context**

This is a blocking issue for the migration to Leptos v0.7. Note that a `Provider` implementation as suggested in #2780 is not possible because if an user wants to provide their own `<Html/>` before or after in the component tree they can't, we would be forcing users to give to our libraries (leptos-fluent and leptos_i18n) the ability to set their own `<Html/>` components in their code.

FYI @Baptistemontan

mondeja:
> Is the issue that you do not want to override a `lang` attribute set by the user with a `lang` attribute set by the internationalization library?

The issue is that currently I don't want to override other attributes that are not `lang` neither `dir`. But, for sure, if the user has provided a `<Html lang=.../>` and/or `<Html dir=.../>` attribute, I at least want to display a warning on `#[cfg(debug_assertions)]` advicing that `leptos-fluent` will override these attributes.

From the conversation at #2780:

[...]

PR #3188, opened 2024-11-03
feat: generalize response data injection

CertainLach:
I have a leptos styled components library, which is using leptos_meta for SSR for a long time.

However, I had some problems with this flow, that are probably solvable by extension of leptos_meta public API, but I decided to generalize leptos_meta server integration instead, now any other library may hook into response data generation.

This PR only implements axum integration, because I'm not sure if there is a better way to implement this feature.

### DIFF
--- a/examples/axum_js_ssr/src/main.rs
+++ b/examples/axum_js_ssr/src/main.rs
@@ -29,6 +29,7 @@ async fn main() {
     use leptos::logging::log;
     use leptos::prelude::*;
     use leptos_axum::{generate_route_list, LeptosRoutes};
+    use leptos_meta::ServerMetaContext;
 
     latency::LATENCY.get_or_init(|| [0, 4, 40, 400].iter().cycle().into());
     latency::ES_LATENCY.get_or_init(|| [0].iter().cycle().into());
@@ -43,7 +44,7 @@ async fn main() {
     let addr = conf.leptos_options.site_addr;
     let leptos_options = conf.leptos_options;
     // Generate the list of routes in your Leptos App
-    let routes = generate_route_list(App);
+    let routes = generate_route_list::<ServerMetaContext, _>(App);
 
     async fn highlight_js() -> impl IntoResponse {
         (
@@ -128,11 +129,11 @@ async fn main() {
 
     let app = Router::new()
         .route("/highlight.min.js", get(highlight_js))
-        .leptos_routes(&leptos_options, routes, {
+        .leptos_routes::<ServerMetaContext, _>(&leptos_options, routes, {
             let leptos_options = leptos_options.clone();
             move || shell(leptos_options.clone())
         })
-        .fallback(leptos_axum::file_and_error_handler(shell))
+        .fallback(leptos_axum::file_and_error_handler::<_, ServerMetaContext, _>(shell))
         .layer(middleware::from_fn(latency_for_highlight_js))
         .with_state(leptos_options);
 
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -19,7 +19,6 @@ futures = "0.3.30"
 leptos = { workspace = true, features = ["nonce", "ssr"] }
 server_fn = { workspace = true, features = ["axum-no-default"] }
 leptos_macro = { workspace = true, features = ["axum"] }
-leptos_meta = { workspace = true, features = ["ssr"] }
 leptos_router = { workspace = true, features = ["ssr"] }
 leptos_integration_utils = { workspace = true }
 once_cell = "1"
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -57,9 +57,8 @@ use leptos::{
     IntoView,
 };
 use leptos_integration_utils::{
-    BoxedFnOnce, ExtendResponse, PinnedFuture, PinnedStream,
+    BoxedFnOnce, ExtendResponse, PinnedFuture, PinnedStream, ServerExtension,
 };
-use leptos_meta::ServerMetaContext;
 #[cfg(feature = "default")]
 use leptos_router::static_routes::ResolvedStaticPath;
 use leptos_router::{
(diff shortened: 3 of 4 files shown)


gbj:
Can you say more about the problem you're trying to solve, here? There are some very significant drawbacks to making a breaking change to every application simply to add more verbosity for 99% of users, so I am wondering if there is another way to solve this than adding the generics you've added.

CertainLach:
Basically, I want to inject a single <style> block to head, where contents of this style block are being populated on the fly, by include_sass! macro.
Currently, without this patch, code looks like the following:

```rust
#[component]
fn PermissionConfirmation(
        callback: ChallengeCallback,
        app_name: String,
        new_permissions: Vec<DisplayPermission>,
        kept_permissions: Vec<DisplayPermission>,
) -> impl IntoView {
        include_sass!(css_module, "permissions.scss");
        let callback = {
                move |_: MouseEvent| {
                        callback.run((
                                control::serialize(&PermissionsConfirmationResponse { confirmed: true }),
                                None,
                        ))
                }
        };
        h!(div[[
                CssModule > { css_module },
                h1 > { app_name },
                table.permissions[[for permission in new_permissions {
                        tr.permission[[
                                td.icon[[img(src = permission.icon.clone())]],
                                td.name[[{ permission.name.clone() }]],
                                td.description[[{ permission.description.clone() }]],
                        ]]
                }]],
                table.permissions[[for permission in kept_permissions {
                        tr.permission[[
                                td.icon[[img(src = permission.icon.clone())]],
                                td.name[[{ permission.name.clone() }]],
                                td.description[[{ permission.description.clone() }]],
                        ]]
                }]],
                ul.login_btn[[
                        li[[button(onclick = callback) > "Confirm"]],
                        li[[button > "Cancel (Or close page)"]],
                ]],
        ]])
}
```

(Don't mind the custom view macro, it is required for the proper css module scoping)
include_sass macro generates a bunch of

```rust
let cl_login_btn = "mangled-class-name";
...
let css_module = DataForCssModule {...};
```

local variables, and the custom view macro uses them

But it also requires an additional CssModule component present, which handles the lifecycle of <Style> component children. (Which is pretty complicated, because if the same style file is imported multiple times - only one Style component should be present)

I have tried to reimplement this with direct usage of MetaContext/ServerMetaContext inside of the hook, but the lifecycle is still not great, and there is no way with the current ServerMetaContext api to be able to merge multiple added style tags into one.

```rust
// expansion of use_sass!
let cl_login_btn = "mangled-class-name";
...
use_css_module(DataForCssModule {...});

#[component]
fn PermissionConfirmation(
        callback: ChallengeCallback,
        app_name: String,
        new_permissions: Vec<DisplayPermission>,
        kept_permissions: Vec<DisplayPermission>,
) -> impl IntoView {
        include_sass!("permissions.scss");
        ...
        h!(div[[
                ...
                ul.login_btn[[
                        li[[button(onclick = callback) > "Confirm"]],
                        li[[button > "Cancel (Or close page)"]],
                ]],
        ]])
}
```

To bypass the limitations of ServerMetaContext api, without the need to implement css modules building blocks in leptos itself, I have decided to implement my own response head injector, like leptos_meta does. And to do that - I need some way to make axum integration to use this response injector, where it already has hardcoded integration for leptos_meta.

I agree that the current approach with ServerExtension generic parameter is not great, but I haven't found a prettier way with the current leptos SSR API. I will re-implement this PR, if someone proposes a better API.
...It is also possible to push this breaking change with 0.7.0 major release, as it also gives some freedom to application shell construction, which gives false sense that leptos_meta is not something tightly coupled with core, and something users can implement themselves.

gbj:
> false sense that leptos_meta is not something tightly coupled with core, and something users can implement themselves.

The problem here is not that `leptos_meta` is tightly coupled to the core, and that users can't implement itself; the problem is that `leptos_axum` is designed as a convenient wrapper around the combination of `leptos_router`, `leptos_meta`, and `leptos`. It's just existed as a convenience for so long that it seems like a necessity; but it's really not.

> It is also possible to push this breaking change with 0.7.0 major release

It's definitely possible. When I call it a "breaking change" I don't mean "it's a breaking change according to semver" I mean "it breaks users' applications" — see the fact that every example fails CI without modification. I do not plan to make this change, whether it is allowed by semver or not.

> there is no way with the current ServerMetaContext api to be able to merge multiple added style tags into one.

Fair enough! Is there a way we can adjust the `leptos_meta` API to support it what you need? It might be useful to compare to the discussion in #2856, too, which I think had similar interests.

What additional functionality do you need to support what you're trying to do? Are there one or two functions that could be added to enable it?

CertainLach:
> It's just existed as a convenience for so long that it seems like a necessity; but it's really not.

The problem is, it is not easy to implement SSR without those helpers, => it is not easy to implement functionality like leptos_meta by yourself.

> see the fact that every example fails CI without modification

I agree that the current approach is bad, but it might be improved.
- Extension parameter may have a default type (Error-prone, as if user wants to specify it - it needs to be specified everywhere)
- Extension might be made a value, instead of being purely type-based. This still require modifications to user code (It might be moved to LeptosOptions, made an optional provider function like additional_context, or redone in other, more convinient way), but also will allow users to provide some values from outside (generalized ResponseOptions?)
- Server functions might be wrapped into a some struct. This way you can even get rid of many <IV> generics everywhere (I.e server view definition can be done in a single place), but it requires users to update their codebases too.

> What additional functionality do you need to support what you're trying to do? Are there one or two functions that could be added to enable it?

It might be solved by adding a couple of functions in public api, but there is no way to make this api look good, and be usable for everyone. My implementation only needs to

- Create a style tag from inside of a hook. This is already possible by exposing `register()` function, sharing information about this style tag might be done by using user-provided context. Style tag needs not to be shared, as e.g my css modules implementation takes control of this tag during hydration
- Append contents to this style tag. Current implementation needs to be extended to not only be able to send new tags to ServerMetaContextOutput, but also to be able to update (append-only?) already sent tags.
- Have some way to provide information from use-hooks to hydration stage. An additional modifiable <script> tag api? Some generic way of providing JSON value in SSR, e.g `use_head().set_kv("a", "b") in SSR/use_head().get_kv("a") in hydration`? Just an ability to set style tag attributes?

luoxiaozero:
@CertainLach I found a way to solve this problem without modifying leptos. Refer to the **`SSRMountStyleProvider`** component in https://github.com/thaw-ui/thaw/pull/335/commits/266dd6dd4a75484501cd8936429669221c1d4ee1.

PR #3192, opened 2024-11-04
`Cargo.lock`

sabify:
Closes https://github.com/leptos-rs/leptos/issues/2881


sabify:
I have excluded examples, projects and benchmarks to prevent PR pollution from Depdendabot. Please let me know if you need any of them to be included.

sabify:
Failed CIs are not related to this PR.

benwis:
I am confused why unused functions would cause compilation to fail, isn't that usually a warning?

gbj:
The dead code warning causes failures because we have CI set to fail on clippy warnings. This has been fixed on main so will not be an issue here.

Looks good, and thanks for taking the time!

PR #3198, opened 2024-11-05
fix: allow `non_snake_case` and `dead_code` lints to run within component functions

yescallop:
Currently, the `non_snake_case` and `dead_code` lints will not run within component functions because the `component` macro emits `allow` attributes for the two lints, which allows room for users to make mistakes. This PR fixes the problem by converting the function name to snake case before prefixing it with `__` and dropping the `allow` attributes for the two lints from the macro.

I have run `cargo check` on the library and all examples and see no additional warnings. We can have the CI confirm this.


yescallop:
The `server_fns_axum` example failed to build because there are multiple component and server functions with the same name but in different case ([`rkyv_example` and `RkyvExample`](https://github.com/leptos-rs/leptos/blob/5cc2f3858dbd9307c6c630f5f25172511bc8c036/examples/server_fns_axum/src/app.rs#L302-L313), for example). I changed component and server macros to emit different prefixes (`__leptos_component_` and `__leptos_server_`) to address the problem.

yescallop:
I don't know why, but I might've run into a `cargo fmt` bug touching seemingly unrelated code. I thought it would be better not to include [those formatting changes](https://github.com/leptos-rs/leptos/pull/3198/files#diff-0cb320a9e87b55dd019bea871ed418089b052a3a7b4a63603bfe4e0074be3a89L661), but apparently the CI would not be happy with that! So I included the changes and force-pushed again. Hopefully everything is OK now.

leptos_macro/src/component.rs
```diff
@@ -1211,7 +1211,10 @@ fn prop_to_doc(
 }
 
 pub fn unmodified_fn_name_from_fn_name(ident: &Ident) -> Ident {
-    Ident::new(&format!("__{ident}"), ident.span())
+    Ident::new(
```
rakshith-ravi, line 1214: Out of curiosity, why was this changed?

rakshith-ravi:
Oh, nevermind, I see what happened here. Perhaps we can have a `#[allow(non_snake_case, dead_code)]` in the declaration, and another `#[warn(non_snake_case, dead_code)]` in the body to bring back the lint instead of changing the name of the fn or struct?

Not sure if that would work, but worth a shot

rakshith-ravi:
Hi @yescallop. Wanted to check in to see if you had time to check out my comments?

yescallop:
Hi @rakshith-ravi. I guess I get your point. Do you mean something like the following?

```rust
#[allow(non_snake_case)]
fn __Example() {
    #[warn(non_snake_case)]
    {
        // body wrapped, this will warn
        let A = ();
    }
}
```

But I notice that an inner attribute `#![allow(non_snake_case)]` on the module will not have an effect on the wrapped body, which is an adverse effect of this solution. What do you think?

rakshith-ravi:
> But I notice that an inner attribute #![allow(non_snake_case)] on the module will not have an effect on the wrapped body, which is an adverse effect of this solution. What do you think?
Oh my. Yeah I didn't think about that. Valid point indeed. It just seems very "hacky" to do a `__leptos_component__{}` 😅 

Any other suggestion you have? If not, I guess this could do for now. We can always clean it up later if we think of a better solution

gbj:
My only feedback would be that the `server_fn_macro` crate is used by/can be used by other framework, so perhaps simply `__server_...` and `__component_...` rather than `__leptos_server` etc. would be a better naming convention. Otherwise looks good to me.

rakshith-ravi:
Changes made @gbj

benwis:
@rakshith-ravi Looks like it's been a bit since this was made. If you can resolve the commits I'll see about merging it

yescallop:
Conflicts resolved @benwis

rakshith-ravi:
I've triggered the CI. Let's see if there's any issues

benwis:
Thanks for the PR!

Issue #3196, opened 2024-11-05
`A` components aria-current attribute not getting set in nested routes


TERRORW0LF:
**Describe the bug**
When placing links with the `A` component inside a nested route the aria-current attribute does not get set on client side navigation.

**Leptos Dependencies**
```toml
leptos = { version = "0.7.0-rc1", features = ["nightly"] }
leptos_axum = { version = "0.7.0-rc1", optional = true }
leptos_meta = { version = "0.7.0-rc1" }
leptos_router = { version = "0.7.0-rc1", features = ["nightly"] }
```

**To Reproduce**
1. Set up the router example in an start-axum-0.7 template (remove tracing and futures)
2. Remove routing progress (causes bugs which alter this one)
3. Run `cargo leptos watch`
4. Open the site and navigate to a contact
5. Open the browser's inspector and select the contact's `a` tag
6. The `aria-current="page"` attribute is not there

**Expected behavior**
The `aria-current="page"` attribute gets applied to the `a` tag of the current contact

codewing:
FYI: Just came across this bug in 0.6.15 as well

PR #3202, opened 2024-11-07
test: fixtures for testing aria-current

metatoaster:
This provide the basic fixtures to check for `aria-current` for links, plus a basic test case to show that it works - can easily be modified to use non-highlighted link to trigger text failure.

This may be merged separately via this pull request or be part of the bigger pull request that includes the fix plus tests.  Navigating around `suspense_test` will demonstrate the particular issue as reported in #3196 so this was as good place as any to set this up.


benwis:
Always happy to have more tests for things that could break, thanks @metatoaster!

PR #3203, opened 2024-11-07
chore: add missing docs for 0.7

gbj:


router/src/ssr_mode.rs
```diff
@@ -20,17 +20,23 @@ use crate::static_routes::StaticRoute;
...
 /// restrictive mode will be used: i.e., if even a single nested route asks for `Async` rendering, the whole initial
 /// request will be rendered `Async`. (`Async` is the most restricted requirement, followed by `InOrder`, `PartiallyBlocked`, and `OutOfOrder`.)
 #[derive(Default, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SsrMode {
+    /// Out-of-order streaming.
     #[default]
     OutOfOrder,
+    /// Partially-blocked out-of-order streaming.
     PartiallyBlocked,
+    /// In-order streaming.
```
rakshith-ravi, line 36:
```suggestion
    /// **In-order streaming** (`InOrder`): Walk through the tree, returning HTML synchronously as in synchronous rendering and out-of-order streaming until you hit a `Suspense`. At that point, wait for all its data to load, then render it, then the rest of the tree.
    ///     - *Pros*: Does not require JS for HTML to appear in correct order.
    ///     - *Cons*: Loads the shell more slowly than out-of-order streaming or synchronous rendering because it needs to pause at every `Suspense`. Cannot begin hydration until the entire page has loaded, so earlier pieces
    ///       of the page will not be interactive until the suspended chunks have loaded.
```

router/src/ssr_mode.rs
```diff
@@ -20,17 +20,23 @@ use crate::static_routes::StaticRoute;
...
 ///
 /// The mode defaults to out-of-order streaming. For a path that includes multiple nested routes, the most
 /// restrictive mode will be used: i.e., if even a single nested route asks for `Async` rendering, the whole initial
 /// request will be rendered `Async`. (`Async` is the most restricted requirement, followed by `InOrder`, `PartiallyBlocked`, and `OutOfOrder`.)
 #[derive(Default, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SsrMode {
+    /// Out-of-order streaming.
     #[default]
     OutOfOrder,
+    /// Partially-blocked out-of-order streaming.
```
rakshith-ravi, line 34:
```suggestion
    /// **In-order streaming** (`InOrder`): Walk through the tree, returning HTML synchronously as in synchronous rendering and out-of-order streaming until you hit a `Suspense`. At that point, wait for all its data to load, then render it, then the rest of the tree.
    ///     - *Pros*: Does not require JS for HTML to appear in correct order.
    ///     - *Cons*: Loads the shell more slowly than out-of-order streaming or synchronous rendering because it needs to pause at every `Suspense`. Cannot begin hydration until the entire page has loaded, so earlier pieces
    ///       of the page will not be interactive until the suspended chunks have loaded.
```

router/src/ssr_mode.rs
```diff
@@ -20,17 +20,23 @@ use crate::static_routes::StaticRoute;
...
-/// 6. **`Static`**:
+/// 6. **`Static`**: Renders the page when the server starts up, or incrementally, using the
+///    configuration provided by a [`StaticRoute`].
 ///
 /// The mode defaults to out-of-order streaming. For a path that includes multiple nested routes, the most
 /// restrictive mode will be used: i.e., if even a single nested route asks for `Async` rendering, the whole initial
 /// request will be rendered `Async`. (`Async` is the most restricted requirement, followed by `InOrder`, `PartiallyBlocked`, and `OutOfOrder`.)
 #[derive(Default, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SsrMode {
+    /// Out-of-order streaming.
```
rakshith-ravi, line 31:
```suggestion
    /// **Out-of-order streaming** (`OutOfOrder`, the default): Serve an HTML shell that includes `fallback` for any `Suspense`. Load data on the **server**, streaming it down to the client as it resolves, and streaming down HTML for `Suspense` nodes.
    ///     - *Pros*: Combines the best of **synchronous** and `Async`, with a very fast shell and resources that begin loading on the server.
    ///     - *Cons*: Requires JS for suspended fragments to appear in correct order. Weaker meta tag support when it depends on data that's under suspense (has already streamed down `<head>`)
```

router/src/ssr_mode.rs
```diff
@@ -20,17 +20,23 @@ use crate::static_routes::StaticRoute;
...
     #[default]
     OutOfOrder,
+    /// Partially-blocked out-of-order streaming.
     PartiallyBlocked,
+    /// In-order streaming.
     InOrder,
+    /// Async rendering.
```
rakshith-ravi, line 38:
```suggestion
    /// **`Async`**: Load all resources on the server. Wait until all data are loaded, and render HTML in one sweep.
     ///     - *Pros*: Better handling for meta tags (because you know async data even before you render the `<head>`). Faster complete load than **synchronous** because async resources begin loading on server.
     ///     - *Cons*: Slower load time/TTFB: you need to wait for all async resources to load before displaying anything on the client.
```

router/src/ssr_mode.rs
```diff
@@ -20,17 +20,23 @@ use crate::static_routes::StaticRoute;
...
     Async,
+    /// Static rendering.
```
rakshith-ravi, line 40:
```suggestion
    /// **`Static`**: Renders the page when the server starts up, or incrementally, using the
    ///    configuration provided by a [`StaticRoute`].
```

rakshith-ravi (approved):
Small nit-pick, but I think it makes sense to include the rendering mode in the individual enums as well, since those are the ones that will be highlighted in LSPs when they're selected

PR #3220, opened 2024-11-10
Fix deprecated parameters js warning (0.6)

paul-hansen:
Same as #3219 but for Leptos 0.6

Feel free to close if you don't want to bother backporting this. I just was using it so thought I'd share. I do plan on upgrading to 0.7 when I can.


paul-hansen:
One thing I'm not sure on is if I should bump wasm-bindgen to 0.2.93 which is the first version that added support for passing the parameters as an object. I guess it would fail if you're still on an older version. Lmk if that's desired.

benwis:
Don't worry, we don't support older wasm-binder versions. Cargo-leptos and the rest are on 0.2.95

Issue #3226, opened 2024-11-10
[0.7] static site generation only generates fallback pages


NiklasEi:
**Describe the bug**
The `static_routing` example generates a static site containing only the fallback pages saying "Page not found." for all blog entries.

**Leptos Dependencies**

Latest main f804a69f2f5bfc2c7580e564730c2dd6878f7257 

**To Reproduce**
1. Navigate to the `static_routing` examples
2. (fix) remove the two `regenerate` calls in the router to get around #3224
3. Run `cargo leptos serve`
4. Check the static site output in `target/site`
    * The pages under "post/post<x>/index.html" are all there, but only contain "Page not found.", not the actual post contents

**Expected behavior**
The build output should contain the same content you get when opening the paths while the server is running.

**Additional context**
The content is correct if we keep the server running and visit the posts in the browser. But, in my opinion, this is not the expected behaviour for a static route. The full content should already be written to the output before any call to the server.

gbj:
Same cause as #3224, see #3228.

Issue #3224, opened 2024-11-10
[0.7] panic in static_routing example


NiklasEi:
**Describe the bug**
When running the `static_routing` example with `cargo leptos watch` it panics with "using static routing, but couldn't find ParamsMap" at `leptos/router/src/static_routes.rs:334:30`

**Leptos Dependencies**

I pulled latest main bb0dff6af57b491ab7372ffd6833024742c7f186 so everything is at `0.7.0-rc1`

**To Reproduce**
1. Navigate to examples/static_routing in your terminal
2. Execute `cargo leptos watch`

**Expected behavior**
The example should start and host the expected pages

**Additional context**
The code path only gets hit when using `regenerate` on the static routes.

gbj:
Same cause as #3226, see #3228. Thanks for noticing this.

PR #3228, opened 2024-11-10
Fix static routing example

gbj:
As far as I can tell, both #3224 and #3226 are really the same problem, caused by the fact that the static path expansion algorithm was accidentally expanding `(StaticSegment(""), StaticSegment("post"), ParamSegment("slug"), StaticSegment("/"))` into routes like `//post/1/` instead of `/post/1/`.


NiklasEi (approved):
This fixes both issues for me, thank you :+1:

gbj:
Thanks for checking!

PR #3229, opened 2024-11-10
Add tests for converting path segments into paths

NiklasEi:
This includes a unit test that will be red until #3228 is merged


gbj:
Thanks for adding the test, much appreciated!

PR #3245, opened 2024-11-15
feat(reactive_stores): add `map_untracked` to `OptionStoreExt`

tqwewe:
Adds an unreactive variant of `OptionStoreExt::map` called `map_untracked`.

I didn't add any tests for this, as I'm not entirely sure what to test for.


tqwewe:
> But in fact it seems to me that the behavior that the documentation describes is actually the desired behavior, right?

I've scratched my head for a while, but don't have a good answer for this really. Only thing I'm wondering is, if it did memoize on Some-ness, then what if you wanted to get reactive changes on inner field updates (like what it does currently)? Might be a problem in that case

PR #3249, opened 2024-11-17
test: aria-current end-to-end tests (#3213)

metatoaster:
End-to-end tests for #3213.

Edit: Not sure why the autofix.ci decided to kick in here but that CI setup probably should be done with the original pull request that introduce this to have a clean slate for this pull request to apply to.


benwis:
@metatoaster It ran because it found a redundant match in your PR, I believe it runs cargo clippy on PRs now

gbj:
The redundant match is in code other than this PR, so I'm not sure why it wasn't caught at some other point, but the other CI did pass on the PR and on the version with the fix. I will merge them as separate commits.

gbj:
(Oh and thanks very much for adding the tests!)

Issue #3155, opened 2024-10-23
`ServerFnError<TargetErr>` should impl `From<ServerFnError<SourceErr>>` where `TargetErr: From<SourceErr>`


nicolas-guichard:
Last one in my series of proposed improvements to `ServerFnError`: we could provide a conversion from `ServerFnError<OldCustErr>` to `ServerFnError<NewCustErr>` if `NewCustErr: From<OldCustErr>`. Something like:
```
impl<OldCustErr, NewCustErr: OldCustErr> From<ServerFnError<OldCustErr>> for ServerFnError<NewCustErr> {
    fn from(value: ServerFnError<NoCustomError>) -> Self {
        match self {
            ServerFnError::<OldCustErr>::WrappedServerError(o) => ServerFnError::<NewCustErr>::WrappedServerError(o.into()),
            ServerFnError::<OldCustErr>::Registration(s) => ServerFnError::<NewCustErr>::Registration(s),
            ServerFnError::<OldCustErr>::Request(s) => ServerFnError::<NewCustErr>::Request(s),
            ServerFnError::<OldCustErr>::Response(s) => ServerFnError::<NewCustErr>::Response(s),
            ServerFnError::<OldCustErr>::ServerError(s) => ServerFnError::<NewCustErr>::ServerError(s),
            ServerFnError::<OldCustErr>::Deserialization(s) => ServerFnError::<NewCustErr>::Deserialization(s),
            ServerFnError::<OldCustErr>::Serialization(s) => ServerFnError::<NewCustErr>::Serialization(s),
            ServerFnError::<OldCustErr>::Args(s) => ServerFnError::<NewCustErr>::Args(s),
            ServerFnError::<OldCustErr>::MissingArg(s) => ServerFnError::<NewCustErr>::MissingArg(s),
        }
    }
}
```

The typical use-case I have is helper functions that can fail in one or two ways, shared by multiple server functions that can also fail in other ways.

gbj:
Again, feel free to make a PR if the compiler will actually allow this.

benwis:
This one also conflicts with an existing trait, this time in convert
```
   Compiling server_fn v0.7.0-rc1 (/celCluster/projects/leptos/server_fn)
error[E0119]: conflicting implementations of trait `std::convert::From<ServerFnError<_>>` for type `ServerFnError<_>`
   --> server_fn/src/error.rs:197:1
    |
[...]

Issue #3154, opened 2024-10-23
`NoCustomError` should be replaced by `Infallible` (or another empty `enum`)


nicolas-guichard:
When matching over a `ServerFnError`, we always have to cover the `WrappedServerError` case, even though it should not be used when `CustErr = NoCustomError`.

This happens because `NoCustomError` is an empty `struct`, which _can_ be constructed.

Instead I think we should use an empty `enum` such as `std::convert::Infallible`, which would:
* prevent from creating that variant altogether
* remove the need to match for that variant (since [Rust 1.82](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#omitting-empty-types-in-pattern-matching))

This would however be a breaking change.

nicolas-guichard:
If/when this is implemented, we could provide an infallible conversion from `ServerFnError<NoCustomError>` to `ServerFnError<CustErr>`:
```rust
impl<CustErr> From<ServerFnError<NoCustomError>> for ServerFnError<CustErr> {
    fn from(value: ServerFnError<NoCustomError>) -> Self {
        match self {
            // not needed since Rust 1.82 if NoCustomError becomes uninhabited:
            // ServerFnError::<NoCustomError>::WrappedServerError(_) => unreachable!(),
            ServerFnError::<NoCustomError>::Registration(s) => ServerFnError::<CustErr>::Registration(s),
            ServerFnError::<NoCustomError>::Request(s) => ServerFnError::<CustErr>::Request(s),
            ServerFnError::<NoCustomError>::Response(s) => ServerFnError::<CustErr>::Response(s),
            ServerFnError::<NoCustomError>::ServerError(s) => ServerFnError::<CustErr>::ServerError(s),
            ServerFnError::<NoCustomError>::Deserialization(s) => ServerFnError::<CustErr>::Deserialization(s),
            ServerFnError::<NoCustomError>::Serialization(s) => ServerFnError::<CustErr>::Serialization(s),
            ServerFnError::<NoCustomError>::Args(s) => ServerFnError::<CustErr>::Args(s),
            ServerFnError::<NoCustomError>::MissingArg(s) => ServerFnError::<CustErr>::MissingArg(s),
        }
    }
}
```

Currently to enable the `?` operator, I parameterize the functions that return `ServerFnError` to take the `CustError` from their caller.

gbj:
Could you provide an example of the problem you're trying to solve?

nicolas-guichard:
[...]

Issue #3153, opened 2024-10-23
`ServerFnError<T>` should impl `From<E: std::error::Error>` for any `T`


nicolas-guichard:
[Currently](https://github.com/leptos-rs/leptos/blob/7b8cd90a6ef113495d0efc5fb8676bd0694a7705/server_fn/src/error.rs#L178), `ServerFnError` has a blanket `From` impl to convert from `E`, but only with the default `NoCustomError` type:
```
impl<E: std::error::Error> From<E> for ServerFnError {
    fn from(value: E) -> Self {
        ServerFnError::ServerError(value.to_string())
    }
}
```

This means we can't easily use `?` to convert other errors into a `ServerFnError<MyCustomError>`.

Instead I think we should provide a blanket impl for any `T`:
```
impl<T, E: std::error::Error> From<E> for ServerFnError<T> {
    fn from(value: E) -> Self {
        ServerFnError::ServerError(value.to_string())
    }
}
```

gbj:
Feel free to make a PR.

benwis:
That trait implementation interferes with the one in core, so Rust won't let that happen. I'm also not sure people want to convert their E into a String instead of using our WrappedError variant. Closing this as well.

PR #3253, opened 2024-11-17
Improvements around custom error handling in ServerFnError

nicolas-guichard:
This implements a fix for #3154 by making `NoDefaultError` unhinabited, and proposes workarounds for #3153 and #3155 to make converting easier:
1. from `Result<T, ServerFnError<NoCustomError>>` to `Result<T, ServerFnError<AnyOtherCustomError>>`
2. from some `Result<T, ServerFnError<SourceCustomError>>` to `Result<T, ServerFnError<TargetCustomError>>` when `TargetCustomError: From<SourceCustomError>`
3. from `Result<T, E>` to `Result<T, ServerFnError<AnyCustomError>>` when `E: std::error::Error`, reflecting the existing `From<E: std::error::Error> for ServerFnError<NoCustomError>`

Usage examples are provided in the docstring for each of these new traits. My typical uses cases are:
- for 1. and 2. shared helper functions returning ServerFnError already, for instance a function that calls `use_context`.
- for 3. external functions that may raise an error, a database call for instance.

As mentioned in #3153 and #3155 I would have preferred implementing `From<ServerFnError<SourceCustomError>> for ServerFnError<TargetCustomError>` and `From<E: std::error::Error> for ServerFnError`, but as @benwis correctly pointed out those conflict with `From<T> for T`, at least until specialization becomes available.

### DIFF
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -31,24 +31,50 @@ impl From<ServerFnError> for Error {
     Clone,
     Copy,
 )]
-#[cfg_attr(
-    feature = "rkyv",
-    derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
-)]
-pub struct NoCustomError;
+pub enum NoCustomError {}
+
+#[cfg(feature = "rkyv")]
+impl rkyv::Archive for NoCustomError {
+    const COPY_OPTIMIZATION: rkyv::traits::CopyOptimization<Self> =
+        rkyv::traits::CopyOptimization::disable();
+    type Archived = ();
+    type Resolver = ();
+    fn resolve(&self, _: Self::Resolver, _: rkyv::Place<Self::Archived>) {
+        match *self {}
+    }
+}
+
+#[cfg(feature = "rkyv")]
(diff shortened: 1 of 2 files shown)


benwis:
I think there are some nice utilities here. Originally I had meant the server_fn_error! macro to convert Error into ServerFnError depending on the traits Error implemented. I suspect that can be expanded to handle some of these conversions here and land on a more general single trait.

server_fn/src/error.rs
```diff
@@ -31,24 +31,50 @@ impl From<ServerFnError> for Error {
...
     }
 }
 
 impl FromStr for NoCustomError {
     type Err = ();
 
-    fn from_str(_s: &str) -> Result<Self, Self::Err> {
-        Ok(NoCustomError)
+    fn from_str(_: &str) -> Result<Self, Self::Err> {
+        Err(())
```
tqwewe, line 77: Ironically, I do think `Infallible` would make sense here as the `Err` type for `FromStr` (since it implements `Error`), but this is a breaking change so not sure if its worth the effort.

gbj:
I'm going to close this in favor of #3274, which has been merged into the `leptos_0.8` branch. If there are other changes you think should be made, feel free to open a new PR against that branch!

PR #3255, opened 2024-11-17
fix: type inference regression due to `From<T> for Signal<T>` impl (closes #3200)

### LINKED ISSUE #3200: Signal type conversion
**Describe the bug**

If the component parameter is of type Signal\<T\>, [impl\<T\> From\<T\> for Signal\<T\>](https://github.com/leptos-rs/leptos/blob/5cc2f3858dbd9307c6c630f5f25172511bc8c036/reactive_graph/src/wrappers.rs#L616) and [impl\<T\> From\<T\> for Signal\<Option\<T\>\>](https://github.com/leptos-rs/leptos/blob/5cc2f3858dbd9307c6c630f5f25172511bc8c036/reactive_graph/src/wrappers.rs#L859) will conflict. 

```
error[E0283]: type annotations needed
  --> src/main.rs:5:9
   |
5  |         view! { <SimpleCounter step=1 /> }
   |         ^^^^^^^^^^^^^^^^^^^^^^^----^^^^^^^
   |         |                      |
   |         |                      required by a bound introduced by this call
   |         cannot infer type for struct `leptos::prelude::Signal<_>`
   |
   = note: multiple `impl`s satisfying `leptos::prelude::Signal<_>: From<i32>` found in the `reactive_graph` crate:
           - impl<T> From<T> for leptos::prelude::Signal<T>
             where T: Send, T: std::marker::Sync, T: 'static;
           - impl<T> From<T> for leptos::prelude::Signal<std::option::Option<T>>
             where T: Send, T: std::marker::Sync, T: 'static;
   = note: required for `i32` to implement `Into<leptos::prelude::Signal<_>>`
```

**Leptos Dependencies**

leptos-v0.7.0-rc1

**To Reproduce**
Run the following code:
```rust
use leptos::prelude::*;

pub fn main() {
    mount_to_body(|| {
        view! { <SimpleCounter step=1 /> }
    })
}

#[component]
pub fn SimpleCounter<T>(#[prop(into)] step: Signal<T>) -> impl IntoView
where
    T: Send + Sync + 'static,
{
    view! {<div></div>}
}
```

**Expected behavior**

Delete `impl<T> From<T> for Signal<Option<T>>` and keep only `impl<T> From<T> for Signal<T>`.

**Screenshots**

**Additional context**

gbj:


zakstucke:
As outlined in https://github.com/leptos-rs/leptos/issues/3200#issuecomment-2481637746, I don't think this is the right way to go.

`SimpleCounter<T>(#[prop(into)] step: Signal<T>)` will always be a broken pattern (as it was in 0.6 anyway inference-wise), the compiler will always have a conflict on some input type that should be conceptually valid.

gbj:
Have you tried `SimpleCounter<T>(#[prop(into)] step: Signal<T>)`? It works on 0.6, and it works on 0.7 with this PR ([see here](https://github.com/leptos-rs/leptos/pull/3255/files#diff-68f7fae7c3dadd5ca1d616c959ca932b9b82af8424aa907fa708774d6e07dc64R6)) so I'm not sure what you mean by saying it "will always be a broken pattern," unless I'm misunderstanding something.

gbj:
I think you're right, see my comment back on the issue :-)

zakstucke:
Awesome, glad we hashed this one out, this PR made me nervous 😅

Issue #3263, opened 2024-11-19
Support `fallback` or `loading_view` prop in `ProtectedRoute`


tqwewe:
**Is your feature request related to a problem? Please describe.**
The new ProtectedRoute using transition behind the scenes is great. However there doesn't seem to be a way for users to customize the fallback view for when the condition is still being resolved asynchronously.

**Describe the solution you'd like**
The `ProtectedRoute` component should have an optional prop for allowing a view to be rendered while the condition is loading. This allows for nice loading screens etc.

gbj:
Should be a 1-2 line change to add a fallback and pass it through to the Transition -- Are you interested in making a PR?

tqwewe:
Yep sure! I'm a little confused about what to name the new prop. `fallback` in the `Routes` component means the view to render if the route isn't matched. Are we fine to use then name `fallback` in the `ProtectedRoute` for the transition fallback?

PR #3264, opened 2024-11-19
feat: add `fallback` prop to `ProtectedRoute` and `ProtectedParentRoute`

tqwewe:
Adds a `fallback` prop to protected routes when the inner suspense is loading or the condition returns `None`.

Closes https://github.com/leptos-rs/leptos/issues/3263


benwis:
I like it! Nice work! I'm happy to merge this, I wonder if any of our examples use it. It'd be a nice addition

PR #3274, opened 2024-11-22
Add FromServerFnError trait to server_fn crate

### LINKED ISSUE #3270: Allow custom error type in server_fn return type for enhanced ergonomics and security
- Any `?` operator in the server function might inadvertently expose sensitive error messages from the server side because the From<E> is implemented for `ServerFnError` which uses `error.to_string()`.
- Functions that return `anyhow::Error` are cumbersome to use because they need explicit conversion (e.g., using the `server_fn_error!` macro).
- It's difficult to share structured error types between server and client because `ServerFnError` has its own serialization strategy utilizing `FromStr` and `Display`.
- Since `ServerFnError` is an external type from applications and has tailored trait implementations, users cannot implement arbitrary traits such as `From<MyError>`.

To address these issues, I propose the following:

1. Make `ServerFnError` only intended for using it in server_fn return type just for the ease of error conversion with `?`.
2. Use `ServerFnErrorErr` instead of `ServerFnError` everywhere internally.
3. Allow users to specify any error type for the server fn result where `E: Serialize + DeserializeOwned + From<ServerFnErrorErr> + Debug + Display`.
4. Depricate `ServerFnError(Err)::WrappedServerError` variant, `ViaError`, `server_fn_error!`, and `NoCustomError`.

This maintains backward compatibility in the typical server_fn use case that uses `ServerFnError`.

An alternative solution could simplify the implementation further but would sacrifice backward compatibility. In this scenario, users can still use `ServerFnError` as the return type of server_fns, but error conversion with `?` from any `E: Error` types is no longer available.

1. Remove the `WrappedServerError` variant, `ServerFnErrorErr`, and error conversion toolings, including ViaError.
2. Implement `Error` for `ServerFnError`. This removes the support of `From<E: Error>` for `ServerFnError`, but it's still allowed to use it as the error type because it fulfills the below bounds.
3. Allow users to specify any error type for the server fn result where `E: Serialize + DeserializeOwned + From<ServerFnError> + Debug + Display`.

**Additional context**

- @ealmloff, @benwis, and @gbj on Leptos Discord: https://discord.com/channels/1031524867910148188/1031525141382959175/1289280751791247367
- This feature request builds on a related proposal I made for the Dioxus project: https://github.com/DioxusLabs/dioxus/issues/3243
- me and @ealmloff on Dioxus Discord: https://discord.com/channels/899851952891002890/1108097327475982386/1308637911813722152

ryo33:
Closes #3270 

- [x] Added `FromServerFnError`
- [x] `cargo clippy -p server_fn --all-features`
- [x] `cargo clippy -p leptos_server --all-features`
- [ ] Update `server_fn_macros` to use `FromServerFnError` trait
- [ ] End to end smoke tests???

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3266,6 +3266,7 @@ version = "0.7.0-rc1"
 dependencies = [
  "actix-web",
  "axum",
+ "base64",
  "bytes",
  "ciborium",
  "const_format",
--- a/leptos_server/src/action.rs
+++ b/leptos_server/src/action.rs
@@ -3,7 +3,10 @@ use reactive_graph::{
     owner::use_context,
     traits::DefinedAt,
 };
-use server_fn::{error::ServerFnErrorSerde, ServerFn, ServerFnError};
+use server_fn::{
+    error::{FromServerFnError, ServerFnErrorSerde},
+    ServerFn,
+};
 use std::{ops::Deref, panic::Location, sync::Arc};
 
 /// An error that can be caused by a server action.
@@ -42,7 +45,7 @@ where
     S: ServerFn + 'static,
     S::Output: 'static,
 {
-    inner: ArcAction<S, Result<S::Output, ServerFnError<S::Error>>>,
+    inner: ArcAction<S, Result<S::Output, S::Error>>,
     #[cfg(debug_assertions)]
     defined_at: &'static Location<'static>,
 }
@@ -52,13 +55,14 @@ where
     S: ServerFn + Clone + Send + Sync + 'static,
     S::Output: Send + Sync + 'static,
     S::Error: Send + Sync + 'static,
+    S::Error: FromServerFnError,
 {
     /// Creates a new [`ArcAction`] that will call the server function `S` when dispatched.
     #[track_caller]
     pub fn new() -> Self {
         let err = use_context::<ServerActionError>().and_then(|error| {
             (error.path() == S::PATH)
-                .then(|| ServerFnError::<S::Error>::de(error.err()))
+                .then(|| S::Error::de(error.err()))
                 .map(Err)
         });
         Self {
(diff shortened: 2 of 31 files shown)


ryo33:
@nicolas-guichard Hi, this fix is similar to #3253, but takes a different approach, addressing #3153 and #3155. It would be great if you could take a look and review whether this approach works.

examples/server_fns_axum/src/app.rs
```diff
@@ -652,32 +653,72 @@ pub fn FileWatcher() -> impl IntoView {
...
+    fn from(value: String) -> Self {
+        MyErrors::Other(value)
+    }
+}
+
+impl FromServerFnError for MyErrors {
+    fn from_server_fn_error(value: ServerFnErrorErr) -> Self {
+        MyErrors::ServerFnError(value)
+    }
+}
```
ryo33, line 715: This is the new way to define a custom error type for server fns. It can used directly in the signature of server fn like `-> Result<(), MyErrors>` instead of `-> Result<(), ServerFnError<MyErrors>`.

gbj:
Thanks for this work!

While I have not had time to manually test it, and have not reviewed every line, this looks like it is the right approach, and it certainly seems to make custom errors more usable, which I think is great. 

It comes at a slightly awkward time in our release cycle as we are planning a 0.7 release this weekend and this makes some breaking changes that won't have time to be tested in the beta/rc process before that release. 

So here's my suggestion:
- @benwis plans to test this out and we will welcome other testers and reviews
- We will release 0.7 without this included
- We will also plan a 0.8 some time in late December/early January, that includes the final version of this PR and any breaking-change-requiring bugfixes from 0.7

ryo33:
Thank you for your feedback! That release schedule is reasonable to me.

gbj:
Thanks so much for your work! I am going to start merging things into a `leptos_0.8` branch, so this can be found there for now.

fsktom:
Not sure I can thank you enough - absolutely amazing changes! 🔥

PR #3279, opened 2024-11-23
Referential context: with_context and update_context

zakstucke:


reactive_graph/src/owner/context.rs
```diff
@@ -319,3 +356,84 @@ pub fn expect_context<T: Clone + 'static>() -> T {
 pub fn take_context<T: 'static>() -> Option<T> {
     Owner::current().and_then(|owner| owner.take_context())
 }
+
+/// Access a reference to a context value of type `T` in the reactive system.
+///
+/// This traverses the reactive ownership graph, beginning from the current reactive
+/// [`Owner`] and iterating through its parents, if any. When the value is found, it is cloned.
```
gbj, line 363: > When the value is found, it is cloned.

Rather, the function that you pass is applied to a reference to it.

reactive_graph/src/owner/context.rs
```diff
@@ -319,3 +356,84 @@ pub fn expect_context<T: Clone + 'static>() -> T {
...
+/// # });
+/// ```
+pub fn with_context<T: 'static, R>(cb: impl FnOnce(&T) -> R) -> Option<R> {
+    Owner::current().and_then(|owner| owner.with_context(cb))
+}
+
+/// Update a context value of type `T` in the reactive system.
+///
+/// This traverses the reactive ownership graph, beginning from the current reactive
+/// [`Owner`] and iterating through its parents, if any. When the value is found, it is cloned.
```
gbj, line 401: Same idea as above.

gbj (commented):
Looks good! And thanks for fixing the double-locking bug in `take_context` that someone mentioned to me in passing three months ago but never made a PR to fix. Two notes on docs and this should be good to go.

zakstucke:
Updated!

gbj:
Thanks very much!

PR #3287, opened 2024-11-24
Fix help message for island macro

NiklasEi:
I stumbled over the help message of the `island` macro using `component` instead of `island`.


gbj:
Thanks!

PR #3304, opened 2024-11-29
Update Leptos to use latest wasm-bindgen version

SleeplessOne1917:
`wasm-bindgen` got another update, so I took the liberty of bumping the version used by Leptos.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1761,9 +1761,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.72"
+version = "0.3.73"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


benwis:
We'll need to keep this in sync with the version in cargo-leptos, we should be able to merge this for the upcoming 0.7 dot release

SleeplessOne1917:
Would it also be possible to backport the wasm-bindgen dependency update to 0.6?

gbj:
These dependencies do not need to be updated; cargo will automatically resolve them to the highest patch version in your application. The only one that needs an update is `cargo-leptos`, and that update will apply to both 0.6 and 0.7 as cargo-leptos is agnostic about leptos version.

SleeplessOne1917:
I can delete this if you don't think it's necessary.

PR #3310, opened 2024-12-01
Remove the Send requirement on a local Action's future (fixes #3309)

### LINKED ISSUE #3309: The future returned from the function passed into Action::new_local has an unnecessary(?) Send requirement
**Describe the bug**

The signature of `Action::new_local` is
```rust
    pub fn new_local<F, Fu>(action_fn: F) -> Self
    where
        F: Fn(&I) -> Fu + 'static,
        Fu: Future<Output = O> + Send + 'static,
```
which makes this function less useful than it could be, since it means that all values held across await points must be `Send`.  I don't think the `Send` requirement on the future is actually needed though?  I forked leptos locally and removed it, and it still seems to compile and work properly.

**Leptos Dependencies**

```toml
leptos = { version = "0.7.0", features = ["csr"] }
```

Also checked my fork with the `ssr` feature in case there was a requirement for `Send`ness that only shows up when server-side rendering is enabled.

**To Reproduce**
```rust
Action::new_local(|()| async move {
  let x = Rc::new(0);
  futures_util::future::ready(()).await;
  *x
})
```

**Expected behavior**
This should compile.

**Additional context**
This came up in code ported from leptos 0.6, which has such a non-Send value being held across an await.

rjmac:


benwis:
LGTM, Thanks for noticing and submitting

Issue #3298, opened 2024-11-29
It's not possible to redirect in a server function if the request comes from a plain HTML form


veigaribo:
I'm trying to be as thorough as possible, so sorry if there's a mistake somewhere.

Also I'm not sure I would consider this a bug, or rather a limitation caused by the `server_fn` error-handling abstraction that is there by design, but recent discussions in the Discord server have led me to think of it as the former.

**Describe the bug**

When a server function is invoked through a plain HTML form, presumably because JS and/or WASM is not being used, the following line of code will set the HTTP status code and the `Location` HTTP header of the response to some values, to allow for server-side rendering of errors (AFAIK):

<https://github.com/leptos-rs/leptos/blob/5947aa299e5299eb3dc75c58e28affb15e79b6ff/server_fn/src/lib.rs#L291>

But, as a consequence, it is not possible to make the server function redirect anywhere else, since the above will end up overwriting the HTTP response in the end.

**To Reproduce**
1. Write a server function that tries to redirect somewhere different from the originating URL, for example using [`leptos_actix::redirect`](https://docs.rs/leptos_actix/0.7.0-rc3/leptos_actix/fn.redirect.html) or [`leptos_axum::redirect`](https://docs.rs/leptos_axum/0.7.0-rc3/leptos_axum/fn.redirect.html);
2. Write a form that invokes that function, for example using an [`ActionForm`](https://docs.rs/leptos/0.7.0-rc3/leptos/form/fn.ActionForm.html) with a [`ServerAction`](https://docs.rs/leptos/0.7.0-rc3/leptos/server/struct.ServerAction.html);
3. Load and submit that form in a browser with JavaScript disabled;
4. Note that the value of the `Location` header of the response is the originating URL, not what you set (presumably your browser will allow you to do that);
5. Optionally try the above with JavaScript enabled and see that the `Location` header is correct.

**Expected behavior**

I think the `Location` response header, as well as the status code, should end up being set to the values that were set in the server function, but I don't know how that would be the case while preserving the current behavior.

**Additional context**

I was originally hoping that the `form-redirects` feature could be turned off from `leptos` entirely, so that I could sidestep this problem by not enabling the default features; but, if this really is a bug, then there must be a better solution. Allowing disabling of `form-redirects` would have its problems too. 

[...]

PR #3311, opened 2024-12-02
fix: prevent multiple location headers on redirect (#3298)

veigaribo:
The problem turned out to be specific to 0.7 Actix, which I did not expect.

This solution is what was proposed by @gbj in #3298, although I did not make the new code part of `extend_response` because `extend_response` is also [called in `handle_static_route`](https://github.com/leptos-rs/leptos/blob/d665dd4b89151e5d797df3db5cd2260cbe1e8fae/integrations/actix/src/lib.rs#L1285), and I don't think it needs this (maybe I'm wrong?).

I believe a better solution would actually be to match the Axum behavior, that uses [`http::HeaderMap::extend`](https://github.com/hyperium/http/blob/5a1a5e8ee54fc2e2f274faf318d9a8b1129c386a/src/header/map.rs#L2081): For merging map A with map B, if map B contains a key that is present in A, all values with such key in A are removed and all values with that key in B are added. That would at least be more consistent.

I also still don't understand why both `server_fn::ServerFn::run_on_server` and `leptos_actix::handle_server_fns_with_context` need to handle redirection to Referer, but I assume there is a reason.

Took the oportunity to fix the surrounding comments too, to the best of my ability.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -375,8 +375,8 @@ pub fn handle_server_fns_with_context(
                                     .take(),
                             );
 
-                            // it it accepts text/html (i.e., is a plain form post) and doesn't already have a
-                            // Location set, then redirect to to Referer
+                            // if it accepts text/html (i.e., is a plain form post) and doesn't already have a
+                            // Location set, then redirect to the Referer
                             if accepts_html {
                                 if let Some(referrer) = referrer {
                                     let has_location =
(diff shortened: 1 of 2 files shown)


benwis:
I think the problem with this one is that headers like Set-Cookie are set multiple times and should not be overwritten this way

veigaribo:
> I think the problem with this one is that headers like Set-Cookie are set multiple times and should not be overwritten this way

I think this solution is pretty hacky, but it should not affect any header other than Location 🤔 because that's what we call `.remove()` on

benwis:
LGTM, seems ideal

Issue #3322, opened 2024-12-04
`Dependabot`


sabify:
`Dependabot` has been removed in 21f26e7a6b2b8aa32b64daff6397977ec1d48f87 and I think it is due to PR pollution caused by this bot. I believe it is still worth keeping it but have fewer opened PRs by this bot. It can easily be achieved by: `open-pull-requests-limit: 3`.

`leptos` has pretty large dependencies and keeping them up-to-date manually can be time-consuming and keeping them up-to-date is essential for security and stability. By limiting the open pull requests to three, we maintain a manageable workflow while staying current with dependency changes. This not only enhances our project's reliability but also frees up time for the team to focus on new features and improvements.

OM-HASE:
@sabify, can I try it.

gbj:
Frankly the reason we removed it was that the PRs were, for the most part, completely useless. We do not need to add a lock file in order to add Dependabot in order to have it make PRs to bump patch releases of dependencies. Cargo will do this on its own and the notification spam made actual project maintenance harder. 

I'm open to other configurations, but the main goal here would be to limit bot PRs to actual updates.

benwis:
I agree, I would be in favor of this as a CI step during releases, but mostly it is a giant pain for the day to day

sabify:
It seems the most useful update strategy doesn't work in cargo version of dependabot, `versioning-strategy: increase-if-necessary`. https://github.com/dependabot/dependabot-core/issues/4009

And because of that some switched to renovate like Astral (https://github.com/astral-sh/uv/pull/2653).

romanstingler:
offtopic: 
lol empty project 
```
/s/r/qwer ❯ cargo clean                                                                                                          161.797s (main|?) 23:50
     Removed 4346 files, 3.2GiB total
 ```

PR #3323, opened 2024-12-04
#3322 Dependabot 

OM-HASE:
#3322 , Open Pull Request Add to 3.


benwis:
I appreciate the desire for Dependabot here, but we've had it before, and it mostly spammed us with useless PRs for minor version updates. I wouldn't mind having something in the release pipeline to check versions, but I don't think we're going to merge this

PR #3333, opened 2024-12-07
feat: add `scroll` prop to `<A/>` component to control scrolling behavior (closes #2666)

### LINKED ISSUE #2666: Add "scroll" prop to <A>
I would like to have a table that has pagination buttons that are proper links and set the `?page=X` query parameter. Unfortunately this causes it to scroll to the top of the page when clicked. Because the table isn't at the top of the page, it's jarring and annoying to have to scroll down to the table again every time.

I'd like to be able to pass `scroll=false` to Leptos' `<A>` component to tell it not to scroll to the top of the page when navigating. 

I've considered just using on:click and use_navigate but that's discouraged because it isn't as accessible according to the [Leptos book](https://book.leptos.dev/router/19_a.html?highlight=scroll#navigating-programmatically):

> You should almost never do something like `<button on:click=move |_| navigate(/* ... */)>`. Any `on:click` that navigates should be an `<a>`, for reasons of accessibility.

**Additional context**
I looked up how this was done in the React world for inspiration and here they have a `scroll` property on their `<link>` component: https://nextjs.org/docs/pages/api-reference/components/link#scroll

gbj:


paul-hansen (approved):
Looks and works great! Thanks!

I don't have my app fully updated yet so couldn't test it there, I added it temporarily to the hackernews_axum example with the next/prev buttons and it stayed a the bottom of the page instead of scrolling to the top as expected. It didn't make sense to keep it that way there, but worked for quick testing. (It makes more sense for pager controls that are at the bottom of the page.)

PR #3337, opened 2024-12-07
Increase number of branch arms for `either!` macro

bicarlsen:
Increased number of possible branch arms for the `either!` macro from 6 to 8.


bicarlsen:
@gbj It seems that these changes haven't yet been included in the `leptos` release. It seems like `leptos` is still using `0.1.1` instead of `0.1.2`.

gbj:
@bicarlsen Ah thanks. I've just published 0.1.2, so `cargo update` should pull it in for you.

PR #3341, opened 2024-12-09
Fix a doc typo and one broken intra-doc link

FreezyLemon:
Noticed the typo while following along with the book, and then got a rustdoc warning for the `ArenaItem` link. Figured that I might as well fix that too.


gbj:
Thanks!

PR #3342, opened 2024-12-09
Adding a project detailing flexible mocking architecture inspired by hexagonal architecture for leptos app

sjud:
As title. There is a detailed read me detailing the purpose of the project and what it's showing.


benwis:
I've certainly skimmed it and it looks like an interesting idea. I'm happy to merge this, with the potential caveat that I'm not sure Greg or I understand this, so if it gets too broken we may remove it.

PR #3343, opened 2024-12-10
Implement signal.into_inner()

### LINKED ISSUE #3068: Signal::into_inner

I am using Leptos to manage large objects, including [WebGPU objects that cannot be cloned](https://docs.rs/wgpu/latest/wgpu/struct.CommandEncoder.html). I would wrap the object in a signal, do some reactive operations with it, *and would subsequently like to get the object back*. 

A very similar case came up while I was trying to implement a `pop` method for `struct SignalVec<T> { ArcRwSignal<Vec<ArcRwSignal<T>>> }` . See https://github.com/leptos-rs/leptos/discussions/3066

I'd love it if there were a function analogous to the standard library [`RwLock/Arc::into_inner`](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.into_inner).

```rust
// for the ArcRwSignal

/// Returns the inner value if this is the only reference to to the signal.
/// Otherwise, returns `None`.
/// # Panics
/// Panics if the inner lock is poisoned.
#[track_caller]
pub fn into_inner(self) -> Option<T> {
    // I believe the subscribers don't matter here, because they aren't holding a reference to the value.
    Arc::into_inner(self.value)?.into_inner().ok()
}
```

Using clunky workarounds, such as wrapping my objects into `Option`s, and taking the values out of the options.

stefnotch:
This implements `into_inner` the way it is implemented for `Arc`s. This fixes #3068 .

However, with the non-arc signals, the behaviour can be rather unintuitive. For example, the following does not succeed.
Why?
Because there's still the original `RwSignal` that was created inside the `signal` function. (`RwSignal::new(value).split()` creates 3 signals)

```rust
#[test]
fn into_inner_non_arc_signal() {
    let owner = Owner::new();
    owner.set();

    let (a, b) = signal(2);
    assert_eq!(a.get(), 2);
    b.dispose(); // strong count is now 2
    assert_eq!(a.into_inner(), Some(2)); // fails
}
```

I'm happy to document this, or to attempt a fix, or to remove the trait for normal signals.

### DIFF
--- a/reactive_graph/src/owner/arc_stored_value.rs
+++ b/reactive_graph/src/owner/arc_stored_value.rs
@@ -1,6 +1,6 @@
 use crate::{
     signal::guards::{Plain, ReadGuard, UntrackedWriteGuard},
-    traits::{DefinedAt, IsDisposed, ReadValue, WriteValue},
+    traits::{DefinedAt, IntoInner, IsDisposed, ReadValue, WriteValue},
 };
 use std::{
     fmt::{Debug, Formatter},
@@ -122,5 +122,14 @@ where
 impl<T> IsDisposed for ArcStoredValue<T> {
     fn is_disposed(&self) -> bool {
         false
+    }
+}
+
+impl<T> IntoInner for ArcStoredValue<T> {
+    type Value = T;
+
+    #[inline(always)]
+    fn into_inner(self) -> Option<Self::Value> {
+        Some(Arc::into_inner(self.value)?.into_inner().unwrap())
     }
 }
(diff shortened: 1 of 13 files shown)


gbj:
I think my changes here to the implementation of `signal()` fix this, and cause the `signal()` test to work. It initially failed because `signal()` was implemented as `RwSignal::new(value).split()`, which means there's another reference to the inner `ArcRwSignal` (because it created an `RwSignal` and stored it in the owner). This version should be strictly better, in that it means only allocating 2 slots in the arena per `signal()` instead of 3, which is better for memory anyway. Thanks for surfacing that with the test, and for implementing this useful trait!

stefnotch:
Sweet, thank you for bringing this over the finish line.

PR #3349, opened 2024-12-11
TypedChildrenFn impl Clone trait

redforks:
TypedChildrenFn wraps `Arc<dyn Fn() -> View<T> + Send + Sync>`, but not derived with `Clone`, I think it is logical for a type based on `Arc`, and type `ViewFn` derived with Clone as well.

### DIFF
--- a/leptos/src/children.rs
+++ b/leptos/src/children.rs
@@ -285,6 +285,13 @@ impl<T> Debug for TypedChildrenFn<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


gbj:
I agree that it should implement `Clone`, but using `#[derive(Clone)]` here will put a `T: Clone` bound on it, which isn't necessary. It should be a manual `Clone` implementation that uses `Arc::clone()` on the inner `Arc` instead.

redforks:
Update the source code,

PR #3351, opened 2024-12-11
docs: clarify `Signal::derive()` behavior

gbj:


benwis (approved):
I like it, a lot clearer verbiage

tqwewe:
Very clear! Definitely an improvement that would clear the initial confusion I had. 

The `(and convert it into a Signal<_>)` part seems a little redundant but definitely just a nit pick

bicarlsen:
There's a small typo on line 101
> /// Values can be access with the same function call ...

Should be
> /// Values can be accessed with the same function call ...

reactive_graph/src/wrappers.rs
```diff
@@ -93,12 +93,29 @@ pub mod read {
...
-    /// or derived signal closure.
+    /// an [`ArcReadSignal`], [`ArcMemo`], [`ArcRwSignal`], or derived signal closure,
+    /// or a plain value of the same type
     ///
-    /// This allows you to create APIs that take any kind of `ArcSignal<T>` as an argument,
-    /// rather than adding a generic `F: Fn() -> T`. Values can be access with the same
-    /// function call, `with()`, and `get()` APIs as other signals.
+    /// This allows you to create APIs that take `T` or any reactive value that returns `T`
+    /// as an argument, rather than adding a generic `F: Fn() -> T`.
+    /// Values can be access with the same function call, `read()`, `with()`, and `get()`
```
bicarlsen, line 101: There is two small typos here.
It should be
> /// Values can be accessed with the same function calls ...

reactive_graph/src/wrappers.rs
```diff
@@ -330,11 +347,29 @@ pub mod read {
     }
 
     /// A wrapper for any kind of arena-allocated reactive signal:
-    /// an [`ReadSignal`], [`Memo`], [`RwSignal`], or derived signal closure.
+    /// an [`ReadSignal`], [`Memo`], [`RwSignal`], or derived signal closure,
```
bicarlsen, line 350: Another small typo. Should be
> a [`ReadSignal`], [`Memo`], ...

reactive_graph/src/wrappers.rs
```diff
@@ -330,11 +347,29 @@ pub mod read {
...
+    /// or a plain value of the same type
+    ///
+    /// This allows you to create APIs that take `T` or any reactive value that returns `T`
+    /// as an argument, rather than adding a generic `F: Fn() -> T`.
+    /// Values can be access with the same function call, `read()`, `with()`, and `get()`
```
bicarlsen, line 355: Same typos as above:
> /// Values can be accessed with the same function calls ...

Issue #3362, opened 2024-12-14
Implement IntoSplitSignal for Field<T>


fiadliel:
IntoSplitSignal seems to be missing for Field<T> (erased Subfield). This would be a small QoL improvement when using data binding.

This compiles and works for me, but I'm not sure if the bounds are correct:

```rust
#[cfg(feature = "reactive_stores")]
impl<T, S> IntoSplitSignal for Field<T, S>
where
    Self: Get<Value = T> + Update<Value = T> + Clone,
    S: Storage<ArcField<T>>,
{
    type Value = T;
    type Read = Self;
    type Write = Self;

    fn into_split_signal(self) -> (Self::Read, Self::Write) {
        (self.clone(), self.clone())
    }
}
```

I could make a PR if interested.

gbj:
A PR would be great.

PR #3364, opened 2024-12-14
Implement IntoSplitSignal for Field.

fiadliel:
Adds `IntoSplitSignal` for `Field`. This allows us to use `bind:value=field` instead of `bind:value=(field, field)`.

Copying some similar trait bounds, so please take a look and see if they make sense.

Fixes https://github.com/leptos-rs/leptos/issues/3362


gbj:
Thanks!

PR #3371, opened 2024-12-16
impl Dispose for Callback types and add try_run to the Callable trait

basro:
PR aims to make the Callback types more consistent with the other stored types in leptos.

* Adds `try_run` to callable which (in line with other stored types) will return `None` if the callback had been disposed.
* Adds `impl Dispose` for both `Callback` and `UnsyncCallback`

Small detail regarding implementation:
The impl of `Callable::run` for `UnsyncCallback` and `Callback` was inconsistent. One used `try_with_value` with expect while the other was using `with_value`.
I chose to make them consistent and use `with_value` in both impls. I'm unsure if I chose the right option, perhaps try_with_value was preferable.

### DIFF
--- a/leptos/src/callback.rs
+++ b/leptos/src/callback.rs
@@ -241,11 +241,9 @@ impl<In: 'static, Out: 'static> Callback<In, Out> {
 
 #[cfg(test)]
 mod tests {
-    use reactive_graph::traits::Dispose;
-
-    use crate::callback::{Callback, UnsyncCallback};
-
     use super::Callable;
+    use crate::callback::{Callback, UnsyncCallback};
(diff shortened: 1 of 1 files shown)


gbj:
I'm very open to adding this.

Note that as it stands, it is a breaking change (adds a required method to a `pub` trait) so can't be merged until 0.8. I'm currently planning on releasing a 0.8 some time in January that includes any very minor-but-breaking changes like this that have turned up as people test out 0.7 further.

gbj:
Thanks again! I'm going to start merging things into a `leptos_0.8` branch, so I will merge this now.

PR #3379, opened 2024-12-17
Update elements.rs to include `popovertarget` and `popovertargetaction` for the `<button>` element

Figments:
Adds `popovertarget` and `popovertargetaction` to the list of valid `<button>` attributes.


gbj:
Thanks!

PR #3382, opened 2024-12-18
feat: allow to destructure props

geovie:
Allows props to be destructured in the component definition e.g.
```rust
    #[prop(name = "data")] UserInfo { email, user_id }: UserInfo,
 ```
 
for that to work a `name` attribute was introduced which is used to define the name of the prop in the generated builder.

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -655,14 +655,44 @@ impl Prop {
                 abort!(e.span(), e.to_string());
             });
 
-        let name = if let Pat::Ident(i) = *typed.pat {
-            i
-        } else {
-            abort!(
-                typed.pat,
-                "only `prop: bool` style types are allowed within the \
-                 `#[component]` macro"
-            );
+        let name = match *typed.pat {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: leptos_macro/src/component.rs [hunk 1]
```diff
@@ -655,14 +655,43 @@ impl Prop {
...
-        } else {
-            abort!(
-                typed.pat,
-                "only `prop: bool` style types are allowed within the \
-                 `#[component]` macro"
-            );
+        let name = match *typed.pat {
+            Pat::Ident(i) => {
+                if let Some(name) = &prop_opts.name {
+                    PatIdent {
```
geovie, line 661: This also allows to use the new `name` attribute on props which are not destructured.   
That means props can have different names inside the component and outside (builder).

We can also error here if we don't want to allow renaming normal props

gbj:
Thanks, this looks good! I updated the tests for macro output with `TRYBUILD=overwrite` — looks mostly like the compiler updated what it underlines in that case. Assuming this all then passes I'll be happy to merge.

PR #3383, opened 2024-12-18
docs: showcase let syntax in for_loop

### LINKED ISSUE #3059: Destructuring `let` attribute in `<For>` component
When iterating over a structured type (i.e. a `tuple` or `struct`) in a `<For>` component, it would be convenient if we could destructure it in the `let` attribute. To do this at the moment requires additional boilerplate.
e.g. 
```rust
struct Item {
  pub name: String,
  pub value: String,
}

#[component]
fn List(list: Vec<Item>) -> impl IntoView {
  view! {
    <For
      each=move || list
      key=|(name, _)| name.clone()
      let:item
    >
      {move || { // Boilerplate to destructure
        let Item { name, value } = &item;
        view! {
          <div>{name} ": " {description}</div>
        }
      }}
    </For>
  }
}
```
```rust
type Point = (i32, i32);

#[component]
fn Points(points: Vec<Point>) -> impl IntoView {
  view! {
    <For
      each=move || points
      key=|point| point
      let:point
    >
      {move || { // Boilerplate to destructure
        let (x, y) = point;
        view! {
          <div>{x} ", " {y}</div>
        }
      }}
    </For>
  }
}
```

It would be nice to allow destructuring in the `let` attribute to remove this additional layer of code.
```rust
#[component]
fn List(list: Vec<Item>) -> impl IntoView {
  view! {
    <For
      each=move || list
      key=|(name, _)| name.clone()
      let:Item { name, value }  // Destructure in `let`
    >
      <div>{name} ": " {description}</div>
    </For>
  }
}
```
```rust
#[component]
fn Points(points: Vec<Point>) -> impl IntoView {
  view! {
    <For
      each=move || points
      key=|point| point
      let:(x, y)  // Destructure in `let`
    >
      <div>{x} ", " {y}</div>
    </For>
  }
}
```

purung:
I thought it would be a boon for DX to showcase the `let` syntax for the `<For>` component. 

I knew there were some improvements, but I erroneously tried the legacy `let:item` style to destructure my struct and had to go looking for docs. It was kind of hard to find so I thought I'd try to contribute :sweat_smile: Please, edit and adjust to your liking. 

I think this should close #3059 

See cousin PR for the book.

### DIFF
--- a/leptos/src/for_loop.rs
+++ b/leptos/src/for_loop.rs
@@ -49,13 +49,30 @@ use tachys::{reactive_graph::OwnedView, view::keyed::keyed};
 /// component, using the `let` syntax:
 ///
 /// ```
-///  <For
-///    each=move || counters.get()
-///    key=|counter| counter.id
-///    let(counter)
-///  >
-///      <button>"Value: " {move || counter.count.get()}</button>
-///  </For>
+/// # use leptos::prelude::*;
+///
+/// # #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+/// # struct Counter {
+/// #   id: usize,
+/// #   count: RwSignal<i32>
+/// # }
+/// #
+/// # #[component]
+/// # fn Counters() -> impl IntoView {
+/// #   let (counters, set_counters) = create_signal::<Vec<Counter>>(vec![]);
+/// #
+///   view! {
+///     <div>
+///         <For
+///           each=move || counters.get()
+///           key=|counter| counter.id
+///           let(counter)
+///         >
+///             <button>"Value: " {move || counter.count.get()}</button>
+///         </For>
+///     </div>
+///   }
+/// # }
 /// ```
 ///
 /// The `let` syntax also supports destructuring the pattern of your data.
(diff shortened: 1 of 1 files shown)


gbj:
Thanks for the PR! I think this is a good idea. 3 of the 4 CI failures here are unrelated (semver, autofix, and any_spawner), but the `leptos` one is caused by the PR and needs a fix: specifically, the doctests don't compile because they're missing imports and the `view` macro.

It's good to have these and they double as regression tests. You should just need to add [the prelude import](https://github.com/leptos-rs/leptos/pull/3383/files#diff-7d35b5f3b5ed44cb41c9fc0f0c033b3757a5fd899b264bd353ffb0cc7c458922R17) and also a `view! {` wrapping the two examples and it should be fine. (You can comment these out with a `#` at the beginning of the line so they don't show up in the docs.) I'd recommend running `cargo test` locally to make sure you've got it right.

purung:
Thank you for the guidance! Tests pass locally now. I chose to keep whole snippets of `view! {}`. Let me know if you'd like me to reconsider the crop.

gbj:
Thank you!

PR #3386, opened 2024-12-18
Remove `From<T>` and add back `From<Fn() -> T>`

mscofield0:
The reason for is that the new `From` impls focus more on making `MaybeSignal` work for `Signal` since now it is removed, but the consequence of it is that:
```rust
#[component]
fn FooComponent(
    #[prop(into)]
    is_true: Signal<bool>,
) -> impl IntoView { ... }

<FooComponent is_true=move || blabla.get().is_true />
```
doesn't work. Ideally, `Signal` should have both `From`s, but I don't see how that's possible, so I believe that supporting the common case (the non-`MaybeSignal` one) should be preferred.

###### WARNING
This commit is a breaking change because it removes and adds a couple of trait impls.

### DIFF
--- a/reactive_graph/src/wrappers.rs
+++ b/reactive_graph/src/wrappers.rs
@@ -643,30 +643,36 @@ pub mod read {
         }
     }
 
-    impl<T: Send + Sync + 'static> From<T> for ArcSignal<T, SyncStorage> {
-        #[track_caller]
-        fn from(value: T) -> Self {
-            ArcSignal::stored(value)
-        }
-    }
-
-    impl<T> From<T> for Signal<T>
+    impl<F, T> From<F> for Signal<T>
     where
+        F: Fn() -> T + Send + Sync + 'static,
         T: Send + Sync + 'static,
     {
         #[track_caller]
-        fn from(value: T) -> Self {
-            Self::stored(value)
+        fn from(value: F) -> Self {
+            Self::derive(value)
         }
     }
 
-    impl<T> From<T> for Signal<T, LocalStorage>
+    impl<F, T> From<F> for Signal<T, LocalStorage>
     where
+        F: Fn() -> T + 'static,
         T: 'static,
     {
         #[track_caller]
-        fn from(value: T) -> Self {
-            Self::stored_local(value)
+        fn from(value: F) -> Self {
+            Self::derive_local(value)
+        }
+    }
+
+    impl<F, T> From<F> for ArcSignal<T, SyncStorage>
+    where
+        F: Fn() -> T + Send + Sync + 'static,
+        T: Send + Sync + 'static,
+    {
+        #[track_caller]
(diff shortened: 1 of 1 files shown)


mscofield0:
Weird CI errors, compiles both on stable and nightly locally for me...

gbj:
Thanks for the PR!

A couple notes:

1. `nightly` errors

>  Weird CI errors, compiles both on stable and nightly locally for me...

No, it doesn't compile with the `nightly` feature. Using the nightly compiler does not activate the nightly features; try `cd reactive_graph && cargo check --features=nightly` (when using the nightly toolchain) and you will see the compiler error. The problem is that *simply* implementing `From<F: Fn() -> T>` causing another set of conflicting implementations, because `From<ReadSignal<T>>` etc. are still implemented, but with the `nightly` feature they also implement `Fn()` themselves. 

The 0.6 implementation relied on the `auto_trait`/`negative_impls` unstable nightly features to work around this; you can consult the `leptos_0.6` branch if you want to see how that worked.

2. We do want to support both `T` and `Fn() -> T` but unfortunately that's not possible. This version *does* support both, for `FooComponent` -- it would be something like `<FooComponent is_true=false/>` and `<FooComponent is_true=Signal::derive(move || ...)/>`. In retrospect, perhaps the better choice would've been to support `<FooComponent is_true=Signal::stored(false)/>` and `<FooComponent is_true=move || .../>` as that removes the breaking change here. This simply wasn't caught by anyone who was testing during the alpha/beta/rc period over the last 6 months, unfortunately.

Note that this change doesn't make this less *expressive*, it just adds a bit of boilerplate in the derived signal case.

I'm open to changing back to that second option in the next version, although unfortunately that will just be another breaking change at that point. It's not clear to me whether the reactive (`move || blabla.get().is_true`) or static (`foo`) option should be the default, but probably the reactive one I guess. However,

3. Unfortunately this can't be merged at the moment anyway because of the breaking change, but I will mark it with the 0.8 milestone to remind myself to come back to this conversation when we're going to have another major release. (I anticipated some things like this coming up and I'm planning on releasing a 0.8 with minor breaking changes like this in late January/early Feb)

benwis:
I think I agree with @zakstucke here, the benefits of the `From<T> for Signal` impl outweigh a slightly worse reactive case. 

I've been working on updating a component library from 0.6 to 0.7, and their solution to this issue is to create their own newtype Signal wrapper and then write the requisite From impls for all their static types as well as most Signal types to make the library user's life easier. With 0.7 I can chuck that wrapper type and get the same benefits.

However that does not resolve the base case of converting a closure to a Signal for props, I don't like the different behavior between closures in the component and in props for the component. But even in 0.6 we didn't get that. Until we get Specialization this appears to be the best of two imperfect solutions, and it's at least consistent.

I'm going to close this as won't fix, Thanks @mscofield0 writing the PR and bringing this to our attention.

PR #3406, opened 2024-12-24
Enable `console` feature of `web-sys` for `reactive_graph`

alexisfontaine:
Without the feature enabled, having a direct dependency to `reactive_graph` fails to `cargo check` under `wasm32-unknown-unknown`.
Just to give you some context, I'm building my own store in a separate library which only needs the reactive primitives.

### DIFF
--- a/reactive_graph/Cargo.toml
+++ b/reactive_graph/Cargo.toml
@@ -25,7 +25,7 @@ async-lock = "3.4.0"
 send_wrapper = { version = "0.6.0", features = ["futures"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
-web-sys = "0.3.72"
+web-sys = { version = "0.3.72", features = ["console"] }
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


alexisfontaine:
This seems to be the only occurrence of `web_sys` in this crate:
https://github.com/leptos-rs/leptos/blob/28af33d5114a17ca8b0436a57020cdb2f26d3a6f/reactive_graph/src/lib.rs#L116

Are there any plans to replace `web_sys` with a logging library?

gbj:
Yeah I mean fair enough

https://github.com/leptos-rs/leptos/blob/28af33d5114a17ca8b0436a57020cdb2f26d3a6f/reactive_graph/src/lib.rs#L102-L117

> Are there any plans to replace `web_sys` with a logging library?

If you read the lines above the `web_sys` part you'll see that there already *is* a logging library, and that this exists solely to provide a fallback so that logs are not swallowed silently if `tracing` isn't enabled. (Based on experience, I can say that this is important to avoid a huge amount of user confusion.)

PR #3408, opened 2024-12-24
Provide custom state in `file_and_error_handler`

spencewenski:
The `file_and_error_handler` provided by leptos_axum does not provide the app's custom state in the context. This means that when the fallback handler tries to render the app, the app will not be able to access the custom state (and will panic if it uses `expect_context`).

The handler defined by `file_and_error_handler` has access to the custom state from Axum, so we can pass the custom state from Axum into the Leptos handler using the `additional_context` parameter of `handle_response_inner`.

I think this might technically be a semver breaking change because we're adding constraints to the `S` type parameter. However, I think it should be fine because any custom state used with this method would need to satisfy these constraints to be used in the first place (e.g., in the [LeptosRoutes trait](https://github.com/leptos-rs/leptos/blob/b91429d4f0d8cd0685a3709e42bfd602bc22aa08/integrations/axum/src/lib.rs#L1635)). However, I'll leave it up to you to decide if a version bump is needed, or if there's a way to do this without requiring changing the type constraints.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -1994,12 +1994,12 @@ pub fn file_and_error_handler<S, IV>(
        + 'static
 where
     IV: IntoView + 'static,
-    S: Send + 'static,
+    S: Send + Sync + Clone + 'static,
     LeptosOptions: FromRef<S>,
 {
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! I'm fine with this, per your point that any `State` that is usable is going to satisfy these.

PR #3414, opened 2024-12-27
adding missing attribute (type) in audio tag

Phosphorus-M:
Someone in Discord found this problem:

imagen


gbj:
The `<audio>` element does not have a `type` attribute that I'm aware of (unless I'm misreading [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#attributes))

Is this perhaps being mixed up with [the `type` attribute on the `<source>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#audio_element_with_source_element)?

Issue #3412, opened 2024-12-27
Missing `mathvariant` attribute for `<mi>` tag in mathml


lnay:
There's a compile error with the following in a `view!` macro in leptos 0.7.2 (but is fine without the attribute):
```html
<mi mathvariant="normal">"Δ"</mi>
```
presumably due to there not being a "mathvariant" attribute here: https://github.com/leptos-rs/leptos/blob/main/tachys/src/mathml/mod.rs#L135.
Although this attribute is essentially useless at the moment and will be [removed soon](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mi), it is generated by [temml](https://temml.org/) which is the easiest way to create mathml from LaTeX, and the error message doesn't lead you to the fix:
```
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate
```

gbj:
Want to make a PR? It's a one-word change [here](https://github.com/leptos-rs/leptos/blob/1c30a4c13188007a40a4e6d58cd10e302b736fda/tachys/src/mathml/mod.rs#L135) to add `mathvariant` to the list of attributes supported on `mi`.

gbj:
Sorry, I misled you due to my lack of familiarity with MathML — `mathvariant` is already supported as [a global attribute](https://github.com/leptos-rs/leptos/blob/1c30a4c13188007a40a4e6d58cd10e302b736fda/tachys/src/mathml/mod.rs#L80).

The issue you are having is not that the method is not defined, but something else. 

A trivial example (see below) compiles correctly, but feel free to provide an example that reproduces the issue and I'm happy to look at it.

```rust
#[component]
pub fn App() -> impl IntoView {
    view! {
        <mi mathvariant="normal">"Δ"</mi>
    }
}
```

lnay:
Sorry yes, I was also overconfident about this being the issue after finding something like this a while ago.

PR #3415, opened 2024-12-27
fix: missing `mathvariant` attribute for `<mi>` mathml tag

lnay:
Small edit to address issue #3412


gbj:
See my comment on #3412 for why this is failing.

Issue #3378, opened 2024-12-17
[Feature]: Rethink Type Safety and Flexibility of the `view!` Macro for Conditional Logic


wiseaidev:
Hiya! 

I've been using Leptos for a while now (ever since building [this crate](https://github.com/GigaDAO/wasi-sol/blob/main/examples/leptos/src/main.rs)), and I've noticed that working with the `view!` macro can be quite cumbersome when adding conditional logic (`if-else`) inside the macro. Let me explain with some examples:

**The Problem**

To add `if-else` logic in `view!`, it always has to be wrapped in a closure (`{move ||}`). Here's a complex example:

```rust
<div class="wallet-info">
    {move ||
        if connected.get() {
            Some(view!{
                {move ||
                    if let Some(key) = phantom_wallet_adapter.get().public_key() {
                        view!{
                            <p>"Connected Wallet: " {move || phantom_wallet_adapter.get().name()} </p>
                            <p>"Connected Public Key: " {move || key.to_string() } </p>
                            <div class="forms">
                                <div class="send-sol-form">
                                    <h2 class="form-title">{ "Transfer SOL" }</h2>
                                    <form on:submit={transfer_sol_phantom}>
                                        <div class="form-group">
                                            <label for="destination-address">
                                                { "Destination Address" }
                                            </label>
                                            <input
                                                id="destination-address"
                                                type="text"
                                                class="form-control"
                                                node_ref={input_dest_ref}
                                                required=true
                                                value=dest
                                            />
                                        </div>
                                        <div class="form-group">
                                            <label for="sol-amount">
                                                { "SOL Amount (in lamports)" }
                                            </label>
                                            <input
                                                id="sol-amount"
                                                type="number"
                                                class="form-control"
                                                node_ref={input_amount_ref}
                                                required=true
                                                value=amount
                                            />
                                        </div>
                                        <button type="submit" class="submit-button">{ "Send" }</button>
                                    </form>
                                </div>
                                <div class="sign-form">
                                    <h2 class="form-title">{ "Sign Message" }</h2>
                                    <form on:submit={sign_msg_phantom}>
                                        <div class="form-group">
                                            <label for="message">
                                                { "Message" }
                                            </label>
                                            <input
                                                id="Message"
                                                type="text"
                                                class="form-control"
                                                node_ref={input_msg_ref}
                                                required=true
                                                value=msg
                                            />
                                        </div>
                                        <button type="submit" class="submit-button">{ "Sign" }</button>
[...]

PR #3417, opened 2024-12-28
feat: either_or combinator

geovie:
This is an idea that came out of [this](https://github.com/leptos-rs/leptos/issues/3378) feature request.

It adds an `EitherOr` trait providing a `either_or` method which can be used to "map" over `bool`, `Option` and `Result` types in a similar manner as e.g. [Option.map_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else).

With this there is yet another way to remove the need for `to_any` or to manually specify the Either::Left/Right:
e.g. 
```rust
if let Some(key) = phantom_wallet_adapter.get().public_key() {
  view!{ ... }.into_any()
} else {
  view!{ ... }.into_any()
}
```
could be then written as
```rust
phantom_wallet_adapter.get().public_key().either_or(|key| view!{ ... }, |_| view!{ ... });
```

### DIFF
--- a/either_of/src/lib.rs
+++ b/either_of/src/lib.rs
@@ -32,6 +32,86 @@ where
     }
 }
 
+pub trait EitherOr {
+    type Left;
+    type Right;
+    fn either_or<FA, A, FB, B>(self, a: FA, b: FB) -> Either<A, B>
+    where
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! I think this is nice.

PR #3420, opened 2024-12-29
Tachy class: impl IntoClass for Cow<'_, str>

sgued:
This simply makes it easier to use custom classes. In my hobby project I have use a lot of `Cow<'static, str>` props/signals since my attribute values are mostly static and sometimes dynamic. Being able to use them directly would be a nice DX win.

I also thought about making the implementation more generic, for example over T: AsRef<str>, but that would change the associated types and be a breaking change it would also conflict with the `Option` impl due to the orphan rules.

Maybe there is something to be done with Borrow or Deref in a future release?

### DIFF
--- a/tachys/src/html/class.rs
+++ b/tachys/src/html/class.rs
@@ -3,7 +3,7 @@ use crate::{
     renderer::Rndr,
     view::{Position, ToTemplate},
 };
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

PR #3422, opened 2024-12-30
Add missing `#[track_caller]`s

mscofield0:
There were missing `#[track_caller]` on a number of trait method definitions and impls. All I could find and that was relevant to my case were added.


gbj:
I'd like to keep unrelated changes in separate PRs -- Could you please remove the latest commit adding Dispose to Store and open it as a separate PR? That way I can squash the other commits in this one into a general "add missing `#[track_caller]`" commit.

PR #3436, opened 2025-01-03
fix failure on try_new() methods which should not fail

kstep:
I stumbled upon a bug while working on https://github.com/Synphonyte/leptos-struct-table.
Specifically these lines: https://github.com/Synphonyte/leptos-struct-table/blob/leptos-0.7/src/components/table_content.rs#L276-L279

```rust
            if let Some(mut rows) = rows.try_write_value() {
                rows.set_sorting(&sorting);
                clear(false);
            };
```

The `rows` is `StoredValue`.

The problem is `try_write_value()` works down to `UntrackedWriteGuard::try_new()`, which calls `ArcRwLockWriteGuardian::take`, which panics if it's impossible to get a write lock. But `try_write_value()` mustn't panic, it must return `None` instead.

Hence the fix.

### DIFF
--- a/reactive_graph/src/signal/guards.rs
+++ b/reactive_graph/src/signal/guards.rs
@@ -104,7 +104,7 @@ impl<T: 'static> Debug for Plain<T> {
 impl<T: 'static> Plain<T> {
     /// Takes a reference-counted read guard on the given lock.
     pub fn try_new(inner: Arc<RwLock<T>>) -> Option<Self> {
-        ArcRwLockReadGuardian::take(inner)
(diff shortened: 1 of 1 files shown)


gbj:
This sounds like a correct change, to me. Thanks!

For the sake of clarity in the descriptive text above -- `::take()` only panics in that case on single-threaded architectures like WASM, not in general. (In general it will block, which is why it panics if there is no thread to put to sleep, behaving like a `RefCell` instead.)

kstep:
That's right. The problem happens on WASM target.

Issue #3437, opened 2025-01-03
Context is not available in timeout


tversteeg:
**Describe the bug**
When a timeout, debounce or other timing related item is called, the contexts or routing state provided earlier are not available anymore.

**Leptos Dependencies**

```toml
leptos = { version = "0.7", features = ["csr"] }
```

**To Reproduce**

```rust
use leptos::prelude::*;
use std::time::Duration;

#[component]
pub fn Demo() -> impl IntoView {
    provide_context(10_i32);

    let (value, set_value) = signal(0_i32);

    set_timeout_with_handle(
        move || {
            set_value.set(expect_context::<i32>());
        },
        Duration::from_millis(1000),
    )
    .expect("could not create timeout");

    view! {
        <div>
            {move || value.get()}
        </div>
    }
}
```

**Expected behavior**

After one second, `value` should be set to `10`.

**Additional context**

This error is thrown in the browser:

```
panicked at ../leptos/reactive_graph/src/owner/context.rs:305:9:
Location { file: "src/lib.rs", line: 12, col: 27 } expected context of type "i32" to be present
```

gbj:
That is true, and working as designed. 

If you want to access context outside the ownership tree (like in a timeout), you need to reestablish what the owner is at that moment. 

Here's how to apply this idea to your example:
```rust
#[component]
pub fn App() -> impl IntoView {
    provide_context(10_i32);

    let (value, set_value) = signal(0_i32);
    let owner = Owner::current().unwrap();

    set_timeout_with_handle(
        move || {
            owner.with(|| {
                set_value.set(expect_context::<i32>());
            });
        },
        Duration::from_millis(1000),
    )
    .expect("could not create timeout");

    view! {
        <div>
            {move || value.get()}
        </div>
    }
}
```

tversteeg:
Thanks for the clarification, this was a bit of a foot gun for me since I didn't know about the ownership tree and the documentation of those functions didn't mention anything about it. So I added #3442.

Another suggestion: wouldn't it make sense to always reestablish the owner before calling the callback inside the function implementations themselves? That way with "normal" use of the API people don't have to consider the ownership tree..

PR #3442, opened 2025-01-04
docs: warn about callbacks outside the ownership tree

tversteeg:
Make it clear what is called outside the ownership tree to prevent other people from having the same issue I had with #3437.


gbj:
This is a good idea, thanks. I've modified the text to make it a bit more explicit *how* to do what you're describing, and modified the formatting because the `<div class="warning">` approach doesn't work with the usual link generation for docs.

Issue #3443, opened 2025-01-04
Axum released version 0.8


johansmitsnl:
Axum released a new version 0.8, bump/update the integration to be compatible with this new version.

gbj:
See comments on #3439 and [here](https://github.com/leptos-rs/leptos/pull/3274#issuecomment-2508672840); this will require a version bump from `leptos_axum`, so can be rolled into the planned 0.8 toward the end of this month. (And of course, users eager to upgrade to Axum 0.8 can use the PR branch as a dependency in the interim.)

gbj:
Merged 0.8 to main; these changes are included in 0.8.0-alpha and -beta.

PR #3439, opened 2025-01-03
chore: upgrade `axum` to `v0.8`

sabify:

PR #3444, opened 2025-01-04
Update axum depend to 0.8.1

johansmitsnl:
Solves #3443

No breaking changes noticed, but I'm not using the Axum integration yet. Do verify this PR.


gbj:
Note that this is, almost by definition, a semver-breaking change — and in fact Axum makes breaking changes within its routing in particular that would cause all `.leptos_routes()` to fail if we simply updated to 0.8 without further changes.

This is a duplicate of #3439, which does update the routing as well, so I'm going to close this one in favor of that one.

PR #3448, opened 2025-01-05
Rkyv feature should propagate properly

thestarmaker:
I suppose enabling `rkyv` feature on `leptos` crate should enable it on all dependencies. 

Otherwise `Resource::new_rkyv()` is not accessible without extra dependency juggling i.e. explicitly bringign in `leptos_server` with `rkyv` feature enabled.


benwis:
LGTM. Thanks!

PR #3449, opened 2025-01-06
Add support for custom patch

mscofield0:
Adds a new `#[patch(|this, new| ...)]` derive helper for `Patch` that allows specifying an arbitrary patch.

### DIFF
--- a/reactive_stores_macro/src/lib.rs
+++ b/reactive_stores_macro/src/lib.rs
@@ -541,8 +541,13 @@ struct PatchModel {
 }
 
 enum PatchModelTy {
-    Struct { fields: Vec<Field> },
-    Enum { variants: Vec<Variant> },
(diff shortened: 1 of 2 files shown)


reactive_stores_macro/src/lib.rs
```diff
@@ -537,61 +537,129 @@ fn variant_to_tokens(
 struct PatchModel {
     pub name: Ident,
     pub generics: Generics,
-    pub fields: Vec<Field>,
+    pub ty: PatchModelTy,
+}
+
```
gbj, line 542: This just wants an `#[allow(dead_code)]` for the `Enum` branch.

mscofield0: Added

gbj:
Thanks!

Issue #3451, opened 2025-01-06
Add a scoped lifetime owner+signal utility


mscofield0:
**Is your feature request related to a problem? Please describe.**
Spamming either `Arc*` clones everywhere or doing:
```rust
let owner = Owner::new();
owner.set();
let signal = <your-signal>::from(<your-arc-signal>);
...
owner.unset();
```
isn't ideal.

**Describe the solution you'd like**
I would like a `Scoped<T>`, where `T` is a signal, utility that does the above code example automatically for us.

gbj:
Could you please say some more about what this would be used for, or what problem it's trying to solve?

It is not clear to me why it should be added to the library. Are there places you think it would simplify existing examples, or other internal code?

mscofield0:
As a utility for the end user, it's not too long and I don't think it's hard to maintain so I would rather put it in the core crate than have my own version somewhere. (+ I couldn't implement `From<ArcRwSignal<T>> for Scoped<RwSignal<T>>` for example)

gbj:
That doesn't answer either of my questions, unfortunately. Let me know if I can clarify what I mean by asking "What would this be used for?"

mscofield0:
Consider:
```rust
play::Event::Joined { player_id, game_id, seat } => {
	let games = state.casual_games().read_untracked();
	let Some(game) = games.get(&game_id).cloned() else {
		log::error!("game '{game_id}' does not exist");
		return;
	};
	let owner = Owner::new();
	owner.set();
	let game = Store::<play::models::CasualGameInfo>::from(game);
	game.clone().info().table().write().insert(seat, player_id);
	// add to tracking game info if you are the one who joined
	if let Some(session) = state.session().get_untracked() {
		if session.user().read_untracked().user_id == player_id {
			let game_info = game.info().get_untracked();
			state.game_info().set(Some(ArcStore::new(game_info)));
		}
	}
	log::info!("user '{player_id}' joined seat {seat} of game '{game_id}'");
},
```
as opposed to:
```rust
play::Event::Joined { player_id, game_id, seat } => {
	let games = state.casual_games().read_untracked();
	let Some(game) = games.get(&game_id).cloned() else {
		log::error!("game '{game_id}' does not exist");
		return;
	};
	let game = Scoped::from(game);
	game.clone().info().table().write().insert(seat, player_id);
	// add to tracking game info if you are the one who joined
	if let Some(session) = state.session().get_untracked() {
[...]

PR #3452, opened 2025-01-06
add Scoped utility for signals

mscofield0:
Solves #3451


mscofield0:
Actually sort of not useful if `Scoped` isn't `Copy` somehow. Open to suggestions, I was thinking of a crude hack using `ArenaItem<Owner>` which may or may not make sense...

Issue #3156, opened 2024-10-24
Inconsistencies with `--cfg=erase_components`


sabify:
**Describe the bug**
By using `--cfg=erase_components` in `RUSTFLAGS`, leptos ignores `attr:` in components and nothing will be applied.

**Leptos Dependencies**
`leptos = "0.7.0-rc0"`

1. use `RUSTFLAGS="--cfg=erase_components" cargo leptos serve`
2. Add an attribute to a component: `<ActionForm attr:class="xyz">`
3. On serve, no class attribute is added

**Expected behavior**
The class attribute should be added to the component.

gbj:
This is unfortunate but known. After much work I was not able to support adding arbitrary attributes to `AnyView` without causing difficult-to-reproduce linker crashes on larger projects. You can read the length discussion beginning [here](https://github.com/leptos-rs/leptos/pull/2905#issuecomment-2395464765).

You can add `#[component(transparent)]` to components that need to support attribute spreading and that should work.

zakstucke:
@gbj maybe for now we should enable the dodgy impl when `--cfg=erase_components` is turned off? At the moment this would affect release builds when `into_any()` is legitimately used.

gbj:
@zakstucke To the best of my knowledge, there has never been a non-`todo!()` implementation of `AddAnyAttr` for `AnyView` that did not lead to linker errors. If anyone has a fix for that, I will be very excited!

gbj:
This has been fixed, closing now while going through old issues.

PR #3458, opened 2025-01-08
another try at adding AddAnyAttr to AnyView

gbj:
@luxalpa had the very reasonable suggestion that perhaps the `#[inline(always)]` annotation that I've removed [here](https://github.com/leptos-rs/leptos/pull/3458/files#diff-dfd7b4995a3df23a931fe30b3c3574799e21fea8ae7da1e7b76115676ca80e2bL133) was the problem with the `AnyView::add_any_attr()` implementation all along. This makes sense to me intuitively as I could see how it would lead to a mutually-recursive inlined function that would be... very big.

I don't have a good test case and can no longer compile the test case that you shared with me @zakstucke, so I'm wondering if some others could try it out to see if this fixes the infinite compile time issues with that change. 

See #3156 and discussion beginning [here](https://github.com/leptos-rs/leptos/pull/2905#issuecomment-2395464765) for context.

### DIFF
--- a/tachys/src/view/any_view.rs
+++ b/tachys/src/view/any_view.rs
@@ -5,7 +5,12 @@ use super::{
     RenderHtml,
 };
 use crate::{
-    html::attribute::Attribute, hydration::Cursor, ssr::StreamBuilder,
+    html::attribute::{
+        any_attribute::{AnyAttribute, IntoAnyAttribute},
+        Attribute,
+    },
+    hydration::Cursor,
+    ssr::StreamBuilder,
 };
 use std::{
     any::{Any, TypeId},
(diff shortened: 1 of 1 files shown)


zakstucke:
Not looking good :(

After 10 mins of compiling:

`cannot encode offset of relocations; object file too large`

This error is raised from this compiler code:
https://llvm.org/doxygen/MachObjectWriter_8cpp_source.html

geoffreygarrett:
Would it be agreeable to feature gate this under `any-view-any-attr` for now, while we work on getting it fully functional? I haven't encountered any issues on my end.

zakstucke:
See referenced PR, different implementation but same problem (I was hoping traits and hoisting the downcast out might fix it). It might narrow down the issue.

geoffreygarrett:
@zakstucke, would you mind giving #3461 a spin on your failing case to see if it helps?

zakstucke:
No change unfortunately - attempt in https://github.com/leptos-rs/leptos/pull/3460 seems to show the issue being `add_any_attr(attr)` rather than nested `into_any()`

geoffreygarrett:
Leaving this here for traceability https://github.com/leptos-rs/leptos/pull/3461/commits/7fc0c416485ea81431b379e8f9366f5d7a445ba8. Let me what effect it has, if any.

gbj:
I'll close this in favor of #3460 or #3461.

PR #3461, opened 2025-01-08
AddAnyAttr experimentation on AnyView

geoffreygarrett:
Another attempt: Add from_box constructor to avoid recursive type expansion when adding attributes to type-erased views.

- #3458

### DIFF
--- a/tachys/src/view/any_view.rs
+++ b/tachys/src/view/any_view.rs
@@ -19,10 +19,25 @@ use std::{
 #[cfg(feature = "ssr")]
 use std::{future::Future, pin::Pin};
 
+// Minimizes deep generics via a single dynamic dispatch
+trait DynValueAttr: DynValue {
+    fn apply_attr(self: Box<Self>, attr: AnyAttribute) -> AnyView;
+}
+
 trait DynValue: Send {
     fn dyn_add_any_attr(self: Box<Self>, attr: AnyAttribute) -> AnyView;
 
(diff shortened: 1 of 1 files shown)


zakstucke:
Same issue :( 

10 minutes or so of compiling ends in:
`cannot encode offset of relocations; object file too large`

zakstucke:
Still no change unfortunately but was a good effort! I think until we get some feedback from the rust team on https://github.com/rust-lang/rust/issues/130729 I'm out of ideas.

That's for the compiler errors arising on leptos stable, but I'm sure this issue will have a similar root cause.

benwis:
@zakstucke can you create a new issue specifically about the AddAnyAttr implementation in that repo with a simple reproduction. The old one got kinda broad

benwis:
> Still no change unfortunately but was a good effort! I think until we get some feedback from the rust team on [rust-lang/rust#130729](https://github.com/rust-lang/rust/issues/130729) I'm out of ideas.
> 
> That's for the compiler errors arising on leptos stable, but I'm sure this issue will have a similar root cause.

@zakstucke Can you create a new rust lang issue about this specific AddAnyAttr implementation, include a reproduction, and then I'll see if I can poke some people I know on the Rust team

zakstucke:
@benwis thaw can repro, see my fork that sets the repro up (see README):
https://github.com/zakstucke/thaw

gbj:
We've achieved this with #3518 and friends.

Issue #3463, opened 2025-01-09
Implement AddAnyAttr trait for Static<V>


geoffreygarrett:
### Is your feature request related to a problem? Please describe.
Yes, the `AddAnyAttr` trait is not implemented for `Static<V>`, resulting in compiler errors when attempting to add attributes to static strings.

```rust
#[component]
fn A(children: Children) -> impl IntoView {
    view! { {children()} }
}
#[component]
fn B() -> impl IntoView {
    view! {
        <A attr:class="x">
            "Test"
        </A>
    }
}
```

### Describe the solution you'd like
Implement the `AddAnyAttr` trait for `Static<V>` so that adding an attribute returns the static string unchanged, ensuring type safety and compatibility.

### Additional context
It's a small edge case that doesn't really *need* action, but for the sake of polish, might be nice.

PR #3464, opened 2025-01-09
Implement AddAnyAttr trait for Static<V>

geoffreygarrett:
Closes: #3463


gbj:
Thanks, this is an interesting approach that I can probably apply to the other string types as well. It has been very odd to me that the compiler couldn't figure this out -- thanks for fixing it.

No action needed, the CI failures are due to a change in `cargo check` behavior that I need to fix elsewhere.

PR #3469, opened 2025-01-09
Add support for effects with custom runtimes

### LINKED ISSUE #3468: Provide executor to Effect at creation
Currently Effects in `reactive_graph` use `any_spawner` to provide a runtime. This is not suitable for environments with no globally set runtime, such as when trying to integrate `reactive_graph` with `bevy` through `bevy-tokio-tasks`.

Versions of the `Effect::new` and `Effect::new_sync` API that allow for a callback to which you can provide custom spawn logic, for example

```rust
Effect::new_sync_with_runtime(
    move || {
        /* do something with effect */
    },
    |task| runtime.spawn(task),
);
```

I am not aware of an alternative way to achieve this result at present, as custom executors in `any_spawner` are required to be `'static`, and this would enable using a runtime bound on a local lifetime (such as one provided by a bevy resource) for the execution context.

**Additional context**
I implemented a hacky version by forking and rewriting the `new_sync` function to accept a reference to a Tokio runtime directly, and it fixed the problems I was having. The callback API would be a more flexible version of this.

```rust
pub fn new_sync_2<T, M>(
    mut fun: impl EffectFunction<T, M> + Send + Sync + 'static,
    rt: &Runtime,
) -> Self
where
    T: Send + Sync + 'static,
{
    let inner = cfg!(feature = "effects").then(|| {
        let (mut rx, owner, inner) = effect_base();
        let mut first_run = true;
        let value = Arc::new(RwLock::new(None::<T>));

        rt.spawn({
            let value = Arc::clone(&value);
            let subscriber = inner.to_any_subscriber();

            async move {
                while rx.next().await.is_some() {
                    if subscriber
                        .with_observer(|| subscriber.update_if_necessary())
                        || first_run
                    {
                        first_run = false;
                        subscriber.clear_sources(&subscriber);

                        let old_value =
                            mem::take(&mut *value.write().or_poisoned());
                        let new_value = owner.with_cleanup(|| {
                            subscriber.with_observer(|| {
                                run_in_effect_scope(|| fun.run(old_value))
                            })
                        });
                        *value.write().or_poisoned() = Some(new_value);
                    }
                }
            }
        });

        ArenaItem::new_with_storage(Some(inner))
    });

    Self { inner }
}
```

LLBlumire:
Fixes https://github.com/leptos-rs/leptos/issues/3468

This is almost certainly not the best way to do it, but I couldn't figure out the generic types so did a pinned box instead, but it demonstrates the functionality that is desired.

### DIFF
--- a/reactive_graph/src/effect/effect.rs
+++ b/reactive_graph/src/effect/effect.rs
@@ -12,7 +12,9 @@ use any_spawner::Executor;
 use futures::StreamExt;
 use or_poisoned::OrPoisoned;
 use std::{
+    future::Future,
     mem,
+    pin::Pin,
     sync::{atomic::AtomicBool, Arc, RwLock},
 };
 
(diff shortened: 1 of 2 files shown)


reactive_graph/src/effect/effect.rs
```diff
@@ -196,6 +198,57 @@ impl Effect<LocalStorage> {
         Self { inner }
     }
 
+    /// Creates a new effect, which runs once on the next “tick”, and then runs again when reactive values
+    /// that are read inside it change.
+    ///
+    /// This spawns a task on the local thread using
+    /// [`spawn_local`](any_spawner::Executor::spawn_local). For an effect that can be spawned on
+    /// any thread, use [`new_sync`](Effect::new_sync).
```
gbj, line 206: Docs should probably be updated to
1) Note that this allows you to pass in a runtime, and 
2) Update the `new_sync` x 2 here to `new_sync_with_runtime`

reactive_graph/src/effect/effect.rs
```diff
@@ -196,6 +198,57 @@ impl Effect<LocalStorage> {
...
+    pub fn new_with_runtime<T, M>(
+        mut fun: impl EffectFunction<T, M> + 'static,
+        pass_to_rt: impl FnOnce(Pin<Box<dyn Future<Output = ()> + 'static>>),
```
gbj, line 209: I'd suggest a name like `spawn` here rather than `pass_to_rt` as that's the phrase we tend to use for "spawn a task" elsewhere

I think the signature can be something like
```
    pub fn new_with_runtime<T, M, F, Fut>(
        mut fun: impl EffectFunction<T, M> + 'static,
        pass_to_rt: F
    ) -> Self 
    where 
        T: 'static,
        F: FnOnce(Fut) + 'static,
        Fut: Future<Output = ()>
```
if you want to avoid the `Pin<Box<_>>` here

gbj (commented):
This looks good to me, overall! Couple small notes on docs, and a suggestion if you do want to use generics rather requiring the function to accept a pinned-box -- although you should probably make that decision based on how you're going to consume it! (i.e., if the current signature is easier to *use* then go with this one)

PR #3478, opened 2025-01-11
feat (`either_of`): Extent API; Implement other iterator methods; Update deps

DanikVitek:


DanikVitek:
This example fails because the nightly version it depends on is too old. The method was stabilized in Rust v1.82. What is the MSRV for Leptos?

DanikVitek:
> What is the MSRV for Leptos?

Ok, I've found the info that it's Rust 1.76, but only from the CI bot. This should be mentioned in the README.md

benwis:
@DanikVitek That's quite the PR! You've really expanded the either_of API

PR #3496, opened 2025-01-17
Add node_ref back to Form

paul-hansen:
Adds a node_ref field to `<Form/>` elements which was there in [0.6](https://docs.rs/leptos_router/0.6.15/leptos_router/fn.Form.html) but removed in [0.7](https://docs.rs/leptos_router/0.7.4/leptos_router/components/fn.Form.html). It was still there on ActionForm, so I mainly copied the implementation from there.

### DIFF
--- a/router/src/form.rs
+++ b/router/src/form.rs
@@ -4,6 +4,7 @@ use crate::{
     location::{BrowserUrl, LocationProvider},
     NavigateOptions,
 };
+use either_of::Either;
 use leptos::{ev, html::form, logging::*, prelude::*, task::spawn_local};
 use std::{error::Error, sync::Arc};
 use wasm_bindgen::{JsCast, UnwrapThrowExt};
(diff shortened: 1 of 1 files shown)


paul-hansen:
CI is correct that this is technically a semver breaking change since a field is added to a public struct. I wasn't sure how you wanted to handle that so I left any version changes out of this PR for now. Allow edits by maintainers is checked or just lmk if you want any changes.

gbj:
The intention over time is to handle pass-through attributes like this using attribute spreading, rather than props, so that the cost is only paid when you use them, not for all users (in terms of binary size in particular)
```rust
#[component]
    let node_ref = NodeRef::new();

    view! {
        <Form action="/foo" {..} node_ref=node_ref>"todo"</Form>
    }
}
```
That said, this example is currently broken, so needs to be special-cased in the view macro I guess.

paul-hansen:
Makes sense, thanks for looking!

gbj:
Fixed in ac3352724bec517e02b4a330c10c3c347511023a, so the `<Form /* ... */ {..} node_ref=node_ref/>` version should now work.

paul-hansen:
Thanks so much! Making changes to the view macro was a bit daunting to me still so I didn't want to make any promises that I'd get to that.

Got my 30k+ line app to compile on 0.7 last night! Still some problems to fix and lots of testing to do, but getting so close!
If you're curious, it's https://startthegamealready.com/ (still on 0.6) not a lot to look at on the public side of it yet, plan to share it around more once we have better explainations of the features on the main pages and a few more rough edges smoothed out. Making it with my brother who's handling the hardware side.

Thanks for making Leptos, it made Rust a viable option for this and has made it infinitely more fun for me.

Issue #3467, opened 2025-01-09
Rkyv is broken on Leptos 7


thestarmaker:
leptos 0.7 depends on rkyv 0.8
leptos 0.7 depends on codee 0.2; 
codee 0.2 (and master) depends on rkyv 0.7

rkyv 0.7 is not compatible with rkyv 0.8

an attempt to use rkyv serialisation in yelds in 

```
the trait `rkyv::Serialize<rkyv::ser::serializers::CompositeSerializer<rkyv::ser::serializers::alloc::AlignedSerializer<rkyv::util::aligned_vec::AlignedVec>, rkyv::ser::serializers::core::FallbackScratch<rkyv::ser::serializers::alloc::HeapScratch<1024>, rkyv::ser::serializers::alloc::AllocScratch>, rkyv::ser::serializers::alloc::SharedSerializeMap>>` is not implemented for `base::api::monkey::Monkey`, which is required by `codee::binary::rkyv::RkyvCodec: codee::traits::Encoder<_>`
     |
help: trait impl with same name found
    --> /Users/<redacted>.rs:6:32
     |
6    | #[derive(Archive, Deserialize, Serialize, Eq, PartialEq, Clone)]
     |                                ^^^^^^^^^
     = note: perhaps two different versions of crate `rkyv` are being used?
     = help: the following other types implement trait `rkyv::Serialize<S>`:
               `()` implements `rkyv::Serialize<S>`
               `Arc<T>` implements `rkyv::Serialize<S>`
               `AtomicBool` implements `rkyv::Serialize<S>`
               `AtomicI16` implements `rkyv::Serialize<S>`
               `AtomicI32` implements `rkyv::Serialize<S>`
               `AtomicI8` implements `rkyv::Serialize<S>`
               `AtomicIsize` implements `rkyv::Serialize<S>`
               `AtomicU16` implements `rkyv::Serialize<S>`
             and 114 others
     = note: required for `codee::binary::rkyv::RkyvCodec` to implement `codee::traits::Encoder<base::api::monkey::Monkey>`
```

thestarmaker:
Attempting to upgrade `codee` dependencies: https://github.com/Synphonyte/codee/pull/3

maccesch:
Merged and released as codee `0.3.0`

gbj:
[...]

PR #3504, opened 2025-01-18
issue-3467 - bumping codee version to support rkyv 8

thestarmaker:
This change happens on the back of https://github.com/leptos-rs/leptos/issues/3467

I've tried this locally, rkyv serialisation seems to work now. Leptos codebase compiles, but re-exports may break something else I am not aware of.


benwis:
Appreciate you testing and PRing this. rkyv is an interesting encoding for sure

PR #3506, opened 2025-01-20
update workspace dependency versions

gbj:


benwis:
Seems like a reasonable merge to me

PR #3508, opened 2025-01-21
Enhanced docs for reactive_stores

dcsturman:
After building with Stores I found there we places the docs could be enhanced.  I took a stab at documenting the things that tripped me up. I also included some docs for the new support for Box.  Added docs on shadow traits, Option, Enum, Vec, and Box usage with Store.

I'm not sure if you have a particular way you want docs done, or even want this sort of detail.  I also realize you might want the docs in the Leptos Book - but I figured I'd start here and then get guidance from the maintainers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2764,6 +2764,7 @@ dependencies = [
  "any_spawner",
  "guardian",
  "itertools",
+ "leptos",
  "or_poisoned",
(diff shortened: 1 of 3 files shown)


benwis:
Always nice to see people add and refine documentation. Thanks @dcsturman!

PR #3517, opened 2025-01-25
impl IntoAttributeValue for TextProp

SleeplessOne1917:
I noticed that I can't pass `TextProp`s directly to HTML elements as attributes. This seems like something it should be able to do, so I made `TextProp` impl `IntoAttributeValue`.


gbj:
Awesome! Thanks.

PR #3524, opened 2025-01-29
fix: correctly notify descendants and ancestors of store fields (closes #3523)

### LINKED ISSUE #3523: Store fields don't notify their children properly
**Describe the bug**
Notifying in stores only seems to work on the first layer.

Given this test:
```rust
    #[tokio::test]
    async fn test_example() {
        use reactive_graph::traits::*;
        _ = any_spawner::Executor::init_tokio();

        #[derive(Debug, Clone, Store, Patch, Default)]
        struct Foo {
            id: i32,
            bar: Bar,
        }

        #[derive(Debug, Clone, Store, Patch, Default)]
        struct Bar {
            bar_signature: i32,
            baz: Baz,
        }

        #[derive(Debug, Clone, Store, Patch, Default)]
        struct Baz {
            more_data: i32,
            baw: Baw,
        }

        #[derive(Debug, Clone, Store, Patch, Default)]
        struct Baw {
            more_data: i32,
            end: i32,
        }

        let store = Store::new(Foo {
            id: 42,
            bar: Bar {
                bar_signature: 69,
                baz: Baz {
                    more_data: 9999,
                    baw: Baw {
                        more_data: 22,
                        end: 1112,
                    },
                },
            },
        });

        Effect::new_sync(move |_| {
            println!("foo: {:?}", store.get());
        });

        Effect::new_sync(move |_| {
            println!("foo.id: {:?}", store.id().get());
        });

        Effect::new_sync(move |_| {
            println!("foo.bar: {:?}", store.bar().get());
        });

        Effect::new_sync(move |_| {
            println!(
                "foo.bar.bar_signature: {:?}",
                store.bar().bar_signature().get()
            );
        });

        Effect::new_sync(move |_| {
            println!("foo.bar.baz: {:?}", store.bar().baz().get());
        });

        Effect::new_sync(move |_| {
            println!(
                "foo.bar.baz.more_data: {:?}",
                store.bar().baz().more_data().get()
            );
        });

        Effect::new_sync(move |_| {
            println!("foo.bar.baz.baw: {:?}", store.bar().baz().baw().get());
        });

        Effect::new_sync(move |_| {
            println!(
                "foo.bar.baz.baw.more_data: {:?}",
                store.bar().baz().baw().more_data().get()
            );
        });

        Effect::new_sync(move |_| {
            println!(
                "foo.bar.baz.baw.end: {:?}",
                store.bar().baz().baw().end().get()
            );
        });

        tick().await;
        store.set(Default::default());
        tick().await;
        store.bar().baz().set(Default::default());
        println!("\n==================");
        tick().await;
    }
```
One would assume that `store.set()` notifies all its children (up to even `store.bar.baz.baw.end`) but it only notifies its immediate children.

One would also assume that `store.bar().baz().set(Default::default())` would notify its children also, but if you look in the console logs, only its parent chain gets notified.

**Leptos Dependencies**
0.7

**To Reproduce**
The test from above I had in `reactive_stores/src/lib.rs` test section. Just run it and you will see what I mean.

**Expected behavior**
One would assume that `store.set()` notifies all its children (up to even `store.bar.baz.baw.end`) but it only notifies its immediate children.

One would also assume that `store.bar().baz().set(Default::default())` would notify its children also, but if you look in the console logs, only its parent chain gets notified.

gbj:


mscofield0:
lgtm

Issue #3521, opened 2025-01-27
feature request: allow raw identifiers in `Params` derive macro


PvlD:
like serde rename probably is better.

![Image](https://github.com/user-attachments/assets/09f635aa-7166-4264-ace3-1a80a1fe4a06)

PR #3525, opened 2025-01-29
feat: allow raw identifiers in Params derive macro

geovie:
Handles raw identifier e.g. `r#type` in Params macro

Fixes https://github.com/leptos-rs/leptos/issues/3521

### DIFF
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -34,6 +34,7 @@ log = "0.4.22"
 typed-builder = "0.20.0"
 trybuild = "1.0"
 leptos = { path = "../leptos" }
+leptos_router = { path = "../router", features= ["ssr"] }
 server_fn = { path = "../server_fn", features = ["cbor"] }
 insta = "1.41"
 serde = "1.0"
--- a/leptos_macro/src/params.rs
+++ b/leptos_macro/src/params.rs
@@ -13,7 +13,13 @@ pub fn params_impl(ast: &syn::DeriveInput) -> proc_macro::TokenStream {
             .named
             .iter()
             .map(|field| {
-				let field_name_string = &field.ident.as_ref().expect("expected named struct fields").to_string();
+				let field_name_string = &field
+                    .ident
+                    .as_ref()
+                    .expect("expected named struct fields")
+                    .to_string()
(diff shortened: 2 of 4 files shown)


leptos_macro/Cargo.toml
```diff
@@ -34,6 +34,7 @@ log = "0.4.22"
 typed-builder = "0.20.0"
 trybuild = "1.0"
 leptos = { path = "../leptos" }
+leptos_router = { path = "../router", features= ["ssr"] }
```
geovie, line 37: I had to add leptos_router to the dev deps to add test for Params

leptos_macro/src/params.rs
```diff
@@ -13,7 +13,13 @@ pub fn params_impl(ast: &syn::DeriveInput) -> proc_macro::TokenStream {
             .named
             .iter()
             .map(|field| {
-				let field_name_string = &field.ident.as_ref().expect("expected named struct fields").to_string();
+				let field_name_string = &field
+                    .ident
+                    .as_ref()
+                    .expect("expected named struct fields")
+                    .to_string()
+                    .trim_start_matches("r#")
```
geovie, line 21: This is also how serde does it: https://github.com/serde-rs/serde/blob/04ff3e8f950eb9862759781d85ec29d906917b70/serde_derive/src/internals/attr.rs#L134

geovie (commented):
And a bit off topic:
The `Params` looks a lot like `TryFrom<leptos_router::params::ParamsMap>`.
Is there any reason why it's its own trait?

e.g. it could be:
```rust
impl TryFrom<leptos_router::params::ParamsMap> for XYZ {
    type Error = leptos_router::params::ParamsError;

    fn try_from(value: leptos_router::params::ParamsMap) -> std::result::Result<Self, Self::Error> {
        todo!()
    }
}
```
and then there would also not be the confusion around the Params trait vs Params macro

gbj:
Looks good, thanks! 

On the trait: yes, I think you're right. I think this has been in there [since `0.0.1`](https://docs.rs/leptos_router/0.0.1/leptos_router/trait.Params.html), when I was simply less good at Rust so didn't realize there was already an obvious trait to use here :smile:

PR #3526, opened 2025-01-29
add file_and_error_handler_with_context

sstepanchuk:
like in leptos_routes_with_context func


gbj:
Looks good to me, thanks!

PR #3528, opened 2025-01-30
Implement PatchField for Option

iradicek:
Adding implementation of `PatchField` (in `reactive_stores`) for `Option<T>`.


iradicek:
Workflows failed previously. I think it should be fixed by rebasing the branch (since semver check failed for a type I did not touch), not sure though.

gbj:
Thanks!

PR #3529, opened 2025-01-31
Leptos 0.8

gbj:
This is meant to prepare for a `0.8.0-alpha` release.

This release is intended to include any semver-breaking changes needed to fix bugs or make small changes after 0.7. It also includes updates to make the error type returned from server functions much more flexible.

It is not intended to require any meaningful updates to user/application code, unlike 0.7.


chrisp60:
any chance we can still squeeze https://github.com/leptos-rs/leptos/pull/3522 into here?. It is breaking. Currently without conflict

gbj:
Yes!

PR #3534, opened 2025-01-31
Add transpose to the OptionStoreExt

mahdi739:
This PR adds a method called "transpose" to the OptionStoreExt which converts a subfield with an option into an option of a subfield. 
I found it useful in pattern matching while we want to keep the reactivity of the subfield.

### DIFF
--- a/reactive_stores/src/option.rs
+++ b/reactive_stores/src/option.rs
@@ -1,5 +1,5 @@
 use crate::{StoreField, Subfield};
-use reactive_graph::traits::{Read, ReadUntracked};
+use reactive_graph::traits::{FlattenOptionRefOption, Read, ReadUntracked};
 use std::ops::Deref;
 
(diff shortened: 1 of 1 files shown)


mahdi739:
Thanks. Actually I wasn't sure about the name so I sent it to get some input. And maybe with the presence of `map` it's kind of redundant. You can close it or change it or tell me if I should make any change.

mahdi739:
I made several changes based on your feedback:
- Implemented the new method as a default implementation in the trait to avoid breaking changes.
- Renamed `transpose` to `invert` to distinguish it from Rust's similar method.
- Replaced `read` with `try_read` in `map` and `map_untracked` to handle cases where the signal was disposed, now it returns `None` in such scenarios.

gbj:
Awesome. Looks great -- thank you!

PR #3539, opened 2025-02-01
Fixing closing brace

thestarmaker:
Most likely confglict merge artefact


benwis:
Yep that should be there

PR #3544, opened 2025-02-03
docs: add note to ElementExt::on about dropping the handle

lukasschlueter:
Dropping the `RemoveEventHandler` returned by [`ElementExt::on`](https://docs.rs/leptos/latest/leptos/prelude/trait.ElementExt.html#tymethod.on) also drops the closure, which invalidates the event listener.

Example code:
```rust
let button_ref: NodeRef<Button> = NodeRef::new();

button_ref.on_load(move |btn| {
    let cb = btn.on(leptos::ev::click, move |ev| {
        /* ... */
    });
    // Delay the drop to keep the event listener alive
    on_cleanup(move || std::mem::drop(cb));
});
```

### DIFF
--- a/tachys/src/html/element/element_ext.rs
+++ b/tachys/src/html/element/element_ext.rs
@@ -44,6 +44,14 @@ pub trait ElementExt {
         S: IntoStyle;
 
     /// Adds an event listener to the element, at runtime.
+    ///
+    /// Dropping the returned handle also drops the closure and invalidates the event listener.
+    /// Consider using [on_cleanup](reactive_graph::owner::Owner::on_cleanup) to delay dropping:
+    ///
+    /// ```rust,ignore
+    /// let remove = element.on(ev::blur, move |_| /* ... */);
+    /// on_cleanup(move || std::mem::drop(remove));
+    /// ```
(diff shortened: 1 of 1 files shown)


gbj:
Running `cargo doc` to test this out gives the following error:
```
warning: unresolved link to `reactive_graph::owner::Owner::on_cleanup`
  --> tachys/src/html/element/element_ext.rs:49:37
   |
49 |     /// Consider using [on_cleanup](reactive_graph::owner::Owner::on_cleanup) to delay dropping:
   |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `reactive_graph` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```
and a broken link.

This is because the `reactive_graph` crate is not *necessarily* included but is an opt-in feature. Perhaps some text similar to the `must_use` message, or maybe just mentioning "If you're using this with Leptos, you can use `on_cleanup`"? I never know exactly how to handle these situations where the crate is designed for more general purpose use but there are Leptos-specific docs that would be useful.

PR #3550, opened 2025-02-06
fix: Return empty iterator instead of panicking when the KeyedSubfield is disposed

mahdi739:


gbj:
Beautiful! Thank you!

PR #3551, opened 2025-02-06
Less panic in stores

mahdi739:


gbj:
Thank you for this one, as well!

PR #3518, opened 2025-01-26
AddAnyAttr working with erase_components

zakstucke:
cc #3156 #3458 #3460 #3461 #3476

TLDR: this PR enables `AddAnyAttr` on `AnyView` when the`--cfg erase_components` feature flag is enabled. Made possible by further type erasing tuples of attrs types to `Vec<AnyAttribute>` before they even became tuples preventing generation of an exponential number of types. `--cfg erase_components` is now feature complete and provides up to 1.5-3x faster compilation speed. Yay!

Unfortunately this means the issues surrounding `AddAnyAttr` and `AnyView` without type erasure are most likely not an issue with rust or the compiler, but too many unique attribute tuple types, monorphisation and codegen, but importantly this missing feature is actually what is making stable leptos compile, by trimming the potential for futher attr types at each `AnyView` boundary and preventing further tuple types of that type being created, and why the feature cannot be enabled without reducing the number of unique types.

Arguably the above **is** an issue with rust/the compiler, but on further analysis it looks like the solution isn't "fix a bug in rust/compiler" it's "how do we reduce the number of unique types without hurting binary size or runtime too much because the current number seems to be insane".

As I've mentioned, the compile bottleneck is only surrounding the static typing of attributes, not the entire view tree, which makes sense considering these are the "leaves" whereas html elements are the "branches". Maybe erasing attributes is a worthwhile cost, whilst leaving the tree itself static (i.e splitting `--cfg erase_components` into 2 parts, the attr part being erased.

The main takeaway is that compiler issues, compile time, the `AnyAttr` feature not working on `AnyView`, are all down to the tuple impls that create an exponential number of types.

Using [thaw](https://github.com/thaw-ui/thaw)'s demo project, pointing leptos to this PR, you can see the effect of erasing the components with [cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines): llvm lines drop by 35% from 11million to 7million, still huge but it get's it down to something that seems manageable.

Without erasure:
```
cd demo
CARGO_PROFILE_RELEASE_LTO=fat cargo +stable llvm-lines --bin demo --release --features ssr | head -50
```

```
  11037943                245438                (TOTAL)
    753224 (6.8%,  6.8%)   14798 (6.0%,  6.0%)  core::ops::function::FnOnce::call_once
    344453 (3.1%,  9.9%)    3661 (1.5%,  7.5%)  <T as tachys::view::any_view::IntoAny>::into_any::{{closure}}
[...]

PR #3553, opened 2025-02-06
AddAnyAttr for AnyView for non-erased

### LINKED ISSUE #3354: Reactive tracking warning not shown correctly
**Describe the bug**

Consider this small repository: <https://github.com/veigaribo/leptos-reactive-tracking-warning-issue>

If it's run as it is (commit ID 5038489), the client JS+WASM will display the following warning:

```
At /.../leptos/reactive_graph/src/wrappers.rs:477:35, you access a reactive_graph::signal::arc_read::ArcReadSignal<core::option::Option<alloc::string::String>> (defined at /.../leptos/reactive_graph/src/signal/rw.rs:180:40) outside a reactive tracking context. This might mean your app is not responding to changes in signal values in the way you expect.
```

Which is of course not very useful, as none of the mentioned code pertains to the repository in question.
However, in `src/app.rs`, if line 25 is commented and line 24 uncommented:

```diff
#[component]
pub fn App() -> impl IntoView {
-    // let (x, set_x) = signal("Test");
-    let (x, set_x) = use_cookie::<String, FromToStringCodec>("test");
+    let (x, set_x) = signal("Test");
+    // let (x, set_x) = use_cookie::<String, FromToStringCodec>("test");

    view! { <button>{x.get()}</button> }
}
```

Then the new code will not produce any warning at all &mdash; which I believe is wrong, because the signal access in line 27 is still not tracked. In fact, the signal access is exactly the same; only the signal definition is different.

**Leptos Dependencies**

```toml
leptos = { version = "0.7.0", features = ["nightly"] }
leptos_meta = { version = "0.7.0" }
leptos_actix = { version = "0.7.0", optional = true }
leptos_router = { version = "0.7.0" }
```

Though I did some testing with
```toml
[patch.crates-io]
leptos = { path = "../../../leptos/leptos" }
leptos_meta = { path = "../../../leptos/meta" }
leptos_actix = { path = "../../../leptos/integrations/actix" }
leptos_router = { path = "../../../leptos/router" }
```

On a repository checked out on tag v0.7.0 

**To Reproduce**

1. Clone the repository at <https://github.com/veigaribo/leptos-reactive-tracking-warning-issue/>
2. Run with e.g. `cargo leptos watch --hot-reload --bin-features=ssr_actix` (see README)
3. Access the application in your browser
4. Open the developer tools console and see the warning
5. Optionally do the mentioned modifications and see the absence of any warning

**Expected behavior**

I expected that the warning would be shown in both cases of the example code, and that the call and definition sites presented on the warning would not point to code internal to the project. 

**Additional context**

`use_cookie` is defined here: <https://github.com/Synphonyte/leptos-use/blob/18cdd812f292ba83443bf868957291aedf1d9585/src/use_cookie.rs#L145>. I kept it from my actual application code, where I found this issue, because I don't yet know what causes it to behave differently.

When I was trying to understand this behavior, I noticed that, if I changed [`ArcRwSignal::read_only`](https://github.com/leptos-rs/leptos/blob/e1e90b859526a8b0f082496415dfe54f81583301/reactive_graph/src/signal/arc_rw.rs#L169) to propagate `self.defined_at`, instead of creating a new Location with `Location::caller()`, the resulting warning does include the correct signal definition, which is in `leptos-use`.

```diff
    /// Returns a read-only handle to the signal.
    #[track_caller]
    pub fn read_only(&self) -> ArcReadSignal<T> {
        ArcReadSignal {
            #[cfg(debug_assertions)]
-           defined_at: Location::caller(),
+           defined_at: self.defined_at,
            value: Arc::clone(&self.value),
            inner: Arc::clone(&self.inner),
        }
    }
```

```
At /.../leptos/reactive_graph/src/wrappers.rs:477:35, you access a reactive_graph::signal::arc_read::ArcReadSignal<core::option::Option<alloc::string::String>> (defined at /.../leptos-use-0.14.0-rc5/src/use_cookie.rs:189:32) outside a reactive tracking context. This might mean your app is not responding to changes in signal values in the way you expect.
```

And also saw that the immediate reason why the patched example repository code does not produce a warning is because it is in the `SpecialNonReactiveZone`. That is all the understanding I gathered.

### LINKED ISSUE #3487: ErrorBoundary ignores error
**Describe the bug**
When getting a resource inside an error boundary the ErrorBoundary ignores the error should the resource return one, if the navigation to the page occurred from within the app.

**Leptos Dependencies**
```toml
leptos = { version = "0.7", features = ["nightly", "hydration"] }
leptos_meta = "0.7"
leptos_router = "0.7"
leptos_axum = { version = "0.7", optional = true }
```

**To Reproduce**
1. Create a project from the start-axum template
2. Copy the attached code into the app
3. (local) host the site and open it
4. Use the hyperlink to navigate to /test
5. Neither the ErrorBoundary fallback nor its children are shown

**Expected behavior**
The ErrorBoundary fallback is shown.

**Additional context**
When accessing the page directly (e.g. reloading the page / opening the app directly on the page) the fallback is correctly shown.
I'm also on 0.7.0 bc the cargo_leptos version in nixpkgs is 0.2.21, and leptos 0.7.1 requires wasm-bingen ^0.2.97 while cargo_leptos 0.2.21 uses 0.2.95. I have not seen any commits that address the issue of this bug report since the 0.7.0 release, so I assume it has not been fixed yet.

**app.rs code**
```rust
use leptos::prelude::*;
use leptos_meta::{provide_meta_context, MetaTags, Stylesheet, Title};
use leptos_router::{
    components::{Route, Router, Routes, A},
    StaticSegment,
};

    view! {
        <!DOCTYPE html>
        <html lang="en">
            <head>
                <meta charset="utf-8"/>
                <HydrationScripts options/>
                <MetaTags/>
            </head>
            <body>
                <App/>
            </body>
        </html>
    }
}

#[component]
    // Provides context that manages stylesheets, titles, meta tags, etc.
    provide_meta_context();

    view! {
        // injects a stylesheet into the document <head>
        // id=leptos means cargo-leptos will hot-reload this stylesheet
        <Stylesheet id="leptos" href="/pkg/bug.css"/>

        // sets the document title

        // content for this welcome page
        <Router>
            <main>
                    <Route path=StaticSegment("test") view=Test/>
                </Routes>
            </main>
        </Router>
    }
}

/// Renders the home page of your application.
#[component]
    view! {
        <A href="test">"test"</A>
    }
}

#[component]
fn Test() -> impl IntoView {
    let data = Resource::new(|| (), |_| get_test());

    view! {
        <Transition fallback=|| view! {"Loading..."}>
            <ErrorBoundary fallback=|_| {
                view! { <span class="error">"Error detected"</span> }
            }>
                {move || {
                    data.get()
                        .map(|data| {
                            data.map(|_| {
                                view! {
                                    <div>
                                        <h1>"No Errors"</h1>
                                    </div>
                                }
                            })
                        })
                }}
            </ErrorBoundary>
        </Transition>
    }
}

#[server]
async fn get_test() -> Result<(), ServerFnError> {
    Err(ServerFnError::new("An Error"))?
}

```

### LINKED ISSUE #3491: Enable Box to be used within structs using Store
Enable `Box` to be used properly within `Store`.  Example code of what I'd want to be able to do:
```rs
use leptos::prelude::*;
use reactive_stores::{Store, Field, OptionStoreExt};

#[derive(Debug, Store, Clone)]
struct BigUglyDS {
    msg: String,
    sub1: Option<Box<BigUglyDS>>
}

#[component]
fn ShowIt(#[prop(into)] big_ugly: Field<BigUglyDS>) -> impl IntoView {
    view! {
        <span>
            <b>{move || big_ugly.msg().get()}</b>
            {move || {
                if big_ugly.sub1().read().is_some() {
                    // This is the key change needed.
                    view! { <ShowIt big_ugly=big_ugly.sub1().unwrap().get() />
                } else {
                    view! { () }.into_any()
                }
            }}
        </span>
    }
}

#[component]
fn App() -> impl IntoView {
    let mut big_ugly = BigUglyDS { msg: "First".to_string(),sub1: None };
    big_ugly.sub1 = Some(Box::new(BigUglyDS { msg: "Second".to_string(), sub1: Some(Box::new(BigUglyDS { msg: "Third".to_string(), sub1: None })) }));

    let big_ugly = Store::new(big_ugly);
    
    view! { <ShowIt big_ugly=big_ugly /> }
}

fn main() {
    mount_to_body(App);
}
```

Just using `big_ugly.sub1().unwrap()` will not work to be passed to `ShowIt` as `ShowIt` is not expecting a `Box`.    If `Box` was understood by `Store` (and really `Field`) then it would "just work".  My workaround right now is as follows, which forces a clone of a large data structure. 
```rs
use leptos::prelude::*;
use reactive_stores::{Store, Field, OptionStoreExt};

#[derive(Debug, Store, Clone)]
struct BigUglyDS {
    msg: String,
    sub1: Option<Box<BigUglyDS>>
}

#[component]
fn ShowIt(#[prop(into)] big_ugly: Field<BigUglyDS>) -> impl IntoView {
    view! {
        <span>
            <b>{move || big_ugly.msg().get()}</b>
            {move || {
                if big_ugly.sub1().read().is_some() {
                    let big_ugly = *big_ugly.sub1().unwrap().get();
                    view! { <ShowIt big_ugly=Store::new(big_ugly) /> }.into_any()
                } else {
                    view! { () }.into_any()
                }
            }}
        </span>
    }
}

#[component]
fn App() -> impl IntoView {
    let mut big_ugly = BigUglyDS { msg: "First".to_string(),sub1: None };
    big_ugly.sub1 = Some(Box::new(BigUglyDS { msg: "Second".to_string(), sub1: Some(Box::new(BigUglyDS { msg: "Third".to_string(), sub1: None })) }));

    let big_ugly = Store::new(big_ugly);
    
    view! { <ShowIt big_ugly=big_ugly /> }
}

fn main() {
    mount_to_body(App);
}
```

Suggested alternative was to take a `Field<Box<Bigugly>>` and then just `Box` the original `Store` on creation.  It does not seem a component is okay with that as we get this error:
`
the method `msg` exists for struct `Field<Box<BigUglyDS>>`, but its trait bounds were not satisfied
items from traits can only be used if the trait is implemented and in scope
`

### LINKED ISSUE #3509: `Suspend` should accept `IntoFuture` instead of `Future`
Consider the following:

```rust
let res = Resource::new(..., ...);
// let s = Suspend:new(res); // Does not work because `res` is not `Future`
let s = Suspend::new(async move { res.await }); // Works because for await `IntoFuture` is sufficient
```

Is there any reason for `Suspend` to not accept `impl IntoFuture`?

zakstucke:
Reattempt of #3548 on 0.8 branch, made a hash of switching branches.

For AddAnyAttr in non-erased mode, the issue is simply the number/size of the types generated once something has already become an AnyView being too much for the compiler to handle.

This PR "truly erases" AnyView by preventing further types being created once something's been erased, utilising an "attribute bag" that will be passed down through the rendering traits, instead of producing a new type each time an attribute is added.

It's definitely the brute-force fix, but it seems to work!

Edit:
The into_owned() method being added to RenderHtml is somewhat separate, but is neccessary to make erased mode not cause new compile errors relating to AnyView not being static, it allows removing the static bound from IntoAny.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,7 +88,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -205,7 +205,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -230,7 +230,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -267,7 +267,7 @@ dependencies = [
  "async-executor",
  "futures",
  "glib",
- "thiserror 2.0.9",
+ "thiserror 2.0.11",
  "tokio",
  "tracing",
  "wasm-bindgen-futures",
@@ -275,9 +275,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.94"
+version = "1.0.95"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
+checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
 
 [[package]]
 name = "async-executor"
@@ -322,7 +322,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -333,13 +333,13 @@ checksum = "8b75356056920673b02621b35afd0f7dda9306d03c79a30f5c56c44cf256e3de"
 
 [[package]]
 name = "async-trait"
-version = "0.1.85"
+version = "0.1.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f934833b4b7233644e5848f235df3f57ed8c80f1528a26c3dfa13d2147fa056"
+checksum = "644dd749086bf3771a2fbc5f256fdb982d53f011c7d5d560304eafeecebce79d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -362,7 +362,7 @@ dependencies = [
  "manyhow",
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -378,7 +378,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "quote-use",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -465,9 +465,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "2.6.0"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
+checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
 
 [[package]]
 name = "block-buffer"
@@ -491,9 +491,9 @@ dependencies = [
 
 [[package]]
 name = "brotli-decompressor"
-version = "4.0.1"
+version = "4.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a45bd2e4095a8b518033b128020dd4a55aab1c0a381ba4404a472630f4bc362"
+checksum = "74fa05ad7d803d413eb8380983b092cbbaf9a85f151b871360e7b00cd7060b37"
 dependencies = [
  "alloc-no-stdlib",
  "alloc-stdlib",
@@ -501,9 +501,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.16.0"
+version = "3.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
+checksum = "1628fb46dfa0b37568d12e5edd512553eccf6a22a78e8bde00bb4aed84d5bdbf"
 
 [[package]]
 name = "bytecheck"
@@ -525,7 +525,7 @@ checksum = "523363cbe1df49b68215efdf500b103ac3b0fb4836aed6d15689a076eadb8fff"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.90",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -536,9 +536,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.9.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "325918d6fe32f23b19878fe4b34794ae41fc19ddbe53b10571a4874d44ffd39b"
+checksum = "f61dac84819c6588b558454b194026eb1f09c293b9036ae9b159e74e73ab6cf9"
 
 [[package]]
 name = "bytestring"
@@ -557,9 +557,9 @@ checksum = "8b96ec4966b5813e2c0507c1f86115c8c5abaadc3980879c3424042a02fd1ad3"
 
 [[package]]
 name = "cc"
-version = "1.2.5"
+version = "1.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c31a0499c1dc64f458ad13872de75c0eb7e3fdb0e67964610c914b034fc5956e"
+checksum = "755717a7de9ec452bf7f3f1a3099085deabd7f2962b861dae91ecd7a365903d2"
 dependencies = [
  "jobserver",
  "libc",
@@ -634,7 +634,7 @@ dependencies = [
  "serde-lite",
  "serde-wasm-bindgen",
  "serde_json",
- "thiserror 2.0.9",
+ "thiserror 2.0.11",
  "wasm-bindgen",
 ]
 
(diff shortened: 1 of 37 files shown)


benwis:
Awesome work as always, I see no harm in merging this one for experimentation purposes. I'll ping @gbj here in case he wants to chat, but let's try it

gbj:
I have not yet had time to look at this in detail, but the implementation involved passing type-erased attributes, or the potential for type-erased attributes, everywhere.

A quick test against the `hackernews` example shows a 426kb release build before this commit, and 454kb afterwards, i.e., this adds 18% to the binary size of the application. For `todomvc`, it's 210kb to 216kb. For `axum_js_ssr`, it's 617kb to 661kb.

To me, this is prohibitive.

Making type erasure more pervasive *will have this bloating affect* and should be avoided if possible: the reason the view tree is typed so thoroughly is to allow the compiler to remove dead code, whereas the type-erased types end up adding significant amounts of unused code.

zakstucke:
@gbj I'm not sure if I'm missing something, but those changes sound like 6.6%, 2.9% and 7.1% respectively?

It's important to remember that this could literally be a side effect of fixing the bug: there are now attributes included that were previously being thrown away. `AnyView` is used internally in `Fragment`, `ChooseView` and the routers for some sections, on top of an direct downstream usage.

I haven't spent anytime looking at binary size, I'll do so this weekend. But even if that doesn't lead to anything, I don't think it takes away from this PR being a good idea right now in that: 
- This is a silent but important bug, that has already been noticed downstream multiple times.
- This provides parity with erase_components, making fast compiles a first-class citizen
- There is as of now no alternative, I've spent quite some time playing around with the issue and the compiler: it's simply just too many types. As it stands I don't see any alternative method to fix the bug. 
- The binary size change isn't drastic, and may be explained away by the "missing code", i.e. the lack of attributes that should've existed but were thrown away prior to this. In my opinion, a serious bug trumps a binary size change (if no alternative way of fixing the bug).
- It's easily reversed in a 0.9 if a better solution arises, or rust / compiler improves performance, the latter I don't hold out much hope for. The breaking change on the average downstream user will be next to none now or in a reversal.

How are you benchmarking binary size? Just so we're on the same page. I'm currently testing with the below, but getting quite different results:
`cargo +stable leptos build --release` then checking `target/site/pkg/$name.wasm`

gbj:
Sure, I just can't do math. Still, those are pretty large changes from my perspective.

I do not think that the changes are due to including attributes that were previously thrown away: in the examples I compared, I haven't noticed that there are elements missing attributes that they should have. If you have noticed missing attributes in those three examples that this fixes, could you point them out to me? I would be glad to fix that.

Like I said, I have not reviewed this PR other than taking a glance and testing the size increases, so I don't have any further substantive thoughts.

zakstucke:
It's no problem, I understand the frustration! I've felt similar trying to solve it 😅 .

> One point of clarification: The bug that we are talking about is, specifically, the fact that attribute spreading onto components that return AnyView doesn't work, right? I just want to make sure I'm on the same page.

Yes. https://github.com/leptos-rs/leptos/pull/3518 provided a fix for the erased branch, making erased work perfectly as a faster-to-compile version of leptos for dev work (currently about 50% faster, I have a few more PR's lined up that make it another 50% faster on-top). It was a much bigger problem on there, because everything becomes an `AnyView` so all attributes would be lost. But this left "stable" leptos in a sticky situation of now having a "bug" that erased does not: the anyviews that do exist do not respect their attributes, albeit far less important now it works on erased but a confusing inconsistency nontheless. This PR is an attempt at fixing the issue on "stable", without does not require vectorising **everything** and the imagined performance issues that would entail (I actually haven't tested erased yet for binary size or runtime speed).

I'm not a big fan of the solution either, but I've been quite deep down the rabbit hole and this is the only way I've gotten it working without vectorising things that weren't intended to be `AnyView`'s, which I think would have bigger downsides. Both the vectorising fix and this one, both exemplify and solve the root issue: too large / many types if they aren't culled at each `AnyView` boundary.

The best tool I've been using is [cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines), it's somewhat limited but it really shows when things blow up - e.g. it's possible to compile with `AddAnyAttr` as originally intended without this fix if you gut `AnyView` of all it's contents, but you then get to see that simply by adding the implementation, the llvm lines shoot up from about 10million to 100million.

Some other things iirc i've attempted: 
- Optimising hot trait implementations, which is where all the monomorphisation bottlenecks are
- Removing const generics and replacing them with functions, which supposedly helps
- dynamic dispatch as an alternative to monorphised erasure

Some of it helps with compile time, but nothing big enough to solve the types issue.

Happy to discuss offline too

PR #3567, opened 2025-02-08
fix(reactive_stores_macro): `store` attribute signature error message

DanikVitek:
The error message did not correspond to the actual signature


gbj:
Thank you!

PR #3571, opened 2025-02-09
Allows non static lifetimes in component macro

### LINKED ISSUE #3564: Components with arguments passed by reference
I would like to be able to create components that take arguments by value:

```rust
#[component]
fn ByRef<'a>(_a: &'a str) -> impl IntoView {
    "static lifetime"
}
```

Which currently expands to something like:

```rust
struct ByRefProps<'a> {
    #[builder(setter(doc = "**_a**: [`&'a str`]"))]
    _a: &'a str,
}
impl<'a> ::leptos::component::Props for ByRefProps<'a> {
    type Builder = ByRefPropsBuilder<'a>;
    fn builder() -> Self::Builder {
        ByRefProps::builder()
    }
}
fn ByRef<'a>(props: ByRefProps<'a>) -> impl IntoView + 'a {
    let ByRefProps { _a } = props;
    ::leptos::reactive::graph::untrack_with_diagnostics(move || __ByRef(_a))
}
pub fn __ByRef<'a>(_a: &'a str) -> impl IntoView {
    "static lifetime"
}
```

With the issue being the `+ 'a` in the `ByRef` fn output.

The `component` macro should not automatically add a `impl IntoView + '_` in it's output, but should leave the lifetimes as specified by the user. Old behaviour should still be available by manually adding the lifetime, while in *most* cases, the `impl IntoView` actually has a static lifetime.

I think, with the way that lifetimes currently interact with the proc-macro, it will never be useful. Therefore, this should not be a breaking change. It seems more like a bug in the current system, rather than a feature to be implemented. It would be great if this could still be included in the 0.8 release :)

jvdwrf:
closes #3564

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -144,8 +144,6 @@ impl ToTokens for Model {
         let (impl_generics, generics, where_clause) =
             body.sig.generics.split_for_impl();
 
-        let lifetimes = body.sig.generics.lifetimes();
-
         let props_name = format_ident!("{name}Props");
         let props_builder_name = format_ident!("{name}PropsBuilder");
         let props_serialized_name = format_ident!("{name}PropsSerialized");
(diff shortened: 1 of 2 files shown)


gbj:
Thanks! Lints are failing in CI because `lifetimes` is now unused

jvdwrf:
Ah thanks, It should be fixed now with the new commit

Issue #3568, opened 2025-02-09
`IntoStyle` for `(&str, Option<T>)`s


luxalpa:
**Is your feature request related to a problem? Please describe.**
Conditionally setting css styles on an HTML element.

**Describe the solution you'd like**
In Leptos 0.6 it was possible to provide an `Option<T>` or an `fn -> Option<T>` to `style:cssprop=` attributes (i.e. `.style(("prop", Some("value")))`. I was somewhat depending on that feature, but it seems it was forgotten during the upgrade to 0.7.

**Describe alternatives you've considered**
The workaround right now is to use style overrides (like `.unwrap_or("auto")` or use just a single style attribute for all styles and render the HTML. However both of these approaches have downsides as they are not able to precisely add and remove the singular style from the element.

**Additional context**
I've tried doing he PR myself but actually the trait relationships with asyncOutput etc are really difficult and I'm not deep enough in the new leptos code base to understand it although I did spend a couple of hours trying to get something to work.

luxalpa:
I didn't know these are different but I actually need `IntoStyle` for the reactive function that returns an `Option<T>` for my use-case.

gbj:
Done for 0.8 in #3618

PR #3576, opened 2025-02-10
Implement IntoStyle for `(str, Option)`

luxalpa:
This is my attempt at allowing `(str, Option<T>)` in IntoStyles for #3568 . As I am not very familiar with the code base there might be things I am misunderstanding.

I also wanted to fix the case for a `ReactiveFunction<Output = Option<T>>` however the way I see it it might require a refactor, due to possible overlapping impls.

Also, I feel like in general something like `trait IntoStyleValue` could be useful.

### DIFF
--- a/tachys/src/html/style.rs
+++ b/tachys/src/html/style.rs
@@ -9,7 +9,7 @@ use crate::{
     renderer::Rndr,
     view::{Position, ToTemplate},
 };
-use std::{future::Future, sync::Arc};
+use std::{borrow::Cow, future::Future, sync::Arc};
 
 /// Returns an [`Attribute`] that will add to an element's CSS styles.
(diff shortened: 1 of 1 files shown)


gbj:
Thanks for your work! Having looked at it a little today, I agree with your idea that fixing it for the reactive case requires larger refactoring. I've made a small start and will try to take a crack at it this weekend -- I'm going to leave this open for now until I see what needs to happen.

gbj:
Okay, I think I was able to do a more general implementation of this in #3618 — I made the PR against the 0.8 branch as it includes some semver-breaking changes I think. (Good idea to use a helper trait.)

luxalpa:
Perfect, thank you!

PR #3579, opened 2025-02-10
Impl into<Signal> for subfields

jvdwrf:


gbj:
Thanks! Technically we could implement something marginally more optimal if `reactive_graph` knew about `reactive_stores` (not possible, circular dependency) or if `Signal` was implemented differently, but this is plenty good for now and is a nice convenience.

jvdwrf:
Yeah I saw that, and would have liked to implement it from the reactive_graph crate. 
I guess it would be something that can be added without backwards compatibility concerns.

Would it be possible to expose some more internal details of `Signal`, s.t. users could create custom signal types that convert to signal without overhead?

PR #3582, opened 2025-02-11
Implement several into traits for store fields

mahdi739:
This PR implements these traits:
- `IntoClass`
- `IntoStyle`
- `IntoProperty`
- `Render`
- `RenderHtml`
- `AttributeValue`
- `InnerHtml`

for these store fields:
- `Store`
- `ArcStore`
- `Field`
- `ArcField`
- `SubField`
- `AtKeyed`
- `KeyedSubField`
- `DerefedField`
- `AtIndex`

Implementation for `Store`, `Field`, `ArcStore`, and `ArcField` was approachable easily by existing macro rules for signals, but for other store fields, I had to write new macro rules. Maybe someone can unify them with a better design.

The implementation of `AddAnyAttr` for signals was empty and marked with `todo!()`. I also added this empty implementation to new macros.

~~I didn't put them under the "reactive_stores" feature flag. Please let me know if I should do so.~~
I put them under the "reactive_stores" feature flag.

Additionally, it implements `IntoSplitSignal` for `AtKeyed`, `AtIndex`, and `DerefedField`.

This PR is written based on the previous PR (#3569), which had rewritten the implementation of signals to use `try_get()` instead of `get()`.

### DIFF
--- a/tachys/src/reactive_graph/bind.rs
+++ b/tachys/src/reactive_graph/bind.rs
@@ -14,17 +14,22 @@ use crate::{
     renderer::{types::Element, RemoveEventHandler},
     view::{Position, ToTemplate},
 };
-#[cfg(feature = "reactive_stores")]
-use reactive_graph::owner::Storage;
 use reactive_graph::{
     signal::{ReadSignal, RwSignal, WriteSignal},
     traits::{Get, Update},
     wrappers::read::Signal,
 };
-#[cfg(feature = "reactive_stores")]
-use reactive_stores::{ArcField, Field, KeyedSubfield, Subfield};
 use send_wrapper::SendWrapper;
 use wasm_bindgen::JsValue;
+#[cfg(feature = "reactive_stores")]
+use {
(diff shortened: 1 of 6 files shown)


benwis:
This is getting out of hand, now there are two of them!

Would we want to merge this one and the other one?
Overall I think using .try_get() is a decent solution here for panic prevention, but I am worried that instead of getting a panic it seems like this implementation just eats the error.  I imagine it would have to emit an error somewhere user visible, whether that's the console in the browser or the server log on the server

benwis:
Copied from the other one you've opened, so apologies if you read that one first. Feel free to reply to either one

Overall I think using .try_get() is a decent solution here for panic prevention, but I am worried that instead of getting a panic it seems like this implementation just eats the error. I imagine it would have to emit an error somewhere user visible, whether that's the console in the browser or the server log on the server

mahdi739:
I close this PR in favor of the new PR: #3658

PR #3588, opened 2025-02-12
Minor: "wasm-bindgen" - Moved the crate definition up to the root workspace

martinfrances107:
"This synchronizes the version number amongst all sub-projects. [Where the definition is "optional" manual adjustment is still required]"

Please forgive my paranoia

Upstream from here in my app I recently had some issues where a major change to the wasmbindgen crate was  being hidden in a "minor update".

With this PR, in all these core sub crates ( and more ), the version number is now in lock step.

leptos/Cargo.toml
leptos_dom/Cargo.toml
leptos_server/Cargo.toml

This is a defensive patch as any bugs due to differing version number brings up are a pain to resolve.


benwis:
Looks good to me

PR #3589, opened 2025-02-12
Remove getrandom

martinfrances107:
This warning 

"warning: Feature js of package getrandom has been obsolete in version 0.3.1"

Now using the feature "wasm_js"

Just a small maintence nudge

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-rustflags = ["--cfg", "getrandom_backend=\"wasm_js\""]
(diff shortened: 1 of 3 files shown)


martinfrances107:
ok so the rustflags need to be platform sepecfic - to take care of windows etc

martinfrances107:
My bad I was not asking the question what uses getrandom ... nothing

benwis:
For this conflict, can you run a `cargo build` so that it updates the Cargo.lock?

martinfrances107:
Fixed

benwis:
Thanks!

PR #3590, opened 2025-02-12
Minor: Bumped version of convert_case to 0.7

martinfrances107:
Just doing a minor periodic spring clean

I want to keep the output of "cargo outdated" to a small set


benwis:
LGTM

PR #3593, opened 2025-02-12
Minor: Bump itertools to "0.14.0"

martinfrances107:
The crate is used by "leptos_macro","reactive_stores" and "tachys"

So the definition of itertools can be centralized up into the root workspace

PS 

Dependabot report 16 vunerabilities .. as a way to resolve this 

I am chippy away at the list provided by "cargo outdated" .. as a way to reasons with  those "potential" vunerabilities


benwis:
LGTM

Issue #3583, opened 2025-02-11
Hydration error when rendering `Vec<AnyView>` between Text Nodes.


luxalpa:
**Describe the bug**
When rendering a `Vec<AnyView>` that returns a text node or that is an empty vec in between 2 other text nodes, there will be a hydration panic.

```
A hydration error occurred while trying to hydrate an element defined at src\app.rs:80:10.

The framework expected a text node, but found this instead:  
<!--  -->
```
 
This does not happen if it's only surrounded by one text node (and possibly an HTML element on the other side), or when the view inside the vec returns other HTML elements.

**Leptos Dependencies**
```toml
[patch.crates-io]
leptos = { git = "https://github.com/leptos-rs/leptos.git", rev = "b17a4c92c745271dba947412bf14b6e0ece1270f" }
leptos_meta = { git = "https://github.com/leptos-rs/leptos.git", rev = "b17a4c92c745271dba947412bf14b6e0ece1270f" }
leptos_router = { git = "https://github.com/leptos-rs/leptos.git", rev = "b17a4c92c745271dba947412bf14b6e0ece1270f" }
leptos_actix = { git = "https://github.com/leptos-rs/leptos.git", rev = "b17a4c92c745271dba947412bf14b6e0ece1270f" }
reactive_graph = { git = "https://github.com/leptos-rs/leptos.git", rev = "b17a4c92c745271dba947412bf14b6e0ece1270f" }
```

**To Reproduce**
```rs
#[component]
fn HomePage() -> impl IntoView {
    // let elements: Vec<AnyView> = vec![]; // <- Empty vec will also reproduce the issue.
    let elements: Vec<AnyView> = vec![view! {
        "hello"
    }
    .into_any()];

    view! {
        <span class="italic">
            "Any text" {elements} "other text"
        </span>
    }
}
```

**Expected behavior**
Not panic

**Additional context**
I want to render a text node inside parentheses as part of a specific markup parser. Currently working around the issue by creating additional `<span>` tags.

```rs
fn into_render(self) -> Self::Output {
        match self {
            MarkupElement::Text(text) => text.into_any(),
            MarkupElement::Symbol(symbol) => {
                let url = symbol.svg_url();
                view! {
                    <img class="cost-symbol" src=url alt=symbol.to_string() />
                }
                .into_any()
            }
            MarkupElement::Italic(elements) => {
                let elements = elements
                    .into_iter()
                    .map(IntoRender::into_render)
                    .collect_view();

[...]

PR #3598, opened 2025-02-12
Strange marker node bug

zakstucke:
This is a rather unorthodox bug report, but I thought this would be the easiest way to review and checkout locally. 

Whilst adding erasure, I think I've stumbled on a bug, or something very weird. 

It seems having `HtmlElement` extend it's children as `Vec<AnyView>`  breaks leptos somehow. With this PR, running `counter_isomorphic` `cargo leptos watch` and checking out the site, console shows: 
```
(index):419 Uncaught SyntaxError: Unexpected token '>' (at (index):419:21)
(index):427 Uncaught SyntaxError: Unexpected token '>' (at (index):427:58)
```
Which is a random `<!>` marker in the html in the second line of this snippet:
```
ws.onclose = () => console.warn('Live-reload stopped. Manual reload necessary.');
 })(3001, 'ws://')<!></script><link rel="modulepreload" href="/pkg/counter_isomorphic.js" nonce="nRIE4NM7UA2nhEng5NCK0g"><link rel="preload" href="/pkg/counter_isomorphic.wasm" as="fetch" type="application/wasm" crossorigin="nRIE4NM7UA2nhEng5NCK0g"><script type="module" nonce="nRIE4NM7UA2nhEng5NCK0g">(function (root, pkg_path, output_name, wasm_output_name) {
	import(`${root}/${pkg_path}/${output_name}.js`)
		.then(mod => {
			mod.default({module_or_path: `${root}/${pkg_path}/${wasm_output_name}.wasm`}).then(() => {
				mod.hydrate();
			});
		})
})
```

I'm pretty confused how it could be anything wrong with this diff, and I've un erased flagged it so this is verifiable in normal mode. Any ideas @gbj? 

I also checked patching 0.8 branch with the solution to #3583, but doesn't help.

### DIFF
--- a/tachys/src/html/element/mod.rs
+++ b/tachys/src/html/element/mod.rs
@@ -7,8 +7,9 @@ use crate::{
     renderer::{CastFrom, Rndr},
     ssr::StreamBuilder,
     view::{
-        add_attr::AddAnyAttr, any_view::AnyView, IntoRender, Mountable,
-        Position, PositionState, Render, RenderHtml, ToTemplate,
+        add_attr::AddAnyAttr, any_view::AnyView, iterators::StaticVec,
+        IntoRender, Mountable, Position, PositionState, Render, RenderHtml,
+        ToTemplate,
     },
 };
 use const_str_slice_concat::{
@@ -106,7 +107,7 @@ where
     NewChild: IntoRender,
     NewChild::Output: RenderHtml,
 {
-    type Output = HtmlElement<E, At, Vec<AnyView>>;
+    type Output = HtmlElement<E, At, StaticVec<AnyView>>;
 
     fn child(self, child: NewChild) -> Self::Output {
         use crate::view::any_view::IntoAny;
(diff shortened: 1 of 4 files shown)


zakstucke:
CI failures show the same issue I'm describing

zakstucke:
@gbj thanks! If my second (tiny) commit is what you were describing, then this seems to have fixed it. Seems to work fine locally. 

The html diff error in the `counters` CI example is weird though, I think it might just be a side-effect of the erasure but not actually a problem.

benwis:
@zakstucke  Looks like it needs a bit of tweaking, you probably didn't mean to insert more `<!>`

zakstucke:
@benwis haven't added any new `<!>` I don't think, just only including them if `escape=true`, which I _think_ is what @gbj meant, and I _think_(?!) makes sense to me.

benwis:
> @benwis haven't added any new `<!>` I don't think, just only including them if `escape=true`, which I _think_ is what @gbj meant, and I _think_(?!) makes sense to me.

I thought he meant escaping the characters itself, like 

> @benwis haven't added any new `<!>` I don't think, just only including them if `escape=true`, which I _think_ is what @gbj meant, and I _think_(?!) makes sense to me.

I can't find the test that failed, but one of them failed with a weird `<!>` before a closing tag. Most of the tests seem to fail like this
```
         assertion `left == right` failed
          left: "<div><button>Clear<!----></button><button>-1<!----></button><span>Value: 0!<!----></span><button>+1<!----></button><!----></div>"
         right: "<div><button>Clear</button><button>-1</button><span>Value: 0!<!----></span><button>+1</button><!----></div>"
```
Which seems like you're not escaping enough? Or differently than we were before, so the tests fail

gbj:
Commit looks right to me -- not sure why it adds escape placeholders in the `counters` example, I can dig in a bit.

zakstucke:
@gbj I think I've found a workaround by just making a `StaticVec<T>` type that wraps `Vec<T>` but doesn't have to worry about marker nodes because the places it's used should always be static (i.e. replacing a static array or tuple type). 

I'll leave this open incase you did want to dig into the difference in the counter example we were seeing before, but feel free to close, I'll open a new PR with this all cleaned up.

zakstucke:
I'll close this, there's no real need to find out what was going on here unless someone actually notices something like this in normal usage. Thanks for the help 🙏

PR #3614, opened 2025-02-15
Internally erase html elements

zakstucke:
As iterated on in #3598. 

I've left out the changes from #3583, they'll be pulled in from main and anything not explicitly related to this to keep simple.


benwis:
Awesome!

PR #3623, opened 2025-02-16
Erased routing, codegen opts

zakstucke:
There's a few things in here (sorry!), to sum up: 
- Routing is erased, most importantly this should (probably) solve any remaining compile errors in rust debug/dev builds
- The `view!{}` macro now erases thoroughly
- A few more hot codegen optimisations of existing logic in traits
- erased containers define their `fn()` functions with concrete `fn foo` declarations rather than closures, TIL this is more codegen efficient
- The [erased](https://docs.rs/erased/latest/erased/) crate is used as a replacement for `Box<dyn Any>` in erased containers, which is more efficient and (I think) avoids a massive `dyn Any` vtable
- `IntoFragment` contains `StaticVec` instead of a `Vec`

This shouldn't have any breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -924,6 +924,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
+name = "erased"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1731451909bde27714eacba19c2566362a7f35224f52b153d3f42cf60f72472"
+
+[[package]]
 name = "errno"
 version = "0.3.10"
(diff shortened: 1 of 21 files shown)


zakstucke:
This is **almost** finished, there's a second part that somehow breaks leptos, which I need to fix, but I've found the change that breaks it, I'll fix it then finish this in like 8 hrs
But after this PR, there's an erase internals of Either PR, which I'll only open if it actually benchmarks gains after this one, then the erasure journey's complete I think

zakstucke:
Should be ready to go, updated description.

zakstucke:
Happy to say my `Either` enum internal erasure PoC, on top of everything else, only provides a minimal codegen optimisation (~1%), so is not worth the breaking change and weirdness of a variant that is actually a function (to handle vari-erasure) etc. 

This means (I think) the erasure journey is now complete 🎉

benwis:
This looks good to me, I'll let @gbj take a look before we merge this. Super excited!

PR #3640, opened 2025-02-19
Erased mode in CI

zakstucke:
Add erase components as part of the test matrix, seems to work first try, although I am slightly surprised by that!

### DIFF
--- a/.github/workflows/ci-changed-examples.yml
+++ b/.github/workflows/ci-changed-examples.yml
@@ -28,5 +28,6 @@ jobs:
     uses: ./.github/workflows/run-cargo-make-task.yml
     with:
       directory: ${{ matrix.directory }}
+      erased_mode: ${{ matrix.erased_mode }}
       cargo_make_task: "ci"
       toolchain: stable
--- a/.github/workflows/ci-examples.yml
+++ b/.github/workflows/ci-examples.yml
@@ -25,5 +25,6 @@ jobs:
     uses: ./.github/workflows/run-cargo-make-task.yml
     with:
       directory: ${{ matrix.directory }}
+      erased_mode: ${{ matrix.erased_mode }}
       cargo_make_task: "ci"
       toolchain: stable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,6 @@ jobs:
     uses: ./.github/workflows/run-cargo-make-task.yml
     with:
       directory: ${{ matrix.directory }}
-      dev_mode: ${{ matrix.dev_mode }}
+      erased_mode: ${{ matrix.erased_mode }}
       cargo_make_task: "ci"
       toolchain: nightly-2025-02-19
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,5 +25,6 @@ jobs:
     uses: ./.github/workflows/run-cargo-make-task.yml
     with:
       directory: ${{ matrix.directory }}
+      dev_mode: ${{ matrix.dev_mode }}
       cargo_make_task: "ci"
       toolchain: nightly-2025-02-19
--- a/.github/workflows/get-changed-examples-matrix.yml
+++ b/.github/workflows/get-changed-examples-matrix.yml
@@ -50,5 +50,5 @@ jobs:
             echo "matrix={\"directory\":${{ steps.changed-dirs.outputs.all_changed_files }}}" >> "$GITHUB_OUTPUT"
           else
             # Create matrix with one item to prevent an empty vector error
-            echo "matrix={\"directory\":[\"NO_CHANGE\"]}" >> "$GITHUB_OUTPUT"
+            echo "matrix={\"directory\":[\"NO_CHANGE\"], \"erased_mode\": [false, true]}" >> "$GITHUB_OUTPUT"
           fi
--- a/.github/workflows/get-examples-matrix.yml
+++ b/.github/workflows/get-examples-matrix.yml
@@ -28,7 +28,7 @@ jobs:
           sed 's/\/$//' |
           jq -R -s -c 'split("\n")[:-1]')
           echo "Example Directories: $examples"
-          echo "matrix={\"directory\":$examples}" >> "$GITHUB_OUTPUT"
+          echo "matrix={\"directory\":$examples, \"erased_mode\": [false, true]}" >> "$GITHUB_OUTPUT"
       - name: Print Location Info
         run: |
           echo "Workspace: ${{ github.workspace }}"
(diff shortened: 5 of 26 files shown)


benwis:
I also think it's working, but I am not the CI guru. Nice!

sabify:
I don't see any issue with the workflow and the configuration flag has been set correctly in dev mode on. The only issue is that it is only running over member crates and not examples. I don't know if it is intended for that.

@benwis can you please elaborate more on "Cache warnings"? I don't see any warnings like that...

sabify:
IMO, naming "dev_mode" is a bit misleading since you can be in dev mode without activating the erase_components cfg flag. Perhaps having its name be more clear here.

benwis:
> I don't see any issue with the workflow and the configuration flag has been set correctly in dev mode on. The only issue is that it is only running over member crates and not examples. I don't know if it is intended for that.
> 
> @benwis can you please elaborate more on "Cache warnings"? I don't see any warnings like that...

If I click into the checks and then click on a specific thing, it switches from passed to failed, and throws a warning. This page is an example of one: https://github.com/leptos-rs/leptos/actions/runs/13422224100

It could be possible that I'm jetlagged enough to not understand it, but it seems like things are wonky

sabify:
You are right. I mentioned that the example CI left unmodified. @zakstucke You need to add the same flags from `get-leptos-matrix.yml` and `ci.yml` to the respective example CI configuration, `ci-changed-examples.yml` and `ci-examples.yml`.

benwis:
Also adding a note here that we should figure out how to verify if it's in erase_components mode in the CI

sabify:
> IMO, naming "dev_mode" is a bit misleading since you can be in dev mode without activating the erase_components cfg flag. Perhaps having its name be more clear here.

I suggested that we may need to rename `dev_mode`.

benwis:
There is a part of me that finds the idea of the CI just up and deciding it's done compiling and to start the tests is hilarious.
```rust
   Compiling wasm-bindgen v0.2.100
+ echo Times up, running tests
Times up, running tests
[cargo-make] INFO - Running Task: test-playwright
+ cd /home/runner/work/leptos/leptos/examples/error_boundary
+ BOLD=\e[1m
+ GREEN=\e[0;32m
+ RED=\e[0;31m
+ RESET=\e[0m
+ project_dir=/home/runner/work/leptos/leptos/examples/error_boundary
+ command -v pnpm
+ PLAYWRIGHT_CMD=pnpm
/home/runner/setup-pnpm/node_modules/.bin/pnpm
+ find . -name playwright.config.ts
+ dirname ./e2e/playwright.config.ts
+ pw_dir=./e2e
+ cd ./e2e
+ pnpm playwright test
   Compiling js-sys v0.3.77
   Compiling yoke-derive v0.7.5
   Compiling zerofrom v0.1.5
```
Probably worth looking at why the compile times changed here

benwis:
This seems suboptimal:
```rust
[tasks.wait-server]
script = '''
for run in {1..12}; do
    echo "Waiting to ensure server is started..."
    sleep 10
done
echo "Times up, running tests"
```

sabify:
> This seems suboptimal:
> 
> ```rust
> [tasks.wait-server]
> script = '''
> for run in {1..12}; do
>     echo "Waiting to ensure server is started..."
>     sleep 10
> done
> echo "Times up, running tests"
> ```

I think a single 10s sleep will be enough...

sabify:
@zakstucke Have you tried transparent route component (#[component(transparent)])? I guess it does work without needing to modify the macro...

Almost all route components in the examples (and failed CIs) are not transparent.

zakstucke:
@sabify this wouldn't work because the parent component, the one calling the nested routing component, is expecting `AnyNestedRoute` in erased mode, and can't handle the generic `impl` return type. It's very specific to that case, and the way I've done it requires no breaking change/interop between erased and normal mode. Couldn't find anything better than a proc macro hack, so I think it's fine for this case.

sabify:
@zakstucke I just replaced `#[component]` with `#[component(transparent)]` over `suspense_tests`' `InstrumentedRoutes()` and it worked.

zakstucke:
@sabify I think you must've checked in non-erased mode, at https://github.com/leptos-rs/leptos/commit/e36d7da1ecec5991beaaa94633ace1c0d2631bf2, setting `InstrumentedRoutes` to `#[component(transparent)]` and running `RUSTFLAGS="--cfg erase_components" cargo test` fails. With current HEAD both erased and non-erased work.

sabify:
@zakstucke This is what I've done and I still don't get any error:
```bash
git clone https://github.com/leptos-rs/leptos
cd leptos/examples/suspense_tests
cargo clean
sed -i '205s/.*/#[component(transparent)]/' src/instrumented.rs
RUSTFLAGS="--cfg=erase_components" cargo test --features=ssr
RUSTFLAGS="--cfg=erase_components" cargo test --features=hydrate
# OR
RUSTFLAGS="--cfg=erase_components" cargo build --features=ssr
RUSTFLAGS="--cfg=erase_components" cargo build --features=hydrate
# OR
RUSTFLAGS="--cfg=erase_components" cargo leptos build
```

zakstucke:
This is on the 0.8 branch, where routing itself is erased, 0.7 wouldn't have this problem

sabify:
Ooops! Thanks for the clarification and your hard work on this feature.

benwis:
Glad to see you two collaborating. Looks like we've only got one failure left to fix!

zakstucke:
The last ones a bit of a headscratcher. It's not an issue with erased mode, but a rust limitation where if you cross compile to a different target e.g. wasm, the proc macro doesn't respect RUSTFLAGS, so it's in a situation where `leptos_macro` is compiling in non-erased, and the rest of the crates in erased, leading to the error. 

The good news is it means testing erased mode for wasm/frontend after this is fixed will probably be faster, bad news is I have absolutely no idea how to solve it 😅 . 

https://github.com/rust-lang/cargo/issues/4423
https://users.rust-lang.org/t/rustflags-ignored-when-cross-compiling-a-proc-macro-crate/92537

sabify:
How about adding `CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS: ${{ inputs.erased_mode && '--cfg erase_components' || '' }}` besides `RUSTFLAGS` in `.github/workflows/run-cargo-make-task.yml`?

zakstucke:
It's actually the other way round, it's the host platform that doesn't respect them (e.g. mac or linux), because rust deems the RUSTFLAGS shouldn't apply to build.rs or proc macros. As far as I can see there's absolutely no workaround that works, it's pretty ridiculous. 

I think I've found a solution though, and internal `__internal__erase_components` feature for `leptos_macro` that get's automatically enabled by the `leptos` crate. This doesn't have the same drawbacks we initially didn't want to use a feature for erasure for, because this is internal and auto enabled.

zakstucke:
@benwis think this is finally there, new CI errors aren't erase specific and look to just be versioning issues.

sabify:
@zakstucke Could you please update `examples/spread/rust-toolchain.toml`'s channel to `nightly-2025-02-19`? This may solve the compiling issue of the failed example.

zakstucke:
Passed! 🎉

benwis:
Woah, one line move past another, nice work!

Issue #2822, opened 2024-08-13
Implement immediate/synchronous effects


gbj:
**Is your feature request related to a problem? Please describe.**

0.7 changes effect scheduling to automatically batch effects, and to run them asynchronously, as determined by whichever async runtime is being used. In the default web case, this is `wasm-bindgen-futures`, which uses `queueMicrotask`. In other words, updating a signal updates its value immediately, but runs effects on the next microtask. This means that an effect does not run immediately, once per signal change.

In general, this is good, because effects are intended to synchronize the current/latest value of a signal with the outside world in some way that is typically orders of magnitude more expensive than actually updating the signal: for example, updating a DOM node.

This has several benefits:
1. Updating a signal multiple times, synchronously, does not run that expensive effect (like updating a DOM node) multiple times in a way that's unnecessary
2. Updating multiple signals read in an effect does not run that effect multiple times. (Currently, users can opt into batching using `batch()`; the number of bug reports in which the solution was "use `batch()`" and that wasn't obvious to the user is more than a handful.)

The canonical example of why this is good would be something like
```rust
let a = RwSignal::new(1);
let b = RwSignal::new(2);

Effect::new(move |_| {
  let value = a.get() + b.get();
  // do something expensive with `value`, like updating the DOM
});

// ... 
on:click=move |_| {
  a.set(3);
  b.set(4);
}

// how many times did you want the effect to run?
// did you really intend there to be an inconsistent state
// where the effect had the old value of `b` and new value of `a`?
```

However, there are some situations in which users want an effect that runs immediately when signals are changed, and that does not batch in this way.

**Describe the solution you'd like**

An alternative effect type whose `ReactiveGraph` and `Subscriber` implementations are designed such that, when marked check or dirty, it simply runs its inner function, instead of triggering a notification to wake up an async task. 

This means `Effect` and render effects (created for closures passed into the view, for example) run in the async, batched way, but users can opt into immediate effects for other cases.

**Describe alternatives you've considered**

1. Making all effects immediate, and requiring opt-in batching. (Status quo ante 0.7)
2. A `flush_sync()` that allows you to tell all effects to run now, not on the next microtask. (Difficult to implement because the async runtimes we piggyback on don't expose what we need to do this.)

connorgmeehan:
[...]

PR #3650, opened 2025-02-24
`ImmediateEffect`

QuartzLibrary:
This introduces `ImmediateEffect`, which executes its effect once upon creation and then every time a dependency changes, immediately.

**This is not what you want in most code**, but is useful for some kinds of double-binding (for example to some syncing mechanism like CRDTs, or a signal to the browser's route state).

I think this can live outside `leptos`, so if you think it's better to keep it out or this version is not quite where you want it, I am happy to close this.

---

A note on `Send` and `Sync`: since the dependency that triggers the effect could be in any thread, this also means the effect *always* has to be `Send` and `Sync`. I still matched the `Effect` interface.
This is unfortunate, but I couldn't figure out a way to avoid that without `unsafe`, and even then it would have 'solved' the problem by panicking if called from another thread.

I left that code in an intermediate commit, but as there is currently no `unsafe` code in `leptos`, I don't think it's worth it even if it's sound (as a casual observer would lose the simple and strong no-unsafe anywhere guarantee).

---

cc: https://github.com/leptos-rs/leptos/issues/2822

### DIFF
--- a/any_error/src/lib.rs
+++ b/any_error/src/lib.rs
@@ -109,7 +109,7 @@ pub fn get_error_hook() -> Option<Arc<dyn ErrorHook>> {
 /// Sets the current thread-local error hook, which will be invoked when [`throw`] is called.
 pub fn set_error_hook(hook: Arc<dyn ErrorHook>) -> ResetErrorHookOnDrop {
     ResetErrorHookOnDrop(
-        ERROR_HOOK.with_borrow_mut(|this| mem::replace(this, Some(hook))),
+        ERROR_HOOK.with_borrow_mut(|this| this.replace(hook)),
     )
 }
 
--- a/any_error/src/lib.rs
+++ b/any_error/src/lib.rs
@@ -8,8 +8,7 @@ use std::{
     cell::RefCell,
     error,
     fmt::{self, Display},
-    future::Future,
-    mem, ops,
+    future::Future, ops,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
@@ -108,9 +107,7 @@ pub fn get_error_hook() -> Option<Arc<dyn ErrorHook>> {
 
 /// Sets the current thread-local error hook, which will be invoked when [`throw`] is called.
 pub fn set_error_hook(hook: Arc<dyn ErrorHook>) -> ResetErrorHookOnDrop {
-    ResetErrorHookOnDrop(
-        ERROR_HOOK.with_borrow_mut(|this| this.replace(hook)),
-    )
+    ResetErrorHookOnDrop(ERROR_HOOK.with_borrow_mut(|this| this.replace(hook)))
 }
 
 /// Invokes the error hook set by [`set_error_hook`] with the given error.
--- a/any_error/src/lib.rs
+++ b/any_error/src/lib.rs
@@ -8,7 +8,8 @@ use std::{
     cell::RefCell,
     error,
     fmt::{self, Display},
-    future::Future, ops,
+    future::Future,
+    ops,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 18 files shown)


gbj (commented):
Nice, thank you! This would close #2822. 

I have looked over the implementation and it certainly looks correct to me. The only thing missing is adding some tests -- my assumption would be that copying some of the tests in `tests/effects.rs`, using immediate effects, and dropping the various `Executor::tick().await;` points (which wait for the current, batched/async effects to run) would be sufficient to test this!

Really nice work, and much appreciated.

gbj:
(Oh, and just to add a note on the Send/Sync bounds: these are okay, and definitely a requirement. I'd say that for this niche use case, I'm perfectly happy to say that users who need to run them on the current thread are very likely doing so in a single-threaded browser context and can use `SendWrapper` to wrap a `!Send` effect, etc.)

QuartzLibrary:
> The only thing missing is adding some tests

Definitely! I wanted to get it up as soon as possible to avoid doing this in the dark.

So, some more design considerations:
- If we allow recursion, we have to use `Fn`s.
  I'm fine with using `Fn`s because it's a more general interface and this is a bit of a 'power user' feature anyway. I have a helper that simulates panic-on-recursion usage, but I'm leaning toward removing it before merging. Skip-inner-recursion is also doable.
- If we allow recursion, then does it make sense to pass in the `Option<T>` to the effect function? On recursion it would be always `None` as the old value is still being used by the level above. I'm leaning toward no. The user can pass in additional state if they want/need it.
- If we allow recursion, the `sources` situations is a bit tricky, as in normal effects we clear them before every run. That might be unexpected under recursion. So, options:
  - As is: clear before every run, keep the last run in full and whatever comes after that in the parent runs.
  - Keep it all: track all sources from all recursion levels (clear the sources only before the top-level run).
  - Keep only the last run: figure out what the last run is, and only track things from that, ignoring any tracking that technically comes after it from the parents.

I think I am leaning toward 'allow recursion', no `Option<T>`, and keeping only the last run, but I'll sit on it a bit and see if that changes.

QuartzLibrary:
Implemented it as described above, I'll probably give it a look over the weekend with fresh eyes. It'll also need some multi-threaded tests.

gbj:
This looks reasonable to me. I understand the recursion count; I'm not entirely clear on the purpose of the thread ID check, so if you could write up a few sentences here or in comments, I'd appreciate it.

(The `spread` example is failing in CI, don't worry about it -- I'm happy to merge this assuming the tests you've written for the new code pass.)

sabify:
@QuartzLibrary @gbj  The `reactive_graph` CI failed because of a typo in this PR:

```diff
// --> reactive_graph/src/effect/immediate.rs:40:5
- use reactive_graph::effect::immediateEffect;
+ use reactive_graph::effect::ImmediateEffect;
```

QuartzLibrary:
Apologies for the delay in getting to this, I started looking at documenting the behavior and further simplifications and then got sidetracked trying to look into #3681.

Of particular relevance was that I was changing the effect to behave more like a `RenderEffect`, where it's not by default cleaned up when the scope is. It seemed more in line with the usage this should actually see (for example if you double-bind to a CRDT, you might want the CRDT-wrapper to own the effect and signal).

Is this still doable before a release, as it's now a breaking change against `main`?

---

As for `ThreadId`, I have added a comment in #3692 to explain:

```
Thread 1:
-------------------------
  ---   ---    =======

Thread 2:
-------------------------
            -----------
```

In the parallel example above, we can see why we need the `ThreadId`. The last run is marked using `=`, but another run in the other thread might also be gathering sources. So we only allow the run from the correct [ThreadId] to push sources.

---

Also, thank you @sabify for #3691, will follow up with further cleanup if #3692 is not merged.

sabify:
@QuartzLibrary Thanks for the follow-up and cleanup.

gbj:
@QuartzLibrary a breaking change change against `main` is not a problem, as it has not yet been released.

PR #3656, opened 2025-02-26
Add websocket support for server functions

ealmloff:
This PR adds websocket support for server functions that simultaneously accept and emit a stream of serializable items. This requires some breaking changes around the server function trait to model the request and response type together.

TODO:
- [x] Websocket protocol implementation
- [x] Http protocol implementation
- [x] Axum integration
- [x] Actix integration
- [x] Server function macro support
- [x] Reqwest integration


gbj:
I just read through this all and it looks amazing.

Hope you don't mind, I just pushed two commits after reviewing, as these were pretty minimal changes just knocking a few things off the CI list. The only substantive change was that we also support Axum running as WASM on JS Fetch based platforms like Deno, but `mio` can't be compiled for that target so the `ws` feature of Axum doesn't work. I've just made it panic under those circumstances, which I think is fine. Your API design here meant that no Axum WS types had to be exposed anyway, so I think this will work fine.

My only significant feedback: The breakages in the `server_fns_axum` example suggest to me that migrating to this will require a bunch of changes like this:
```rust
// from this
#[server(input = Postcard, output = Postcard)]
// to this
#[server(input = PostPostcard, output = PostPostcard)]
```
It would be nice if there were a way to do this in the macro automatically, given that the macro already has a default method (I think?)

Thanks for all your work.

ealmloff:
> I just read through this all and it looks amazing.
> 
> Hope you don't mind, I just pushed two commits after reviewing, as these were pretty minimal changes just knocking a few things off the CI list. The only substantive change was that we also support Axum running as WASM on JS Fetch based platforms like Deno, but `mio` can't be compiled for that target so the `ws` feature of Axum doesn't work. I've just made it panic under those circumstances, which I think is fine. Your API design here meant that no Axum WS types had to be exposed anyway, so I think this will work fine.

Great! WASM compatibility was one of the areas I'm wasn't not sure about. I think now you should be able to implement the server trait yourself if you want to use websockets with wasm on the server

> My only significant feedback: The breakages in the `server_fns_axum` example suggest to me that migrating to this will require a bunch of changes like this:
> 
> ```rust
> // from this
> #[server(input = Postcard, output = Postcard)]
> // to this
> #[server(input = PostPostcard, output = PostPostcard)]
> ```
> 
> It would be nice if there were a way to do this in the macro automatically, given that the macro already has a default method (I think?)
> 
> Thanks for all your work.

I'm hesitant to hardcode too much into the server macro. Looking at that server function call if Postcard only implement `Encodes` and `Decodes`, I would assume `input` works for any struct that implements `Encodes<T> + Decodes<T>`. For now, I just renamed the encoding variants to `*Encoding` like `PostcardEncoding` so this is less breaking.

server_fn/src/codec/json.rs
```diff
@@ -1,221 +1,36 @@
...
-///
-/// ## Browser Support for Streaming Input
-///
-/// Browser fetch requests do not currently support full request duplexing, which
-/// means that that they do begin handling responses until the full request has been sent.
-/// This means that if you use a streaming input encoding, the input stream needs to
-/// end before the output will begin.
-///
-/// Streaming requests are only allowed over HTTP2 or HTTP3.
-pub struct JsonStream<T, E>(Pin<Box<dyn Stream<Item = Result<T, E>> + Send>>);
```
ealmloff, line 108: Note json streaming http support is removed entirely. The browser doesn't respect the http chunk boundary which makes decoding this unreliable: https://github.com/DioxusLabs/dioxus/issues/3695

gbj:
Thanks for the changes. I think this is in a good state, I've just been ticking off the last random CI things. Unless you have any other thoughts, I'm happy to merge once it's passing.

server_fn/src/codec/json.rs
gbj: Yeah, unfortunate but understood. The websockets will be a good replacement for that in any case, I think.

ealmloff:
> I've just been ticking off the last random CI things. Unless you have any other thoughts, I'm happy to merge once it's passing.

Thanks! Looks good to me. I'm planning on some follow up PRs to make the macro logic more modular in the next few weeks

PR #3658, opened 2025-02-27
Implement several into traits for store fields (0.8)

mahdi739:
This PR implements these traits:

- `IntoStyle`
- `IntoStyleValue`
- `IntoProperty`
- `Render`
- `RenderHtml`
- `AttributeValue`
- `InnerHtmlValue`

for these store fields:

- `Store`
- `ArcStore`
- `Field`
- `ArcField`
- `SubField`
- `AtKeyed`
- `KeyedSubField`
- `DerefedField`
- `AtIndex`

It also implements `IntoSplitSignal` for these traits:

- `AtKeyed`
- `AtIndex`
- `DerefedField`

I created a new macro rule that was flexible enough to handle the signals, so I replaced the previous macro with this new one for all signals.

I didn't implement `IntoClass` because it also had an implementation for `(&'static str, $sig<bool, S>)`, which my new macro wasn't flexible enough to handle. However, I thought that if you create something like `IntoStyleValue` for `IntoClass`, it can have a unified implementation similar to `IntoStyle`. So I waited for your response about this. (Or I can write a separate macro for this tuple.)

Another thing about `IntoStyleValue` is that you implemented it only for `Arc` signals, but in the new macro, it is implemented for all signals. I wasn't sure of the true meaning of this, but the compiler didn't complain, so I didn't touch it.

In the `mod.rs` (the implementation of `AttributeValue`, `RenderHtml`, `AddAnyAttr`, and `Render`) there was an argument for the current macros called `$dry_resolve`. I wasn't sure what value should I pass for the new store fields, so I passed `false` for them.

The implementation of `AddAnyAttr` for signals was empty and marked with `todo!()`. I also added this empty implementation to the new macros.

### DIFF
--- a/tachys/src/reactive_graph/bind.rs
+++ b/tachys/src/reactive_graph/bind.rs
@@ -14,17 +14,22 @@ use crate::{
     renderer::{types::Element, RemoveEventHandler},
     view::{Position, ToTemplate},
 };
-#[cfg(feature = "reactive_stores")]
-use reactive_graph::owner::Storage;
 use reactive_graph::{
     signal::{ReadSignal, RwSignal, WriteSignal},
     traits::{Get, Set},
     wrappers::read::Signal,
 };
-#[cfg(feature = "reactive_stores")]
-use reactive_stores::{ArcField, Field, KeyedSubfield, Subfield};
 use send_wrapper::SendWrapper;
 use wasm_bindgen::JsValue;
+#[cfg(feature = "reactive_stores")]
+use {
+    reactive_graph::owner::Storage,
+    reactive_stores::{
+        ArcField, AtIndex, AtKeyed, DerefedField, Field, KeyedSubfield,
+        StoreField, Subfield,
+    },
+    std::ops::{Deref, DerefMut, IndexMut},
+};
 
 /// `group` attribute used for radio inputs with `bind`.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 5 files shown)


gbj:
Thanks -- This all looks good to me! Sorry it took me a while to get around to reviewing.

mahdi739:
> Thanks -- This all looks good to me! Sorry it took me a while to get around to reviewing.
Thanks for checking.
I wrote a separate macro for tuples in `class.rs`. Should I commit it or wait for an `IntoClassValue`
Actually I looked more deeply in style.rs and to me it was impossible to do the same thing for `IntoClass` because in tuple values for class, the second field can only be bool.
But you may know about way to make them unify.

gbj:
I would suggest a separate macro for `class.rs`. `IntoStyleValue` exists to support multiple different types in the second value of the tuple: so you can have `style:font-weight="bold"` (which is `("font-weight", "bold")`) or you can have `style:font-weight=Some("bold")` (which is `(font-weight, Some("bold"))`), and that second value could also be a function or signal that returns that type . For classes, the tuple syntax only takes booleans (or functions/signals that return booleans) as the second value, so I don't think there's a need for an additional trait indirection. (It doesn't need to support both `bool` and `Option<bool>` for example.)

So if I'm reading you correctly, I'd suggest that implementing this for `(&'static str, {concrete store fields}<bool>)` with a macro is the right approach. Feel free to make a separate PR.

mahdi739:
> I would suggest a separate macro for `class.rs`. `IntoStyleValue` exists to support multiple different types in the second value of the tuple: so you can have `style:font-weight="bold"` (which is `("font-weight", "bold")`) or you can have `style:font-weight=Some("bold")` (which is `(font-weight, Some("bold"))`), and that second value could also be a function or signal that returns that type . For classes, the tuple syntax only takes booleans (or functions/signals that return booleans) as the second value, so I don't think there's a need for an additional trait indirection. (It doesn't need to support both `bool` and `Option<bool>` for example.)
> 
> So if I'm reading you correctly, I'd suggest that implementing this for `(&'static str, {concrete store fields}<bool>)` with a macro is the right approach. Feel free to make a separate PR.

Thanks. Ok so I created a separate PR for `IntoClass`: #3670
And to me this PR is complete. Tell me if I should change anything.

gbj:
Honestly I thought I'd merged it already when I left my comment above, which was why I suggested a separate PR for classes. Oops! Merged now.

PR #3683, opened 2025-03-07
wip: Update axum to 0.8

leo91000:
Update axum to 0.7


gbj:
This was already done in #3439 and is available on the `leptos_0.8` branch, because it requires us to release a new major version (so is being bundled with some other breaking changes).

PR #3692, opened 2025-03-08
`ImmediateEffect` follow up

QuartzLibrary:
Follow up to #3650.

### DIFF
--- a/reactive_graph/src/effect/immediate.rs
+++ b/reactive_graph/src/effect/immediate.rs
@@ -176,6 +176,20 @@ mod inner {
         run_done_max: usize,
         /// The [ThreadId] of the run with the highest id.
         /// Used to prevent over-subscribing during parallel execution with the last run.
+        ///
+        /// ```text
+        /// Thread 1:
+        /// -------------------------
+        ///   ---   ---    =======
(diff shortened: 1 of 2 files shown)


reactive_graph/src/effect/immediate.rs
```diff
@@ -214,6 +176,20 @@ mod inner {
         run_done_max: usize,
         /// The [ThreadId] of the run with the highest id.
         /// Used to prevent over-subscribing during parallel execution with the last run.
+        ///
+        /// ```
```
sabify, line 180: Please use `/// ```no_run` instead.

reactive_graph/src/effect/immediate.rs
```diff
@@ -214,6 +176,20 @@ mod inner {
         run_done_max: usize,
         /// The [ThreadId] of the run with the highest id.
         /// Used to prevent over-subscribing during parallel execution with the last run.
+        ///
+        /// ```no_run
```
sabify, line 180: Sorry for that! Please use `text` instead as it is not code. With `no_run` attribute, the compiler tries to compile the block.

PR #3700, opened 2025-03-11
fix(reactive_stores_macro): Make tuple struct field locator in `impl Patch` be `syn::Index` instead of `usize`

DanikVitek:
This change future-proofs the generated code, as syntax like `tuple.0usize`, `tuple.1usize` etc. was allowed by the compiler accidentally and is, therefore, temporary


gbj:
Thanks!

PR #3701, opened 2025-03-11
feat(reactive_stores): Replace `AsRef` bound of `StoreFieldIterator` blanket impl with `Len` bound

DanikVitek:
This bound was overly restrictive, as some collection types are not contiguous in memory, and therefore do not implement `AsRef<[T]>`. This change extracts the `len` method into a separate `pub trait Len`, lifting the restriction.


gbj:
Fine by me. I do think that this is a breaking change so I'm going to merge it into the 0.8 branch.

PR #3711, opened 2025-03-14
refactor: make `shell` parameter in `file_and_error_handler*` generic

tversteeg:
This is more consistent with the other Axum integration functions, which are also generic. This also allows passing a closure.

The reason I made this PR is that I'm trying to create a private library that wraps the Leptos Axum integration, but the function pointer is giving me a lot of trouble.


gbj:
Thanks!

PR #3715, opened 2025-03-15
Remove SendWrapper from the external interface of LocalResource

zakstucke:
Man this was hard to figure out.. 

The tricky part is the previous setup used `Option<SendWrapper<T>>`, but these means you couldn't deref to `&Option<T>` like you need to. This now internalises a `MaybeSendWrapper` in `ArcAsyncDerived` that means it's layed out as `SendWrapper<Option<T>>`, so can be dereffed. 

Had the side effect of neatening up the internals of both `LocalResource` and `ArcAsyncDerived`.

### DIFF
--- a/leptos_server/src/local_resource.rs
+++ b/leptos_server/src/local_resource.rs
@@ -7,16 +7,16 @@ use reactive_graph::{
         AnySource, AnySubscriber, ReactiveNode, Source, Subscriber,
         ToAnySource, ToAnySubscriber,
     },
-    owner::use_context,
+    maybe_send_wrapper::MaybeSendWrapper,
+    owner::{use_context, LocalStorage},
     signal::{
-        guards::{AsyncPlain, ReadGuard},
+        guards::{AsyncPlain, Mapped, ReadGuard},
         ArcRwSignal, RwSignal,
     },
     traits::{
         DefinedAt, IsDisposed, ReadUntracked, Track, Update, With, Write,
     },
 };
-use send_wrapper::SendWrapper;
 use std::{
     future::{pending, Future, IntoFuture},
     panic::Location,
@@ -24,7 +24,7 @@ use std::{
 
 /// A reference-counted resource that only loads its data locally on the client.
 pub struct ArcLocalResource<T> {
-    data: ArcAsyncDerived<SendWrapper<T>>,
+    data: ArcAsyncDerived<T>,
     refetch: ArcRwSignal<usize>,
     #[cfg(any(debug_assertions, leptos_debuginfo))]
     defined_at: &'static Location<'static>,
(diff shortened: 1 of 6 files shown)


gbj:
The problem with `SendWrapper<Option<T>>` is that it actually constructs a `SendWrapper(None)` on the server, which will cause panics under load for Axum/anything using the Tokio multithreaded runtime.

If you want a simple test case:
```rust
#[server]
pub async fn does_something() -> Result<i32, ServerFnError> {
    tokio::time::sleep(std::time::Duration::from_millis(125)).await;
    Ok(42)
}

#[component]
    let takes_time = Resource::new(|| (), |_| does_something());
    let local = LocalResource::new(|| async { "hello!" });
    view! {
        <Suspense>
            {move || takes_time.get()}
            {move || local.get().map(|n| n.to_string())}
        </Suspense>
    }
}
```
testing this with something like
```sh
wrk -t8 -c16 -d10s http://localhost:3000
```
will yield a bunch of panics as the `SendWrapper(None)`s created on the server are dropped. Whereas `Option<SendWrappr>` never creates the wrapper so there's no problem on the `leptos_0.8` branch currently.

(I am open to better solutions than `SendWrapper` as it has proven quite annoying to navigate, but it's the least bad option I've seen so far.)

zakstucke:
@gbj hmm I can't reproduce this, using that snippet with counter_isomorphic yields no panics when running `wrk -t8 -c16 -d10s http://localhost:3000`, and that command just outputs:
```
Running 10s test @ http://localhost:3000
  8 threads and 16 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   128.05ms  825.60us 132.04ms   72.76%
    Req/Sec    16.51      4.79    20.00     69.78%
  1248 requests in 10.08s, 20.76MB read
Requests/sec:    123.75
Transfer/sec:      2.06MB
```

Am I doing something wrong? 

In any case, I think I can workaround this concern in the same way by just rewrapping as `Option<SendWrapper<Option<T>>>` where the inner Option is always unwrapped.

gbj:
`counter_isomorphic` uses Actix, which pins requests/responses to a single thread in any case: try it with an Axum example (`ssr_modes_axum` etc.)

zakstucke:
This should fix it, the real issue being repro'd was the change to `new_mock()` for `AsyncDerived` to use `AsyncDerived<T, LocalStorage>` rather than `AsyncDerived<T>`. 

I've fixed that, and updated `MaybeSendWrapper` -> `MaybeSendWrapperOption` that'll make sure not to construct a `SendWrapper` in the `None` case.

zakstucke:
Huh CI seems to have suddenly exploded from something unrelated

zakstucke:
@gbj @benwis looks like this might be quite important to deal with ASAP, it looks like env secrets from this repro could potentially be comprimised.

zakstucke:
There's a soundness hole in the `DerefMut` implementation, seeing if there's a fix.

benwis:
> @gbj @benwis looks like this might be quite important to deal with ASAP, it looks like env secrets from this repro could potentially be compromised.

The good news here is we have no env values or secrets for this or cargo-leptos, so we're all set

gbj:
Looks good! I just added a commit with a safety comment for the Send/Sync impl here, but it all looks right to me. 

I think I managed to merge in one of the earlier commits already accidentally when trying to merge in changes from some other branch (idk), so apologies for any git messiness.

zakstucke:
Awesome, thanks! I think that was me with that extra commit, I must've branched incorrectly

PR #3716, opened 2025-03-15
view!{} macro optimisation: don't wrap string types in closures when passing to ToChildren

zakstucke:
This optimisation improves thaw demo compile time in erased mode by 25% and non-erased mode by 10%, and shouldn't be a breaking change. Albeit this is probably a larger improvement than for the average project. For thaw, this optimises ~2500 unique string nodes.

It seems the unique closures for each component's children prop, when passing to `ToChildren::to_children(move || #children)` is a significant bottleneck for the compiler. 

We can identify strings inside the `view!{}` macro, so we can provide an optimised path that doesn't add the closure. 

It does bring up the idea of replacing all `FnOnce()` children types with this, assuming `AnyView` could be made `Clone`, but I don't see how this would be possible without a massive breaking change, so I'm not looking into it.

### DIFF
--- a/leptos/src/children.rs
+++ b/leptos/src/children.rs
@@ -91,6 +91,9 @@ pub trait ToChildren<F> {
     fn to_children(f: F) -> Self;
 }
 
+/// Compiler optimisation, can be used with certain type to avoid unique closures in the view!{} macro.
+pub struct ChildrenOptContainer<T>(pub T);
+
 impl<F, C> ToChildren<F> for Children
 where
     F: FnOnce() -> C + Send + 'static,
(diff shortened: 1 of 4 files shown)


gbj:
Awesome! Thanks, Zak, this is another great PR. I'm going to merge it into 0.8 because while it won't be a breaking change for normal use, it is technically semver-breaking for... weird use. For example, it breaks this (weird) code:
```rust
#[component]
    view! {
        <BreakingChange>"foo"</BreakingChange>
    }
}

pub struct WeirdChildren(&'static str);

impl<F> ToChildren<F> for WeirdChildren
where
    F: Fn() -> &'static str,
{
    fn to_children(f: F) -> Self {
        WeirdChildren(f())
    }
}

#[component]
pub fn BreakingChange(children: WeirdChildren) -> impl IntoView {
    children.0
}
```

Issue #3719, opened 2025-03-16
Unwanted linear growth of extraneous serverfn API calls in direct navigation to route with `ParamSegment`


metatoaster:
**Describe the bug**
If user opens a link in the application referencing some component on some `Route` (without going through any sibling `Route`s) that contains a `ParamSegment`, where that derived `Params` is hooked to a resource that invokes a server function, there will be a linear growth of extraneous serverfn API calls that will happen as the user invokes that affected resource.  The following is a version that shows it most unambiguously (directly in CSR, without involving hydration, even though I do all of my testing via `axum` or `actix`):

<details>
<summary>Expand for the omitted preamble:</summary>

```rust
use leptos::prelude::*;
use leptos_meta::{MetaTags, *};
use leptos_router::{
    components::{ParentRoute, Route, Router, Routes, A},
    hooks::use_params,
    nested_router::Outlet,
    params::Params,
    path, ParamSegment, SsrMode, StaticSegment,
};

pub fn shell(options: LeptosOptions) -> impl IntoView {
    view! {
        <!DOCTYPE html>
        <html lang="en">
            <head>
                <AutoReload options=options.clone()/>
                <HydrationScripts options/>
                <MetaTags/>
            </head>
            <body>
                <App/>
            </body>
        </html>
    }
}

#[server]
async fn foo_ident(arg: String) -> Result<String, ServerFnError> {
    Ok(arg)
}

#[component]
fn HomePage() -> impl IntoView {
    view! {
        <h1>"Home Page"</h1>
    }
}

#[derive(serde::Serialize, serde::Deserialize, PartialEq, Clone, Debug)]
struct FooResult(String);

#[component]
fn Foo() -> impl IntoView {
    view! {
        <h1>"Foo"</h1>
        <Outlet/>
    }
}
```
</details>

```rust
#[component]
pub fn App() -> impl IntoView {
    provide_meta_context();
    let fallback = || view! { "Page not found." }.into_view();
    view! {
        <Title text="Demo"/>
        <Router>
            <main>
                <nav>
                    <ul>
                        <li><A href="/">"Root"</A></li>
                        // <li><A href="/foo1">"Foo 1"</A></li>
                        // <li><A href="/foo2">"Foo 2 "</A></li>
                        // <li><A href="/foo3">"Foo 3"</A></li>
                        <li><A href="/foo">"Foo"</A></li>
                        <li><A href="/foo/1">"Foo 1"</A></li>
[...]

Issue #3699, opened 2025-03-11
Hydration issue caused by `Resource` provided via `expect_context` inside a component situated before `Routes`


metatoaster:
**Describe the bug**

Given the following `App` and the resource to be provided (encapsulated in the newtype `Ctx`):

```rust
#[derive(Clone, Debug)]
struct Ctx(Option<Resource<Result<String, ServerFnError>>>);

#[component]
pub fn App() -> impl IntoView {
    provide_meta_context();
    let (rs, ws) = signal(Ctx(None));
    provide_context(rs);
    provide_context(ws);
    let fallback = || view! { "Page not found." }.into_view();
    view! {
        <Title text="Demo"/>
        <Router>
            <main>
                <CtxView/>
                <Routes fallback>
                    <Route path=path!("") view=HomePage ssr=SsrMode::Async/>
                    <Route path=path!("/foo") view=Foo ssr=SsrMode::Async/>
                    <Route path=path!("/bar") view=Bar ssr=SsrMode::Async/>
                </Routes>
                // If CtxView is here instead above Routes, there would be no issues with hydration
                // <CtxView/>
            </main>
        </Router>
    }
}
```

Where the minimal implementation for `CtxView` that would `expect_context` on `Ctx` may look like:

```rust
#[component]
fn CtxView() -> impl IntoView {
    let rs = expect_context::<ReadSignal<Ctx>>();
    view! {
        <Transition>{
            move || Suspend::new(async move {
                let ctx = rs.get();
                leptos::logging::log!("ctx = {ctx:?}");
                if let Some(resource) = ctx.0 {
                    let value = resource.await?;
                    leptos::logging::log!("returning actual view");
                    Ok::<_, ServerFnError>(Some( 
                        view! { <div>"The value is: "{value}</div> } 
                    )) 
                } else {
                    leptos::logging::log!("returning empty view");
                    Ok(None) 
                }
            })
        }</Transition>
    }
}
```

`Foo` and `Bar` are otherwise identical saved for the name (this is for testing CSR navigation), they look like:

```rust
#[component]
fn Bar() -> impl IntoView {
    let set_ctx = expect_context::<WriteSignal<Ctx>>();

    on_cleanup(move || {
        leptos::logging::log!("Running on_cleanup in Bar");
        Effect::new(move || {
            leptos::logging::log!("set_ctx with None in Effect of Bar on_cleanup");
            set_ctx.set(Ctx(None));
        });
    });

    let hook = move || {
[...]

PR #3720, opened 2025-03-16
test: regression from #3502

metatoaster:
Also as reported in #3719, which has an actual minimum example.

To provided additional detail that I missed there (even though I mentioned this briefly in #3699), these tests in this pull request should have flagged #3502 which the cdee2a9476109b10f5d8da268684bd1257c6ce9b being the first source of the regression, and that this following patch is as close to the source of the issue as I can find:

```patch
diff --git a/router/src/nested_router.rs b/router/src/nested_router.rs
index 27269b3c..3fa6ed0e 100644
--- a/router/src/nested_router.rs
+++ b/router/src/nested_router.rs
@@ -655,8 +655,8 @@ where
                                         ScopedFuture::new(view.choose()),
                                     );
                                     let view = view.await;
-                                    let view =
-                                        MatchedRoute(matched.0.get(), view);
+                                    // let view =
+                                    //     MatchedRoute(matched.0.get(), view);
                                     OwnedView::new(view).into_any()
                                 })
                                     as Pin<
```

Obviously that's not the fix as the router in islands mode probably depends on that, I thought I should at least narrow down the source of this regression to as close as possible.

I am setting the merge base to `main` as that probably will have the immediate benefit, and that these tests should hopefully pass as I feel getting a pass from this first from CI and then pulling the resulting `main` into `leptos_0.8` to see the test failures may be better from a provenance standpoint. (Also maybe the final merge can be a 3-way merge sourced from `main` (with these tests), `leptos_0.8` plus the pull request branch that will provide the fix, if we want to make this fancy, I don't mind either way.)

### DIFF
--- a/examples/suspense_tests/e2e/features/check_instrumented_issue_3719.feature
+++ b/examples/suspense_tests/e2e/features/check_instrumented_issue_3719.feature
@@ -1,0 +1,99 @@
+@check_instrumented_issue_3719
+Feature: Using instrumented counters to test regression from #3502.
+    Check that the suspend/suspense and the underlying resources are
+    called with the expected number of times.  If this was already in
+    place by #3502 (5c43c18) it should have caught this regression.
+    For a better minimum demonstration see #3719.
+
+    Background:
+
+        Given I see the app
+        And I select the mode Instrumented
+
+    Scenario: follow all paths via CSR avoids #3502
+        Given I select the following links
+            | Item Listing         |
+            | Item 1               |
+            | Inspect path2        |
+            | Inspect path2/field3 |
+        And I click on Reset CSR Counters
+        When I select the following links
+            | Inspect path2/field1 |
+            | Inspect path2/field2 |
+        And I go check the Counters
+        Then I see the following counters under section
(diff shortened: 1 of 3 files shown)


metatoaster:
Took the opportunity to fix a silly mistake from when I first put this in because I hadn't learned the lesson about using context at components and move into the callbacks, and not using contexts directly in the callbacks (like `on_cleanup`).

I also changed the way the tests are spawned by cucumber to load individual features with their sets of scenarios one at a time, rather than letting cucumber running all 57 scenarios from all the features (as of now) all in parallel which necessitates all those retries which may or may not complete correctly.

Also, given that this will get merged (or rebased) on top of `leptos_0.8`, I included a chore commit which normalized the version in `main` branch with the one in `leptos_0.8`.

gbj:
Sure — LGTM and this does pass. Thanks for narrowing down the commit related to the issue, this is extremely helpful. Will merge this into the main branch, make a merge commit from main into 0.8, and then fix on the 0.8 branch of course as that's the only place where it's broken.

PR #3731, opened 2025-03-19
docs(`Effect::watch`): refer to `dependency_fn` and `handler` args

jmevel:
`deps` and `callback` were argument names used in the deprecated `watch` function


gbj:
Thanks!

PR #3733, opened 2025-03-20
feat: add patch http method codec

ChosunOne:
# Description
Adds `PATCH` method support for `#[server]` functions via a `Patch` codec.  Includes a note about potential browser support issues when JS/WASM is not enabled.  Useful when defining `REST` apis via `#[server]` functions.  

# Why do we need this?
Per the current Leptos documentation:

> Isomorphic: Leptos provides primitives to write isomorphic [server functions](https://docs.rs/leptos_server/0.2.5/leptos_server/index.html), i.e., functions that can be called with the “same shape” on the client or server, but only run on the server. This means you can write your server-only logic (database requests, authentication etc.) alongside the client-side components that will consume it, and call server functions as if they were running in the browser, without needing to create and maintain a separate REST or other API.
Web: Leptos is built on the Web platform and Web standards. The [router](https://docs.rs/leptos_router/latest/leptos_router/) is designed to use Web fundamentals (like links and forms) and build on top of them rather than trying to replace them.

However, in the current implementation you *must* maintain a separate REST API if you want to define your server functions with `#[server]` *and* provide a RESTful API.  It also encourages non-Web standard practices such as using `POST` for requests that are clearly not suited for that method, such as fetching data or updating existing resources.  Since this imposes duplication on the backend public API and does not adhere to web standards, we should provide tools so that the use of `#[server]` functions can leverage existing RESTful apis.

## What about graceful degradation?
A concern with supporting `PATCH`, `PUT`, `DELETE`, etc. for `#[server]` functions is their limited support in the case the browser does not have JS/WASM enabled.  This is a real concern, but the current library's choice not to provide these methods is severely limiting on the backend side, forcing duplication of API endpoints and not allowing the backend to adhere to web standards for defining its public APIs.  Ultimately, it should be up to the developer to decide if they need their functionality to require working without JS/WASM.  The defaults are not changed, so existing applications are not affected, and a note has been provided that use of `PATCH` on a `#[server]` function may require JS/WASM to be enabled to function.

### DIFF
--- a/server_fn/src/codec/cbor.rs
+++ b/server_fn/src/codec/cbor.rs
@@ -1,3 +1,4 @@
+use super::Patch;
 use super::Post;
 use crate::{ContentType, Decodes, Encodes};
 use bytes::Bytes;
@@ -36,3 +37,5 @@ where
 
 /// Pass arguments and receive responses using `cbor` in a `POST` request.
 pub type Cbor = Post<CborEncoding>;
+/// Pass arguments and receive responses using `cbor` in a `PATCH` request.
+pub type PatchCbor = Patch<CborEncoding>;
--- a/server_fn/src/codec/cbor.rs
+++ b/server_fn/src/codec/cbor.rs
@@ -1,5 +1,4 @@
-use super::Patch;
-use super::Post;
+use super::{Patch, Post};
 use crate::{ContentType, Decodes, Encodes};
 use bytes::Bytes;
 use serde::{de::DeserializeOwned, Serialize};
--- a/server_fn/src/codec/json.rs
+++ b/server_fn/src/codec/json.rs
@@ -34,4 +34,5 @@ where
 
 /// Pass arguments and receive responses as JSON in the body of a `POST` request.
 pub type Json = Post<JsonEncoding>;
+/// Pass arguments and receive responses as JSON in the body of a `PATCH` request.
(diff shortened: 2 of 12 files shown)


gbj (commented):
Thanks for taking the time to work on this and make a PR! I think it's a perfectly reasonable use case, and it will be a nice way to answer the question of other HTTP verbs/methods in the future.

As far implementation: It looks to me like `try_new_post` and `try_new_patch` are identical except for the `post()` vs `patch()` method calls in their respective client libraries. This makes sense. However, it occurs to me that this will cause the relevant traits to get bigger and bigger with mostly duplicate code, as more methods like `DELETE` are added.

I would suggest refactoring so that there are `try_new_req()`, `try_new_req_bytes()`, and `try_new_req_form_data()` (or whatever names) which all take `method: Method`, and then either adapt the `Post` and `Patch` encoding methods to use those, or do default implementations of `try_new_post()` and so on that call out to `try_new_req()` etc... Did I explain that clearly enough?

(Also note: I will be merging the 0.8 branch into main in a few minutes, so this PR can be retargeted there in the near future.)

ChosunOne:
Makes total sense! I didn't want to cause any breakages with existing setups but when I get back from vacation next week I'll open a larger refactoring pr to avoid duplication. Thanks for the feedback!

PR #3735, opened 2025-03-20
fix(CI): free-up disk, properly gate nightly feature and pre-install deps

sabify:


benwis:
This one is interesting, I'm curious what drove the creation of the rustc-nightly flag and build.rs for a bunch of the flags

sabify:
When running `cargo clippy --all-features` on stable rustc (which is part of the workflow), it will enable "nightly" features and it will enable nightly only rustc features which will be a hard error with `-D warning`. I've added build scripts and a custom flag to automatically detect nightly rustc and then match it with "nightly" features, so both should be true for the nightly only features to be enabled. 

This approach also reduces user errors if one of the conditions weren't satisfied. Users should use both `cargo +nightly` and the `"nightly"` feature to enable nightly-only features.

sabify:
The other thing is that I limited the number of feature combinations to 2 for speeding up CI workflow builds and tests with `cargo-all-features`. I couldn't find any feature combination more than 2 in the codebase. Mostly one feature is enabled, and in a few places (I think 1-2) there were two features enabled at once, which will be covered by this configuration.

PR #3739, opened 2025-03-20
fix(CI): cancel in-group inflight and pending jobs on new pushes in pull requests

sabify:


benwis:
Another that makes sense to me, thanks!

PR #3741, opened 2025-03-20
ArcLocalResource fix (0.8)

zakstucke:
`ArcLocalResource` doesn't use the `new_mock` variant in `ssr` like `LocalResource`, which breaks it before this fix.


benwis:
Yeah that is an oversight, to be sure

Issue #3742, opened 2025-03-20
Post fixing CI in PR #3537 issues


sabify:
@gbj @benwis @zakstucke; After merging #3735, some shared errors are surfacing, can you please look at the errors at https://github.com/leptos-rs/leptos/actions/runs/13974404883/job/39133322985?

sabify:
A fix need to be provided for `lepto_macro` in stable rustc without nightly rustc features.

In `leptos_macro` directory, run `cargo +stable test-all-features`

gbj:
Thanks. I've also just merged `leptos_0.8` into `main` (and created a new `leptos_0.7` branch that was the 0.7 version of main), so I expect there are a bunch of CI issues on main right now which we can chip away it. I'm logging off for the day soon but will take a look over the weekend.

Apologies to GitHub for demolishing your servers, we love you :tada:

PR #3747, opened 2025-03-21
chore: no need to filter out "nightly" feature as of #3735

sabify:


benwis:
@sabify There are almost too many CI tickets. Is this one ready? I see the issues in the examples, although I can't say I understand why

sabify:
@benwis The macro should be fixed. Please look at #3742. Everything here is OK.

PR #3750, opened 2025-03-21
feat: switch `extract()` helper to use `ServerFnErrorErr` (closes #3745)

### LINKED ISSUE #3745: `leptos_axum::extract` returning `ServerFnError` makes using it unergonomic in server functions in 0.8
Because `leptos_axum::extract` returns `ServerFnError`, there's no obvious way to use it in a server function returning a custom error without first converting it manually to a `ServerFnErrorErr` using `match` between each variant and then using `.into_app_error()` or `CustomError::from_server_fn_error` on that.

I intend to produce a PR for this in the near future, unless somebody beats me to it, but feel free to also let me know if this is not something you want "fixed" if I'm just misunderstanding something.

ilyvion:
This provides much better compatibility with server functions using custom errors while still working as before with server functions using `ServerFnError`.

This can be verified with the `server_fns_axum` which uses the `extract()` function and a `ServerFnError` without needing to be edited in any way, and I've also tested it against my own server functions that use a custom error and that also works as expected with the `?` operator.

Resolves #3745

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -41,9 +41,8 @@ use leptos_router::{
 use once_cell::sync::Lazy;
 use parking_lot::RwLock;
 use send_wrapper::SendWrapper;
-use server_fn::{
-    redirect::REDIRECT_HEADER, request::actix::ActixRequest, ServerFnError,
-};
+use server_fn::error::ServerFnErrorErr;
+use server_fn::{redirect::REDIRECT_HEADER, request::actix::ActixRequest};
 use std::{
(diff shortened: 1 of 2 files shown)


ilyvion:
There were a bunch of failures of unexpected outputs for something UI related in the CI, which I don't understand as I didn't touch any UI stuff.

gbj:
Yep thanks -- This is clearly a good idea. Don't worry about whatever was going on with CI, I'm about to merge a bunch of things, leave things in a bit of mess, and then clean it up :smile:

PR #3755, opened 2025-03-22
feat: support aliased `Result` return types for `server_fn`

ifiokjr:
### Description

Add support for aliased result types to the `server_fn` macro. 

```rust
use leptos::prelude::*;

type AliasedResult = Result<String, ServerFnError>;

#[server]
pub async fn full_alias_result() -> AliasedResult {
    Ok("full".to_string())
}
```

Partial aliases can also be used. 

```rust
use leptos::prelude::*;

type PartialResult<T> = Result<T, ServerFnError>;

#[server]
pub async fn partial_alias_result() -> PartialResult<String> {
    Ok("partial".to_string())
}
```

This also supports custom error types added in `0.8.0-alpha`. 

- Enhanced error messages for better clarity when server functions do not meet the expected return type requirements.
- Added `trybuild` tests to ensure behaviour is not broken in future releases.
- Add `ServerFnErrorErr` and `FromServerFnError` to prelude.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3279,6 +3279,7 @@ dependencies = [
  "tokio-tungstenite",
  "tower",
  "tower-layer",
+ "trybuild",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3267,6 +3267,7 @@ dependencies = [
  "reqwest",
  "rkyv",
  "rmp-serde",
+ "rustversion",
  "send_wrapper",
  "serde",
  "serde-lite",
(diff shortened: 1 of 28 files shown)


ifiokjr:
@gbj my previous (smaller) PR was auto closed (I think) when deleting the `leptos_0.8` branch. 

https://github.com/leptos-rs/leptos/pull/3744

tetuaoro:
Hi, thank you for this pr. I also use alias types for my apps but `#[server]` does not support that.

gbj:
Thanks! This looks good overall. There are a few other server fn PRs to merge so I will clean up any outstanding issues in the CI when that's done.

PR #3758, opened 2025-03-22
fix(CI): optimize CI workflow

sabify:
We are currently wasting too much resource by using `cargo-cmake` to build and test against workspace in each member/examples directory (wrongly!). I'm trying to fix this in this PR so that each test is isolated to its corresponding package. This should significantly improve performance for all workflows.

It will also tries to test all members and examples against stable and nightly rustc and erase mode separately.


sabify:
This PR is ready. @benwis/@gbj Please merge this to save CI time by several orders of magnitude!

You can safely ignore the current CI status in this PR. It doesn't reflect the changes.

benwis:
Thanks so much for keeping at this

sabify:
This will take effect on new pull requests.

PR #3761, opened 2025-03-22
Re-export the codee crate

zakstucke:
Prevents having to keep `codee` and `leptos` in sync because `codee` has a different versioning scheme. Used in common case of changing encoding for resources, so makes sense to be exported.


zakstucke:
Might be worth patching 0.7 with this too

PR #3766, opened 2025-03-22
feat: allow using different error types for req/resp with WebSockets, closes #3724

### LINKED ISSUE #3724: 0.8.0-alpha: Decouple websocket protocol error types

The current generic bounds for `Websocket` protocol in `0.8.0-alpha` enforce usage of the same error type for all three errors: in the request stream, initial response and response stream. It makes it pretty inconvenient to use with the better error handling introduced in the same alpha.
```rust
    Server: crate::Server<E>,
    E: FromServerFnError + Send,
    Client: crate::Client<E>,
```

Define 3 different generic error types for `Websocket` protocol all bound by `FromServerFnError + Send`.

Quickly trying to implement myself to make sure it is possible, but it seems like it requires changes to `Server` and `Client` traits as well.

**Additional context**
...

myypo:
Hopefully it makes sense and I haven't messed up the types anywhere, among other things.

Here is a random example:
```rust
use leptos::{prelude::*, server_fn::{codec::JsonEncoding, BoxedStream, Websocket}, task::spawn_local};
use leptos_meta::MetaTags;
use server_fn::error::FromServerFnError;

    view! {
        <!DOCTYPE html> 
        <html lang="en">
            <head>
                <meta charset="utf-8" />
                <meta name="viewport" content="width=device-width, initial-scale=1" />
                <HydrationScripts options />
                <MetaTags />
            </head>
            <body>
                <App />
            </body>
        </html>
    }
}

#[derive(Serialize, Deserialize)]
struct Request {
    name: String
}

#[derive(Serialize, Deserialize)]
enum Response {
    Scream(String),
    Whisper(String)
}

#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
enum InputError {
    Test,
}

impl FromServerFnError for InputError {
    fn from_server_fn_error(_: server_fn::error::ServerFnErrorErr) -> Self {
        Self::Test
    }
}

#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
enum OutputError {
    Test,
}

impl FromServerFnError for OutputError {
    fn from_server_fn_error(_: server_fn::error::ServerFnErrorErr) -> Self {
        Self::Test
    }
}

#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
enum ConnectionError {
    Test,
}

impl FromServerFnError for ConnectionError {
    fn from_server_fn_error(_: server_fn::error::ServerFnErrorErr) -> Self {
        Self::Test
    }
}

#[server(protocol = Websocket<JsonEncoding, JsonEncoding>)]
async fn echo_websocket(
    input: BoxedStream<Request, InputError>,
) -> Result<BoxedStream<Response, OutputError>, ConnectionError> {
    use futures::channel::mpsc;
    use futures::{SinkExt, StreamExt};
    let mut input = input; 

    let (mut tx, rx) = mpsc::channel(1);

    if  5 < 3 {
        return Err(ConnectionError::Test);
    }

    tokio::spawn(async move {
        while let Some(msg) = input.next().await {
            match msg {
                Ok(msg) =>  {
            let msg = Response::Scream(msg.name.to_ascii_uppercase());
            tx.send(Ok(msg)).await.unwrap();
                },
                Err(_) => tx.send(Err(OutputError::Test)).await.unwrap(),
            }

        }
    });

    Ok(rx.into())
}

#[component]
    use futures::channel::mpsc;
    use futures::StreamExt;
    let (mut tx, rx) = mpsc::channel(1);
    let (latest,set_latest) = signal(None);

    if cfg!(feature = "hydrate") {
        spawn_local(async move {
            match echo_websocket(rx.into()).await {
                Ok(mut messages) => {
                    while let Some(Ok(msg)) = messages.next().await {
                        set_latest(Some(match msg {
                            Response::Scream(msg) => msg,
                            Response::Whisper(msg) => msg,
                        }));
                    }
                }
                Err(e) => leptos::logging::warn!("{:?}", e),
            }
        });
    }

    view! {
        <input
            type="text"
            on:input:target=move |ev| {
                tx.try_send(
                        Ok(Request {
                            name: ev.target().value(),
                        }),
                    )
                    .unwrap();
            }
        />
        <p>{latest}</p>
    }
}
```


leptos/src/form.rs
```diff
@@ -97,6 +97,7 @@ where
     ServFn: Send + Sync + 'static,
     ServFn::Output: Send + Sync + 'static,
     ServFn::Error: Send + Sync + 'static,
+    <ServFn as ServerFn>::Client: Client<<ServFn as ServerFn>::Error>,
```
myypo, line 100: These seem to be necessary.

benwis:
@myypo I'd be curious on what your thoughts are of @sabify's PR #3811

benwis:
@myypo I'm curious if you've considered unifying your error types into one, or creating a nested error structure with enums? It seems to me like ConnectionError, InputError, and OutputError could all be in one like this
```rust
struct AppError{
ConnectionError(ConnectionError),
InputError(InputError),
OutputError(OutputError),
}
```

Unless I'm missing something, wouldn't that work just as well without having to add more websockets error types?

myypo:
>  I'm curious if you've considered unifying your error types into one, or creating a nested error structure with enums?

Yes, I have considered nesting. This is a workaround that works in the current version. The downside of this approach is that you end up having a redundant pattern matching everywhere (or creating helper functions) to extract the expected error type and handle the unlikely matching failure.

Unifying, is, in my opinion, worse, because most variants make sense only in one of the three cases.

This is rather about ergonomics than making possible something that was not before.

> @myypo I'd be curious on what your thoughts are of @sabify's PR https://github.com/leptos-rs/leptos/pull/3811

Unless I am missing something obvious, it seems to me like the encoding changes introduced in this PR do not solve the mentioned ergonomics issue. I don't really see how this could be solved without additional generics. 

Obviously, a problem with this addition is that it changes lots of traits. I am not a good judge on how bad these rippling changes are and whether the ergonomics for users justify them, because I am not as familiar with the codebase.

benwis:
Thanks @myypo for the PR, I think it's reasonable to allow different error types for the request v the response. I'm hoping to release 0.8 proper soon with this included

PR #3782, opened 2025-03-24
fix(CI): run only the exact examples on the only examples change

sabify:
And removed duplicated workflow.


sabify:
This PR is ready.

sabify:
@benwis No, this won't change any main workflow and it is limited to only changed examples. If only the examples change, it will find the exact examples that changed and then run CI against them without running other unchanged examples, as examples are isolated, making it much more efficient.

Here is the workflow:
- If any leptos members change, CI runs against all leptos members and examples.
- If any leptos members and examples change, the same as above and CI runs against all leptos members and examples.
- If any examples change, CI runs only against the changed examples, not all of them.

sabify:
@benwis Please check the run workflow in this PR. I changed only one example and it ran just against the same example and not anymore. Previously, it was running against all examples which was irrelevant.

Issue #3780, opened 2025-03-23
Portal example test issue


sabify:
@gbj / @benwis Can you please look at the following failed test in portal example?

https://github.com/leptos-rs/leptos/actions/runs/14020826630/job/39253039071

After this and #3779 we would have a green CI finally.

benwis:
I wonder if this is locally reproducible

PR #3785, opened 2025-03-25
fix: portal example

sabify:
Closes: #3780

### DIFF
--- a/examples/cargo-make/wasm-test.toml
+++ b/examples/cargo-make/wasm-test.toml
@@ -1,7 +1,3 @@
-[tasks.test]
-env = { RUN_CARGO_TEST = false }
-condition = { env_true = ["RUN_CARGO_TEST"] }
-
 [tasks.post-test]
 dependencies = ["test-wasm"]
 
--- a/examples/cargo-make/wasm-test.toml
+++ b/examples/cargo-make/wasm-test.toml
@@ -1,6 +1,6 @@
-[tasks.test]
-env = { RUN_CARGO_TEST = false }
-condition = { env_true = ["RUN_CARGO_TEST"] }
+# [tasks.test]
+# env = { RUN_CARGO_TEST = false }
+# condition = { env_true = ["RUN_CARGO_TEST"] }
 
 [tasks.post-test]
 dependencies = ["test-wasm"]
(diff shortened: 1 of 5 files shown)


benwis:
I'm not sure we want opt level s for the dev profile

gbj:
Note: I'm pretty sure the difference between the expected and actual values here only come from the difference in the whitespace included in raw text nodes (capturing the full whitespace being supported only by nightly, I think). As a result, it should be equally possible to fix the tests for this purpose by simply quoting the relevant text nodes. I'm open to feedback on which of these two approaches people think is better.

```diff
diff --git a/examples/portal/src/lib.rs b/examples/portal/src/lib.rs
index d5bfc2689..6e6bb9b0d 100644
--- a/examples/portal/src/lib.rs
+++ b/examples/portal/src/lib.rs
@@ -8,7 +8,7 @@ pub fn App() -> impl IntoView {
     view! {
         <div>
             <button id="btn-show" on:click=move |_| set_show_overlay.set(true)>
-                Show Overlay
+                "Show Overlay"
             </button>
 
             <Show when=move || show_overlay.get() fallback=|| ()>
@@ -17,14 +17,14 @@ pub fn App() -> impl IntoView {
                     <div style="position: fixed; z-index: 10; width: 100vw; height: 100vh; top: 0; left: 0; background: rgba(0, 0, 0, 0.8); color: white;">
                         <p>This is in the body element</p>
                         <button id="btn-hide" on:click=move |_| set_show_overlay.set(false)>
-                            Close Overlay
+                            "Close Overlay"
                         </button>
                         <button id="btn-toggle" on:click=move |_| set_show_inside_overlay.set(!show_inside_overlay.get())>
-                            Toggle inner
+                            "Toggle inner"
                         </button>
 
                         <Show when=move || show_inside_overlay.get() fallback=|| view! { "Hidden" }>
-                            Visible
+                            "Visible"
                         </Show>
                     </div>
                 </Portal>
```

sabify:
I think having a simple minifier would be somehow a bullet proof over future changes. Please let me know if you want me to apply your proposed changes.

gbj:
Fair enough. I think this is fine, thanks for the fix!

PR #3787, opened 2025-03-26
fix(examples): fix compilation errors in examples

bioinformatist:
Hi Greg,

The examples `hacknews` and `hacknews_axum`, who can be used within both `csr` and `ssr` mode, raise error during `trunk serve` with latest `cargo`:

```
2025-03-25T09:12:13.675539Z  INFO 🚀 Starting trunk 0.21.9
2025-03-25T09:12:14.274171Z  INFO 📦 starting build
warning: output filename collision.
The bin target `hackernews` in package `hackernews v0.1.0 (/home/ysun/leptos/examples/hackernews)` has the same output filename as the lib target `hackernews` in package `hackernews v0.1.0 (/home/ysun/leptos/examples/hackernews)`.
Colliding filename is: /home/ysun/leptos/examples/hackernews/target/wasm32-unknown-unknown/debug/hackernews.wasm
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The bin target `hackernews` in package `hackernews v0.1.0 (/home/ysun/leptos/examples/hackernews)` has the same output filename as the lib target `hackernews` in package `hackernews v0.1.0 (/home/ysun/leptos/examples/hackernews)`.
Colliding filename is: /home/ysun/leptos/examples/hackernews/target/wasm32-unknown-unknown/debug/hackernews.wasm.dwp
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.

(Compilation details omitted...)

Caused by:
    0: HTML build pipeline failed (1 errors), showing first
    1: error from asset pipeline
    2: running cargo build
    3: found more than one target artifact: ["hackernews", "hackernews"]:
        * consider adding `<link data-trunk rel="rust" data-bin={bin} />` to the index.html to build only the specified binary
        * or adding `<link data-trunk rel="rust" data-target-name={artifact} />` to select the specific artifact by name
```

My output of `cargo --version`:

```
cargo 1.85.1 (d73d2caf9 2024-12-31)
```

While the error also occurs with `1.81.0-nightly(a2b58c3da 2024-07-16)`.

Though [it's raised by `trunk`](https://github.com/trunk-rs/trunk/blob/7e0a0117b4bd3c8e4b272fb342264ee4a0f61b3b/src/pipelines/rust/mod.rs#L511-L518), people believe it's a cargo problem: https://github.com/trunk-rs/trunk/issues/740.

It seems cargo used to produce duplicate artifacts but the checking rules may be changed at some time: https://github.com/rust-lang/cargo/issues/6313#issuecomment-2222146333

To fix it, I explicitly defined the binary artifact name and specified the name in `index.html`. I'm not sure if it's the best way though 😅 

Also, the `simple_logger` was removed from the example `counter_isomorphic`, which indeed was not needed before, but somehow cause another compilation error:

2415fc6feb97a4e3c1814895bfb3fd1

### DIFF
--- a/examples/counter_isomorphic/Cargo.toml
+++ b/examples/counter_isomorphic/Cargo.toml
@@ -27,7 +27,6 @@ once_cell = "1.19"
 gloo-net = { version = "0.6.0" }
 wasm-bindgen = "0.2.93"
 serde = { version = "1.0", features = ["derive"] }
-simple_logger = "5.0"
 tracing = { version = "0.1.40", optional = true }
 send_wrapper = "0.6.0"
 
--- a/examples/hackernews/Cargo.toml
+++ b/examples/hackernews/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2021"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[[bin]]
+name = "hackernews_bin"
+path = "src/main.rs"
+
 [profile.release]
 codegen-units = 1
 opt-level = "z"
--- a/examples/hackernews/index.html
+++ b/examples/hackernews/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 	<head>
-		<link data-trunk rel="rust" data-wasm-opt="z" data-cargo-features="csr"/>
+		<link data-trunk rel="rust" data-wasm-opt="z" data-cargo-features="csr" data-bin="hackernews_bin"/>
 		<link data-trunk rel="css" href="./style.css"/>
 	</head>
 	<body></body>
--- a/examples/hackernews_axum/Cargo.toml
+++ b/examples/hackernews_axum/Cargo.toml
@@ -37,11 +37,7 @@ log = { version = "0.4", optional = true }
 
 [features]
 default = ["csr"]
-csr = [
-  "leptos/csr",
-  "dep:console_log",
-  "dep:log",
-]
+csr = ["leptos/csr"]
 hydrate = ["leptos/hydrate"]
 ssr = [
   "dep:axum",
--- a/examples/hackernews_axum/Cargo.toml
+++ b/examples/hackernews_axum/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2021"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[[bin]]
+name = "hackernews_axum_bin"
+path = "src/main.rs"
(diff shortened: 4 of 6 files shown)


gbj:
Just double-checking: You say it's an error, but when I try it (on `1.87.0-nightly 2025-03-05`), it's a warning, and the log output you post has it as a warning, right? At present it works completely as intended, but tells you it might break in the future?

To be honest I'd be surprised if the change to convert it from a warning to a hard error ever landed.

bioinformatist:
Sorry it's my fault, I haven't clarified what happened.

My cargo version is `1.85.1 (d73d2caf9 2024-12-31)` and the trunk version is `0.21.9`.

The warnings are thrown by `cargo`:

image

While the error is raised by `trunk`:

image

All these abnormalities occurs within `csr` mode by `trunk serve`.

We have `main()` function in the source code, so there will be a binary named as `hackernews` after compiling. Meanwhile, our `ssr` mode need `cdylib` as one of the crate types, which may produce another `hackernews` artifact. I know it's not binary but it has an ambiguous name. Actually, the `csr` mode only need the binary one, so I renamed it with `hackernews_bin` and specify this name in `index.html` for `trunk`. That's it.

Sorry for wasting your time before.

gbj:
Oh, weird. `trunk` does not have an error; if I go to `hackernews` and try `trunk serve --open` it works just fine.

I guess if `trunk` is actually erroring for some users then it makes sense to make the change.

Could you revert the change adding `console_log`/`log` to the Axum one however, and just remove that initialization line in `main`. `log` is not actually used anywhere in the example.

bioinformatist:
Done 😄

bioinformatist:
Currently the PR fails to pass the CIs. I can reproduce it locally by `cargo leptos build`:

image

It seems like a `cargo-leptos` problem, which is similar to: https://github.com/leptos-rs/cargo-leptos/issues/459

I'm trying to figure the source out.

benwis:
This seems to be an issue with rust's latest nightly, which as mentioned in the mentioned commits. There is an artifact collision warning but that shouldn't prevent compilation. I'll close this as a separate issue, but if you feel the warning is an issue, let's track that elsewhere

PR #3791, opened 2025-03-26
add track_caller to OptionStoreExt

mscofield0:


mscofield0:
Something's still missing, no idea where the missing `#[track_caller]` should go.

mscofield0:
Seems to be the reason:
<https://doc.rust-lang.org/reference/attributes/codegen.html#r-attributes.codegen.track_caller.decay>

gbj:
Happy to try to help -- But could you provide an example that currently isn't working as expected, so I can explore how to get the right lines to show up in a helpful way?

gbj:
Thanks. Yes, I agree that this won't work. The only way I can think of to get it to work would be to turn these from function pointers into boxed closures, so that they can close over the location; but that then adds some overhead in all cases in order  to provide a better error message in the case of incorrect code, so it's not necessary ideal.

PR #3793, opened 2025-03-26
Improved handling of `<Title/>`

gbj:


ilyvion:
Don't merge this yet, I've grabbed the code from this branch and testing it against my project and I'm encountering some new issues with it that weren't there before. I'm working my way to get to the source of it, should have a MRE in like an hour at most, probably.

ilyvion:
It seems like using any kind of signal with `Title` now causes a panic when a page is being navigated away from; I can reproduce it as easily as putting `<Title text=Signal::derive(move || "test") />` on a page and then navigating to another. You can use the Axum starter template, and stick this in as the `App` and related pages:
```rs
#[component]
    provide_meta_context();

    view! {
        <Stylesheet id="leptos" href="/pkg/hydration_error_mre.css" />

        <Router>
            <main>
                    <Route path=StaticSegment("") view=HomePage />
                    <Route path=StaticSegment("/other") view=OtherPage />
                </Routes>
            </main>
        </Router>
    }
}

#[component]
    let test_resource = Resource::new(|| {}, move |_| async move { test(69).await });

    view! {
        <Title text=Signal::stored("home") />
        <h1>"Welcome to Leptos!"</h1>
        <A href="/other">Go to other</A>
    }
}

#[component]
fn OtherPage() -> impl IntoView {
    let test_resource = Resource::new(|| {}, move |_| async move { test(69).await });

    view! {
        <Title text=Signal::stored("other") />
        <h1>"This is the other page!"</h1>
        <A href="/">Go to home</A>
    }
}
```
`cargo leptos serve` and then visit either page and click its link to the other page; you'll get this in the JS console:
```
panicked at C:\Users\alexs\Documents\Projects\leptos\reactive_graph\src\traits.rs:388:39:
At C:\Users\alexs\Documents\Projects\leptos\reactive_graph\src\nightly.rs:149:1, you tried to access a reactive value which was defined at src\app.rs:71:21, but it has already been disposed.
```
Obviously, in my "real" project I'm doing something more advanced, but even a "static" signal triggers the behavior, so.

gbj:
Thanks @ilyvion, I hadn't tested this. Latest commit should fix that particular issue; let me know if there are other things you notice in your real use case.

ilyvion:
Yeah, I can no longer cause this to panic in my project. 👍

PR #3798, opened 2025-03-27
derive_local for ArcSignal<T, LocalStorage>

zakstucke:
Seems to be accidentally forgotten

### DIFF
--- a/reactive_graph/src/wrappers.rs
+++ b/reactive_graph/src/wrappers.rs
@@ -225,32 +225,14 @@ pub mod read {
         /// Wraps a derived signal. Works like [`Signal::derive`] but uses [`LocalStorage`].
         #[track_caller]
         pub fn derive_local(derived_signal: impl Fn() -> T + 'static) -> Self {
-            let derived_signal = SendWrapper::new(derived_signal);
-            #[cfg(feature = "tracing")]
-            let span = ::tracing::Span::current();
-
-            let derived_signal = move || {
-                #[cfg(feature = "tracing")]
-                let _guard = span.enter();
-                derived_signal()
-            };
-
-            Self {
-                inner: SignalTypes::DerivedSignal(Arc::new(derived_signal)),
-                #[cfg(any(debug_assertions, leptos_debuginfo))]
-                defined_at: std::panic::Location::caller(),
-            }
+            Signal::derive_local(derived_signal).into()
         }
 
         /// Moves a static, nonreactive value into a signal, backed by [`ArcStoredValue`].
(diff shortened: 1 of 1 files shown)


zakstucke:
Or actually, `local_` variants aren't setup for any arc signal types, and the way I've done it here doesn't make it `Send` correctly. 

This does compile right now:
```rust
    fn foo() -> impl Send {
        let arc_signal: ArcSignal<_, LocalStorage> =
            Signal::derive_local(|| NonNull::<usize>::dangling()).into();
    }
```

zakstucke:
Maybe just calling `.into()` internally is fine and reduces duplicate logic?

gbj:
Yeah the only reason there is even a storage type generic on `ArcSignal<_>` is to support `ArcMemo`, and the only reason `Memo` has it is because local memos need to have a special storage type, and rather than duplicating all the memo logic I just had all memos use that.
 
I think your `.into()` version is fine.

zakstucke:
@gbj actually this doesn't quite work:
compiles:
```rust
        let signal = ArcSignal::derive_local(std::ptr::null::<usize>);
        std::thread::spawn(move || {
            signal.track();
        });
```
`*const usize` cannot be sent between threads safely`:
```rust
        let signal = ArcSignal::derive_local(std::ptr::null::<usize>);
        let signal: ArcSignal<_, LocalStorage> = signal.into();
        std::thread::spawn(move || {
            signal.track();
        });
```
Isn't this a valid usecase for all `Arc` signals to have `_local` variants? Not just:
> Yeah the only reason there is even a storage type generic on ArcSignal<_> is to support ArcMemo

As far as I can see, the usecase is identical to the need for all normal signals to have `_local` variants. If you want to not be limited to the owner lifecycle and still need a hidden `SendWrapper`.

gbj:
Thinking a bit more about it...

None of the `Arc` variants have `_local` forms because the `_local`-ness of the `_local` ones for normal arena versions refers to whether the *arena* is thread-local or not, not the inner value. 

i.e,. if you create an `RwSignal::new_local()` for some `T: !Send` it creates an `ArcRwSignal<T>`, which is `!Send`. Then it stores that in a thread-local arena, and gives you an `RwSignal<T>`, which *is* `Send` but can only be accessed from that same thread. But it does not create a `SendWrapper` that's internal to the `ArcRwSignal`.

Adding `_local` variants for all of the various `Arc` forms seems to me like a bunch of extra code to maintain for a relatively uncommon use case that can be be implemented in userland already (by wrapping an `ArcRwSignal<SendWrapper<T>>` if you want an API that doesn't expose the send wrapper, or just using that if you don't care) I'm open to being wrong, though.

PR #3801, opened 2025-03-28
fix: allow repeated `class=` for all tuples, not only static ones (closes #3794)

### LINKED ISSUE #3794: Repeat `class=(name, value)` attributes
I want to assign a static classname, `names::one`, and conditionally a second class name, `names::two`, where `one` and `two` are `&'static str` constants (produced by `stylance`)

```rust
let value = RwSignal::new(false);
view! {
<div class=names::one class=(names::two, value)>
</div>
}
```

I can ofc manually format my class string, its HTML 'under the hood' anyway, but I was surprised that this seemingly common use case wasn't already working

**Additional context**
Using leptos 0.7

gbj:


ActuallyHappening:
Thx for fixing this, I never got around to it
Your time is not unnoticed :)

PR #3802, opened 2025-03-28
fix: prevent race condition in executor initialization + docs, optimization and tests

sabify:
What's changed:
- Fixed race conditions in executor initialization.
- Optimized `spawn`, `spawn_local`, and `tick` usage with inlining and branch optimization.
- Covered tests for any executor feature.
- ~~Changed orderings of atomic variables to *Sequentially Consistent* (`SeqCst`) to ensure the orderings will not be optimized out by the compiler and hardware.~~
- Corrected tick usage in `reactive_stores`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,6 +267,7 @@ version = "0.2.1"
 dependencies = [
  "async-executor",
  "futures",
+ "futures-lite",
  "glib",
  "serial_test",
  "thiserror 2.0.12",
@@ -392,9 +393,9 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "axum"
-version = "0.8.1"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d6fd624c75e18b3b4c6b9caf42b1afe24437daaee904069137d8bab077be8b8"
+checksum = "de45108900e1f9b9242f7f2e254aa3e2c029c921c258fe9e6b4217eeebd54288"
 dependencies = [
  "axum-core",
  "base64",
@@ -430,12 +431,12 @@ dependencies = [
 
 [[package]]
 name = "axum-core"
-version = "0.5.0"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df1362f362fd16024ae199c1970ce98f9661bf5ef94b9808fee734bc3698b733"
+checksum = "68464cd0412f486726fb3373129ef5d2993f90c34bc2bc1c1e9943b2f4fc7ca6"
 dependencies = [
  "bytes",
- "futures-util",
+ "futures-core",
  "http 1.3.1",
  "http-body",
  "http-body-util",
@@ -2312,9 +2313,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
-version = "1.21.2"
+version = "1.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2806eaa3524762875e21c3dcd057bc4b7bfa01ce4da8d46be1cd43649e1cc6b"
+checksum = "42f5e15c9953c5e4ccceeb2e7382a716482c34515315f7b03532b8b4e8393d2d"
 
 [[package]]
 name = "openssl"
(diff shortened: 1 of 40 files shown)


sabify:
@gbj I've tested the `reactive_stores` in four different machines and platforms, and all tests were passed, but in the CI the two `reactive_stores`  tests fail. Can you please look at the error and let me know what the issue is? I think there should be some edge cases.

https://github.com/leptos-rs/leptos/actions/runs/14138282599/job/39614957708

sabify:
The fail also happened in last merged PR in the main.

sabify:
This PR is ready regardless of the failed CI, as the issue was introduced in the PR https://github.com/leptos-rs/leptos/pull/3799.

sabify:
The problem with proving if the weaker ordering works as expected is the actual problem. Incorrect use can lead to extremely subtle and hard-to-debug race conditions or incorrect behavior that might only manifest under specific interleavings. Weaker orderings are cache friendly and compiler and hardware (CPU) can easily change the execution order of the atomic variables on read and write!

Unless we have a clear understanding of the memory model and can prove the correctness of our algorithms using `Acquire`/`Release` or `Relaxed` (the least favorite), we should prefer the safest and strongest ordering.

I replaced all because it was the safest option. The "Relax" ordering may be used in counters which you may not be relying on the value if it was incorrect at the read operation (duplicated, the value may not be set in the previous write operation). In some examples, there were read and write operations next to each other.

Tests (besides the application logic) should always be correct and I think keeping them there also is valuable.

Please let me know if you are still interested in the weaker atomic orderings and tell me which one I should replace.

gbj:
Re: atomic ordering — I think it's outside the scope of this PR, which I'm otherwise happy to accept.

I am happy to have a longer conversation about it in an issue or a PR that's dedicated to that topic. I do not want to include it as a global search-and-replace in an unrelated PR.

sabify:
I just reverted the atomic ordering changes.

ilyvion:
So... how are we supposed to do tests now? This change broke all my tests. I have this:
```rs
use leptos::prelude::Owner;
use leptos::task::Executor;
use tokio::task::LocalSet;

pub async fn with_leptos_owner<F: Future<Output = ()>>(f: F) {
    _ = Executor::init_tokio();
    let owner = Owner::new();
    owner.set();
    LocalSet::new().run_until(f).await;
}
```
and then in my tests, I do things like
```rs
#[tokio::test]
async fn countdown_decreases_as_time_advances() {
    with_leptos_owner(async {
        let (now, set_now) = signal(Utc::now());
        let (error, set_error) = signal(None::<SignInError>);

        let remaining = use_sign_in_throttle_seconds_remaining(now.into(), error.into());

        let base_time = now.get_untracked();
        set_error.set(Some(SignInError::Throttled(TimeDelta::seconds(5))));
        Executor::tick().await;
        assert_eq!(remaining.get_untracked(), Some(5));

        set_now.set(base_time + TimeDelta::seconds(2));
        Executor::tick().await;
        assert_eq!(remaining.get_untracked(), Some(3));

        set_now.set(base_time + TimeDelta::seconds(6));
        Executor::tick().await;
        assert_eq!(remaining.get_untracked(), None);
    })
    .await;
}
```

sabify:
Use `any_spawner::Executor` directly like what we have in tests throughout the codebase. Replace `use leptos::task::Executor;` with `use any_spawner::Executor;` and include it under your `Cargo.toml` dependencies.

ilyvion:
> The problem with proving if the weaker ordering works as expected is the actual problem. Incorrect use can lead to extremely subtle and hard-to-debug race conditions or incorrect behavior that might only manifest under specific interleavings. Weaker orderings are cache friendly and compiler and hardware (CPU) can easily change the execution order of the atomic variables on read and write!
> 
> Unless we have a clear understanding of the memory model and can prove the correctness of our algorithms using `Acquire`/`Release` or `Relaxed` (the least favorite), we should prefer the safest and strongest ordering.

I just want to say that you (the general "any one of you") may feel free to rope me in if you ever feel like you need to go over your atomic orderings. I have a good understanding of them and can offer my input.

To quickly summarize what matters though: atomic ordering only really matters when it comes to happens-before and happens-after relationships with other memory accesses *within the same execution thread*[^not-across-threads]. The canonical example is a mutex, where you need a happens-before relationship between the lock and the contents when acquiring the lock and a happens-after relationship between the contents and releasing the lock.

In most cases where such a strict relationship isn't required for correct behavior, `Relaxed` is the right choice.

[^not-across-threads]: I.e., not in code running in different threads/on different CPUs/CPU cores -- an atomic counter e.g. will always be strictly incrementing across multiple threads even with `Relaxed` reads/writes as long as the atomic operations themselves are correct.

PR #3804, opened 2025-03-29
Migrate from Tailwind 3 to Tailwind 4 for the axum example.

pico-bolero:
Migrates to filenames that are used in the official tailwind documentation e.g., input.css and output.css.
Uses @tailwindcss/cli tool to build the output.css.
Removes no longer needed tailwind.config.js as that configuration can be embedded in input.css.
Removes src/style.css as it only included tailwind 3 configurations that are no longer needed.


gbj:
Thanks!

PR #3805, opened 2025-03-30
Add track_caller to store field methods

jvdwrf:
Improves error messages by pointing to the actual location that the signal was accessed.


gbj:
Thanks!

PR #3811, opened 2025-04-01
feat: enhancing server_fn errors

sabify:
This PR provides custom encoding and decoding for returned error types from server_fns.

The new syntax:
```rust
#[derive(
    thiserror::Error,
    Debug,
    Clone,
    Display,
    Serialize,
    Deserialize,
    rkyv::Archive,
    rkyv::Serialize,
    rkyv::Deserialize,
)]
pub enum MyErrors {
    ServerFnError(ServerFnErrorErr),
    Other(String),
}


impl FromServerFnError for MyErrors {
    type Encoder = RkyvEncoding;

    fn from_server_fn_error(value: ServerFnErrorErr) -> Self {
        MyErrors::ServerFnError(value)
    }
}
```

### DIFF
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -3,7 +3,7 @@
 use crate::{ContentType, Decodes, Encodes, Format, FormatType};
 use base64::{engine::general_purpose::URL_SAFE, Engine as _};
 use bytes::Bytes;
-use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 use std::{
     fmt::{self, Display, Write},
     str::FromStr,
@@ -263,13 +263,7 @@ impl FormatType for ServerFnErrorEncoding {
 
 impl<CustErr> Encodes<ServerFnError<CustErr>> for ServerFnErrorEncoding
 where
-    CustErr: std::fmt::Debug
-        + Display
-        + Serialize
-        + DeserializeOwned
-        + 'static
-        + FromStr
-        + Display,
+    CustErr: Display,
 {
     type Error = std::fmt::Error;
 
@@ -311,13 +305,7 @@ where
 
 impl<CustErr> Decodes<ServerFnError<CustErr>> for ServerFnErrorEncoding
 where
-    CustErr: std::fmt::Debug
-        + Display
-        + Serialize
-        + DeserializeOwned
-        + 'static
-        + FromStr
-        + Display,
+    CustErr: FromStr,
 {
     type Error = String;
 
@@ -361,13 +349,7 @@ where
 
 impl<CustErr> FromServerFnError for ServerFnError<CustErr>
 where
-    CustErr: std::fmt::Debug
-        + Display
-        + Serialize
-        + DeserializeOwned
-        + 'static
-        + FromStr
-        + Display,
+    CustErr: std::fmt::Debug + Display + FromStr + 'static,
 {
     type Encoder = ServerFnErrorEncoding;
 
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -1,12 +1,13 @@
 #![allow(deprecated)]
 
+use crate::{ContentType, Decodes, Encodes, Format, FormatType};
 use base64::{engine::general_purpose::URL_SAFE, Engine as _};
+use bytes::Bytes;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
     fmt::{self, Display, Write},
     str::FromStr,
 };
-use thiserror::Error;
 use throw_error::Error;
 use url::Url;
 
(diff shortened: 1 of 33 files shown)


gbj:
Hm... Could you clarify what this enables that a custom of the `ser()` and `de()` methods in `FromServerFnError` would not already support? I might be missing something — this looks like it uses a base64-encoded string anyway, so how is it different from providing custom implementations rather than the `serde_json`-based defaults for those methods?
```rust
#[derive(
    thiserror::Error,
    Debug,
    Clone,
    Display,
    Serialize,
    Deserialize,
    rkyv::Archive,
    rkyv::Serialize,
    rkyv::Deserialize,
)]
pub enum MyErrors {
    ServerFnError(ServerFnErrorErr),
    Other(String),
}

impl FromServerFnError for MyErrors {
    fn from_server_fn_error(value: ServerFnErrorErr) -> Self {
        MyErrors::ServerFnError(value)
    }

    fn ser(&self) -> String {
        // do whatever you want
    }

    /// Deserializes the custom error type from a [`&str`]. Defaults to deserializing from JSON.
    fn de(data: &str) -> Self {
       // do whatever you want
    }
    ```

**Comment by sabify** [comment]
Using base64 depends on the protocol and the encoder format and we only encode/decode to base64 when a protocol needs the UTF-8 representation of the binary data. This implementation unifies communication between server and client, and some users may not want to get JSON type for every error type.

Without interface change, we have to use base64 encoder for any binary format type in protocols that support sending binary format type and also in regards to DX, implementing `ser/de` methods for every error type is cumbersome and it's not a clean approach.

This implementation supports sending errors as bytes and is protocol context-aware that can safely represent the binary formats for UTF-8 compatible protocols whenever needed.

**Comment by sabify** [comment]
Regardless of the benefits I've mentioned earlier, you can compare these two in developer experience perspective:

```rust
#[derive(
    thiserror::Error,
    Debug,
    Clone,
    Display,
    Serialize,
    Deserialize,
    rkyv::Archive,
    rkyv::Serialize,
    rkyv::Deserialize,
)]
pub enum MyErrors {
    ServerFnError(ServerFnErrorErr),
    Other(String),
}
```

Current implementation:

```rust
impl FromServerFnError for MyErrors {
    fn from_server_fn_error(value: ServerFnErrorErr) -> Self {
        MyErrors::ServerFnError(value)
    }

    fn ser(&self) -> String {
        let encoded = RkyvEncoding::encode(self).unwrap_or_else(|e| {
            RkyvEncoding::encode(&Self::from_server_fn_error(
                ServerFnErrorErr::Serialization(e.to_string()),
            ))
            .expect(
                "error serializing should success at least with the \
                 Serialization error",
            )
        });

        STANDARD_NO_PAD.encode(encoded)

    }

    /// Deserializes the custom error type from a [`&str`]. Defaults to deserializing from JSON.
    fn de(data: &str) -> Self {
       let decoded = STANDARD_NO_PAD.decode(data).map(|data| data.into())..unwrap_or_else(|e| {
            ServerFnErrorErr::Deserialization(e.to_string()).into_app_error()
       })?;
       let decoded = RkyvEncoding::decode(decoded).unwrap_or_else(|e| {
            ServerFnErrorErr::Deserialization(e.to_string()).into_app_error()
        });

        decoded
    }
}
```

The new implementation:

```rust
impl FromServerFnError for MyErrors {
    type Encoder = RkyvEncoding;

    fn from_server_fn_error(value: ServerFnErrorErr) -> Self {
        MyErrors::ServerFnError(value)
    }
}
```

sabify:
This PR also supersedes #3766.

server_fn/src/lib.rs
```diff
@@ -445,8 +446,7 @@ where
 
         // if it returns an error status, deserialize the error using FromStr
```
benwis, line 447: This comment seems outdated

benwis:
@sabify It's an interesting idea, replacing our generic Error type in the trait system with Bytes and then working off it. I'll need to chew on the implications a bit. I think we'll need to add a page to the book about custom error handling for this.

@ealmloff Dioxus uses this too, so I'd love your input on this change

ealmloff:
> @ealmloff Dioxus uses this too, so I'd love your input on this change

This looks like it would work for dioxus as long as we can serialize and deserialize a custom data type. Serializing the data type into a string or bytes would both be fine

benwis:
> This PR may also supersede #3766.

I'm a little puzzled by this, how does it apply to #3766 ?

sabify:
> I'm a little puzzled by this, how does it apply to #3766 ?

My bad! I had a very quick look at the PR and didn't realize that these are two different domains. You can merge #3766 first and then I will try to rebase based on the new changes.

> @sabify It's an interesting idea, replacing our generic Error type in the trait system with Bytes and then working off it. I'll need to chew on the implications a bit. I think we'll need to add a page to the book about custom error handling for this.

Bytes will carry out the serialized data and the generics are still in there and doing their job for providing the se/der data.

sabify:
> I'm a little puzzled by this, how does it apply to #3766 ?

My bad! I had a very quick look at the PR and didn't realize that these are two different domains. You can merge #3766 first and then I will try to rebase based on the new changes.

> @sabify It's an interesting idea, replacing our generic Error type in the trait system with Bytes and then working off it. I'll need to chew on the implications a bit. I think we'll need to add a page to the book about custom error handling for this.

Bytes will carry out the serialized data and the generics are still in there and doing their job of providing the ser/de data.

benwis:
> > I'm a little puzzled by this, how does it apply to #3766 ?
> 
> My bad! I had a very quick look at the PR and didn't realize that these are two different domains. You can merge #3766 first and then I will try to rebase based on the new changes.
> 
> > @sabify It's an interesting idea, replacing our generic Error type in the trait system with Bytes and then working off it. I'll need to chew on the implications a bit. I think we'll need to add a page to the book about custom error handling for this.
> 
> Bytes will carry out the serialized data and the generics are still in there and doing their job of providing the ser/de data.

@sabify You're absolutely right I wasn't absolutely clear, and that sounds like a plan. I've merged #3766 and once this one is fixed I'll merge it as well. I'm hoping to release 0.8 soon with both changes included

benwis:
Thanks @sabify as always!

server_fn/src/error.rs
```diff
@@ -505,34 +526,54 @@ impl<E> From<ServerFnUrlError<ServerFnError<E>>> for ServerFnError<E> {
...
+        };
+        let err = E::de(bytes);
+        Ok(Self(err))
     }
 }
 
 /// A trait for types that can be returned from a server function.
 pub trait FromServerFnError:
-    std::fmt::Debug + Serialize + DeserializeOwned + 'static
+    std::fmt::Debug + Sized + Display + 'static
```
ilyvion, line 565: Did we really have to go back to the `Display` bound? I was so happy when I learned that 0.8 wasn't going to require the custom errors to have `Display` anymore, and now this obnoxious requirement is back? 😩

zakstucke:
Hey @sabify this PR removed support for `E` on `ByteStream`'s which makes streaming apis much less ergonomic in that all errors need to impl `Into<Bytes>` and would then need manual deserialization on the client. Why was this necessary, or can we revert that? 

Change here: [server_fn/src/codec/stream.rs](https://github.com/leptos-rs/leptos/pull/3811/files#diff-92f9735065dc4dade43747e27c98b38d7e047d469902310af84f1871fb46cf70)

sabify:
Hi @zakstucke, could you please provide an example of your use case?

zakstucke:
@sabify pseudocode but:

```rust
#[server]
async fn api_fn() -> Result<leptos::server_fn::codec::ByteStream, ServerFnError<MyErr>> {
    let (tx, rx) = tokio::sync::mpsc::channel::<Result<Vec<u8>, MyErr>>(5);
    Ok(leptos::server_fn::codec::ByteStream::new(
        tokio_stream::wrappers::ReceiverStream::new(rx),
    ))
}

// Client:
let mut stream = api_fn().await.unwrap().into_inner();

while let Some(result) = stream.next().await {
    match result {
        Ok(bytes) => {
            // Do something with the bytes
        }
        Err(e) => {
            // Handle the error
        }
    }
}
```

sabify:
@zakstucke Can you please check out #3869?

zakstucke:
@sabify replied over there, that sorts it :) nice work

PR #3812, opened 2025-04-01
fix: remove `SendOption` from public API of actions

gbj:
Reverts the API of the `.input()` and `.value()` signals on actions back to the previous API (a signal that returns some value without needing to be dereferenced) while maintaining the thread-safety benefits of #3762.


reactive_graph/src/signal/mapped.rs
```diff
@@ -0,0 +1,358 @@
...
+            inner: {
+                let this = ArcRwSignal::from(inner);
+                StoredValue::new_with_storage(ArcMappedSignal::new(
+                    this, map, map_mut,
+                ))
+            },
+        }
+    }
+}
+
```
ilyvion, line 260:
```suggestion
impl<T> Copy for MappedSignal<T> {}

impl<T> Clone for MappedSignal<T> {
    fn clone(&self) -> Self {
        *self
    }
}
```

ilyvion (commented):
I realize it's probably too late for this now, since this has already been merged, but without these impls, using a `MappedSignal` (like the one returned from `ServerAction::value()`) is rather painful.

reactive_graph/src/signal/mapped.rs
gbj: Thanks! Not at all too late, will just make a separate PR. Easy to fix.

gbj: #3827

PR #3815, opened 2025-04-03
Fix: update wasm preload and wasm fetch to use cross-origin same-site

### LINKED ISSUE #3814: Generated wasm file is loaded twice
**Describe the bug**
The wasm file is loaded twice and there is a console warning stating:

```
[Warning] The resource http://localhost:3000/pkg/example.wasm was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it wasn't preloaded for nothing.
```

This appears to be caused by a [CORS mismatch](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#cors-enabled_fetches) between the preload link and the fetch of the wasm file.

**Leptos Dependencies**

For example:
```toml
[dependencies]
leptos = { version = "0.7.0" }
leptos_router = { version = "0.7.0" }
axum = { version = "0.7", optional = true }
console_error_panic_hook = { version = "0.1", optional = true}
leptos_axum = { version = "0.7.0", optional = true }
leptos_meta = { version = "0.7.0" }
tokio = { version = "1", features = ["rt-multi-thread"], optional = true }
wasm-bindgen = { version = "=0.2.100", optional = true }
```

**To Reproduce**
1. Create a blank leptos project. I used https://github.com/leptos-rs/start-axum
2. Run the app
3. Open a new browser window with dev tools opened to the Network tab (I am using safari)
4. Visit localhost:3000
5. Note the duplicate wasm file requests and the warning on the dev tools Console tab.

```bash
% cargo leptos new --git https://github.com/leptos-rs/start-axum                

🤷   Project Name: example
✔ 🤷   Use nightly features? · No
% cd example 
% cargo leptos watch                                            

   Compiling proc-macro2 v1.0.94

   ( Omitting for brevity )

   Compiling leptos_server v0.7.8
   Compiling leptos_dom v0.7.8
   Compiling leptos v0.7.8
   Compiling leptos_meta v0.7.8
   Compiling leptos_router v0.7.8
   Compiling example v0.1.0 (/Users/paul/Desktop/example)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 31.33s
       Cargo finished cargo build --package=example --lib --target-dir=/Users/paul/Desktop/example/target/front --target=wasm32-unknown-unknown --no-default-features --features=hydrate
       Front generating JS/WASM with wasm-bindgen
    Finished generating JS/WASM for front in 188.114375ms
   Compiling unicode-ident v1.0.18

   ( Omitting for brevity )

   Compiling leptos_dom v0.7.8
   Compiling leptos_server v0.7.8
   Compiling leptos v0.7.8
   Compiling leptos_router v0.7.8
   Compiling leptos_meta v0.7.8
   Compiling leptos_integration_utils v0.7.8
   Compiling leptos_axum v0.7.8
   Compiling example v0.1.0 (/Users/paul/Desktop/example)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.02s
       Cargo finished cargo build --package=example --bin=example --no-default-features --features=ssr
     Serving at http://127.0.0.1:3000
    Creating ignore list from '.gitignore' file
listening on http://127.0.0.1:3000
```
**Expected behavior**
The wasm file should only be loaded once.

**Screenshots**

Image

Image

**Additional context**
The preload link looks like this:

```html
<link rel="preload" href="/pkg/psbot.wasm" as="fetch" type="application/wasm" crossorigin="3wWvZEHBTLHCxexEm0dK0g">
```

That is not a valid value for [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin). I believe that attribute should be `nonce`.

prsutherland:
Passes a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) object instead of a string URL to [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) so that we can define the CSP as `same-origin`. 

This is done indirectly in `leptos/src/hydration/hydration_script.js` because the actual call to `fetch` is buried in code generated by `wasm-bindgen`.

Fixes #3814 

Open questions:

1. Is `same-origin` the behavior we want? Will that cause headaches for anyone hosting static files on a CDN?
2. Any additional changes needed like test coverage? I'm fairly new to leptos, but happy to follow conventions.
3. What degree should this be back ported? I've tested on 0.8 and 0.7.

### DIFF
--- a/leptos/src/hydration/hydration_script.js
+++ b/leptos/src/hydration/hydration_script.js
@@ -1,7 +1,12 @@
 (function (root, pkg_path, output_name, wasm_output_name) {
 	import(`${root}/${pkg_path}/${output_name}.js`)
 		.then(mod => {
-			mod.default({module_or_path: `${root}/${pkg_path}/${wasm_output_name}.wasm`}).then(() => {
+			mod.default({
+				module_or_path: new Request(
+					`${root}/${pkg_path}/${wasm_output_name}.wasm`,
+					{ "credentials": "same-origin" }
+				)
+			}).then(() => {
 				mod.hydrate();
 			});
 		})
--- a/leptos/src/hydration/mod.rs
+++ b/leptos/src/hydration/mod.rs
@@ -115,7 +115,8 @@ pub fn HydrationScripts(
                 href=format!("{root}/{pkg_path}/{wasm_file_name}.wasm")
                 r#as="fetch"
                 r#type="application/wasm"
-                crossorigin=nonce.clone().unwrap_or_default()
+                nonce=nonce.clone().unwrap_or_default()
+                crossorigin
             />
             <script type="module" nonce=nonce>
                 {format!("{script}({root:?}, {pkg_path:?}, {js_file_name:?}, {wasm_file_name:?});{islands_router}")}


prsutherland:
Ugh, I did not test well enough and a typo caused a different issue which masked fixing #3814.

So, closing this PR as it  does not work as intended. The root problem is a Safari bug that can be reproduced with this minimal example:

**index.html**:
```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <link rel="preload" href="/example.dat" as="fetch" crossorigin>
        <script type="module">(
            function (url) {
                // fetch(url).then(function(response) {
                //     console.log(response.ok)
                // })
                const req = new XMLHttpRequest();
                req.addEventListener("load", function() {
                    console.log(this.status);
                });
                req.open("GET", url);
                req.send();
            })("/example.dat");
        </script>
    </head>
    <body>Testing preload fetch</body>
</html>
```

```bash
$ dd bs=1048576 seek=1 of=example.dat count=0
$ python3 -m http.server 3000 
```

PR #3831, opened 2025-04-08
More information in "outside reactive tracking context" warning.

ISibboI:
As of now, the "outside reactive tracking context" warning often only points to internals of `reactive_graph`, making it hard to figure out where in the user code this was caused. This commit adds a backtrace that is printed after the warning. It has the caveat that it is a WASM backtrace, and hence contains less information than a Rust backtrace. However, [it is currently impossible to get a nice Rust backtrace out of WASM](https://users.rust-lang.org/t/what-is-the-current-state-of-backtraces-in-wasm/128002/2?u=isibboi), hence this seems to be the best option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2777,6 +2777,7 @@ dependencies = [
  "tokio",
  "tokio-test",
  "tracing",
+ "wasm_backtrace",
  "web-sys",
 ]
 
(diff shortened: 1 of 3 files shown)


rakshith-ravi:
Does this require the `--keep-debug` flag on `wasm-bindgen` (as mentioned in the link)? If so, we'll need to update cargo-leptos to do that. Also, could you please help us with a screenshot of the backtrace? This seems pretty useful. Might be worth including in the changelog.

PR #3840, opened 2025-04-12
chore: correctly un/pin deps version + upgrades

sabify:


leptos/.cargo/config.toml
rakshith-ravi, line 1: Not sure if it makes sense to keep this in leptos itself. Would this trickle down to sub-crates if we do this? Don't want to mess with a downstream crate that doesn't require this flag

rakshith-ravi:
Also, @sabify - does it make sense to pin `wasm-bindgen` to a specific version instead of keeping the version requirement open? Given how unstable features in bindgen broke earlier versions, I'm wondering if it makes sense to keep the version pinned to a specific one.

CC @gbj

leptos/.cargo/config.toml
gbj: My understanding from past attempts to do it was that libraries cannot actually enable `cfg` flags on downstream crates. Based on that, I would not have expected adding this flag here to do anything for user applications. I don't have a broken example at hand to test it against, though.

gbj:
Oh and no, I don't think pinning `wasm-bindgen` to a specific version at a library level would not help with the issues there. At present, a user can always pin the version for their application if needed, to match a particular `cargo-leptos` version (by matching the `wasm-bindgen-cli-support` version included in that). Pinning it at a library level would mean we need to make a new patch release each time they do, not only of the CLI tool but of the whole library, and then it would make it impossible for users to set any other version.

rakshith-ravi:
Oh my. Okay yeah that's much worse. How do we deal with unstable features breaking our code though? Do we just swallow the pill and deal with it till bindgen makes them stable?

sabify:
> Oh my. Okay yeah that's much worse. How do we deal with unstable features breaking our code though? Do we just swallow the pill and deal with it till bindgen makes them stable?

You can just pin it in your app if you have a mismatched version for the CLI and the library.

leptos/.cargo/config.toml
sabify: You are right! I just add it in `build.rs`.

sabify: The idea of adding `.cargo/config.toml` was not good (because they are local and user preferences). I just changed it to be part of `build.rs`. It works at least for the CI (members, examples, tests, even isolated tests I've tried). We need a section in the book and documentation to educate the users to enable a `getrandom` WASM backend if there's anything wrong with the published library.

gbj:
> How do we deal with unstable features breaking our code though? Do we just swallow the pill and deal with it till bindgen makes them stable?

It's not a question of unstable features that need to be stabilized. The `wasm-bindgen` team has decided that the `wasm-bindgen-cli-support` crate is not supposed to be used by anything other than their own `wasm-bindgen` CLI tool, and to therefore say that every patch change is a breaking change by definition. (This is also why the error messages talk about updating `wasm-bindgen` even though it is built into `cargo-leptos` so they don't do anything.)

Long term the only solution is likely for us to shell out to `wasm-bindgen` instead of compiling it into `cargo-leptos`. There are obvious downsides to this (like requiring installing another dependency.) @benwis have talked about it back and forth a few times.

sabify:
I just changed this PR to `chore: upgrade rand and getrandom`.

PR #3842, opened 2025-04-13
fix: correctly decode base64-encoded server action error messages stored in URL

gbj:
#3811 introduced a bug when decoding `ServerAction` errors from the URL (when JS/WASM is turned off, or in islands mode), because errors were no longer decoded properly. Conveniently, fixing it along removed a bunch of code by correctly using the new helper method.

### DIFF
--- a/leptos_server/src/action.rs
+++ b/leptos_server/src/action.rs
@@ -4,7 +4,7 @@ use reactive_graph::{
     traits::DefinedAt,
 };
 use server_fn::{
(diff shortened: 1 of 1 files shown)


sabify:
I think the issue was introduced earlier in #3274! Please check out https://github.com/leptos-rs/leptos/pull/3274/files#diff-2dad071e80104a0fd5503d3cc9d2d4b8d4c0aa8a044fd4febdcd1991e8cb1d15L61 (directly deserializing the error).

gbj:
@sabify Ah yes you're correct, my mistake!

PR #3843, opened 2025-04-13
fix: reactive_graph keymap impl and clippy warnings

sabify:
This PR improves the `reactive_stores` `KeyMap` implementation by using concurrent `DashMap` on non-wasm and lock-free std `HashMap` on wasm. This also solves two Clippy errors, one mentioned earlier and the other was enum large stack usage.


benwis:
Nice fix! I'll have to remember to consider dashmap instead of `Arc<RwLock<...>>` for things that might have concurrent access

PR #3847, opened 2025-04-15
fix: remove `Display` requirement from `FromServerFnError`

ilyvion:
As I lamented @ https://github.com/leptos-rs/leptos/pull/3811#pullrequestreview-2766764718 I find the returned requirement of `Display` on all `FromServerFnError` implementations problematic. It turns out that it's there only so that `to_string()` can be called in a few places where web sockets are involved.

Since I don't use websockets and I don't ever display error messages "verbatim" to my end users, this `Display` requirement is just extra churn and noise for me.

We don't necessarily have to go with the exact solution I've provided here, but I'm opening this PR to at least "force" a discussion on the matter.

### DIFF
--- a/server_fn/src/client.rs
+++ b/server_fn/src/client.rs
@@ -202,13 +202,15 @@ pub mod browser {
                             Err(err) => {
                                 let err = InputStreamError::de(err);
                                 web_sys::console::error_1(
-                                    &js_sys::JsString::from(err.to_string()),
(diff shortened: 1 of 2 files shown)


ilyvion:
@sabify Ah, if your PR is accepted, we can close this one. I'll leave it open for now, just in case, but yeah, if the `Display` requirement goes away for another reason, this PR becomes redundant.

PR #3848, opened 2025-04-15
 fix: send/receive websocket data

sabify:
With the current implementation, you are not able to send errors over WebSockets and you will be trapped into an infinite loop. This PR solves it and you can properly send stream data over WebSocket. As an effect, there is no longer a requirement to have `Display` bound on the `FromServerFnError` trait as we are no longer using it.

Supersedes: #3847

PR origin: #3656

### DIFF
--- a/examples/server_fns_axum/src/app.rs
+++ b/examples/server_fns_axum/src/app.rs
@@ -945,9 +945,7 @@ pub fn CustomClientExample() -> impl IntoView {
                             Item = Result<server_fn::Bytes, server_fn::Bytes>,
                         > + Send
                         + 'static,
-                    impl Sink<Result<server_fn::Bytes, server_fn::Bytes>>
-                        + Send
-                        + 'static,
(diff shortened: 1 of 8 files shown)


benwis:
Interesting, moving from `Result<Bytes>` to a `Bytes` in the serialized format. If I understand correctly, any non-leptos user accessing an endpoint now has to decode `[u8, Bytes]` instead of Bytes. You did make these functions public, which I appreciate.

I don't currently understand the how of how that switch fixes the infinite loop described above. I'm also a touch hesitant as it seems like we're coming up with our own implementation of how to serialize/de-serialize a Result

gbj:
Thanks, @sabify, this looks good to me. I think the custom ser/de for Results is the right call. @ealmloff tagging you FYI  for Dioxus integration purposes, but I think the effects here are mostly contained with `server_fn` itself as the broader `BoxedStream<T, E>` interface hasn't changed.

PR #3849, opened 2025-04-15
fix: Use stabilized ClipboardEvent

feathecutie:
`ClipboardEvent` has been stabilized in `web-sys` for a while now, and is widely available according to MDN, so this PR uses `ClipboardEvent` for the appropriate events (`copy`, `cut`, `paste`).

- `web-sys` changelog: https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/CHANGELOG.md#changed-7
  (Note that the version number in the changelog refers to `wasm-bindgen` instead of `web-sys`, but `ClipboardEvent` is indeed stabilized in the version of `web-sys` that `tachys` uses)
- MDN page: https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent

### DIFF
--- a/tachys/src/html/event.rs
+++ b/tachys/src/html/event.rs
@@ -733,9 +733,9 @@ generate_event_types! {
   // =========================================================
   // DocumentAndElementEventHandlersEventMap
   // =========================================================
-  copy: Event, // ClipboardEvent is unstable
-  cut: Event, // ClipboardEvent is unstable
-  paste: Event, // ClipboardEvent is unstable
+  copy: ClipboardEvent,
+  cut: ClipboardEvent,
+  paste: ClipboardEvent,
 
   // =========================================================
   // DocumentEventMap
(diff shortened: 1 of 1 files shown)


rakshith-ravi:
Hey. Thanks for contributing to Leptos! This looks good to me. If all the tests pass, should be good to go from my end.

If it's not too much trouble, could I trouble you for a PR to [leptos-use](https://github.com/Synphonyte/leptos-use) as well?

CC @maccesch

feathecutie:
I just rebased onto latest `main` since CI was failing unrelated to this PR, hopefully it should work this time.

> If it's not too much trouble, could I trouble you for a PR to [leptos-use](https://github.com/Synphonyte/leptos-use) as well?

What do you mean by that? From what I can gather, it doesn't seem like any part of leptos-use should be affected by this PR at all

rakshith-ravi (approved):

gbj:
Thanks!

PR #3852, opened 2025-04-16
chore(nightly): update proc-macro span file name method name

gbj:
Latest nightly `2025-04-15` removes the [`.source_file()`](https://doc.rust-lang.org/proc_macro/struct.Span.html#method.source_file) method on `proc_macro::Span` (used in our hot-reloading code) and [splits it into `.file()` and `local_file()`](https://github.com/rust-lang/rust/issues/54725#issuecomment-2796806678). 

The bad news: this means leptos nightly no longer compiles on latest nightly.

The good news: this is a step toward stabilization for other parts of the API, which is *great*.

I've made the decision to just change to the newer method name, and require nightly users to use latest nightly.


benwis:
Thanks @gbj, seems reasonable for the nightly folks

Issue #3850, opened 2025-04-16
`ErrorBoundary` cannot handle multiple consequent `server_fn` errors


sabify:
**Describe the bug**
When multiple consecutive errors are returned from `server_fn` in `ErrorBoundry`, the `ErrorBoundry` state will become misaligned and unable to recover from it.

**Leptos Dependencies**

Using the `counter_isomorphic` example on latest commit ([e0bf5ec](https://github.com/leptos-rs/leptos/commit/e0bf5ec480d5e440410e071035ba89ccac5a9fb3))

**To Reproduce**
1. In `counter_isomorphic` example, change the two functions:
```rust
#[server]
#[cfg_attr(feature = "ssr", instrument)]
pub async fn get_server_count() -> Result<i32, ServerFnError> {
    use ssr_imports::*;
    let new = COUNT.load(Ordering::Relaxed);
    if new % 3 == 1 {
        return Err(ServerFnError::Registration("error".into()));
    }
    if new % 4 == 1 {
        return Err(ServerFnError::Registration("error2".into()));
    }
    Ok(new)
}

#[component]
pub fn Counter() -> impl IntoView {
    let dec = Action::new(|_: &()| adjust_server_count(-1, "decing".into()));
    let inc = Action::new(|_: &()| adjust_server_count(1, "incing".into()));
    let clear = Action::new(|_: &()| clear_server_count());
    let counter = Resource::new(
        move || {
            (
                dec.version().get(),
                inc.version().get(),
                clear.version().get(),
            )
        },
        |_| get_server_count(),
    );

    view! {
        <div>
            <h2>"Simple Counter"</h2>
            <p>
                "This counter sets the value on the server and automatically reloads the new value."
            </p>
            <div>
                <button on:click=move |_| {
                    clear.dispatch(());
                }>"Clear"</button>
                <button on:click=move |_| {
                    dec.dispatch(());
                }>"-1"</button>
                <ErrorBoundary fallback=|errors| move || format!("Error: {:#?}", errors.get())>
                    <span>"Value: " <Suspense>{counter} "!"</Suspense></span>
                </ErrorBoundary>
                <button on:click=move |_| {
                    inc.dispatch(());
                }>"+1"</button>
            </div>
        </div>
    }
}
```
2. Run the example with `cargo leptos serve`.
3. On the first and fourth '+1' button clicks, you see an expected error and the expected values between them.
[...]

PR #3853, opened 2025-04-16
feat(examples): add WebSocket example

sabify:
The example taken from the release page was modified to reflect other aspects of the protocol.

This can be considered a PoC for #3848.

@gbj The `ErrorBoundry` (#3850) issue can also be seen here as another MRE.


sabify:
The infinite loop in the current implementation:

https://github.com/leptos-rs/leptos/actions/runs/14496255301/job/40664693207?pr=3853

The CI and tests will not pass without #3848.

benwis:
Appreciate the example and the tests, I'm happy to merge this once we resolve #3848. It's good to have a websocket example so that it gets tested and to help catch regressions

PR #3722, opened 2025-03-16
chore: ran cargo outdated.

martinfrances107:
I noticed the list of outliers was getting large...

PR #3854, opened 2025-04-17
chore: unify all deps with min occurrences of 2

sabify:
This has been done with [`cargo-easy-dep`](https://crates.io/crates/cargo-easy-dep).

### DIFF
--- a/integrations/axum/Cargo.toml
+++ b/integrations/axum/Cargo.toml
@@ -11,11 +11,11 @@ edition.workspace = true
 [dependencies]
 any_spawner = { workspace = true, features = ["tokio"] }
 hydration_context = { workspace = true }
-axum = { version = "0.8.1", default-features = false, features = [
+axum = { default-features = false, features = [
   "matched-path",
-] }
-dashmap = "6"
-futures = "0.3.31"
+] , workspace = true }
+dashmap = { workspace = true, default-features = true }
+futures = { workspace = true, default-features = true }
 leptos = { workspace = true, features = ["nonce", "ssr"] }
 server_fn = { workspace = true, features = ["axum-no-default"] }
 leptos_macro = { workspace = true, features = ["axum"] }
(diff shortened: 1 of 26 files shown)


martinfrances107:
In terms of review, I think this is a good idea.

Having say the version number of the log crate made uniform over all worksapce items is a good idea.
Bugs in this area are really frustrating...

Asking and answering a questions for myself .. in public 

For various examples tokio needs to be build with a diverse collection of features.

So in the PR I went looking for these cases and I can confirm it seems to be handled correctly.

Below I am just copying and pasting two changes of tokio with different sets of features..

```
-  tokio = { version = "1.43", features = ["rt-multi-thread", "macros"] }
-  tokio-test = { version = "0.4.4" }
+ tokio = { features = ["rt-multi-thread", "macros"] , workspace = true }
+ tokio-test = { workspace = true }
```

```
-tokio = { version = "1.41", default-features = false, features = [
+tokio = { default-features = false, features = [
  "rt",
  "macros",
  "time",
] }
] , workspace = true }
```

Also I think this should be good for build times...

integrations/axum/Cargo.toml
```diff
@@ -11,28 +11,28 @@ edition.workspace = true
 [dependencies]
 any_spawner = { workspace = true, features = ["tokio"] }
 hydration_context = { workspace = true }
-axum = { version = "0.8.1", default-features = false, features = [
+axum = { default-features = false, features = [
```
rakshith-ravi, line 14: This introduces a very subtle bug. The workspace dependency is enabled with default features, but here we import it without default features. Since cargo features are only additive, this might enable the default feature for this dependency. Almost all dependencies in the workspace level should be set to `default-features = false` and the default features should be enabled on a per-member level.

You might want to confirm this isn't the case here

rakshith-ravi:
Could you please merge / rebase the changes? I think #3722 might be the cause of this conflict

integrations/axum/Cargo.toml
sabify: Thanks for mentioning it! The only problematic part is `default-features` as it is already turned on by default, and members cannot turn it off. We can simply inspect it with : `cargo metadata --format-version=1 | jq '.packages[] | select(.name == "leptos_axum") | .dependencies[]'`.

sabify:
Why I used `default-features = false` for all in the workspace and reenable them based on each member's preferences: https://github.com/sabify/cargo-easy-dep/tree/v0.1.1#notes

sabify:
@gbj / @benwis Have you been able to review this PR? 

We can also have all the dependencies in the workspace Cargo.toml and reuse them in members even though some may have only one occurrence in members. This can give us a clean view of overall dependencies and on updates, only one file will be changed which is more convenient to review.

gbj:
Thanks! I think this is great.

PR #3856, opened 2025-04-17
fixed compilation error in leptos_macro on nightly

dev-inigmas:
fixed compilation error in leptos_macro on nightly.

verified fix in project where I had discovered the errors. hot reload appears to work correctly.


dev-inigmas:
Fixed #3862

gbj:
Oh thanks, typo when I made the change for 0.7 I guess.

PR #3861, opened 2025-04-18
Complete the migration of examples to Tailwind 4

nnmm:
The tailwind_axum example was migrated in https://github.com/leptos-rs/leptos/pull/3804. However, the tailwind_csr example was not working yet with version 4, and the tailwind_actix example still had the no-longer-needed config file.

By renaming the CSS file to `input.css`, this also brings the three tailwind examples back in sync, mostly. To be honest, I don't know why the axum example has a packages.json and packages-lock.json file, but it does not seem to be needed.


gbj:
Thanks!

PR #3863, opened 2025-04-18
Prevent ScopedFuture stopping owner cleanup

zakstucke:
It seems `ScopedFuture` now prevents an `Owner` being cleaned up, I think this is the right fix?

MRE: (the `local` variant fails, the owner isn't cleaned up correctly, after this MR it passes) 
```rust

pub(crate) struct OnDrop<F>
where
    F: FnOnce(),
{
    f: Option<F>,
}

impl<F> OnDrop<F>
where
    F: FnOnce(),
{
    pub fn new(f: F) -> Self {
        Self { f: Some(f) }
    }
}

impl<F> Drop for OnDrop<F>
where
    F: FnOnce(),
{
    fn drop(&mut self) {
        if let Some(f) = self.f.take() {
            f();
        }
    }
}


    #[rstest]
    #[tokio::test]
    async fn test_drop_semantics(#[values(false, true)] local: bool) {
        tokio::task::LocalSet::new()
            .run_until(async move {
                let owner = Owner::default();

                macro_rules! with_tmp_owner {
                    ($body:block) => {{
                        let tmp_owner = owner.child();
                        tmp_owner.set();
                        let result = $body;
                        tmp_owner.unset();
                        owner.set();
                        result
                    }};
                }

                let dropped = with_tmp_owner! {{
                    let dropped = Arc::new(AtomicBool::new(false));
                    let on_drop = Arc::new(OnDrop::new({
                        let dropped = dropped.clone();
                        move || {
                            dropped.store(true, Ordering::Relaxed);
                    }}));
                    if local {
                        ArenaItem::<_, SyncStorage>::new_with_storage(ArcAsyncDerived::new_unsync(
                            move || {
                                let _on_drop = on_drop.clone();
                                async move {
                                }
                            })
                        );
                    } else {
                        ArenaItem::<_, SyncStorage>::new_with_storage(ArcAsyncDerived::new(
                            move || {
                                let _on_drop = on_drop.clone();
                                async move {
                                }
                            })
                        );
                    }
                    assert!(!dropped.load(Ordering::Relaxed));
                    dropped
                }};
                Executor::tick().await;
                assert!(dropped.load(Ordering::Relaxed));
            })
            .await;
    }
```

### DIFF
--- a/reactive_graph/src/computed/async_derived/arc_async_derived.rs
+++ b/reactive_graph/src/computed/async_derived/arc_async_derived.rs
@@ -560,7 +560,7 @@ impl<T: 'static> ArcAsyncDerived<T> {
         };
         let initial_value = SendOption::new_local(initial_value);
         let (this, _) = spawn_derived!(
-            crate::spawn_local_scoped,
+            crate::spawn_local,
             initial_value,
             fun,
             true,
@@ -595,7 +595,7 @@ impl<T: 'static> ArcAsyncDerived<T> {
             async move { SendOption::new_local(Some(fut.await)) }
         };
         let (this, _) = spawn_derived!(
-            crate::spawn_local_scoped,
+            crate::spawn_local,
             initial,
             fun,
             false,
--- a/reactive_graph/src/computed/async_derived/mod.rs
+++ b/reactive_graph/src/computed/async_derived/mod.rs
@@ -6,7 +6,7 @@ mod future_impls;
 mod inner;
 use crate::{
     graph::{AnySubscriber, Observer, WithObserver},
-    owner::Owner,
+    owner::{Owner, WeakOwner},
 };
 pub use async_derived::*;
 pub use future_impls::*;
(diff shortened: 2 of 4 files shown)


zakstucke:
I've tracked down the regression to https://github.com/leptos-rs/leptos/pull/3802, so this is quite broken on rc1 and rc2, but was fine on the initial beta.

zakstucke:
After seeing the specific cause [here](https://github.com/leptos-rs/leptos/pull/3802/files#diff-ffb35209ebfc25bbc9b74eeac4cf25069314d04721f502a52528292c758fafdaR563), switching the unsync `ArcAsyncDerived` spawners from `Executor::spawn_local` to `crate::spawn_local_scoped`, the real issue is just `ScopedFuture` is being used and shouldn't have been. 

Updated PR with a much cleaner fix.

sabify:
@zakstucke Thanks for the catch and sorry for the introduced issue; I should've taken care of it. I was localizing the executor functionalities but totally forgot this one and used the best candidate!

gbj:
Thanks!

Issue #3851, opened 2025-04-16
`ErrorBoundary` cannot handle `server_fn` errors on initialization


sabify:
**Describe the bug**
 When a `server_fn` returns an error on first access, the `ErrorBoundry` and runtime panic and are unable to recover from it.

**Leptos Dependencies**

Using the counter_isomorphic example on latest commit ([e0bf5ec](https://github.com/leptos-rs/leptos/commit/e0bf5ec480d5e440410e071035ba89ccac5a9fb3))

**To Reproduce**
1. In `counter_isomorphic` example, change the two functions:
```rust
#[server]
#[cfg_attr(feature = "ssr", instrument)]
pub async fn get_server_count() -> Result<i32, ServerFnError> {
    use ssr_imports::*;
    let new = COUNT.load(Ordering::Relaxed);
    if new % 3 == 0 {
        return Err(ServerFnError::Registration("error".into()));
    }
    Ok(new)
}

#[component]
pub fn Counter() -> impl IntoView {
    let dec = Action::new(|_: &()| adjust_server_count(-1, "decing".into()));
    let inc = Action::new(|_: &()| adjust_server_count(1, "incing".into()));
    let clear = Action::new(|_: &()| clear_server_count());
    let counter = Resource::new(
        move || {
            (
                dec.version().get(),
                inc.version().get(),
                clear.version().get(),
            )
        },
        |_| get_server_count(),
    );

    view! {
        <div>
            <h2>"Simple Counter"</h2>
            <p>
                "This counter sets the value on the server and automatically reloads the new value."
            </p>
            <div>
                <button on:click=move |_| {
                    clear.dispatch(());
                }>"Clear"</button>
                <button on:click=move |_| {
                    dec.dispatch(());
                }>"-1"</button>
                <ErrorBoundary fallback=|errors| move || format!("Error: {:#?}", errors.get())>
                    <span>"Value: " <Suspense>{counter} "!"</Suspense></span>
                </ErrorBoundary>
                <button on:click=move |_| {
                    inc.dispatch(());
                }>"+1"</button>
            </div>
        </div>
    }
}
```
2. Run the example with `cargo leptos serve`.
3. Open the browser's console and you will see an unrecoverable error.

**Expected behavior**
The `ErrorBoundy` and runtime should handle the error without panic.

gbj:
Thanks! I spent some time investigating this morning. The interaction between ErrorBoundary and Suspense is very complicated in general, so I'm not surprised.

[...]

PR #3870, opened 2025-04-21
Error boundary fixes

gbj:
This PR should address issues #3850 (clearing the previous error when rebuilding a node from `Err(_)` to `Err(_)`) and #3851 (`ErrorBoundary` with child `Suspense` that contains a value that starts in the `Err(_)` state).

@sabify if you distilled those MREs from a larger or more complex example, please feel free to let me know if there are additional, unaddressed issues! 

The fix for #3851 in particular is a behavior change: it effectively turns any `<ErrorBoundary/>` that contains a `<Suspense/>` in its children into a `<Suspense/>` itself, so that it can wait for the children to load in order to know whether it needs to render its error fallback. It seems to work but I imagine there are also edge cases here that I have not foreseen.

### DIFF
--- a/leptos/src/error_boundary.rs
+++ b/leptos/src/error_boundary.rs
@@ -418,7 +418,9 @@ where
                         }
                         StreamChunk::OutOfOrder { chunks } => {
                             let chunks = chunks.await;
-                            my_chunks.extend(chunks.take_chunks());
+                            my_chunks.push_back(StreamChunk::OutOfOrder {
+                                chunks: Box::pin(async move { chunks }),
+                            });
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


sabify:
Thanks for the quick fix!

The original `counter_isomorphic` example doesn't work anymore. The runtime will panic when returning `Ok(new)` as the first value. Otherwise, both issues have been resolved with this PR.

I will try another MRE after I’m able to run it normally with `Ok` as the first value.

gbj:
Oops -- I was testing with `SsrMode::Async`. Latest commit should fix out-of-order.

sabify:
I can confirm that all issues related to `ErrorBoundary` which I've encountered have been resolved with this PR. 

Thanks again!

PR #3873, opened 2025-04-22
Added header generation method to BrowserResponse

rakshith-ravi:
Comes in useful when you need headers for client side stuff (when implementing ServerFn manually).

Would something like this make sense for Leptos?


gbj:
Looks good to me, thanks!

PR #3874, opened 2025-04-24
fixed the borrow error on console

boubaca:


gbj:
To me, this looks like it would prevent a runtime panic by returning a potentially-incorrect value. Do you have an example of the problem this is intended to solve?

boubaca:
this related to https://github.com/Synphonyte/leptos-use/issues/238 and also was not properly handled

benwis:
While true that this was improperly handled, we rely on the behavior of current() here to panic if we're doing something wrong here. Returning None here is definitely incorrect. There could be an argument to add a `try_current()` method that returns a Result, but that's outside this scope. 

I believe this is a leptos-use error and so I'm closing this

PR #3881, opened 2025-04-26
Fixed AsyncTransition

audioXD:
Fixed the current implementations race condition.
```Rust
#[tokio::test]
async fn test_logic() {
    use std::time::Duration;
    use or_poisoned::OrPoisoned as _;
    use reactive_graph::transition::AsyncTransition;

    tokio::join!(
        AsyncTransition::run(move || async move {
            tokio::time::sleep(Duration::from_millis(1)).await;
            tokio::time::sleep(Duration::from_millis(1)).await;
        }),
        AsyncTransition::run(move || async move {
            tokio::time::sleep(Duration::from_millis(2)).await;
            tokio::time::sleep(Duration::from_millis(2)).await;
        }),
    );

    // Made global_transition() and TransitionInner #[doc(hidden)] for this to work
    assert!(reactive_graph::transition::global_transition().write().or_poisoned().is_none(), "A MPSC is still registered")
}
```

### DIFF
--- a/reactive_graph/src/transition.rs
+++ b/reactive_graph/src/transition.rs
@@ -1,16 +1,53 @@
 //! Utilities to wait for asynchronous primitives to resolve.
 
 use futures::{channel::oneshot, future::join_all};
-use or_poisoned::OrPoisoned;
-use std::{
-    future::Future,
-    sync::{mpsc, OnceLock, RwLock},
-};
+use pin_project_lite::pin_project;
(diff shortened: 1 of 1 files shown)


audioXD:
I just realized that we could just use a plain Vec<T> instead of mpsc now...

audioXD:
And I now see that there's an issue where it might not pierce into a nested spawn(async move {}); The old implementation kinda did that by accident (at least if the spawned task finished before the main future resolved)

Issue #3889, opened 2025-04-27
impl Into<Signal> for MappedSignal


jvdwrf:
**Describe the bug**
`MappedSignal` does not implement `Into<Signal>`.

I'm not sure if there is a valid reason for this, but updating from `beta` to `rc3` caused some miscompilations to occur. It would be nice if this conversion would work, since working with resources is a very common thing :)

gbj:
Would you like to make a PR?

gbj:
(Was quick enough just to make my own.)

PR #3897, opened 2025-04-29
feat: `impl From<MappedSignal<T>> for Signal<T>`

gbj:
I experimented with a more complicated version, which used another branch of the enum instead of using a derived signal. Pros: would support `.read()` directly without cloning, would support non-`Clone` types. Cons: additional branch logic => additional binary size, significant additional code to maintain. I think this solution is fine.


gbj:
See #3889.

PR #3899, opened 2025-04-29
Fix leptos_debuginfo

zakstucke:
Seems to have broken recently. Not sure if there's a good way to check for this in CI without increasing the matrix size @gbj @sabify? It's probably quite easy to break accidentally without some CI checking


benwis:
@gbj I think this is probably a pretty minimal penalty to enable in all cases, do you have any metrics on that?

gbj:
Sorry, I just forgot about this change @zakstucke so I'm sure that any code I've written has not included it since the original PR. It's fine to add it of course. I don't have a good idea for how to ensure it remains there in CI.

PR #3915, opened 2025-05-02
reactive_stores: implement PartialEq and Eq for Store

wrl:
StoredValue also has these implemented and does the same thing (i.e. comparing the inner `ArenaItem` directly).


wrl:
not sure what's up with the test failure – seems spurious?

gbj:
Thanks!

PR #3916, opened 2025-05-02
Check counter_isomorphic compiles with leptos_debuginfo

zakstucke:
~~Don't merge yet, making sure CI fails~~

Quick check that with `--cfg leptos_debuginfo` the counter_isomorphic example builds, which'll hopefully catch any breakages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,11 @@ jobs:
     uses: ./.github/workflows/run-cargo-make-task.yml
     with:
       directory: ${{ matrix.directory }}
+  test-leptos-debuginfo:
+    name: Test leptos_debuginfo
+    needs: [get-leptos-changed]
+    if: ${{ success() && needs.get-leptos-changed.outputs.leptos_changed == 'true' }}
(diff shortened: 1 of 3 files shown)


sabify:
@zakstucke The current implementation won't work and there is no need for another workflow to test it out. I will come up with a simpler approach PR.

sabify:
@zakstucke Please check out #3918.

zakstucke:
Thanks @sabify just commented on #3918, out of interest why wouldn't this have worked? I had checked CI runs and would fail if there was a compile error with `--cfg leptos_debuginfo` (by making it fail in a temporary commit)

sabify:
> Thanks @sabify just commented on #3918, out of interest why wouldn't this have worked? I had checked CI runs and would fail if there was a compile error with `--cfg leptos_debuginfo` (by making it fail in a temporary commit)

I meant that it won't work as expected to be part of the workflow (not correctness of running it), more emphasizing on the costly redundant workflow.

gbj:
Merged #3918, so closing this -- thanks to you both.

PR #3917, opened 2025-05-02
fix(CI): prevent regreession from nightly clippy in autofix

sabify:
Please check out https://github.com/rust-lang/rust-clippy/issues/14722.


sabify:
@gbj We have mismatch `stderr`s on stable and nightly, I just limited the macro tests on nightly compiler.

gbj:
Thanks!

PR #3921, opened 2025-05-03
Disable default features for Actix

dbanty:
The Actix integration currently pulls in a few dependencies that I don't want and I don't _think_ are required.


gbj:
I think that's probably fine, but would be a semver-breaking change so it's probably something we'll have to delay for a while, having just released a new major version.

dbanty:
@gbj I made a change so that the default features remain enabled by default, and when using the _existing_ Cargo features, but using `default-features = false` should disable the default Actix features... I think.

Let me know what you think, if we could get that into a patch of 0.8 😁

PR #3947, opened 2025-05-06
Fix cargo-leptos stylesheet caching (#3927)

### LINKED ISSUE #3927: Cargo Leptos sometimes hot-reloads an old style (caching issue)
The bug is caused by https://github.com/leptos-rs/leptos/blob/dc1885ad9222b75c5310ece1542a659c1d4f811b/leptos/src/hydration/reload_script.js#L10, the `version` identifier there only takes the milliseconds of the date. When doing a lot of changes (like for example trying to figure out which color to choose) in your styles, it eventually happens that it loads old stylesheets instead of updating them.

I think this can be fixed by using `Date.now()` instead.

luxalpa:
This should theoretically fix #3927 . I have tested that the style updating still works, however I have not tested that there's no more issues with it (as that is fairly difficult to find out). However, from the code I'm pretty confident that this should completely eliminate the issue.

### DIFF
--- a/leptos/src/hydration/reload_script.js
+++ b/leptos/src/hydration/reload_script.js
@@ -7,7 +7,7 @@ ws.onmessage = (ev) => {
 		let found = false;
 		document.querySelectorAll("link").forEach((link) => {
 			if (link.getAttribute('href').includes(msg.css)) {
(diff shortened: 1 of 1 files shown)


gbj:
Thanks, this seems better to me.

PR #3956, opened 2025-05-08
Fix deprecated parameters js warning

matchish:
Nothing but copy/paste of this fix [PR](https://github.com/leptos-rs/leptos/pull/3219). In 0.8.2 with turned on islands we have the same warning as in the PR


gbj:
Thanks!

PR #3960, opened 2025-05-09
fix: add namespace to g in svg portals (closes #3958)

### LINKED ISSUE #3958: Children of SVG Portals are not visible
**Describe the bug**
Using `<Portal is_svg=true/>` will correctly move the children to the mount target, but the children will not be visible on the page.

```jsx
    <Portal mount={svg} is_svg=true>
        // this circle will show up in the DOM tree, but it won't be rendered
        <circle r="40" cx="50" cy="50" stroke="black" fill="white" stroke-width="10"/>
    </Portal>
```

I think the cause is: Portal wraps its children in a `<g>` , but the `<g>` element is created without a namespace, so it is considered an `HTMLUnknownElement`. Changing [the Portal code](https://github.com/leptos-rs/leptos/blob/09894aaca94a76d4ee3289e20a7716ad05b00bdb/leptos/src/portal.rs#L47) to use `document().create_element_ns(Some("http://www.w3.org/2000/svg"), "g")` seems to fix the issue.

**Leptos Dependencies**

```toml
leptos = { version = "0.7.0", features = ["csr"] }
```
Also tested with:
```toml
leptos = { git = "https://github.com/leptos-rs/leptos.git", features = ["csr"] }
```

**To Reproduce**
1. Go to [this codesandbox](https://codesandbox.io/p/devbox/playground-forked-ywwzzw?file=%2Fsrc%2Fmain.rs)
2. Inspect the page to see two SVGs - The first is visible. The second uses a portal and shows nothing, yet seemingly has the same DOM structure as the first

**Screenshots**

Image

**Next Steps**
[x] I will make a PR
[ ] I would like to make a PR, but need help getting started
[ ] I want someone else to take the time to fix this

uber5001:
Should fix the issue with invisible `<g>` elements created by Portals #3958

### DIFF
--- a/leptos/src/portal.rs
+++ b/leptos/src/portal.rs
@@ -42,11 +42,15 @@ where
         let children = children.into_inner();
 
         Effect::new(move |_| {
-            let tag = if is_svg { "g" } else { "div" };
-
-            let container = document()
-                .create_element(tag)
-                .expect("element creation to work");
+            let container = if is_svg {
+                document()
+                    .create_element_ns(Some("http://www.w3.org/2000/svg"), "g")
+                    .expect("SVG element creation to work")
+            } else {
+                document()
+                    .create_element("div")
+                    .expect("HTML element creation to work")
+            };
 
             let render_root = if use_shadow {
                 container


gbj:
Thanks!

Issue #3968, opened 2025-05-11
Error in how routing handles static path segments


mskorkowski:
**Describe the bug**

If I define a routes

```
<Routes fallback=|| "Not found. fallback">
    <Route path=path!("/") view=|| view!{ "Home"} />
    <Route path=path!("/users") view=|| view!{ "Users" } />
    <Route path=path!("/users/id") view=|| view!{ "user profile" } />
    <Route path=path!("/*any") view=|| view! { <h1>"Not Found any"</h1> }/>
</Routes>
```

Table below describes which route is being selected by the router.

| Path |  Selected by leptos 0.8.2 | Expected
|:-------|:------------------|:-----------------|
| `/users` | `/users`    | `/users` |
|`/usersid` |`/users/id` | `/*any` |
|`/users/id` | `/users/id` | `/users/id` |

 Second line in table shows an issue

**Leptos Dependencies**

[package]
name = "app"
version = "0.1.0"
edition = "2021"

[dependencies]
leptos = { version = "0.8.2", features = ["csr"] }
leptos_router = { version = "0.8.2" }
leptos_meta = { version = "0.8.2" }
console_error_panic_hook = "0.1.7"

**To Reproduce**

```rust
use leptos::prelude::*;
use leptos_meta::MetaTags;
use leptos_router::components::Route;
use leptos_router::components::Router;
use leptos_router::components::Routes;
use leptos_router::path;

pub fn shell(options: LeptosOptions) -> impl IntoView {
    view! {
        <!DOCTYPE html>
        <html lang="en">
            <head>
                <meta charset="utf-8"/>
                <meta name="viewport" content="width=device-width, initial-scale=1"/>
                <AutoReload options=options.clone()/>
                <HydrationScripts options/>
                <MetaTags/>
            </head>
            <body>
                <App/>
            </body>
        </html>
    }
}

#[component]
pub fn App() -> impl IntoView {
    view! {
        <Router>
            <main>
                <Routes fallback=|| "Not found. fallback">
                    <Route path=path!("/") view=|| view!{ "Home"} />
                    <Route path=path!("/users") view=|| view!{ "Users" } />
                    <Route path=path!("/users/id") view=|| view!{ "user profile" } />
                    <Route path=path!("/*any") view=|| view! { <h1>"Not Found any"</h1> }/>
                </Routes>
            </main>
        </Router>
    }
}

pub fn main() {
    console_error_panic_hook::set_once();

    mount_to_body(App);
}
```

[...]

PR #3973, opened 2025-05-13
fix: smooshed static segments no longer matches the path #3968

mskorkowski:
Fixes matching the path containing a smooshed static segments. 

`/barbaz` will no longer match `path!("/bar/baz")`

### DIFF
--- a/router/src/matching/horizontal/mod.rs
+++ b/router/src/matching/horizontal/mod.rs
@@ -14,6 +14,22 @@ pub use static_segment::*;
 pub trait PossibleRouteMatch {
     fn optional(&self) -> bool;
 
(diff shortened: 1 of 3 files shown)


gbj (approved):
This implementation looks good to me.

Thanks for the additional comments, as well.

Per your comment [here](https://github.com/leptos-rs/leptos/issues/3968#issuecomment-2880275110), I'm pretty comfortable saying that `(StaticSegment("/"), StaticSegment("/"))` counts as "holding it wrong" and is not intended to be supported (and I'd be surprised if someone else.) I guess in theory that's a breaking change, but to me, breaking something weird that's not intended to be supported in order to fix a bug is fine.

PR #3985, opened 2025-05-16
fix: remove non-existent feature dep in leptos_macro

sabify:


gbj:
I'm a little confused: `leptos` [does have an `ssr` feature](https://github.com/leptos-rs/leptos/blob/main/leptos/Cargo.toml#L77). Am I misunderstanding?

sabify:
The `leptos` is not part of the `lepos_macro` dependencies.

gbj:
It's in `dev-dependencies` though, for tests (IIRC). I'm not sure if there's a separate way to activate features of dev dependencies, or whether it's even necessary at this point (or any point?) but I think that was the intent. Possibly due to tests in an earlier version in which `leptos_macro`'s SSR output required `leptos/ssr` also to be activated. I guess if tests currently pass that's probably just no longer the case, as I've tried to remove more of that conditional compilation over time.

sabify:
You are right. Tests are passing, so we no longer need it. I think even if we need to keep it, we may need a separate integration test crate to support that.

gbj:
Thanks, works for me.

PR #3987, opened 2025-05-17
 feat(CI): add checking minimal-versions on release

### LINKED ISSUE #3617: Check minimal versions in CI
## Problem

I saw that it's been a pain point for Leptos where forgetting (or not realizing it's needed etc) to update internal versions can cause errors for users until they run a full `cargo update`.
This was discussed here: [discord/#general/2-15-25/10:12](https://discord.com/channels/1031524867910148188/1031524868883218474/1340295652357312614)
> These sorts of errors happen when one of the crates in the workspace depends on another crate in the workspace, which adds a new function or feature, and I forget to update the version number of the dependency in the workspace `Cargo.toml`. It always works locally but fails to tell `cargo` that it needs the later version. It is annoying and my fault but doesn't show up in CI and can always be fixed with a `cargo update`

## Proposed Solution

There's this new `-Zminimal-versions` flag for cargo that might help with making sure our minimum versions are kept up to date.
https://doc.rust-lang.org/cargo/reference/unstable.html#minimal-versions
https://doc.rust-lang.org/cargo/reference/unstable.html#direct-minimal-versions

## Open Questions

- Does this work with internal workspace dependencies? Do we still want it if it doesn't?
- Do we use minimal-versions or direct-minimal-versions?
  Sounds like direct-minimal-versions is possibly going to become minimal-versions? So probably use that? I've only skimmed the issue though.

## Previous Work

Examples of this being used in GitHub actions: 
- https://github.com/tokio-rs/console/pull/596/files
  Using `cargo hack --remove-dev-deps` first in this one is interesting
- [carloskiki/leptos-icons/.github/workflows/test.yml#L47-L84](https://github.com/carloskiki/leptos-icons/blob/ccf67db72b3bef3229738dab555bbb7a0ad92e76/.github/workflows/test.yml#L47-L84)

sabify:
Closes #3617.

As a side effect of having minimal-versions checked, we should have exact versioning for all dependencies. We may include regular `cargo update` and `cargo upgrade` processes as part of the CI to check dependencies every week. This may also lead to more frequent minor releases of the leptos ecosystem to keep up with dependencies.


sabify:
You can see the action in https://github.com/leptos-rs/leptos/actions/runs/15087357300/job/42411615694?pr=3987#step:25:941.

sabify:
Please keep in mind that whoever releases the crates should first publish them to crates.io and then create a tag release commit (like `v0.8.3`) to ensure that the `minimal-versions` will work properly.

This PR is ready.

sabify:
The failed CIs are not related to this PR.

gbj:
Thanks! I think I've introduced a merge conflict while working through other PRs, and it won't let me fix it because I don't think I can edit the PR. Resolving the conflict will also let the CI run again, which should now be passing, so that can't hurt.

sabify:
I just re-added `dependabot` with weekly grouped updates. There won't be PR pollution anymore...

gbj:
Thank you!

PR #3988, opened 2025-05-17
feat: wasm-splitting library support for future cargo-leptos integration

gbj:


wasm_split/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "wasm_split"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-once-cell = { version = "0.5.3", features = ["std"] }
```
sabify, line 7: Since  PR #3987, we may want to unify all member deps in workspace root for better maintainability.

```suggestion
async-once-cell = { default-features = true, workspace = true, features = ["std"] }
```

In workspace `Cargo.toml`:
```toml
async-once-cell = { default-features = false, version = "0.5.3" }
```

wasm_split_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+base16 = "0.2.1"
+digest = "0.10.7"
+quote = "1.0.36"
+sha2 = "0.10.8"
+syn = "2.0.59"
+wasm-bindgen = "0.2.92"
```
sabify, line 12: The same as above.

gbj: Thanks! Done.

PR #3990, opened 2025-05-18
Provide error message if file hashing is enabled but no hash file is …

Bytekeeper:
If file hashing is active and one forgets to deploy the hash.txt file, no warning/error is logged. Although this can be tracked down pretty quick, a QoL improvement is to log an error.

I decided to use an error rather than a warning because the app will be basically unusable.


gbj:
Thanks!

Issue #3996, opened 2025-05-20
Custom static file handling: handle_response_inner is private, leading to excessive code duplication


NCura:
Hi! I'm using Axum with Leptos and wanted to customize how static files are served, specifically to add a `Cache-Control` header.

I saw that I could replace:

```rust
.fallback(leptos_axum::file_and_error_handler::<AppState, _>(shell))
```

with my own custom version. After checking the source, I found this helpful comment in `file_and_error_handler`:

> A reasonable handler for serving static files (like JS/WASM/CSS) and 404 errors.
> This is provided as a convenience, but is a fairly simple function. If you need to adapt it, simply reuse the source code of this function in your own application.

That’s exactly what I needed. I wanted to tweak the `get_static_file` function to include a `CACHE_CONTROL` header, so I:

1. Created a `fileserv.rs` file.
2. Copy-pasted the implementations of:

   * `file_and_error_handler`
   * `file_and_error_handler_with_context`
   * `get_static_file`

However, I quickly hit a snag, `file_and_error_handler_with_context` calls a **private** function: `handle_response_inner`. That meant I also had to copy that function, and now I’m up to 241 lines of mostly unchanged code just to make a minor adjustment to one function.

It works well, but I’m wondering:

* Is there a specific reason `handle_response_inner` is private?
* Would you accept a PR making it public so customization like mine requires less duplication?
* Or is the current expectation to copy-paste all relevant source if a change is needed?

Thanks for your time!

benwis:
I suspect it’s private mostly because we haven’t really decided that’s a public api. In a broad sense it has grown to do more over time.

Happy to accept a PR to make it public if it’s helpful to you

PR #3998, opened 2025-05-20
Make handle_response_inner public to be used in custom file_and_error (see #3996)

NCura:
See #3996.

It's my first time writing a doc comments, so I wasn't too sure of what to write 😅

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -879,7 +879,8 @@ where
     }
 }
 
-fn handle_response_inner<IV>(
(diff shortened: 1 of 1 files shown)


benwis:
So you definitely won't want to document "Made Public" in the comment, since it'll be public and that won't be helpful. Perhaps something like "Can be used in conjunction with a custom [file_and_error_handler_with_context](link here) to process an Axum [Request](link here) into an Axum [Response](link here)"

NCura:
🤦‍♂️ you are right
I improved the comment

gbj:
Thanks!

Issue #3042, opened 2024-09-28
Context provided inside suspense doesn't reach `Outlet`


zakstucke:
**Describe the bug**
Trying to implement recommended approach we were discussing, providing context inside suspend doesn't seem to work.

Tested in the `counter_isomorphic` example.

**Leptos Dependencies**
Git

**To Reproduce**
```rust
#[derive(Debug, Default, Clone)]
struct CtxTopLevel;

#[derive(Debug, Default, Clone)]
struct CtxOutsideSuspense;

#[derive(Debug, Default, Clone)]
struct CtxInsideSuspense;

#[component]
pub fn Counters() -> impl IntoView {
    provide_context(CtxTopLevel);
    view! {
        <main>
            <Router>
                <Routes fallback={|| {}}>
                    <ParentRoute
                        path={path!("/")}
                        view={AuthProvider}
                    >
                        <Route
                            path={path!("/")}
                            view={|| {
                                expect_context::<CtxTopLevel>();
                                expect_context::<CtxOutsideSuspense>();
                                expect_context::<CtxInsideSuspense>();
                                view! { <p>HOME</p> }
                            }}
                        />
                    </ParentRoute>
                </Routes>
            </Router>
        </main>
    }
}

#[component]
pub fn AuthProvider() -> impl IntoView {
    let r_auth_state = Resource::new_blocking(
        || {},
        |_| {
            // AuthManager::refresh_token()
            async { "user_details...".to_string() }
        },
    );

    provide_context(CtxOutsideSuspense);

    let inner_view = move || Suspend::new(async move {
        let _ = r_auth_state.await;
        provide_context(CtxInsideSuspense);
        view! { <Outlet /> }
    });
    view! { <Suspense>{inner_view}</Suspense> }
}
```

```
listening on http://127.0.0.1:3000
thread 'actix-rt|system:0|arbiter:1' panicked at /Users/zak/z/code/leptos/reactive_graph/src/owner/context.rs:265:9:
Location { file: "src/counters.rs", line: 54, col: 33 } expected context of type "counter_isomorphic::counters::CtxInsideSuspense" to be present
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

gbj:
The example I gave in the other issue (#3038) does work, though, which is odd. Maybe worth investigating what the difference is between those two cases?

```rust
#[component]
pub fn App() -> impl IntoView {
    view! {
        <Router>
            <Routes fallback=|| {
[...]

Issue #3729, opened 2025-03-19
Panic: the `sandboxed-arenas` feature is active, but no Arena is active, under axum/tokio multithread runtime using signals of resource


metatoaster:
**Describe the bug**

This issue was previously reported in #3671 under a faulty usage of context, but this time the `on_cleanup` is done like this:

```rust
    let ws = expect_context::<WriteSignal<Option<OnceResource<String>>>>();
    on_cleanup(move || {
        ws.try_set(None);
    });
```

The rate which this trigger is relatively low, roughly 0.06% chance using the `app.rs` below (it's been slightly "malformed" to increase the odds of failure, more on that later):

```
thread 'tokio-runtime-worker' panicked at /leptos/reactive_graph/src/owner/arena.rs:53:17:
at /leptos/reactive_graph/src/owner/arena.rs:56:21, the `sandboxed-arenas` feature is active, but no Arena is active
```

<details>
<summary><code>Stack backtrace:</code>
</summary>

```
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: reactive_graph::owner::arena::Arena::with::{{closure}}
             at /leptos/reactive_graph/src/owner/arena.rs:53:17
   3: core::option::Option<T>::unwrap_or_else
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/option.rs:1017:21
   4: reactive_graph::owner::arena::Arena::with
             at /leptos/reactive_graph/src/owner/arena.rs:52:13
   5: <reactive_graph::owner::arena_item::ArenaItem<T,S> as reactive_graph::traits::IsDisposed>::is_disposed
             at /leptos/reactive_graph/src/owner/arena_item.rs:128:9
   6: <reactive_graph::signal::write::WriteSignal<T,S> as reactive_graph::traits::IsDisposed>::is_disposed
             at /leptos/reactive_graph/src/signal/write.rs:144:9
   7: <T as reactive_graph::traits::Set>::try_set
             at /leptos/reactive_graph/src/traits.rs:547:12
   8: simple::app::__SettingSignal::{{closure}}
             at ./src/app.rs:73:9
   9: core::ops::function::FnOnce::call_once{{vtable.shim}}
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/ops/function.rs:250:5
  10: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
[...]

PR #4015, opened 2025-05-24
Fix context issues with nesting routing

gbj:
@benwis provided me another reproduction similar to the issue #3042, which I simplified to this:
```rust
#[component]
    view! {
        <Router>
            <main>
                <Routes fallback=|| "Not found">
                    <ParentRoute path=StaticSegment("") view=|| view! {
                        <Provider value=42i32>
                            <Outlet/>
                        </Provider>
                    }>
                    </ParentRoute>
                </Routes>
            </main>
        </Router>
    }
}

#[component]
    let value = use_context::<i32>();

    view! {
        <p>{format!("{value:?}")}</p>
    }
}
```

The issue here is that currently an `Outlet` is slotted into the ownership tree as the child of its parent's `view`, not in the location where the `Outlet` itself is consumed. This is important to the way `Outlet` itself works, because it uses context to determine which route it's supposed to be rendering. But it causes the issue that context provided inside something like a `Provider` or a `move || Suspend::new()` does not reach the `Outlet`, because the `Outlet` is a child of the parent `view`. So you would expect the tree to be
```
parent view
|_ Provider
  |_ Outlet
```
but really it is
```
parent view
|_ Provider
|_ Outlet
```

This PR adds a new concept to the ownership model, which is the idea of "joining owners" solely for the sake of context. All this does is allow you to add additional owners that will be searched for context. 

I then used this to join together the two owners here: the owner of the route itself, and the owner where it is used. This means that context will as you expect it, even in cases where you conditionally/dynamically render the `Outlet`, which is much better.

This would benefit from further testing but I'm about to log off to eat dinner!

cc: @benwis @zakstucke

### DIFF
--- a/reactive_graph/src/owner.rs
+++ b/reactive_graph/src/owner.rs
@@ -167,6 +167,7 @@ impl Owner {
                     .map(|parent| parent.read().or_poisoned().arena.clone())
                     .unwrap_or_default(),
                 paused: false,
+                joined_owners: Vec::new(),
             })),
             #[cfg(feature = "hydration")]
             shared_context,
@@ -201,6 +202,7 @@ impl Owner {
                 #[cfg(feature = "sandboxed-arenas")]
                 arena: Default::default(),
                 paused: false,
+                joined_owners: Vec::new(),
             })),
             #[cfg(feature = "hydration")]
             shared_context,
@@ -226,6 +228,7 @@ impl Owner {
                 #[cfg(feature = "sandboxed-arenas")]
                 arena,
                 paused,
+                joined_owners: Vec::new(),
             })),
             #[cfg(feature = "hydration")]
             shared_context: self.shared_context.clone(),
@@ -455,6 +458,7 @@ pub(crate) struct OwnerInner {
     #[cfg(feature = "sandboxed-arenas")]
     arena: Arc<RwLock<ArenaMap>>,
     paused: bool,
+    joined_owners: Vec<Owner>,
 }
 
 impl Debug for OwnerInner {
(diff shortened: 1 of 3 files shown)


benwis:
This seems to work perfectly for my test case, I look forward to not having to have this conversation again with confused developers!

lukexor:
Until this lands in a release, will the `outer_owner.with(|| provide_context(...))` work-around work for this case?

benwis:
It will work fine in this case now and after the release

metatoaster:
While testing the effect of this on #3729 I've discovered that this pull request causes a memory leak with every request.  Basically doing 1M requests on the simple example under that issue blows memory usage up to ~30GB.

I ran `git-bisect` until it hit this:

```
$ git bisect skip                                                    
There are only 'skip'ped commits left to test.                                                                  
The first bad commit could be any of:                                                                           
06dfa37eee174253b9ab5912d35e474f82520274                                                                        
5ba45bb1edc4902fca0978ae14fdb1065cc0d103                                                                        
82ccbbf806577fa5da21cf1d89d7ec227bc6dc7a                                                                        
We cannot bisect more!
```

~~All three are in here, as at this stage I didn't try to retroactively apply the typo fix to the two other commits to allow bisection be done further, I will report back.~~

I `git rebase` to reorder these these commits such that 82ccbbf806577fa5da21cf1d89d7ec227bc6dc7a is flattened onto 06dfa37eee174253b9ab5912d35e474f82520274 before 5ba45bb1edc4902fca0978ae14fdb1065cc0d103 is applied, and testing again confirms that 5ba45bb1edc4902fca0978ae14fdb1065cc0d103 is the source of the memory leak.

PR #4023, opened 2025-05-28
remove unnecessary where-clauses

lcnr:
Uncovered during a crater run for `-Znext-solver` in https://github.com/rust-lang/rust/pull/133502.

These where-clauses cause `tachys` to fail with an overflow error when compiling with the next-generation trait solver right now: https://github.com/rust-lang/trait-system-refactor-initiative/issues/89. We're still unsure whether we actually have to break this when stabilizing the new solver, so it may not be an issue in the end. This breakage is a really subtle issue and quite unfortunate.

Luckily, this PR should be neutral or even desirable by itself, so I think it's fine to merge it regardless of whether it'll be actually necessary to avoid future breakage.

In case this looks good to you, it would be awesome if you could release this change as a point-release for version 0.2 and maybe even backport it to 0.1. It would really help us with our work on the new trait solver.

:rabbit:

### DIFF
--- a/tachys/src/reactive_graph/mod.rs
+++ b/tachys/src/reactive_graph/mod.rs
@@ -615,10 +615,7 @@ macro_rules! reactive_impl {
             fn add_any_attr<NewAttr: Attribute>(
                 self,
                 _attr: NewAttr,
-            ) -> Self::Output<NewAttr>
-            where
-                Self::Output<NewAttr>: RenderHtml,
-            {
+            ) -> Self::Output<NewAttr> {
                 todo!()
             }
         }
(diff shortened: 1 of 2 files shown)


benwis:
@lcnr Love to see the work on the new trait solver and that leptos/tachys is being tested in it! I'm happy to merge and release a new version of tachys to include the changes once the CI passes. I'll ping you here once that's done

benwis:
@lcnr tachys v0.2.3 has been released with your changes included

lcnr:
Awesome, thank you!

If you've got the time, please also cherry-pick this commit on top of the last 0.1 release and do a point release for that one as well.

gbj:
@lcnr I've just released an updated 0.1.9 as well.

PR #4033, opened 2025-05-30
Update `session_auth_axum` example

gbj:
There's a bunch of ceremony in the current version of the example that is no longer necessary in the present version of the framework: specifically the custom handlers are not needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3519,7 +3519,7 @@ dependencies = [
 
 [[package]]
 name = "tachys"
-version = "0.2.2"
+version = "0.2.3"
 dependencies = [
  "any_spawner",
  "async-trait",
--- a/projects/session_auth_axum/Cargo.toml
+++ b/projects/session_auth_axum/Cargo.toml
@@ -19,9 +19,9 @@ leptos_router = { path = "../../router" }
 log = "0.4.0"
 simple_logger = "5.0"
 serde = { version = "1.0", features = ["derive"] }
-axum = { version = "0.7.0", optional = true, features = ["macros"] }
-tower = { version = "0.4.0", optional = true }
-tower-http = { version = "0.5.0", features = ["fs"], optional = true }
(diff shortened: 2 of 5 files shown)


ricketyboo:
I was just about to make a stab at updating this project and then saw you'd already done so! Thank you!

FWIW if you want a safety check, I've pulled this branch and run it everything seems okay bar one thing that is probably pre-existing

-  :heavy_check_mark: Guest can add Todo
-  :heavy_check_mark: Todo is associated to Guest when created by Guest
-  :heavy_check_mark: Guest can delete Todo
-  :heavy_check_mark: Guest can create an account
-  :heavy_check_mark: Guest can log into to account
-  :heavy_check_mark: Logged in user can add Todo
-  :heavy_check_mark: Todo is associated to logged in user when created by Guest
-  :heavy_check_mark: Logged in user can delete Todo
-  :heavy_check_mark: Users in different browsers can interact with the same running app and see each others updates after refresh
- :question: Remember me session cookie is set to expire ~3 months from today regardless of what was selected. This means closing the browser doesn't actually log the user out as the cookie doesn't clear and the session is kept alive on reopening. I suspect this isn't due to your changes and is just a configuration/implementation detail with axum_session_auth itself but I can't actually run the prior code to check so just thought I'd mention it!

gbj:
Thanks! I'm just going to leave it as the "Remember me" session cookie as is, because I didn't touch it here, and am just trying to update the preexisting example that someone else contributed. If you want to make a PR to fix it that's very welcome!

ricketyboo:
After digging into axum_session, and spending some more time thinking about the UX around dropping a session  on browser close vs aligning to server session timeout, I've decided this is really dependent on application needs (security vs user convenience) and there's no "right" pattern here.

Juggling the conditionals to deal with server session time **and** a browser session is probably more complicated than this example needs to be so I won't touch it, as the project is doing its job fine as-is.

PR #4035, opened 2025-05-31
Remove unnecessary Option wrapping

mohe2015:


mohe2015:
It would be great if somebody could take a look!

benwis:
Hi @mohe2015! I don't see anything wrong with this, we'll see what the tests say. Because this changes our API, I imagine it'll have to wait for a breaking release. Thanks for the PR

mohe2015:
Hi @benwis, I could not find where this changes the API in a breaking way. Everything should be (indirectly) `pub(crate)` or private. `cargo semver-checks` seems to agree (if I used it correctly)

benwis:
Looking at this a bit more closely, I think you're right. I'm fine with merging and then we'll see if anything shakes up.
Thanks for the PR!

PR #4056, opened 2025-06-08
Bugfixes to `reactive_stores`

elias098:
Fixes the following:
- `StoreField::triggers_for_path`: Too small vec capacity calculation potentially causing reallocations
- `StoreField::triggers_for_path`: Duplicate trigger when path is empty
- `Subfield::track_field`: Duplicate track on `this` trigger

Also rewrites `Subfield::track_field` to skip 1 `get_trigger`


gbj:
Thanks!

Issue #3372, opened 2024-12-16
Cleanup "Resource B" depends on "Resource A"


dev-inigmas:
**Is your feature request related to a problem? Please describe.**

The Leptos 0.7 Release Notes shows:

```rust
let user = Resource::new(|| (), |_| user_id());
let posts = Resource::new(
    move || user.get(),
    move |_| async move {
        let user = user.await?;
        get_posts(user).await
    },
);

view! {
    <Suspense>
        ...
```

However, this causes Leptos to throw warnings in development mode - which may obscure warnings that a developer would need to see.

Putting the Resource definitions in a `<Suspense>` causes hydration issues.
Using `.track()` rather than `.get()` also doesn't work, because Resources are memoized.

**Describe the solution you'd like**

Either squelch warnings emitted while using this pattern. Or, allow `.track()` to be sufficient within the `source` function. Maybe, there could be a different Resource generating function that doesn't require unique return values from the `source` function for memoization.

**Describe alternatives you've considered**

I guess I could return a random value from the `source` function...

dev-inigmas:
This hack appears to work. I suspect that's possible for it to not work perfectly every time though:

```rust
  let my_resource: Resource<MyResource> = Resource::new(
    move || {
      resource_a.track();
      resource_b.track();
      RandomState::new().build_hasher().finish()
    },
    move |_| fetch_my_resource(resource_a, resource_b),
  );
```

gbj:
The correct answer here in the short term is definitely just to suppress the false-positive warning. (My reading of the issue is that the example in the release notes works as expected, but issues a warning, which is why I say "false positive." Let me know if this is incorrect.)

In the medium-to-long term removing the memoization in resources so that `.track()` can work may be the right idea (the current API has generated some confusion, see e.g., #3308), but it would be a breaking change so couldn't happen until 0.8 in any case.

dev-inigmas:
I'm interested in contributing. Though my time is very limited. I'm also hoping to reproduce some errors for other tickets.

I use Leptos at work - and right now, management is much more interested in getting new features delivered. So, I've been prioritizing finding workarounds for issues as they come up. I got things working well enough that I merged my migration into `main`.

Maybe I can get some work done for Leptos itself this weekend.

Issue #3308, opened 2024-11-30
Regression in Resource reloading on dep change


zakstucke:
**Describe the bug**
82526559595bee3b487038c9a4bbc80ab90a2a51 causes a regression in the following test:

```rust
    use leptos::prelude::*;
    
    use any_spawner::Executor;
    use hydration_context::SsrSharedContext;
    use std::sync::atomic::AtomicUsize;
    use std::sync::atomic::Ordering;
    use std::sync::Arc;

    #[tokio::test]
    async fn test_resource_busting() {
        _ = Executor::init_tokio();
        let ssr_ctx = Arc::new(SsrSharedContext::new());
        let owner = Owner::new_root(Some(ssr_ctx.clone()));
        owner.set();

        let buster = RwSignal::new(1);
        let fetch_calls = Arc::new(AtomicUsize::new(0));
        let resource = Resource::new(
            move || {
                buster.track();
            },
            {
                let fetch_calls = fetch_calls.clone();
                move |_: ()| {
                    fetch_calls.fetch_add(1, Ordering::Relaxed);
                    async move { 2 }
                }
            },
        );
        assert_eq!(resource.await, 2);
        assert_eq!(fetch_calls.load(Ordering::Relaxed), 1);

        Executor::tick().await;

        // Shouldn't recall despite ticking:
        assert_eq!(resource.await, 2);
        assert_eq!(fetch_calls.load(Ordering::Relaxed), 1);

        buster.set(2);
        Executor::tick().await;

        // Was busted so should've refetched:
        assert_eq!(resource.await, 2);
        assert_eq!(fetch_calls.load(Ordering::Relaxed), 2); // FAIL: assertion `left == right` failed. left: 1, right: 2
    }

```
This test passes before this commit, as I think it should. This commit causes it to fail on the last line with:
```
assertion `left == right` failed
 left: 1
 right: 2
```

gbj:
This is actually *not* intended to work; I think it even the "working" version would only "bust" the resource once. 

Resources memorize the first argument so simply tracking a signal inside it won't work. Return the value of that signal and it should (.get() instead of .track())

gbj:
See #2661 as well -- I think what you're saying here is the same as that idea, right?

zakstucke:
Thanks Greg, memoization over tracking raw signals makes sense for resources.

Updating the test to `get()` and return the buster, rather than just tracking it, fixes the issue.

Issue #2661, opened 2024-07-01
Trigger source won't cause a Resource to refetch


cs2dsb:
**Describe the bug**
Passing a trigger as the source of a Resource  doesn't work because Resource creates a memo from the source. 

**Leptos Dependencies**
```toml
leptos_reactive = { version = "0.6.12", features = ["csr", "nightly"] }
```

**To Reproduce**
```rust
use leptos_reactive::{create_resource, create_signal, create_trigger, SignalUpdate};

fn main() {
    let (sig, set_sig) = create_signal(0_usize);
    let trigger = create_trigger();
    
    let _sig_resource = create_resource(
        move || sig(), 
        |_| async move { 
            println!("signal resource fetched");
        }
    );

    let _trigger_resource = create_resource(
        move || trigger(),
        |_| async move { 
            println!("trigger resource fetched");
        }
    );
    println!();
    
    println!("setting signal");
    set_sig.update(|v| *v += 1);
    println!();

    println!("setting trigger");
    trigger.notify();
}
```

**Expected behavior**
Expected to be able to trigger a resource to refetch using Trigger::notify()

**Additional context**
It seems more like a usability/documentation issue than an actual bug. It makes perfect sense when I went and looked at the code of the Resource but it wasn't the behaviour I expected. 

As a minimum, updating the documentation of Resource/create_resource to point out that the source is memoized would be good.

One possible way to fix it (assuming you agree Triggers should always trigger Memos and Resources) would be to have the Trigger return a private type that always returns false to PartialEq.

gbj:
As you note this is 100% a  documentation issue; resource memoize the input signal. In fact this is not specific to triggers: since calling `.track()` on any signal returns `()`, this would be true for any other `signal.track()` used in the input signal. The canonical solution is to trigger resources with some kind of incrementing `usize` signal instead. 

The proposed solution to return a ZST that is always `!=` to itself from `Track` is an interesting one. I'd have to think about what other things it might break, of course -- and it would definitely be a breaking change that may affect user code.

PR #4061, opened 2025-06-09
fix: do not track anything inside the `async` block of a `Resource` (closes #4060)

### LINKED ISSUE #4060: Resource that depends on another resource that has an `.await` point is executed twice
**Describe the bug**
Initially it was thought that it was any resources with dependencies on most other resources, then dependencies on resources that uses server functions, I seem to have isolated the issue down to resources that have an `.await` point.

**Leptos Dependencies**

Tested against main branch.

**To Reproduce**

Simply render the following component in either SSR or CSR:

```rust
#[server]
async fn hello() -> Result<String, ServerFnError> {
    tokio::time::sleep(std::time::Duration::from_millis(1)).await;
    Ok("Hello world!".to_string())
}               

#[component]
pub fn Activate() -> impl IntoView {
    let server_hello = Resource::new_blocking(
        || (),
        move |_| async move {
            let result = hello().await;
            leptos::logging::log!("server_hello: {result:?}");
            result
        },
    );

    let res = Resource::new(
        || (),
        move |_| {
            async move {
                let result = server_hello.await.ok();
                leptos::logging::log!("res: {result:?}");
                result
            }
        }
    );

    let res2 = Resource::new(
        || (),
        move |_| {
            async move {
                let result = res.await;
                leptos::logging::log!("res2: {result:?}");
                result
            }
        }
    );

    let local_hello = Resource::new_blocking(
        || (),
        move |_| async move {
            // emulating a server function only here
            // #[cfg(feature = "ssr")]
            // tokio::time::sleep(std::time::Duration::from_millis(1)).await;
            let result = Some("Hello world!".to_string());
            leptos::logging::log!("local_hello: {result:?}");
            result
        },
    );

    let res3 = Resource::new(
        || (),
        move |_| {
            async move {
                leptos::logging::log!("res3: waiting on local_hello");
                let result = local_hello.await;
                leptos::logging::log!("res3: {result:?}");
                result
            }
        }
    );

    view! {
        <Suspense>{
            move || {
                Suspend::new(async move {
                    res2.await;
                    res3.await;
                })
            }
        }</Suspense>
        <p>"Goodbye!"</p>
    }
}
```

This logging output is produced:

```
local_hello: Some("Hello world!")
res3: waiting on local_hello
res3: Some("Hello world!")
server_hello: Ok("Hello world!")
res: Some("Hello world!")
res: Some("Hello world!")
res2: Some("Hello world!")
res2: Some("Hello world!")
```

I would have expected `res` and `res2` to not be executed twice (demonstrated by duplicated logging output) given no original inputs should have signaled any changes, much like `res3` only got executed once.

However, allowing the await sleep to happen under `local_hello` yields an interesting effect - now `res3` also gets executed twice.  An example run (only applicable under SSR given the feature gate on the `sleep`).

```
res3: waiting on local_hello
local_hello: Some("Hello world!")
server_hello: Ok("Hello world!")
res: Some("Hello world!")
res3: Some("Hello world!")
res: Some("Hello world!")
res3: waiting on local_hello
res3: Some("Hello world!")
res2: Some("Hello world!")
res2: Some("Hello world!")
```

gbj:
Resources are not supposed to track anything inside the `async` block: rather, anything you need to track is tracked in the first argument. This allows us to only track the first argument on the client, and not run the `async` block again until something in the first argument has changed.

However, #4060 shows that when you read from another resource via `.await`, and that first resource is not actually ready, it was actually getting tracked and causing the second (dependent) resource to re-run when the first resource completed, which is not correct.

This corrects that behavior, fixing the problem in #4060.

I think this falls into the awkward category in which it 1) fixes a bug, 2) doesn't change the type/won't break compilation of existing code, but 3) could hypothetically break (incorrect-but-compiling) existing code that depends on the bugged behavior. However I think it's worth making the fix (assuming I didn't break something else obvious), because you almost certainly don't actually want to trigger those `async` blocks multiple times unnecessarily, as they are typically used for something that has real side effects (like calling a server function, etc.)

### DIFF
--- a/reactive_graph/src/computed/async_derived/arc_async_derived.rs
+++ b/reactive_graph/src/computed/async_derived/arc_async_derived.rs
@@ -518,7 +518,9 @@ impl<T: 'static> ArcAsyncDerived<T> {
     {
         let fun = move || {
             let fut = fun();
-            async move { SendOption::new(Some(fut.await)) }
+            ScopedFuture::new_untracked(async move {
+                SendOption::new(Some(fut.await))
+            })
         };
         let initial_value = SendOption::new(initial_value);
         let (this, _) = spawn_derived!(
--- a/reactive_graph/src/computed/async_derived/mod.rs
+++ b/reactive_graph/src/computed/async_derived/mod.rs
@@ -42,6 +42,18 @@ impl<Fut> ScopedFuture<Fut> {
             fut,
         }
     }
+
+    /// Wraps the given `Future` by taking the current [`Owner`] re-setting it as the
+    /// active owner every time the inner `Future` is polled. Always untracks, i.e., clears
+    /// the active [`Observer`] when polled.
+    pub fn new_untracked(fut: Fut) -> Self {
+        let owner = Owner::current().unwrap_or_default();
+        Self {
+            owner,
+            observer: None,
+            fut,
+        }
+    }
 }
 
 impl<Fut: Future> Future for ScopedFuture<Fut> {


metatoaster:
I have used resource chaining in suspense tests (hence the fails), but I am going to need to make sure I didn't depend on the incorrect behavior there also.  In any case I had planned to provide tests for this change, I am going to go ahead to review `suspense_tests` and do a pr for that (plus additional tests).

gbj:
You are the best!

metatoaster:
So for the sake of completeness: If there are two resources, with resource `a` provides some data that resource `b` might use, should it look something like this?

```rust
let (rv, sv) = signal(0);
let res_a = Resource::new(
    move || rv.get(),
    move |c| async move { 
        server_call(c).await
    } 
);
let res_b = Resource::new(
    move || res_a.track(),
    move |_| async move { 
        leptos::logging::log!("res_b running");
        let result = res_a.await.ok()?;
        ws.set(Some(result.clone()));
        Some(result)
    } 
);
```

However, that doesn't actually get `res_b` to track `res_a` (pushing the button in the final example caused the re-fetch, but a simple `.track()` does not work).  I could call `res_a.track()` inside the `async` block, it would result in `res_b running` printed twice (back to where we started).  Likewise we could do something like:

```rust
let res_b = Resource::new(
    move || res_a.get(),
    move |a| async move { 
        leptos::logging::log!("res_b running");
        let result = a?.ok()?;
        ws.set(Some(result.clone()));
        Some(result)
    } 
);
```

Which has the equivalent behavior, though not kosher given the info message about `reading a resource in `hydrate` mode outside a <Suspense/> or <Transition/> or effect`.  I guess what I am after is being able to only track resources that are in fact ready, maybe have that signal emit on ready.  Essentially what I am after is this ability under SSR (conceivably, it doesn't really be limited to SSR):

-  a resource that is pending shouldn't trigger a dependent resource to run.
-  a resource that is ready can then signal/trigger its dependent resources.

Anyway, the final example that I want to get `res_b` reactive in the desired manner:

```rust
#[server]
async fn server_call(count: u64) -> Result<String, ServerFnError> {
    tokio::time::sleep(std::time::Duration::from_millis(1)).await;
    Ok(format!("Hello {count}!"))
}

#[component]
pub fn Signal() -> impl IntoView {
    let (rv, sv) = signal(0);
    let res_a = Resource::new(
        move || rv.get(),
        move |c| async move {
            server_call(c).await
        }
    );
    let res_b = Resource::new(
        // I only want to track, only run the fetcher below only once when `res_a` is no longer pending
        move || res_a.track(),
        move |a| async move {
            leptos::logging::log!("res_b running");
            let result = res_a.await.ok()?;
            Some(result)
        }
    );

    view! {
        <div>
            {move || view! { <span>"count: " {rv.get()}</span> }}
            <button on:click=move |_| sv.update(|value| *value += 1)>"+1"</button>
        </div>
        <Suspense>{
            move || Suspend::new(async move {
                res_b.await
            })
        }</Suspense>
    }
}
```

metatoaster:
Durrrr I am dumb, I forgot feature gating is a thing, this does exactly what is desired:

```rust
let res_b = Resource::new(
    move || (),
    move |_| { 
        #[cfg(not(feature = "ssr"))]
        res_a.track();
        async move { 
            leptos::logging::log!("res_b running");
            let result = res_a.await.ok()?;
            ws.set(Some(result.clone()));
            Some(result)
        } 
    } 
);
```

My suspicion is that multiple runs of the same Resource especially under SSR causes the weirdness that was previously reported, and this is a good enough workaround that I think solves my particular use case, but nonetheless there could still be some breakage that are obvious that may still lurk...

gbj:
The first argument of a Resource is memoized, so calling `.track()` won't do it -- you do need to return a value there, instead of `()` (see #2661, #3308)

The warning in your second version (with `.get()`) is a false positive (see #3372); this *is* how you're supposed to depend on one resource in another, if you want to respond to changes in the first one.

Setting a write-signal in the `async` block seems odd to me -- unless the intended behavior is that this write-signal is *only* set on the server. (The `async` block will not run on the client during the initial load.)

metatoaster:
> Setting a write-signal in the `async` block seems odd to me -- unless the intended behavior is that this write-signal is _only_ set on the server. (The `async` block will not run on the client during the initial load.)

The reason behind that is due to how `Suspense` don't run in parallel and so it wouldn't run to write a value which something earlier in the view tree may depend on, hence this forces the usage of signal being set on the server first.  That said I left that in unintentionally for the example snippets of just `res_b` as that's an unrelated test to the one I am reporting here.

> The warning in your second version (with `.get()`) is a false positive (see #3372); this _is_ how you're supposed to depend on one resource in another, if you want to respond to changes in the first one.

This re-introduces the behavior where both the `None` and `Some(T)` triggers execution under SSR. My suspicion is that if a resource is executed twice which not only causes unwanted side-effects, it actively messes up SSR output under a work-stealing task scheduler.

With #4061 in place, with the full example from above ([in this comment](#issuecomment-2958934649)), changing `res_b` resource to this:

```rust
    let res_b = Resource::new(
        move || res_a.get(),
        move |a| {
            async move {
                let result = a?.ok()?;
                Some(result)
            }
        }
    );
```

Using a single instance of my test request program, 10k requests, we get the following output sizes:

```
sizes = {17378: 1092, 17386: 471, 17391: 11, 17426: 8398, 17434: 28}
```

Same program, without work stealing (i.e. `#[tokio::main(flavor = "current_thread")]`):

```
sizes = {17426: 10000}
```

Anyway, `17378` length output missing `Hello 0!` output,  `17386` causes hydration error.  This does look to me having resources re-run under SSR in a work-stealing task scheduler is problematic.

PR #4062, opened 2025-06-10
Updating/adding tests for #4061

metatoaster:


benwis:
Always appreciate new tests for interesting cases. I'll let @gbj review the content of this, it looks good to me

PR #4069, opened 2025-06-12
enh: hygienize usages of re-exports

mondeja:
The crate `leptos` re-exports several dependencies to use them in the `#[component]` macro:

- `wasm_bindgen`
- `web_sys` - There is a direct usage of it [in an example](https://github.com/leptos-rs/leptos/blob/671ada36ab95217dd19f661144e706d990df873c/examples/directives/src/lib.rs#L52-L64).
- `tracing`
- `serde`
- `serde_json`

This is problematic because it could hide the definition of direct dependencies of a project in Leptos, which can cause hard to debug issues with version mismatches of duplicated dependencies and duplication of dependencies itself. Additionally, it confuses to developer consumers of Leptos about if they must use, for some reason, the dependencies from this path.

Leptos has an example with this problem going unnoticed that serves to understand the problem.

1. Comment the line `use leptos::wasm_bindgen::JsCast;` in `examples/directives/src/lib.rs`.
2. Run `trunk serve` in the `examples/directives` directory.

Rustc will show an error like this:

```text
error[E0599]: no method named `unchecked_into` found for struct `Element` in the current scope
  --> src/lib.rs:54:17
   |
54 |     let el = el.unchecked_into::<web_sys::HtmlElement>();
   |                 ^^^^^^^^^^^^^^
   |
  ::: /.../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasm-bindgen-0.2.100/src/cast.rs:78:8
   |
78 |     fn unchecked_into<T>(self) -> T
   |        -------------- the method is available for `Element` here
   |
   = help: items from traits can only be used if the trait is in scope
help: trait `JsCast` which provides `unchecked_into` is implemented but not in scope; perhaps you want to import it
   |
1  + use leptos::wasm_bindgen::JsCast;
   |
help: there is a method `unchecked_into_f64` with a similar name
   |
54 |     let el = el.unchecked_into_f64::<web_sys::HtmlElement>();
   |                               ++++
```

Note that the error message suggests to import `JsCast` from `leptos::wasm_bindgen`, which is not a good suggestion for the reasons explained above.

The solution included in this PR is to hide re-exports of `leptos` in a `__reexports` internal public module. This is the solution suggested in <https://github.com/rust-lang/rust/issues/63687#issuecomment-526254208> for hygienic usage of libraries in proc-macro generated code.

> There is a pattern to use std or any other third party crate in a hygienic way. You define a `#[doc(hidden)] pub mod __reexports { pub use std; }` in lib.rs and then use `$crate::__reexports::std::...` to refer to std items in the macro.

You can see linked PRs in the issue of the linked comment examples of the same solution in other libraries. Are called different to `__reexports`, sometimes `__export` (liballoc), `__macro_support`... but it is consistent that rustc will not suggest to import from a module starting with `__`.

### DIFF
--- a/examples/directives/Cargo.toml
+++ b/examples/directives/Cargo.toml
@@ -9,8 +9,8 @@ log = "0.4.22"
 console_log = "1.0"
 console_error_panic_hook = "0.1.7"
 web-sys = { version = "0.3.70", features = ["Clipboard", "Navigator"] }
+wasm-bindgen = "0.2.93"
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.42"
-wasm-bindgen = "0.2.93"
 web-sys = { version = "0.3.70", features = ["NodeList"] }
--- a/examples/directives/src/lib.rs
+++ b/examples/directives/src/lib.rs
@@ -50,7 +50,7 @@ impl From<()> for Amount {
 }
 
 pub fn add_dot(el: Element, amount: Amount) {
-    use leptos::wasm_bindgen::JsCast;
+    use wasm_bindgen::JsCast;
     let el = el.unchecked_into::<web_sys::HtmlElement>();
 
     let handle = el.clone().on(click, move |_| {
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -331,17 +331,20 @@ pub mod task {
     };
 }
 
-// these reexports are used in islands
-#[cfg(feature = "islands")]
+// Re-exports used in macros. Don't use these directly in your code.
 #[doc(hidden)]
-pub use serde;
-#[cfg(feature = "islands")]
-#[doc(hidden)]
-pub use serde_json;
-#[cfg(feature = "tracing")]
-#[doc(hidden)]
-pub use tracing;
-#[doc(hidden)]
-pub use wasm_bindgen;
-#[doc(hidden)]
-pub use web_sys;
+pub mod __reexports {
+    #[cfg(feature = "islands")]
+    #[doc(hidden)]
+    pub use serde;
+    #[cfg(feature = "islands")]
+    #[doc(hidden)]
+    pub use serde_json;
+    #[cfg(feature = "tracing")]
+    #[doc(hidden)]
+    pub use tracing;
+    #[doc(hidden)]
+    pub use wasm_bindgen;
+    #[doc(hidden)]
+    pub use web_sys;
+}
--- a/leptos_dom/src/macro_helpers/tracing_property.rs
+++ b/leptos_dom/src/macro_helpers/tracing_property.rs
@@ -4,8 +4,8 @@ use wasm_bindgen::UnwrapThrowExt;
 /// Use for tracing property
 macro_rules! tracing_props {
     () => {
-        ::leptos::tracing::span!(
-            ::leptos::tracing::Level::TRACE,
+        ::leptos::__reexports::tracing::span!(
+            ::leptos::__reexports::tracing::Level::TRACE,
             "leptos_dom::tracing_props",
             props = String::from("[]")
         );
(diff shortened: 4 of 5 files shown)


gbj:
This is also a breaking change to existing user code that uses the re-exports, so that is an important trade-off to consider in removing the re-exports, rather than making a decision about what would be ideal if we were starting from scratch.

mondeja:
> This is also a breaking change to existing user code that uses the re-exports, so that is an important trade-off to consider in removing the re-exports, rather than making a decision about what would be ideal if we were starting from scratch.

So it's a breaking change, I'm OK with it.

gbj:
Maybe you could expand on the reasoning in favor of the change, then? I don't understand either of the problems you describe.

Specifically:

> This is problematic because it could hide the definition of direct dependencies of a project in Leptos, which can cause hard to debug issues with version mismatches of duplicated dependencies 

Of the 5 you list here (`serde`, `serde_json`, `tracing`, `web-sys`, `wasm-bindgen`) none have had a new major version/semver-breaking release in many years, as far as I know. Are you thinking of a particular example in which there has been a version mismatch with one of these crates? They are very stable. Worst case, upgrading a major version of one of them would require a version bump. (Like what happens when `axum` bumps its version and we need to release a new `leptos_axum` version.)

> and duplication of dependencies itself. 

`cargo` will deduplicate the dependencies unless there are two different major versions required (which hasn't happened in the framework's history -- see above)

> Additionally, it confuses to developer consumers of Leptos about if they must use, for some reason, the dependencies from this path.

I'm not sure I understand the confusion. In the example you give, you remove the line `use leptos::wasm_bindgen::JsCast;` and the compiler suggests you add the line `use leptos::wasm_bindgen::JsCast;` This seems pretty straightforward to me. (In fact, it seems better to me than requiring a user to `cargo add wasm_bindgen` in order to do something that's relatively common, but maybe that's a matter of opinion.)

***

If I were creating the library again in a vacuum, would it be worth `#[doc(hidden)] pub mod __reexports { pub use std; }`? Maybe! But I don't understand why they are bad enough that we should remove them, when they've all been reexported since 0.1.

mondeja:
> This seems pretty straightforward to me. (In fact, it seems better to me than requiring a user to `cargo add wasm_bindgen` in order to do something that's relatively common, but maybe that's a matter of opinion.)

That is a matter of opinion, it's the whole point of this PR.

If the framework wants to bundle in their public API `wasm_bindgen` or `web_sys`, like `web_sys` itself does with `wasm_bindgen` and `js_sys`, it should be documented or we'll end increasing the maintainment efforts for things so much naive like knowing what dependency is using an app. Especially for locked installs of apps that are pinning to an exact version of these dependencies, which is not possible to do so if someone is using a dependency like `wasm_bindgen` from leptos in a third party library.

> Are you thinking of a particular example in which there has been a version mismatch with one of these crates? They are very stable.

Yes, the web is plenty of cases where wasm-bindgen has been pinned to a specific version in *Cargo.toml* files to solve a problem because the bidgen format is unstable and all releases are like major version changes. See https://stackoverflow.com/questions/76452749/leptos-with-axum-issue-with-wasm-bindgen-versions If wasm-bindgen is used in an application from leptos and as direct dependency when leptos is pinning it to `^0.2.100` currently, how could will not end the bundle with two duplicated versions?

I don't really care so much. These dependencies are not there because is easier to import from leptos itself than installing them properly. Are there because were needed to be added for generating proc-macro code in an unhygienic way. If you're fine with keeping them reexported in that way, I'm OK, and will write a lint to forbid leptos reexports because I don't want to deal with these maintainment nightmares. I was going to write the lint at first, just opened this in case it could be solved in a better way. Feel free to close.

mondeja:
For the people interested on this, I've written a lint to forbid usages of these re-exports from leptos. See [`leptos_reexports`](https://github.com/mondeja/rs-apps-lints/tree/main/lints/leptos-more/leptos_reexports#readme) of [mondeja/rs-apps-lints](https://github.com/mondeja/rs-apps-lints/) and https://github.com/simple-icons/simple-icons-website-rs/pull/381

PR #4074, opened 2025-06-13
Fix unresolved path in server side redirect

### LINKED ISSUE #4054: Diferent behaviour in SRR and CSR redirect from page without trailing slash to relative page
**Describe the bug**
When using the `Redirect` component with a relative link, in SSR, on a page that has no trailing slash (e.g. you call `your.site/redirect` with redirect to `this/subpage`) you get send to `your.site/this/subpage`. In CSR you get send to `your.site/redirect/this/subpage`

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.8.2", features = ["nightly"] }
leptos_axum = { version = "0.8.2", optional = true }
leptos_meta = { version = "0.8.2"}
leptos_router = { version = "0.8.2"}
```

**To Reproduce**
1. Use the start-axum template
2. Create a route with a redirect to a relative subroute (e.g. route `/redirect` to `this/subpage`)
3. Create a link to `/redirect` from the homepage
4. Open the page on `/redirect`
5. Observe the url changing to `/this/subpage`
6. Open the page on `""`
7. Click on a link to `/redirect`
8. Observe the url chaning to `/redirect/this/subpage`

**Next Steps**
- [ ] I will make a PR

**Additional context**
Looking at the underlying redirect implementation for SSR, it just sets the `Location` header, so I'm not sure how / if to fix it. However the SSR and CSR results still differ, which they definitely shouldn't.

TERRORW0LF:
Fixes #4054 by resolving relative paths before parsing them to the redirect function. As `use_matched` returns the route with the base, we can get away with setting the base to `""` when calling  `resolve_path`.

### DIFF
--- a/router/src/components.rs
+++ b/router/src/components.rs
@@ -4,7 +4,7 @@ use crate::location::RequestUrl;
 pub use crate::nested_router::Outlet;
 use crate::{
     flat_router::FlatRoutesView,
-    hooks::use_navigate,
+    hooks::{use_matched, use_navigate},
     location::{
         BrowserUrl, Location, LocationChange, LocationProvider, State, Url,
     },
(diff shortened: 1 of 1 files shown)


gbj:
Looks good, thanks!

PR #4090, opened 2025-06-16
Fix compilation because of concurrent conflicting change

mohe2015:
https://github.com/leptos-rs/leptos/pull/4035 and https://github.com/leptos-rs/leptos/pull/4074 were conflicting


gbj:
Thanks!

PR #4114, opened 2025-06-27
A place to put e2e tests for regression, plus reporting issue caused by #4091.

metatoaster:
Before I get into what's about the title of this, there is an another reason for this pull request. I am hoping to see that there may be a place to put regression tests in a more formal way, and since `suspense_test` is already a precedent I am more or less following on that idea.  This pull request with the commits roughly gives an idea on how might these tests be added/named, but I supposed a convention can be formalized as more of these are collected.

Now for the subject.  In #4091 the fixes were quite useful, however, b37900ec556b325035a524946e4ea3db08c16968 happens to introduce a change that breaks `on_cleanup` from being run under CSR, as demonstrated by the feature included with this regression test, e.g

```
  Scenario: The result returns to empty due to on_cleanup | Retry attempt: 1/2
   ✔  Given I see the app
   ✔  And I can access regression test 4091
   ✔  When I select the following links
       | test1     |
       | 4091 Home |
   ✘  Then I see the result is empty
      Step failed:
      Defined: features/pr_4091.feature:16:9
      Matched: tests/fixtures/world/check_steps.rs:5:1
      Step panicked. Captured output: assertion failed: `(left == right)`
      
      Diff < left / right > :
      <Test1
```

if the `on_cleanup` had ran, it would have popped that extra text  off and reverted the result to the starting state (which is an empty string).

Applying this patch will correct the above, as the following reverts b37900ec556b325035a524946e4ea3db08c16968:

```diff
diff --git a/router/src/nested_router.rs b/router/src/nested_router.rs
index e2de2c49..2b351dd1 100644
--- a/router/src/nested_router.rs
+++ b/router/src/nested_router.rs
@@ -824,7 +824,7 @@ where
                                     let matched = matched.clone();
                                     Suspend::new(Box::pin(async move {
                                         let view = SendWrapper::new(
-                                            owner_where_used.with(|| {
+                                            owner.with(|| {
                                                 provide_context(child.clone());
                                                 provide_context(params);
                                                 provide_context(url);
```

### DIFF
--- a/examples/regression/e2e/Cargo.toml
+++ b/examples/regression/e2e/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "regression_e2e"
+version = "0.1.0"
+edition = "2021"
+
+[dev-dependencies]
+anyhow = "1.0"
+async-trait = "0.1.81"
+cucumber = "0.21.1"
+fantoccini = "0.21.1"
+pretty_assertions = "1.4"
+serde_json = "1.0"
+tokio = { version = "1.39", features = ["macros", "rt-multi-thread", "time"] }
+url = "2.5"
+
+[[test]]
+name = "app_suite"
+harness = false    # Allow Cucumber to print output instead of libtest
--- a/examples/regression/e2e/Makefile.toml
+++ b/examples/regression/e2e/Makefile.toml
@@ -1,0 +1,20 @@
+extend = { path = "../../cargo-make/main.toml" }
+
+[tasks.test]
+env = { RUN_AUTOMATICALLY = false }
+condition = { env_true = ["RUN_AUTOMATICALLY"] }
+
+[tasks.ci]
+
+[tasks.test-ui]
+command = "cargo"
+args = [
+  "test",
+  "--test",
+  "app_suite",
+  "--",
+  "--retry",
+  "2",
+  "--fail-fast",
+  "${@}",
+]
--- a/examples/regression/e2e/README.md
+++ b/examples/regression/e2e/README.md
@@ -1,1 +1,34 @@
-# TODO
+# E2E Testing
+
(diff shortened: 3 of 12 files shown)


gbj:
Awesome, thanks! I have historically been very bad about adding regression tests and done a lot of it manually, which scales pretty poorly. I have made a fix for this bug, and will try to include additional regression tests for the issues the PR was supposed to address in the first place!

metatoaster:
I actually found another thing broken from that pull request but I haven't took the opportunity to bisect this particular one, but this can be reproduced by starting from the main page, click on 4091, hit back, hit forward, and somehow only whatever inside the `<Outlet/>` is rendered.  I don't think that is related to #4091 but if I have time I can bisect this, but for now I will make an additional pull request to add a couple more QoL helpers to make building further e2e a bit easier.

gbj:
Thanks. No need to bisect--the PR was really a series of WIP commits so I don't think finding which one was responsible is as useful as providing a test case for it would be! (If not that then I can try to construct one)

gbj:
@metatoaster after tearing my hair out (metaphorically) for several hours (literally) I tested a hunch and found that the back/forward issue can be toggled by switch from debug to release/by toggle `disable-erase-components` from `false` to `true`. i.e., it's some bug in one of the type-erased types. I don't think it has to do with the router, which suggests it may not have to do with that PR. I've added a (failing) test for it to #4115 and will try to minimize it so I can rule out the router and then fix it.

metatoaster:
Oof.  Then I guess that other `Suspense` bug that I encountered under release mode (that I also referenced in the design-internals chat) may also actually that.  I wish I had the time to provide further tests towards that type erasure angle because I did notice issues with `StaticVec` but I didn't quite realize that relates to type erasure.

Issue #4088, opened 2025-06-15
use_context fails in nested outlets


jhoobergs:
**Describe the bug**

`use_context` fails when using it in a components which is rendered in an Outlet, which itself is also rendered in an Outlet. This used to work fine in 0.6.0 and now fails in my project after updating to 0.8, even with the latest changes on main.

I created a (somewhat) minimal example from the Leptos Axum Starter Template: https://github.com/jhoobergs/leptos_issue_context_outlet 

The main code is shown below. See `AssignmentsForTeam` for a description of the issue.

```rust
#[component]
pub fn App() -> impl IntoView {
   ...

    view! {
       ...
        <Router>
            <main>
                <Routes fallback=|| "Page not found.".into_view()>
                    <ParentRoute path=path!("") view=||view!{<LoggedIn/>}>
                        <AssignmentRouter />
                    </ParentRoute>
                </Routes>
            </main>
        </Router>
    }
}

#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct UserInfo {
    pub id: usize
}

#[server]
pub async fn get_user_info() -> Result<Option<UserInfo>, ServerFnError> {
    Ok(Some(UserInfo { id: 1 }))
}

#[component]
pub fn LoggedIn() -> impl IntoView {
    let user_info_resource = Resource::new(|| (), move |_| async { get_user_info().await });

    view!{
      <Transition fallback=move || view!{ 
            "loading"
            }
        >
        {move || {
            user_info_resource.get()
                .map(|a|
                    match a {
                        Ok(Some(a)) => Either::Left(view! {
                            <LoggedInContent user_info={a} />
                        }),
                        _ => Either::Right(view!{
                            <Redirect path="/not_logged_in"/>
                        })
                    })
        }}
        </Transition>
    }
}

#[component]
/// Component which provides UserInfo and renders it's child
/// Can also contain some code to check for specific situations (e.g. privacy policies accepted or not? redirect if needed...)
pub fn LoggedInContent(
    user_info: UserInfo, 
) -> impl IntoView {
    provide_context(user_info.clone());

    if user_info.id == 1 {
        Either::Left({
            Outlet()
        })
    } else { 
        Either::Right(view! { <Redirect path="/somewhere" options={NavigateOptions::default()}/> })
    }
}

#[component(transparent)]
pub fn AssignmentRouter() -> impl MatchNestedRoutes + Clone {
    view! {
[...]

PR #4115, opened 2025-06-27
Clean up nested routing ownership and add regression tests

gbj:
This should fix the ownership/cleanup issue identified in #4114, fixing the failing test in `examples/regression`.

Todo
- [x] Add regression test for #4015/#3042
- [x] Add regression test for #4088


examples/regression/e2e/tests/fixtures/find.rs
```diff
@@ -19,3 +17,8 @@ pub async fn link_with_text(client: &Client, text: &str) -> Result<Element> {
         .expect(format!("Link not found by `{}`", text).as_str());
     Ok(link)
 }
+
```
martinfrances107, line 20: While reading this .. I noticed a little simplification ... its in this PR 

https://github.com/leptos-rs/leptos/pull/4121

Issue #3685, opened 2025-03-07
`paste` dependency no longer maintained


tomtom5152:
RUSTSET has just published [RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436) detailing how `paste` is no longer maintained and has been archived. This is used by both the latest 0.7.7 release and the 0.8 branch.

gbj:
This is quite a frustrating RUSTSEC advisory, which is going to have [much larger repercussions](https://github.com/rustsec/advisory-db/pull/2215#issuecomment-2707037305).

Given that `itertools`, `wasm-bindgen`, `sqlx`, `pyo3-ffi`, and [many other huge crates](https://crates.io/crates/paste/reverse_dependencies) use this as a dependency, my assumption would be that someone else is going to end up taking over maintenance. 

There are no listed alternatives and dtolnay (creator of `paste`) commented on the RUSTSEC issue [asking them to remove the alternatives that had initially been listed](https://github.com/rustsec/advisory-db/pull/2215#discussion_r1938305206), which are almost certainly all less reliable.

Anyway, just wanted to post a note to say I'd seen this, and that my plan is to wait and see where the ecosystem lands.

martinfrances107:
I agree with the wait and see approach 

So this is a minor point.. just to add a voice to the discussion.

> This is quite a frustrating RUSTSEC advisory, which is going to have much larger repercussions

I think I reason differently with the  "frustrating" aspect.. it is reasonable for a contingent of the community to say "I want nothing to do with unmaintained code". Yep this consequential, is going to trigger activity in many popular crates.

It really tough to find a position that serves all contingents.

Please excuse my sense of humour, but to end on a happy note, at least we are not the much more diverse linux-rust community where a recently push back from different voices  was hey we can't accept this rust patch -- we will wait it is not secure beacuse rust does not yet have a formal specification...

gbj:
To clarify, here's why I find it frustrating:
1. There's no meaningful action we can take: removing `paste` from our dependencies would not remove it from any user's dependency tree, it would simply add another dependency, increasing both compile times and the security threat surface.
2. Because proc-macro crates can run arbitrary code during compile time (even during rust-analyzer bootup/check time!) they do present a real security risk. The advice to migrate away from a crate owned by a major pillar of the Rust language ecosystem, which has been archived because it is complete, to a fork by someone nobody knows and who there's no reason to trust, seems unwise to me.

[...]

PR #4118, opened 2025-06-28
Paste is dead, long live Pastey

martinfrances107:
The crate paste is no longer maintained, and has a replacement. On that basis only my IDE is reporting "paste" as a security hazard.

"pastey is the fork of paste and is aimed to be a drop-in replacement with additional features for paste crate" <https://crates.io/crates/pastey>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -924,7 +924,7 @@ checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 name = "either_of"
 version = "0.1.6"
 dependencies = [
- "paste",
+ "pastey",
  "pin-project-lite",
 ]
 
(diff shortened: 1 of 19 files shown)


martinfrances107:
For review purposes, I want to bring a potential critisism of this patch to the fore.

"paste" has a really high volume of daily downlaods ~500,000 and I expect a steady fall from here on.

"pastey" is ramping up quickly but is currently peak is only  lowly 3,000 daily downloads.

If a serious security problem becomes unveiled in "paste" the stock answer will always be move to "pastey" we will fix it there.. 

My question is "what is the resonable response if a new security flaw appears  in the  'pastey' crate" ... the low volume of daily downloads means the community is far less likley to see it, report it.

Anyway, I just thought I should mention the hazard.

gbj:
See discussion in #3685

martinfrances107:
From the discussion mentioned 

> Anyway, just wanted to post a note to say I'd seen this, and that my plan is to wait and see where the ecosystem lands.

A resonable attitude ...

PR #4126, opened 2025-07-02
chore: examples - bumped version numbers for sqlx and this error.

martinfrances107:
Making a small extremely narrowly focued issue, there are lots of other outdated crates in the examples... 

This is quixotic - but my way of contributing to open source...is what I do to my own projects 
I do the projects I love/ rely on


gbj:
Thanks!

PR #4127, opened 2025-07-02
chore: server_fns - ran cargo outdated.

martinfrances107:


martinfrances107:
This breaks on "cf-worker tests".

which are fixed here. #4128 --- well  main is still not passing tests

martinfrances107:
The failing test, was failing on head.  Now the block has been cleared by #4140 

So the tests just need to be re-run

martinfrances107:
This issue has become superseeded by fixes elsewehere

PR #4128, opened 2025-07-03
examples/cf-worker: Broken test investigation

martinfrances107:
Top of the tree is currently failing tests.
Failing on the following build

```bash
cd examples/cf-worker$
cargo build  --lib --release --target wasm32-unknown-unknown
```

* builds Ok without the --target specified
* builds Ok with --target specified BUT after cargo clean

so something is sticky, and messy crate dependency issue

My IDE reports "worker" and "worker-macros" are outdated That is the first thing that needs to be resolved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -651,9 +651,9 @@ dependencies = [
 
 [[package]]
 name = "collection_literals"
-version = "1.0.1"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "186dce98367766de751c42c4f03970fc60fc012296e706ccbb9d5df9b6c1e271"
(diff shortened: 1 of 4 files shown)


martinfrances107:
I have a couple of questions  for review... which are a the edge of my understanding!

It look like for each platform/target_arch we need to explicity state what is the source of randomness

For tauri -- is the client using "wasm32-*-*" if not they this will break their stuff.. I am not sure how to cover things to make those devs happy

Are people building leptos apps for android ..For them the sourcre of randomness will also have  to be set.

martinfrances107:
It looks to me as if we have a Cargo.toml settings which is being silently dropped....

This next commit  is just paranoia 

in cf-worker/Cargo.toml getrandom is being specified 3 times, one for a default which MUST be  being overwritten by one of the other two definitions ... for clairty sake I am removing the default definition.

martinfrances107:
removing cf-worker ... (#4140)  will fix the problem

PR #4130, opened 2025-07-03
fix: schedule `ActionForm` submit handler to run after user submit handlers (closes #3872)

### LINKED ISSUE #3872: `ActionForm` does NOT respect `on:sumbit` event
**Describe the bug**
`ActionForm`'s `on:sumbit` event will not be called prior to the `ActionForm`'s inner `on_sumbit` closure. (broken from `0.7`)

**Leptos Dependencies**

The latest commit (b45f982)

**To Reproduce**
1. Modify the `ActionFrom` in `leptos/src/form.rs`:
```rust
    let on_submit = {
        move |ev: SubmitEvent| {
            if ev.default_prevented() {
                crate::logging::log!("---1");
                return;
            }

            ev.prevent_default();

            crate::logging::log!("---2");

            match ServFn::from_event(&ev) {
                Ok(new_input) => {
                    action.dispatch(new_input);
                }
                Err(err) => {
                    crate::logging::error!(
                        "Error converting form field into server function \
                         arguments: {err:?}"
                    );
                    action.value().set(Some(Err(
                        ServerFnErrorErr::Serialization(err.to_string())
                            .into_app_error(),
                    )));
                    action.version().update(|n| *n += 1);
                }
            }
        }
    };
```
2. Modify the `WithActionFrom` in `server_fn_axum` example:
```rust
#[component]
pub fn WithActionForm() -> impl IntoView {
    let action = ServerAction::<AddRow>::new();
    let row_count =
        Resource::new(move || action.version().get(), |_| get_rows());

    view! {
        <h3>Using <code>"<ActionForm/>"</code></h3>
        <p>
            <code>"<ActionForm/>"</code>
            "lets you use an HTML "
            <code>"<form>"</code>
            "to call a server function in a way that gracefully degrades."
        </p>
        <ActionForm
            action
            on:sumbit=move |ev: SubmitEvent| {
                leptos::logging::log!("+++1");
                ev.prevent_default();
                leptos::logging::log!("+++2");
            }
        >
            <input
                // the `name` of the input corresponds to the argument name
                name="text"
                placeholder="Type something here."
            />
            <button>Submit</button>
        </ActionForm>
        <p>You submitted: {move || format!("{:?}", action.input().get())}</p>
        <p>The result was: {move || format!("{:?}", action.value().get())}</p>
        <Transition>
            archive underaligned: need alignment 4 but have alignment 1
            <p>Total rows: {row_count}</p>
        </Transition>
    }
}
```
3. Run the example and submit an input in `Using <ActionFrom>`.
4. See the log in the console, you will see that only `---2` is logged which is inside `ActionFrom` and the example's `on:submit` has been ignored.

**Next Steps**
[ ] I will make a PR
[ x ] I would like to make a PR, but need help getting started
[ x ] I want someone else to take the time to fix this

gbj:


paul-hansen (changes requested):
Looks like calling prevent_default inside request_animation_frame doesn't work since request_animation_frame isn't blocking, so the on submit handler completes and calls the browser's default submit before request_animation_frame's callback is called. I tested this branch locally and confirmed it caused duplicated requests due to browser's default handler still being called and also caused a navigation to happen outside the Leptos router.

My fault for [suggesting delaying the on submit](https://github.com/leptos-rs/leptos/issues/3872#issuecomment-3020462152) without thinking through the fact that we would still need to prevent the browser's default behavior (which can't be done in a delayed way) and calling prevent_default would prevent the user's handler.

I don't know of a workable solution to this at the moment. Thinking on it.

gbj:
@paul-hansen Just checking, which browser were you testing this under?

paul-hansen:
Firefox, though I just tested under chrome and I get the same thing. (Don't forget to check the "preserve log" on the network tab in chrome or it will hide the extra request because the page load resets the network log by default.)

In case any of this matters:
Arch Linux
Chrome Version 138.0.7204.92
Firefox 140.0.2 (64-bit)

Lmk and I can try making a minimal repro if you aren't seeing the same. Currently just testing with my project so maybe something is affecting it there.

gbj:
Huh. Weird. I just tried it with this diff in the `server_fns_axum` and it seems to work as intended in Firefox and in Chrome: i.e., no network request is made, the server function doesn't run, and the log appears in the console.
```diff
--- a/examples/server_fns_axum/src/app.rs
+++ b/examples/server_fns_axum/src/app.rs
@@ -224,7 +224,7 @@ pub fn WithActionForm() -> impl IntoView {
             <code>"<form>"</code>
             "to call a server function in a way that gracefully degrades."
         </p>
-        <ActionForm action>
+        <ActionForm action on:submit=|ev| {
+            ev.prevent_default();
+            leptos::logging::log!("did it submit?");
+        }>
             <input
                 // the `name` of the input corresponds to the argument name
                 name="text"
```

paul-hansen:
Looks like it's because I'm using a browser confirmation dialog.
I tried your example and you are right but when I add a confirmation dialog like this: 
```rust
        <ActionForm action on:submit=|ev| {
            let success = window()
                .confirm_with_message("hello world")
                .unwrap_or_default();
            if !success {
                ev.prevent_default();
                leptos::logging::log!("default prevented");
            }
        }>
```
When I submit and choose "ok", I see two requests, one with a 302 status and one with 200.

paul-hansen:
Came up with a potential alternative in #4131 feel free to use parts of it or anything if you have better ideas.

PR #4142, opened 2025-07-12
Fixes #4136: Compile time errors while deriving store

### LINKED ISSUE #4136: Compile time errors while deriving store for a struct with generic parameters
**Describe the bug**

When deriving `Store` for a struct which takes a generic argument the definition of the StoreField trait doesn't include the original generics definition.

For struct like

```rust
use reactive_stores::Store;

#[derive(Store)]
struct Data<I> {
    value: I,
}
```

The macro expands into

```rust

use reactive_stores::Store;
use uuid::Uuid;
struct Data<I> {
    value: I,
}
trait DataStoreFields<AnyStoreField>
where
    AnyStoreField: reactive_stores::StoreField<Value = Data<I>>,
{
    fn value(self) -> reactive_stores::Subfield<AnyStoreField, Data<I>, I>;
}
impl<AnyStoreField, I> DataStoreFields<AnyStoreField> for AnyStoreField
where
    AnyStoreField: reactive_stores::StoreField<Value = Data<I>>,
{
    fn value(self) -> reactive_stores::Subfield<AnyStoreField, Data<I>, I> {
        reactive_stores::Subfield::new(
            self,
            0usize.into(),
            |prev| &prev.value,
            |prev| &mut prev.value,
        )
    }
}
```

while the `DataStoreFields` trait should be defined as

```rust
trait DataStoreFields<AnyStoreField, I>
where
    AnyStoreField: reactive_stores::StoreField<Value = Data<I>>,
{
    fn value(self) -> reactive_stores::Subfield<AnyStoreField, Data<I>, I>;
}
```

(Notice the second generic argument `I` in the generic arguments of the trait).

**Leptos Dependencies**

```toml
reactive_stores = { version = "0.2.2" }
```

**Next Steps**

- [x] I will make a PR

mskorkowski:
Fixes the case when struct had a generic arguments by adding missing generic arguments into the generated trait and the said trait implementation.

### DIFF
--- a/reactive_stores_macro/src/lib.rs
+++ b/reactive_stores_macro/src/lib.rs
@@ -111,10 +111,8 @@ impl ToTokens for Model {
         } = &self;
         let any_store_field = Ident::new("AnyStoreField", Span::call_site());
         let trait_name = Ident::new(&format!("{name}StoreFields"), name.span());
-        let generics_with_orig = {
-            let params = &generics.params;
-            quote! { <#any_store_field, #params> }
-        };
+        let params = &generics.params;
+        let generics_with_orig = quote! { <#any_store_field, #params> };
         let where_with_orig = {
             generics
                 .where_clause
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

PR #4148, opened 2025-07-13
We're kinda prod-ready

rakshith-ravi:
I think we're in a place where we don't have to deter users from using Leptos in production. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Yes, I’m sure there are. You can see from the state of our issue tracker over 
 
 This may be the big one: “production ready” implies a certain orientation to a library: that you can basically use it, without any special knowledge of its internals or ability to contribute. Everyone has this at some level in their stack: for example I (@gbj) don’t have the capacity or knowledge to contribute to something like `wasm-bindgen` at this point: I simply rely on it to work.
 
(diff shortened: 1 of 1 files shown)


lukexor:
I'd agree with this. In my work we've migrated a moderately sized project from React to Leptos for a 1.0 release in under 2 months with great success. There were some bumps and rough edges, but overall very happy with the maturity of the framework and the few supporting libraries we used

rakshith-ravi:
Out of my own personal curiosity, any feedback you have? What would you like to see in the core framework? What can we do better?

lukexor:
> Out of my own personal curiosity, any feedback you have? What would you like to see in the core framework? What can we do better?

Just a general thanks to all the contributors and the amount of work poured into making things as ergonomic as possible. I've been following the project for awhile with excitement.

Most, if not all of the bumps we encountered were raised and addressed and the rest was just learning curve.

My only other feedback is more documentation that meets web developers of varying skill and experience where they are. E.g. having some React or angular focused cookbooks focused on common patterns in those frameworks and how they may be translated or modified to be idiomatic in Rust/leptos.

Many of the issues and discussions I've seen are some form of learning curve gap so anything that helps reduce the friction and frustration toward adoption is a win!

Also I've seen a bit of tension with third-party libs integrating with leptos. I'm not sure the best way to solve the issues I've seen, but prime example is thaw having their own Model type that allows a wider range of signal-based props for a component library. Similarly leptos-use has some custom utilities. Maybe some of these primitives can be generalized into the core framework allowing the ecosystem to settle on more consistent patterns

gbj:
Thanks! Works for me.

Issue #4146, opened 2025-07-13
Cannot compile simple SVG component anymore


arnaudpoullet:
**Describe the bug**
```
error[E0433]: failed to resolve: could not find `InertElement` in `svg`
  --> app/src/component/svg.rs:46:5
   |
46 | /     view! {
47 | |         <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="size-6">
48 | |           <path stroke-linecap="round" stroke-linejoin="round" d="M15.75 19.5 8.25 12l7.5-7.5" />
49 | |         </svg>
50 | |
51 | |     }
   | |_____^ could not find `InertElement` in `svg`
   |
   = note: this error originates in the macro `view` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I went over the different PRs in leptos and leptos-icons but could not find if I should declare my SVGs differently since the introduction of `InertElement` in `v0.8.3` so it looks like a bug.

I ran `cargo clean` but cache wasn't the issue

**Leptos Dependencies**

```toml
leptos = { version = "0.8.3" }
leptos_meta = { version = "0.8.3"}
leptos_router = { version = "0.8.3" }
leptos_axum = { version = "0.8.3", optional = true }

[features]
default = []
hydrate = ["leptos/hydrate"]
ssr = ["leptos/ssr", "leptos_meta/ssr", "leptos_router/ssr","dep:leptos_axum"]
```

**To Reproduce**
add a component with a SVG element
run `leptos cargo watch`

**Screenshots**

**Next Steps**

- [ ] I will make a PR
- [ ] I would like to make a PR, but need help getting started
- [x] I want someone else to take the time to fix this
- [ ] This is a low priority for me and is just shared for your information

**Additional context**
Here is one of my SVG components:
```rust
#[component]
pub fn ChevronLeft() -> impl IntoView {
    view! {
        <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="size-6">
          <path stroke-linecap="round" stroke-linejoin="round" d="M15.75 19.5 8.25 12l7.5-7.5" />
        </svg>

    }
}
```

gbj:
I have once again forgotten to bump workspace dependency version numbers (#4149). `cargo update` should fix this for you.

PR #4150, opened 2025-07-14
fix(CI): check latest commit for version release instead of version tag

sabify:
This will trigger check minimal versions. Related #4149, #4146.

@gbj, Please first publish the crates to crates.io and then push commit a version release like `v0.8.4` to make sure the CI and `check-minimal-versions` work/trigger properly.

### DIFF
--- a/.github/workflows/run-cargo-make-task.yml
+++ b/.github/workflows/run-cargo-make-task.yml
@@ -169,7 +169,9 @@ jobs:
           cd '${{ inputs.directory }}'
           cargo make --no-workspace --profile=github-actions ci
           # check the direct-minimal-versions on release
-          if [[ "${{ github.ref_name }}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          COMMIT_MSG=$(git log -1 --pretty=format:'%s')
+          # Supports: v1.2.3, v1.2.3-alpha, v1.2.3-beta1, v1.2.3-rc.1, etc.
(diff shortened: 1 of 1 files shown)


gbj:
An ignorant question about how this is supposed to work: once the crates have been published to crates.io, isn't it essentially too late? 

My typical workflow:
- update versions in crates
- `git commit -m "v0.8.3"` (cargo won't publish with uncommitted changes)
- `cargo publish`
- `git push origin main`
- create a GitHub release tagging the new version

If `check-minimal-versions` requires the crates already to be published, then isn't it essentially too late by the time it runs? Or does it simply detect that there's an issue, in which case I can fix it and publish another patch version with the updated dependencies?

sabify:
@gbj Here is what should happen in order:

- Update versions in crates
- `git commit -m "v0.8.5"`
- `cargo publish --allow-dirty` (`allow-dirty` can be used to publish unpushed commits.)
- Wait for 5 minutes to make sure the crates are propagated in the registries.
- `git push origin main` (this needs to happen last.)

The reason for this order is that `check-minimal-version` first removes the `path = ...` declaration in all internal dependencies in `Cargo.toml` files and then starts to build and test packages against the minimum available crate versions declared and published. So, if you forgot to increment some crates versions with API changes, you will notice it for published crates.

If you change the order, two things happen:

- If you *have updated* the crate versions before publishing them, it's an error and won't compile.
- If you *haven't updated* the crate versions before publishing them, it will test the `minimal-versions` against old published packages which is nonsense.

PR #4158, opened 2025-07-16
docs: add warning for reading hash on the server

### LINKED ISSUE #4059: Hydration error when using `use_location().hash`
**Describe the bug**
On SSR loaded pages using `use_location().hash` to determine dom structure results in a hydration error. This is due to the hash / fragment always being `""` on the server as it can't be sent in the request while the client has the "correct" value.

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.8.2", features = ["nightly"] }
leptos_axum = { version = "0.8.2", optional = true }
leptos_meta = { version = "0.8.2"}
leptos_router = { version = "0.8.2"}
```

**To Reproduce**
1. Create a project from the start-axum template
2. Replace HomePage with this:
```html
#[component]
pub fn HomePage() -> impl IntoView {
    let hash = use_location().hash;
    view! {
        <Show when=move || !hash.get().is_empty()>
            <p> "Hydration error </p>
        </Show>
    }
}
```
3. Navigate to /#test from the browser.
4. Open the console and see a hydration error.

**Next Steps**
- [ ] I will make a PR

**Additional context**
This is theoretically expected if you know that fragments are not sent in requests but it might still be worth it to prevent the hydration error in the framework.

TERRORW0LF:
Closes #4059 

There is currently no check on if the app is running in dev or prod mode, since I couldn't figure out how to access the current LeptosOptions to check against the env. If there's a way to get the current options or if you want to check against the build type instead I'm happy to implement those.

From my (admittedly limited) testing this does not appear to produce false positives.

### DIFF
--- a/router/src/location/mod.rs
+++ b/router/src/location/mod.rs
@@ -67,7 +67,7 @@ impl Url {
     }
 
     pub fn hash(&self) -> &str {
-        #[cfg(feature = "ssr")]
+        #[cfg(all(feature = "ssr", any(debug_assertions, leptos_debuginfo)))]
         {
             #[cfg(feature = "tracing")]
             tracing::warn!(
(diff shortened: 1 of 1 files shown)


gbj:
The annotation we use consistently to mark dev mode is
```rust
#[cfg(any(debug_assertions, leptos_debuginfo))]
```

gbj:
Thanks!

PR #4160, opened 2025-07-16
feat: add `debug_log!`, `debug_error!`, `console_debug_log` and `console_debug_error` 

mahdi739:
This PR adds debug-only macros and methods for `log` and `error` levels, aligning with the existing warning level implementation. I guess the debug-only methods contain duplicated code, but I’ve kept them as-is to match the existing pattern. Additionally, I noticed that `console_error` is defined inline, while the other two methods are not. I left this unchanged to avoid potential impacts on compile times. Let me know if any changes are needed.

### DIFF
--- a/leptos_dom/src/logging.rs
+++ b/leptos_dom/src/logging.rs
@@ -23,6 +23,19 @@ macro_rules! error {
     ($($t:tt)*) => ($crate::logging::console_error(&format_args!($($t)*).to_string()))
 }
 
+/// Uses `println!()`-style formatting to log something to the console (in the browser)
+/// or via `println!()` (if not in the browser), but only if it's a debug build.
+#[macro_export]
+macro_rules! debug_log {
+    ($($x:tt)*) => {
+        {
+            if cfg!(debug_assertions) {
+                $crate::log!($($x)*)
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


gbj:
I'd suggest that all of the `console_debug_X` functions should simply be `if cfg!(debug_assertions) { console_X(...) }`. That would also fix the missing `#[allow]` that's causing `console_debug_log` to fail right now ([see here](https://github.com/leptos-rs/leptos/pull/4160/files#diff-cb82789e379a487dc6546c99bc3cf7e20d52e61747e9d70d6c3949fea5ce334bL49) for what's missing) (We set up warnings to prevent logging to make sure nothing is accidentally shipped that allows printing to console -- obviously we opt out for these.)

mahdi739:
Sure. Can I make `console_log` and `console_warn` inline? because `console_error` is inline but those two are not.

mahdi739:
For now, I’ve applied the changes you suggested. Let me know if `console_*` can be inlined as well.

gbj:
My comment was meant to say that probably *none* of them should be inline. Is there a reason that inlining them all makes sense?

gbj:
(I'm about to do a patch release and would like to include these, so for now I'm going to merge the PR!)

PR #4162, opened 2025-07-19
fix(hot-reload): implement Myers diffing algorithm

shadr:
This PR implements Myers diff algorithm for `diff_children` method. This fixes two bugs and one small issue:

- inserting multiple tags at once results in incorrect ordering of elements
- being unable to insert new tags in fragments
- the previous algorithm produces unnecessary patches when inserting new child between two tags

Now patches apply immediately instead of after processing every patch for a view, and actualChildren is rebuilt before each patch. These changes were needed for the new algorithm, because it relies on indexes that reference newly added elements.

I tested it mostly on simple cases, so it could break some edge cases with deeply nested elements or combination of fragments and components. If that happens, I will fix them in the future.

# MRE
Here is the code that you can test changes on, with comments describing what happens with current version of leptos:
```rust
use leptos::prelude::*;
use leptos_meta::{provide_meta_context, MetaTags, Title};
use leptos_router::{
    components::{Route, Router, Routes},
    StaticSegment,
};

    view! {
        <!DOCTYPE html>
        <html lang="en">
            <head>
                <meta charset="utf-8"/>
                <AutoReload options=options.clone()/>
                <HydrationScripts options/>
                <MetaTags/>
            </head>
            <body>
                <App/>
            </body>
        </html>
    }
}

#[component]
    provide_meta_context();

    view! {
        <Router>
            <main>
                </Routes>
            </main>
        </Router>
    }
}

#[component]
    view! {
        <div>
            <AddMultipleTags />
            <br />
            <UnnecessaryPatches />
            <br />
            <ModifiyingFragments />
        </div>
    }
}

#[component]
fn AddMultipleTags() -> impl IntoView {
    // after uncommenting you will see
    // Text 1
    // Text 2
    // Text 3
    // Button
    // Text 4
    view! {
        <div>
            <p>"Text 1"</p>
            <p>"Text 2"</p>
            <p>"Text 3"</p>
            <p>"Text 4"</p>
            <button>"Button"</button>
        </div>
    }
}

#[component]
fn UnnecessaryPatches() -> impl IntoView {
    // this will produce following patches:
    // InsertChildAfter <p>Text 3</p> after <p>Text 3</p>
    // SetText "Text 3" to "Text 2"
    view! {
        <div>
            <p>"Text 1"</p>
            // <p>"Text 2"</p>
            <p>"Text 3"</p>
        </div>
    }
}

#[component]
fn ModifiyingFragments() -> impl IntoView {
    // adding tags to fragments throws an error
    view! {
        <p>"Text 1"</p>
        <p>"Text 2"</p>
        <p>"Text 3"</p>
        <p>"Text 4"</p>
    }
}
```

### DIFF
--- a/leptos_hot_reload/src/diff.rs
+++ b/leptos_hot_reload/src/diff.rs
@@ -305,7 +305,7 @@ impl LNode {
                             action: PatchAction::InsertChild {
                                 before: i,
                                 child: new[j - 1]
-                                    .to_replacement_node(&old_children),
+                                    .to_replacement_node(old_children),
                             },
                         });
                         j -= 1;
--- a/leptos_hot_reload/src/diff.rs
+++ b/leptos_hot_reload/src/diff.rs
@@ -254,6 +254,7 @@ impl LNode {
             let width = old.len() + 1;
             let height = new.len() + 1;
             let mut mat = vec![0; width * height];
+            #[allow(clippy::needless_range_loop)]
             for i in 1..width {
                 mat[i] = i;
             }
--- a/leptos_hot_reload/src/diff.rs
+++ b/leptos_hot_reload/src/diff.rs
@@ -251,93 +251,66 @@ impl LNode {
                 action: PatchAction::ClearChildren,
             }]
         } else {
-            let mut a = 0;
-            let mut b = std::cmp::max(old.len(), new.len()) - 1; // min is 0, have checked both have items
+            let width = old.len() + 1;
+            let height = new.len() + 1;
+            let mut mat = vec![0; width * height];
+            for i in 1..width {
+                mat[i] = i;
+            }
+            for i in 1..height {
+                mat[i * width] = i;
+            }
+            for j in 1..height {
+                for i in 1..width {
+                    if old[i - 1] == new[j - 1] {
+                        mat[j * width + i] = mat[(j - 1) * width + (i - 1)];
+                    } else {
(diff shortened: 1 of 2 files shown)


shadr:
Clippy complains about range loop
```diff
-             for i in 1..width {
+             for (i, <item>) in mat.iter_mut().enumerate().take(width).skip(1) {
```
is that 4 method long chain really more readable than a simple 1..width range ? Can I  add `#[allow(clippy::needless_range_loop)]` or should I apply the suggested change ?

sabify:
> Clippy complains about range loop
> 
> ```diff
> -             for i in 1..width {
> +             for (i, <item>) in mat.iter_mut().enumerate().take(width).skip(1) {
> ```
> 
> is that 4 method long chain really more readable than a simple 1..width range ? Can I add `#[allow(clippy::needless_range_loop)]` or should I apply the suggested change ?

The suggestion can bypass boundary checks and is more performant regardless of readability. However, since hot-reload is used in debug mode, the optimization may not be applied or may be negligible, so either approach can be used.

gbj:
Definitely `#[allow(clippy::needless_range_loop)]`. This is a fairly recent clippy lint (I think?) and I have silenced it in a couple other places recently.

shadr:
You are welcome! Thank you for creating amazing frontend rust library! I created in a hurry new PR #4167 , I really want to have a fix for hot-reload stopping working in the next leptos_ release, but I'm not feeling good today, so I couldn't test it properly and write a proper PR.

PR #4167, opened 2025-07-20
fix(hot-reload): hot-reload stops working when number of views changes in a file + small fixes

shadr:
This PR fixes four bugs:

- hot-reload stops updating view macro map when number of view macro in a file mismatch with previous state of a file
- fragment couldn't be replaced with a single tag
```rust
view! {
<p>"Text 1"</p>
<p>"Text 1"</p>
<p>"Text 1"</p>
}
```
```rust
view! {
<p>"Text 1"</p>
}
```
- unable to fully clear fragment view 
```rust
view! {
<p>"Text 1"</p>
<p>"Text 1"</p>
<p>"Text 1"</p>
}
```
```rust
view! {
}
```
- unable to append children to an empty tag
```rust
view! {
<div></div>
}
```
```rust
view! {
<div>
<p>"Text"</p>
</div>
}
```

### DIFF
--- a/leptos_hot_reload/src/lib.rs
+++ b/leptos_hot_reload/src/lib.rs
@@ -121,6 +121,10 @@ impl ViewMacros {
                     }
                     diffs
                 } else {
+                    // TODO: instead of simply returning no patches, when number of views differs,
+                    // we can compare views content to determine which views were shifted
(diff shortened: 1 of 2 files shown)


gbj:
Thanks for this, as well! I am happy to continue merging small changes to the hot-reload code, so don't feel any rush. There's nothing wrong with additional patch releases.

PR #4168, opened 2025-07-20
Preparing to publish oco_ref 0.2.1

martinfrances107:
In a way this is a field report from a leptos app.

There are duplicate crates derived solely from within the eco-system of leptos crates.

I have traced down why thiserror is still duplicated even though there has been a recent efforts to unify things by referencing a single crate definitions in the root workspace.

The recent changes do not appear in the published version of oco_ref and so the ecosystem at
 large is still relying on stale versions ( crates.io reports oco_ref is about a year old. )

In short, bump oco_ref.


gbj:
Published. Thanks!

PR #4172, opened 2025-07-22
Enhancing members’ versioning

sabify:
This PR introduces two changes:

- Decouple versioning of members from the workspace. If there are no changes to a member, there shouldn't be a version bump and a release.
- To tackle the difficulty of tracking changes in the members, I wrote and added a `bump.sh` script that tracks the changes from the last release tag and provides a simple interface to update member versions. The changes will be applied to both the workspace and the crate's `Cargo.toml`.

Here is the usage:

```bash
# First please install cargo-edit: cargo install cargo-edit
# Then run the script in workspace directory

$ ./scripts/bump.sh

Changes detected in reactive_graph (/Users/sabify/dev/leptos/reactive_graph)
1) patch
2) minor
3) major
Select version bump for reactive_graph: 1
   Upgrading reactive_graph from 0.2.3 to 0.2.4
    Updating workspace's dependency from 0.2.3 to 0.2.4
reactive_graph bumped to patch

# and so on for every changed member that has been detected...
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ members = [
 exclude = ["benchmarks", "examples", "projects"]
 
 [workspace.package]
-version = "0.8.5"
 edition = "2021"
 rust-version = "1.88"
 
--- a/integrations/actix/Cargo.toml
+++ b/integrations/actix/Cargo.toml
@@ -4,7 +4,7 @@ authors = ["Greg Johnston"]
 license = "MIT"
 repository = "https://github.com/leptos-rs/leptos"
 description = "Actix integrations for the Leptos web framework."
-version = { workspace = true }
+version = "0.8.5"
 rust-version.workspace = true
 edition.workspace = true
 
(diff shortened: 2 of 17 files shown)


gbj:
I agree with making this change in general: as the number of crates grows and the release cycles aren't necessarily in sync, I think it's a good idea to decouple the version numbers.

I'm not sure I'm understanding the bump script correctly, though. Shouldn't it be testing whether there have been changes to any given package since the last change in that package's version number, not since the last tag for the whole repo? Or is the plan to make a tag for each new patch release of each crate?

For example:
- current tag is v0.8.5
- I update some code in `wasm_split_macros` and release it as 0.1.2
- next time I run `bump.sh`, it will tell me there are changes to `wasm_split_macros`, unless I added a tag for `v0.1.2`

I might be misunderstanding.

gbj:
(That said, this is super handy and I am going to do a quick patch release right now due to some changes, so -- thank you! I'm going to merge and can manually opt out of anything where issues arise.)

sabify:
The script works best with tag releases as we need to have a reference to rely on. So if you cut a release for a member without releasing a tag, the script cannot track the changes. If that's the case, I'm providing a better version of the script.

PR #4174, opened 2025-07-22
feat: implement IntoProperty for Oco

veigaribo:
Since `Cow<'static, str>` implements `IntoProperty`, I don't see why `Oco<'static, str>` wouldn't. This makes it easier to use a `TextProp` as a property, for instance.

I wondered whether I should put this in `tachys/src/oco.rs` alongside its other implementations of `IntoClass` and `IntoStyle` or in `tachys/src/html/property.rs` alongside the `IntoProperty` implementations of `Cow` and friends...


gbj:
Thanks! Apologies for the delayed response, I've been on vacation :-)

PR #4177, opened 2025-07-24
feat: Allow configuration to be read from option_env!

lizidev:
Allows us to specify the corresponding configuration at compile time, and no longer needs to provide environment variables when executing executable files (exe generated in ssr mode).

```
// Before
1. Use runtime environment variables. (`std::env::var`)
2. Default value.

// After
1. Use runtime environment variables. (`std::env::var`)
2. Use compile-time environment variables. (`option_env!`)
3. Default value.
```

### DIFF
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -10,6 +10,31 @@ use std::{
 };
 use typed_builder::TypedBuilder;
 
+macro_rules! read_env_config {
+    ($key:expr) => {
+        match ::std::env::var($key) {
+            Ok(val) => Ok(Some(val)),
+            Err(::std::env::VarError::NotPresent) => {
+                Ok(option_env!($key).map(ToString::to_string))
+            }
+            Err(e) => Err($crate::errors::LeptosConfigError::EnvVarError(
+                format!("{}: {e}", $key),
+            )),
+        }
+    };
+    ($key:expr, $default:expr) => {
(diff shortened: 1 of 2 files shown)


benwis:
Hi @lizidev, thanks for the PR.

I'm not sure how I feel about this PR. If I'm reading this correctly, if an environment variable is set during compilation(like with cargo_leptos), then when the binary is executed later it will use those values and not the defaults if run in, say, a deployment environment?

To me that seems suboptimal, as deployment variables will in most cases be different than the ones set during compilation. But if you're building in a docker environment you could set production values, but often that environment is separate from the web host one too. I'm not sure I can think of another web framework that behaves in a similar way here.

I'll have to think on this one a bit

lizidev:
It reads the runtime environment variables first, so it doesn't affect setting new environment variables in the deployment environment.

This PR primarily allows me to customize variables at compile time, and then execute the executable directly in the deployment environment (saves the step of setting environment variables).

> I'm not sure I can think of another web framework that behaves in a similar way here.

Currently, both [Vite](https://vite.dev/guide/env-and-mode.html#env-variables-and-modes) and [Dioxus CLI](https://github.com/DioxusLabs/dioxus/blob/a8230d8ff0e769d3c1fee3452b8ea47ec885ecc1/packages/cli-config/src/lib.rs#L75C1-L77C86) behave in this way.

PR #4179, opened 2025-07-26
Demo of `hash-files = true` causing lazy islands to break

spencewenski:


spencewenski:
I didn't mean to open this as a PR.

PR #4182, opened 2025-07-26
fix: support file hashing when using lazy loading

gbj:
This supports the combination of lazy-loading and file hashing by inserting a placeholder hash in the name of `__wasm_split.js`. (This is necessary because changing the length of this string invalidates the WASM binary, so it's much easier to just replace the placeholder!) 

Needs to be used alongside https://github.com/leptos-rs/cargo-leptos/pull/553


gbj:
Okay, cool. CI on this should fail until I release a new patch version of `cargo-leptos`, and then it should work. (Famous last words?)

PR #4187, opened 2025-07-27
feat: improving the bump script

sabify:
Related: #4172 

The script now detects if any changes were made to a member from a commit release up to the latest tag.

@gbj, if you are going to release and publish a member crate without releasing a tag, please push a commit in the following format to keep track of the changes: `<member-name>-v<member-version>`. For example, `oco_ref-v0.2.1`.

Please keep in mind to use the member package name defined in the `Cargo.toml`, not the directory name.

Here is the logic:

- If there is a **release commit** newer than the latest tag for the member crate, check if there are any changes made to it from the **release commit** and request a bump.
- Otherwise, use the **latest tag** and check if there are any changes made to the member crate and request a bump.

### DIFF
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -12,8 +12,16 @@ for PKG in $PACKAGES; do
     MANIFEST_PATH="${PKG##*:::}"
     DIR=$(dirname "$MANIFEST_PATH")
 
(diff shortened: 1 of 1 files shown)


gbj:
Thanks!

Issue #3932, opened 2025-04-29
Improve ergonomics for HashedStylesheet


lukexor:
It feels like there's some awkward ergonomics around `HashedStylesheet` vs `Stylesheet` components when dealing with dev vs release.

`Stylesheet` reloads the css in watch mode when changed which is ideal for development where hashes aren't that important.

`HashedStylesheet` doesn't reload but provides hashing so is more ideal for release.

Supporting both simultaneously currently requires the application to check `options.hash_files` unless I'm missing something.

Is there a better alternative that could bring these under the same code path (for consumers anyways?). I know the APIs are different since `HashedStylesheet` requires an options prop, so it has to go in the shell, whereas normally `Stylesheet` doesn't have this constraint.

For example, this is what I'm currently using, but it would be more ergonomic (imo) if this was provided by leptos with an option to default to `Stylesheet` in development and `HashedStylesheet` when built with `--release`.

```rust
/// Renders either a HashedStylesheet or Stylesheet based on configured option for `hash_files`.
/// Set LEPTOS_HASH_FILES=true for release builds.
#[component]
fn RootStylesheet(options: LeptosOptions) -> impl IntoView {
    if options.hash_files {
        Either::Left(view! { <HashedStylesheet id="style" options /> })
    } else {
        let href = format!(
            "{pkg_path}/{css_name}.css",
            pkg_path = &options.site_pkg_dir,
            css_name = &options.output_name
        );
        Either::Right(view! { <Stylesheet id="style" href /> })
    }
}
```

Then, because I'm running the server as a systemd unit from /usr/local/bin I have to set the following vars for build/runtime:
LEPTOS_SITE_ROOT="/usr/share/www/"
LEPTOS_HASH_FILES="true"
LEPTOS_HASH_FILE_NAME="${LEPTOS_SITE_ROOT}/hash.txt"

luxalpa:
I think there needs to be a way for `cargo leptos watch` to check / watch for the HashedStylesheet (with the proper filename).

In fact, I wasn't even able to get your workaround running. The `<Stylesheet>` straight up doesn't render for me at all if I don't put it into the `<App />` components where `<HashedStylesheet>` can't be used due to leptos options missing.

lukexor:
> I think there needs to be a way for `cargo leptos watch` to check / watch for the HashedStylesheet (with the proper filename).
> 
[...]

PR #4189, opened 2025-07-27
fix css hydration with hash_files

rvdende:
This fixes the css loading of tailwind output file when hash_files is enabled for production cache busting.

related issue https://github.com/leptos-rs/leptos/issues/3932


benwis:
Hi @rvdende! Thanks for the PR. I see the desire to automatically reload css files on cargo watch. It seems like this solution automatically includes the leptos produced CSS on every page, which I imagine some folks will not want.

 We used to have a solution where leptos or cargo-leptos looked for an id on the link tag for the css file if present and updated that with JS if it was present. Have you considered that? You can see it on the todo_app_axum example
```
     <link rel="stylesheet" id="leptos" href="/pkg/todo_app_sqlite_axum.css"/>
```

I imagine we'd want to edit the `AutoReload` component and maybe the `patch.js` file we use to do this

rvdende:
This is for release builds with hash-files true. It has no effect when doing development locally when using leptos watch and hot reload.

leptos/src/hydration/mod.rs
```diff
@@ -144,6 +149,7 @@ pub fn HydrationScripts(
             r#type="application/wasm"
             crossorigin=nonce.clone().unwrap_or_default()
         />
+        <link rel="stylesheet" id="leptos" href=format!("{root}/{pkg_path}/{css_file_name}.css") />
```
benwis, line 152: People may or may not want their CSS to apply to all pages, I don't think we can add this here by default

johansmitsnl:
Any progress on this? Would the option to inline css also be a option?

rvdende:
for now just going to close this as this approach doesn't seem to have much support. I haven't found an alternative method yet that works for us though.

johansmitsnl:
@rvdende isn't this the solution to your issue: https://docs.rs/leptos_meta/latest/leptos_meta/fn.HashedStylesheet.html

PR #4190, opened 2025-07-28
Add name attribute to details element

adoyle0:
When trying to use the `name` attribute of the `<details>` element I noticed it was missing. This is a relatively new attribute that is younger than Leptos so it was probably just never added. Basically if you have a group of `<details>`s that share a name then the browser handles only displaying one at a time. I would like to use this in my component library.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details#browser_compatibility) it is fully supported by all browsers, since September 2024.

I think it makes sense to add it.


gbj:
Thanks!

PR #4194, opened 2025-07-31
Fixes #4193: Adding `command` and `commandfor` attributes for button

### LINKED ISSUE #4193: `command` and `commandfor` attributes for `button`
[Invoker API][invokerapi] allows the button to have two additional attributes [`command`][command] and [`commandfor`][commandfor].

[invokerapi]:https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API
[command]:https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#command
[commandfor]:https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#commandfor

I was playing around with invoker API 

I think both attributes should be added into the leptos directly since they are standard enough that team behind tailwind decided to provide alpha version of their components using just them.

**Additional context**

I'll create a PR for that

mskorkowski:


gbj:
Thanks!

PR #4204, opened 2025-08-02
make is_server and is_browser public

Alxandr:
Referring to discussion #4200, this allows easier branching based on server/client that is not based on cargo features. This is usefull for cases where all the code can compile for both environments, but there is no point in doing the extra work on the server/client.


benwis:
Thanks for the PR, looks good to me

PR #4212, opened 2025-08-06
Avoid panic for set_timeout_with_handle() and set_interval_with_handle() when Duration::as_millis() > i32::MAX

zakstucke:


benwis:
Hi Zak! I'm not sure the answer to someone trying to set a timeout that cannot be represented by an i32 is to silently change that timeout to something that can

zakstucke:
I think in this specific case it can no? `i32::MAX` equals roughtly 24.85 days in this instance, a nonrealistic length of time for a timeout or interval, this change simply removes a footgun I experienced, where I was just using a `Duration::MAX` as a placeholder, which the interface implies would've been fine.

benwis:
It could, but semantically I think this is a question of how it should behave. The library code should provide clear behavior, if someone sets an invalid value it should panic or throw an error, not change it to a valid but unexpected value behind the scenes. 

In this case I think your fix to have leptos_fetch not set such a ridiculously high value is probably the correct one, I'd even go further and throw an error in the devtools instead of reducing it down behind the scenes, but that's your domain :)

zakstucke:
Fair enough, I still think this would be the correct change here, I'd agree with you if `set_timeout_with_handle` took an `i32`, but it takes a `Duration`, the semantically correct thing here in my mind is the largest possible supported value, or a breaking change to the interface, no where in the interface is it clear there will be a panic with higher durations than `i32::MAX`. 

Won't push though, I worked around it downstream anyway.

benwis:
I'd even say this is a bug to take a Duration here if we can't hold the full value. But yeah, ce est la vie

benwis:
Closing this in favor of #4226

Issue #4209, opened 2025-08-05
Server functions do not set content-type header when an error is returned


spencewenski:
**Describe the bug**
I'm using a custom error for my server functions and I'm setting `type Encoder = leptos::server_fn::codec::JsonEncoding` in my `impl FromServerFnError`. I would expect this to set the `content-type: application/json` header like leptos does for successful API responses, but the header is not set at all. I noticed that this also impacts the built-in `ServerFnError` -- I would expect it to set the header to `text/plain`, but it's missing.

**Leptos Dependencies**

Leptos `0.8.6` and the latest `main` branch. Example repro: https://github.com/spencewenski/leptos/commit/3cd8c37b7712dd6d96a305e20fde6f0b8334ae73

**To Reproduce**
1. Update a server function to return an error. In the above repro, `clear_server_count` is updated.
2. Run the `counter_isomorphic` example in in the above repro.
3. Click the `Clear` button
5. Observe that the response does not include the `content-type`.

**Screenshots**

<img width="1293" height="417" alt="Image" src="https://github.com/user-attachments/assets/c195d5f5-d95a-4399-a7e9-d22a9f40454e" />

**Next Steps**

- [x] I will make a PR
- [ ] I would like to make a PR, but need help getting started
- [ ] I want someone else to take the time to fix this
- [ ] This is a low priority for me and is just shared for your information

**Additional context**

spencewenski:
I was looking into this a bit, and so far I've narrowed it down to the implementations of `Res` not setting the content type. I confirmed by hard-coding the content type in these responses. E.g.:

- https://github.com/leptos-rs/leptos/blob/e2e28ef180bd5d81e4532eafaab302c2c1716cee/server_fn/src/response/actix.rs#L77
- https://github.com/leptos-rs/leptos/blob/e2e28ef180bd5d81e4532eafaab302c2c1716cee/server_fn/src/response/generic.rs#L96
- https://github.com/leptos-rs/leptos/blob/e2e28ef180bd5d81e4532eafaab302c2c1716cee/server_fn/src/response/http.rs#L56

[...]

PR #4215, opened 2025-08-06
fix: Set content type header for server function errors

spencewenski:
Problem
-------
When a server function returns an error, the content-type header is not set.

Solution
--------
In order to set the content-type header in the response, the `FromServerFnError::Encoder::CONTENT_TYPE` associated constant needs to make its way to where the response object is constructed. This PR achieves this with the following changes:

- Add `content_method` to the `Res` trait to allow setting the `Content-Type` header for server function error responses.
- Call the `Res::content_method` from the server function handler when an error is returned from the app's server fn impl.

Note: As mentioned in https://github.com/leptos-rs/leptos/issues/4209, this issue also affects Leptos's middleware handlers. Unfortunately, resolving this for middleware requires a semver-incompatible change. The following PR contains a potential approach: https://github.com/leptos-rs/leptos/pull/4249 


server_fn/src/lib.rs
```diff
@@ -307,16 +307,18 @@ pub trait ServerFn: Send + Sized {
...
                                 Self::InputStreamError,
                                 Self::OutputStreamError,
                             >>::Response::error_response(
                                 Self::PATH, e.ser()
-                            ),
-                            Some(e),
-                        )
+                            );
+                        let content_type =
+                    <Self::Error as FromServerFnError>::Encoder::CONTENT_TYPE;
```
spencewenski, line 319: This formatting is kind of weird, but it's what `cargo fmt` wants.

server_fn/src/response/mod.rs
```diff
@@ -39,7 +39,8 @@ where
 pub trait Res {
     /// Converts an error into a response, with a `500` status code and the error text as its body.
     fn error_response(path: &str, err: Bytes) -> Self;
-
+    /// Set the content type header for the response.
+    fn content_type(&mut self, #[allow(unused_variables)] content_type: &str) {}
```
spencewenski, line 43: Using `#[allow(unused_variables)]` here instead of naming the parameter `_content_type` so that when IDEs generate a placeholder implementation, the parameter name is `content_type` instead of `_content_type`.

benwis:
Have you looked at how the Errors are typically returned? If I recall correctly, they're usually not encoded/decoded using the standard logic and may not be valid JSON

spencewenski:
> Have you looked at how the Errors are typically returned? If I recall correctly, they're usually not encoded/decoded using the standard logic and may not be valid JSON

I think they're encoded by the `FromServerFnError::Encoder` associated type now, right? At least for errors returned in the body of the server function.

https://github.com/leptos-rs/leptos/blob/b986fe11dc63a803d026f9ccae804e880c2a702b/server_fn/src/error.rs#L572

server_fn/src/lib.rs
spencewenski: I would also prefer that, but I was trying to avoid introducing a breaking change. Would you prefer to modify `error_response` even though it would be a breaking change?

benwis: Yeah, it'll be better to have the better api and do the breaking change.

spencewenski: Alright, sounds good. I’ll update when I get a chance.

server_fn/Cargo.toml
```diff
@@ -25,6 +25,7 @@ send_wrapper = { features = [
   "futures",
 ], optional = true, workspace = true, default-features = true }
 thiserror = { workspace = true, default-features = true }
+bon = { workspace = true }
```
spencewenski, line 28: I'd be open to using `TypedBuilder` instead, or manually writing the necessary builder code.

server_fn/src/lib.rs
spencewenski: Updated the PR, please take a look when you get a chance. I made a few additional breaking changes that I thought made sense and also resolve the middleware issue I originally tried fixing in https://github.com/leptos-rs/leptos/pull/4216.

spencewenski:
I think the following semver error is a false positive; I was able to implement it in one of the example crates.

```
--- failure trait_newly_sealed: pub trait became sealed ---

Description:
A publicly-visible trait became sealed, so downstream crates are no longer able to implement it
        ref: https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.42.0/src/lints/trait_newly_sealed.ron

Failed in:
  trait server_fn::middleware::Service in file /home/runner/work/leptos/leptos/server_fn/src/middleware/mod.rs:46
```

https://github.com/leptos-rs/leptos/actions/runs/16836198701/job/47702412558?pr=4215#step:4:320

spencewenski:
Hey, @benwis , can you take a look at the updated PR when you get a chance? Thanks!

benwis:
Sorry it's been a few days, had some work to tackle. I'm somewhat puzzled by the direction this took, I didn't think another trait like `ServerFnErrorResponseParts` should be needed. In the previous Iteration I believe you attached the content-type header by referencing the already set value from the currently selected codec.
```rust
    let content_type =
                    <Self::Error as FromServerFnError>::Encoder::CONTENT_TYPE;
```
Why not just pass the above as a value to `error_response()` here: https://github.com/leptos-rs/leptos/pull/4215/commits/9ac211f61d7d49738096311608b8606a45e0f572
It's quite possible I've missed something though.

spencewenski:
> Sorry it's been a few days, had some work to tackle. I'm somewhat puzzled by the direction this took, I didn't think another trait like `ServerFnErrorResponseParts` should be needed. In the previous Iteration I believe you attached the content-type header by referencing the already set value from the currently selected codec.
> 
> ```rust
>     let content_type =
>                     <Self::Error as FromServerFnError>::Encoder::CONTENT_TYPE;
> ```
> 
> Why not just pass the above as a value to `error_response()` here: [9ac211f](https://github.com/leptos-rs/leptos/commit/9ac211f61d7d49738096311608b8606a45e0f572) It's quite possible I've missed something though.

`ServerFnErrorResponseParts` is a struct returned by `FromServerFnError::ser` that contains the data required to return the error response, including the serialized error as `Bytes` and the content type. I wanted to create a struct to contain this data instead of adding a `content_type` parameter to `Res::error_response` so that more data could be returned by `FromServerFnError::ser` and passed to `Res::error_response` in the future without needing another API-breaking change. The content type is still coming from `<Self::Error as FromServerFnError>::Encoder::CONTENT_TYPE`, it's just coming from the `FromServerFnError::ser` directly in the `ServerFnErrorResponseParts` struct instead of needing to get it separately.

Returning the data from `FromServerFnError::ser` also resolves the related issue with the server fn middleware from my [other PR](https://github.com/leptos-rs/leptos/pull/4216) in a cleaner way.

server_fn/src/response/generic.rs
```diff
@@ -92,11 +92,12 @@ where
...
 impl Res for Response<Body> {
-    fn error_response(path: &str, err: Bytes) -> Self {
+    fn error_response(path: &str, err: ServerFnErrorResponseParts) -> Self {
         Response::builder()
             .status(http::StatusCode::INTERNAL_SERVER_ERROR)
             .header(SERVER_FN_ERROR_HEADER, path)
-            .body(err.into())
+            .header(header::CONTENT_TYPE, err.content_type)
+            .body(err.body.into())
             .unwrap()
```
benwis, line 101: This is another of those things where it isn't necessarily your changes but as we add more things to the error response I wonder if we should be using `.unwrap()` here. Perhaps this is better as an fallible method in the `TryRes` trait or we need to add additional traits or a default response returned here instead of a panic

spencewenski: Another note is that `tower-http` has the [CatchPanic layer](https://docs.rs/tower-http/latest/tower_http/catch_panic/struct.CatchPanic.html) that can be used to return an  HTTP error response if anything panics, including this `Res::error_response` handler. I have this enabled in my projects.

benwis (approved):

benwis:
I did, but it doesn’t have to be part of this PR. I think this is basically ready to merge now. I don’t know if we have a section for custom errors for server fns in the book, but I’d put it there. Maybe in the server_fns_axum example

gbj:
In general I'm fine with this. I would prefer to use `typed-builder` rather than adding `bon` to the dependency tree, since `typed-builder` is already used in any leptos project. 

It can't be merged until 0.9, given the breaking changes, and there are no plans for 0.9 in the near future, since releasing new breaking versions causes churn that can take the ecosystem a while to catch up to and is therefore kind of annoying for users.

I do think it's possible to do it with relatively fewer changes, by using default trait methods. See #4247 for a non-breaking approach, which we *could* merge into 0.8. (Assuming `cargo-semver-checks` doesn't disagree with me, which is always possible.)

spencewenski:
> In general I'm fine with this. I would prefer to use `typed-builder` rather than adding `bon` to the dependency tree, since `typed-builder` is already used in any leptos project.
> 
> It can't be merged until 0.9, given the breaking changes, and there are no plans for 0.9 in the near future, since releasing new breaking versions causes churn that can take the ecosystem a while to catch up to and is therefore kind of annoying for users.
> 
> I do think it's possible to do it with relatively fewer changes, by using default trait methods. See #4247 for a non-breaking approach, which we _could_ merge into 0.8. (Assuming `cargo-semver-checks` doesn't disagree with me, which is always possible.)

Thanks, @gbj , the non-breaking approach you mentioned is essentially what I did originally. I would like for this to be able to be released relatively soon, so if 0.9 is not planned any time soon I’d prefer to take the non-breaking approach. The breaking approach could be done later when if/when 0.9 is planned. Any objections on that, @benwis ?

benwis:
> > In general I'm fine with this. I would prefer to use `typed-builder` rather than adding `bon` to the dependency tree, since `typed-builder` is already used in any leptos project.
> > It can't be merged until 0.9, given the breaking changes, and there are no plans for 0.9 in the near future, since releasing new breaking versions causes churn that can take the ecosystem a while to catch up to and is therefore kind of annoying for users.
> > I do think it's possible to do it with relatively fewer changes, by using default trait methods. See #4247 for a non-breaking approach, which we _could_ merge into 0.8. (Assuming `cargo-semver-checks` doesn't disagree with me, which is always possible.)
> 
> Thanks, @gbj , the non-breaking approach you mentioned is essentially what I did originally. I would like for this to be able to be released relatively soon, so if 0.9 is not planned any time soon I’d prefer to take the non-breaking approach. The breaking approach could be done later when if/when 0.9 is planned. Any objections on that, @benwis ?

It makes sense to me, sorry to make this a bit complicated. I'm hoping to get a patch release out soon and would love to have at least this partial release in it.

spencewenski:
I updated the PR to go back to the original semver-compatible approach and included some (reworded) comment's from @gbj's [sample PR](https://github.com/leptos-rs/leptos/pull/4247). I also opened a [separate PR](https://github.com/leptos-rs/leptos/pull/4249) for the semver-<em>in</em>compatible approach. It's based on `main` so it will need an update to remove the `Res::content_type` method. Feel free to take that PR and update it or let me know if you start working on 0.9 and I'd be happy to update it. Or, ignore it and take a different approach to solving the issue. 😂 

If things look good please merge when you get the chance because I do not have merge permissions.

spencewenski:
The semver check is failing due to a [different change](https://github.com/leptos-rs/leptos/pull/4224) that’s already in `main`.

```
variant ReloadWSProtocol:Auto in /home/runner/work/leptos/leptos/leptos_config/src/lib.rs:286
semver requires new major version
```

https://github.com/leptos-rs/leptos/actions/runs/17198851352/job/48793426173?pr=4215

spencewenski:
Looks like the breaking change was reverted in `main`, so I synced my branch and all checks are passing now.

spencewenski:
Hey @benwis , can you merge this PR when you get a chance? I don't have merge permissions.

gbj:
Thanks for all your work, and for your patience and flexibility!

PR #4216, opened 2025-08-06
fix!: Set content type if error occurs in Leptos middleware

spencewenski:
Problem
-------
If an error occurs
[in Leptos's middleware chain](https://github.com/leptos-rs/leptos/blob/main/server_fn/src/middleware/mod.rs#L77), the content-type header of the returned error response is not set.

Solution
--------
Save the error's content-type from the boxed service in `BoxedService`, and pass the content type into `Service::run`.

Note that this requires the following breaking changes:
- Add a new field to `BoxedService`.
- Update the `Service::run` trait method to accept the error's content type as a parameter.

However, since this requires breaking changes anyway, this PR achieves
the following in a more extensible way (but still API-breaking):
- Replace the `ser` field in `BoxedService` with a `ServiceRunConfig` that contains the `ser` error serializer and the error's content type string.
- Mark `BoxedService` as `#[non_exhaustive]` so new fields can be added in the future without breaking the API.
- Replace the `ser` parameter of `Service::run` with the `ServiceRunConfig` This allows `Service::run` to remain the same even if new fields are added to the `ServiceRunConfig` in the future.
 
Note: this is written on top of [another PR](https://github.com/leptos-rs/leptos/pull/4215) that's currently open, so there are some changes in this PR that are originally from that PR.

### DIFF
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -307,16 +307,18 @@ pub trait ServerFn: Send + Sized {
                     .await
                     .map(|res| (res, None))
                     .unwrap_or_else(|e| {
-                        (
+                        let mut response =
                             <<Self as ServerFn>::Server as crate::Server<
                                 Self::Error,
                                 Self::InputStreamError,
                                 Self::OutputStreamError,
                             >>::Response::error_response(
                                 Self::PATH, e.ser()
-                            ),
-                            Some(e),
-                        )
+                            );
+                        let content_type =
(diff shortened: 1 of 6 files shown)


benwis (changes requested):
As mentioned in the other PR, let's refactor this to edit error_response()

spencewenski:
This is fixed by https://github.com/leptos-rs/leptos/pull/4215 now since it was decided in that PR that it should be resolved in an API-breaking way.

PR #4224, opened 2025-08-13
feat: add default "auto" live reload protocol option

veigaribo:
The current behavior is to always try to connect to the live reload endpoint via `ws://`, but I think that's not very smart since, if the website is being served via HTTPS, the browser will actively prevent `ws://` connections. Thus defaulting to `wss://` in those situations seems reasonable, as there is not really another option (unless the browser is configured to allow `ws://` from `https://` somehow...).

Thus I introduce a new option for LEPTOS_RELOAD_WS_PROTOCOL: `auto`. That will make it so that the actual protocol to be used from the client will be determined by the protocol used to access the originating website: if the website was accessed via `https://`, then the live reload protocol is `wss://`. Otherwise it is `ws://`.

This was the only option on #1548, but #1613 removed it in favor of static configuration. Now both should be possible.

I also made `auto` the default option since:

- If one is not using HTTPS at all, nothing changes;
- If one is using HTTPS, they should also be using WSS, since the browser would block the connection otherwise, so nothing changes.

So it shouldn't break anything; and, since it allows for both options (HTTP and HTTPS) at the same time, it should be more flexible.

### DIFF
--- a/leptos/src/hydration/mod.rs
+++ b/leptos/src/hydration/mod.rs
@@ -27,6 +27,7 @@ pub fn AutoReload(
             None => options.reload_port,
         };
         let protocol = match options.reload_ws_protocol {
+            leptos_config::ReloadWSProtocol::Auto => "null",
             leptos_config::ReloadWSProtocol::WS => "'ws://'",
             leptos_config::ReloadWSProtocol::WSS => "'wss://'",
         };
--- a/leptos/src/hydration/reload_script.js
+++ b/leptos/src/hydration/reload_script.js
@@ -1,4 +1,9 @@
 let host = window.location.hostname;
+
+if (protocol === null) {
+	protocol = window.location.protocol === 'https:' ? 'wss://' : 'ws://';
+}
+
 let ws = new WebSocket(`${protocol}${host}:${reload_port}/live_reload`);
 ws.onmessage = (ev) => {
 	let msg = JSON.parse(ev.data);
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -153,7 +153,7 @@ impl LeptosOptions {
                 None => None,
             },
             reload_ws_protocol: ws_from_str(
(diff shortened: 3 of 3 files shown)


veigaribo:
> Thanks for the PR! A small nit, then I see no issues with this. Have you looked at how cargo-leptos handles this?

Hm I'm not sure I understood your question, but I tested this change against my project, which uses cargo-leptos. In it, an NGINX reverse proxy is forwarding both to the SSR application and to the live reload server with `ssl`, as that's the use-case this should facilitate. It seems to be working, as far as I can tell.

Do you mean something else? or do you suggest I do something else?

benwis:
No you got it, it's a thing with semver. Thanks for the PR!

gbj:
Thanks for the PR. Because of the semver breakage and associated problems with multiple versions in the 0.8 series (see #4254), I'm going to revert the change later and migrate it over to a `leptos_0.9` branch so it can be included in a future, breaking change. 

Alternately, if it's simply impossible for the browser to use `ws://` if the current page protocol is `https://`, then can't this be done without changing the configuration at all? i.e., simply overwriting the `Ws` preference if that is not possible to use.
```js
if (protocol === 'ws://' && window.location.protocol === 'https://') {
	protocol = 'wss://';
}
```

veigaribo:
> Alternately, if it's simply impossible for the browser to use ws:// if the current page protocol is https://, then can't this be done without changing the configuration at all? i.e., simply overwriting the Ws preference if that is not possible to use.

Well, it is impossible with a regular stock modern browser, but that's hackable (since we're talking about a feature for development, maybe that's not irrelevant...).

I think overwriting the protocol is a worse solution, since it gives less freedom to the user, but realistically yes, it should be fine.

Perhaps we can overwrite the protocol for now (as you suggested) to not introduce a breaking change but still keep this `auto` option in the 0.9 branch for when that comes out?

veigaribo:
For the future, I think `ReloadWSProtocol` (and probably others) could become `non_exhaustive` so adding new variants is not a problem. I find it unlikely that someone would match on `ReloadWSProtocol` anyway...

PR #4227, opened 2025-08-15
`<ShowLet>` component similar to `<Show>` but for `Option`

maccesch:
I've long wanted some kind of `<Show>` component that also sort of unwraps an `Option` or a `Result`. You can do it in code, of course, but that leads to a lot of braces with the closures and nested `view!` macros.

I thought about adding it to Leptos-Use but I think this is such a common use case that it should be in Leptos proper.

In the end I only did this for `Option` because with `Result` the compiler just didn't want to infer the type of the children parameter so you had to write it in code which defeats the purpose.

### Before

Please note the abundance of curly braces 😀

```rust
let (some_value, set_some_value) = signal(Some(42));

view! {
    <div>
        {move || {
            some_value.get().map(|value| view! {
                <div>"There is a value: " {value}</div>
            })
        }}
    </div>
}
```

### After

IMHO much more readable and convenient to write.

```rust
let (some_value, set_some_value) = signal(Some(42));

view! {
    <div>
        <ShowLet some=some_value let:value>
            <div>"There is a value: " {value}</div>
        </ShowLet>
    <div>
}
```

You can also specify a fallback.

```rust
let (some_value, set_some_value) = signal(Some(42));

view! {
    <ShowLet 
        value=some_value
        let:value 
        fallback=|| "Got nothing"
    >
        "We have an ok value: " {value}
    </ShowLet>
}
```

And, finally, you can use closures instead of signals.

```rust
let (some_value, set_some_value) = signal(Some(42));

view! {
    <div>
        <ShowLet some=move || some_value.get().map(|i| i * 2) let:value>
            <div>"There is a value: " {value}</div>
        </ShowLet>
    <div>
}
```

## Naming

I'm not sure if I'm happy with the name of the component. I thought about `<Map value=...>` or `<IfLet some=...>`. But then I thought that the current components don't necessarily reflect the rust programming idioms. And `map` doesn't really reflect it that well because it also unwraps it. Especially with a fallback which is like `.unwrap_or_else(fallback)`. Finally I thought it's closely related to `<Show>`. So the current naming is a mix of `<Show>` and `<IfLet>`.

I'm very open to suggestions. The current name kind of feels neither here nor there. What do you think?


benwis:
I'm curious what this would do with something like `Option<Option<String>>` or similar. Can I nest these?

maccesch:
Yes you should be able to nest these arbitrarily. Although at that point I think flattening might make  sense

```rust
view! {
    <ShowLet some=move || some_value.get().flatten() let:value>
        <div>"There is a value: " {value}</div>
    </ShowLet>
}
```

benwis:
I like it, and thanks for the PR!

jimbrankelly:
I like it.  ShowLet as the name works for me.

PR #4236, opened 2025-08-21
made `<Show>` accept signals in addition to closures

maccesch:
On the heels of `<ShowLet>` I added the same sort of prop conversion to `<Show>`.

Before (in stable):

```rust
let (value, set_value) = signal(true);

view! {
    <Show when=move || value.get()>
        ...
    </Show>
}
```

After:

```rust
let (value, set_value) = signal(true);

view! {
    <Show when=value>
        ...
    </Show>
}
```

I think the latter is cleaner (it still works also with closures) and more in line like you can use signals in HTML Attributes.

If you think this is useful I'm going to add it to `<AnimatedShow>` as well which, curiously, currently only supports signals and not closures.

### DIFF
--- a/leptos/src/show.rs
+++ b/leptos/src/show.rs
@@ -1,30 +1,115 @@
 use crate::{
     children::{TypedChildrenFn, ViewFn},
+    prelude::{FunctionMarker, SignalMarker},
     IntoView,
 };
 use leptos_macro::component;
 use reactive_graph::{computed::ArcMemo, traits::Get};
+use std::{marker::PhantomData, sync::Arc};
 use tachys::either::Either;
 
+/// Shows its children whenever the condition `when` prop is `true`.
+/// Otherwise it renders the `fallback` prop, which defaults to the empty view.
+///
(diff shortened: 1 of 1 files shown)


benwis:
I think it makes sense to me, and I'd love to see some stability in that UI.

maccesch:
Cool. Getting on it...

maccesch:
The failures seem to be weird github runner failures that have nothing to do with my changes.

From my POV this is ready to be merged. Unless you want me to change sth of course.

benwis:
I agree the ci is broken again, thanks for the PR

gbj:
This PR does not compile when the `nightly` feature is enabled. See #4254. I thought at least one of our examples was supposed to be tested on nightly in CI, not sure what went awry there.

maccesch:
Sorry about that! Just fixed it in a new PR

maccesch:
> This PR does not compile when the `nightly` feature is enabled. See #4254. I thought at least one of our examples was supposed to be tested on nightly in CI, not sure what went awry there.

I just did a quick search and it seems none of the example use nightly currently.

PR #4247, opened 2025-08-24
Set correct `Content-Type` for server fn errors (non-breaking) (closes #4209)

### LINKED ISSUE #4209: Server functions do not set content-type header when an error is returned

gbj:
cf. #4215. This is an alternate approach that shouldn't be semver-breaking and therefore can be merged into 0.8.

The only case it doesn't cover is the case in which a *middleware* throws the error, which would require modifying `BoxedService` as in #4215.

### DIFF
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -568,7 +568,7 @@ pub trait FromServerFnError: std::fmt::Debug + Sized + 'static {
     /// Converts a [`ServerFnErrorErr`] into the application-specific custom error type.
     fn from_server_fn_error(value: ServerFnErrorErr) -> Self;
 
-    /// Converts the custom error type to a [`String`].
+    /// Serializes the custom error type to bytes, according to the encoding given by `Self::Encoding`.
     fn ser(&self) -> Bytes {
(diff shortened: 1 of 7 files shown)


gbj:
Closed in favor of modified #4215.

PR #4249, opened 2025-08-25
fix!: Set content type header for server function errors

spencewenski:
Problem
-------
When a server function returns an error, the content-type header is not set.

Solution
--------
In order to set the content-type header in the response, the `FromServerFnError::Encoder::CONTENT_TYPE` associated constant needs to make its way to where the response object is constructed. This PR achieves this with the following changes:

- **[breaking]** Update the `FromServerFnError#ser` trait method to return the content type as well as the serialized error response bytes. This is returned in a new `ServerFnErrorResponseParts` struct that contains both values and is marked `#[non_exhaustive]` to allow adding new fields in the future without breaking semver.
- **[breaking]** Update the type of the `BoxedService#ser` field to be a function that returns `ServerFnErrorResponseParts` to match the updated `FromServerFnError#ser` trait method. The field is also renamed to `err_ser`.
- **[breaking**] Mark `BoxedService` as `#[non_exhaustive]` to allow adding new fields in the future without breaking semver.
- **[breaking]** Update the `Res#error_response` trait method to take `ServerFnErrorResponseParts` instead of `Bytes`.
- Update the implementations of `Res#error_response` to set the content-type header using the value from `ServerFnErrorResponseParts`.
- Add a dependency on `bon` to auto-generate a builder API for the `ServerFnErrorResponseParts` struct.

Note: I wasn't entirely sure how to handle `ServerFnErrorResponseParts` being used in stream/websocket responses, so I simply extracted the `body` field in order to match the existing behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -499,6 +499,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "bon"
+version = "3.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33d9ef19ae5263a138da9a86871eca537478ab0332a7770bac7e3f08b801f89f"
+dependencies = [
+ "bon-macros",
+ "rustversion",
+]
+
+[[package]]
+name = "bon-macros"
+version = "3.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "577ae008f2ca11ca7641bd44601002ee5ab49ef0af64846ce1ab6057218a5cc1"
+dependencies = [
+ "darling",
+ "ident_case",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "rustversion",
+ "syn 2.0.104",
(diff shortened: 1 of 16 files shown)


spencewenski:
Todo: Use `typed-builder` instead of `bon` as noted in this comment: https://github.com/leptos-rs/leptos/pull/4215#issuecomment-3218304021

gbj:
Just FYI @spencewenski, I have now opened a `leptos_0.9` branch to start collecting changes for an eventual 0.9. If you're happy with the current state of things after the non-breaking PR, great. Otherwise, if you'd like to update this PR (resolving merge conflicts, using `typed-builder`, etc.) that would be great too.

spencewenski:
Thanks, @gbj! I'll pick this back up soon.

spencewenski:
@gbj, @benwis, I updated to the 0.9 branch so this should be ready to re-review.

PR #4257, opened 2025-08-26
fix: use wss for live reload if on https

veigaribo:
Non-breaking version of #4224, as suggested by @gbj.

This builds on top of #4255.

I would like to make another PR reintroducing the `auto` option for an eventual 0.9 release, as I think it's a more elegant solution.


gbj:
Thanks for the new PR! I completely agree with you that adding an `auto` option in a breaking 0.9 release is more elegant. There is no short-term timeline for that but I will tag it with the 0.9 milestone to make sure it doesn't get lost!

PR #4258, opened 2025-08-26
impl From<RwSignal/ReadSignal/Memo> for ArcSignal

zakstucke:
Some extra conversion impls for signal types to `ArcSignal` that seem to have just been missed.


gbj:
Cheers!

PR #4265, opened 2025-08-27
Reintroduce `auto` live reload protocol option

veigaribo:
Substitutes #4257 for #4224, which is a breaking change intended for an eventual 0.9 release.

Additionally makes the `leptos_config::ReloadWSProtocol` and `leptos_config::Env` enums non exhaustive, since simply adding configuration options should probably not be a breaking change. However, this makes it so that the `match` in the `AutoReload` component needs to have a default pattern case for unknown variants, since it's from a different crate than the enum. This means that adding new variants to `ReloadWSProtocol` without handling it in the component will not cause any error and will silently **not** work. I hate this, but I don't see a solution.


veigaribo:
@gbj is this still meant for 0.9?
I don't know if there's a way to change the branch this PR is targeting

gbj:
Thanks for the reminder, apologies for the delay!

PR #4267, opened 2025-08-28
Preserve owner in Actions and event listeners

zakstucke:
Migrating code from 0.6, it's a common painpoint that context doesn't work in `Action`s or event listeners.

I took a look into this today and it was actually pretty easy to support unless there are some downsides I'm missing!

With this PR, the following works:

```rust
    #[derive(Debug, Clone)]
    struct Ctx;
    provide_context(Ctx);

    let a = Action::new(move |_arg: &()| {
        expect_context::<Ctx>();
        async move {
            expect_context::<Ctx>();
        }
    });
    let a_local = Action::new_local(move |_arg: &()| {
        expect_context::<Ctx>();
        async move {
            expect_context::<Ctx>();
        }
    });
    Effect::new(move || {
        spawn_local(async move {
            a.dispatch(());
            a_local.dispatch(());
        });
    });

    view! {
        <button
            on:click={move |_| {
                expect_context::<Ctx>();
            }}
        >
            "Click me"
        </button>
    };
```

### DIFF
--- a/reactive_graph/src/actions/action.rs
+++ b/reactive_graph/src/actions/action.rs
@@ -207,10 +207,10 @@ where
             version: Default::default(),
             dispatched: Default::default(),
             action_fn: Arc::new(move |input| {
-                Box::pin(ScopedFuture::new_untracked_with_owner(
-                    owner.with(|| action_fn(input)),
-                    owner.clone(),
-                ))
+                Box::pin(
+                    owner
+                        .with(|| ScopedFuture::new_untracked(action_fn(input))),
+                )
             }),
             #[cfg(any(debug_assertions, leptos_debuginfo))]
             defined_at: Location::caller(),
@@ -386,10 +386,8 @@ where
             dispatched: Default::default(),
             action_fn: Arc::new(move |input| {
                 Box::pin(SendWrapper::new(
-                    ScopedFuture::new_untracked_with_owner(
-                        owner.with(|| action_fn(input)),
-                        owner.clone(),
-                    ),
+                    owner
+                        .with(|| ScopedFuture::new_untracked(action_fn(input))),
                 ))
             }),
             #[cfg(any(debug_assertions, leptos_debuginfo))]
--- a/reactive_graph/src/actions/action.rs
+++ b/reactive_graph/src/actions/action.rs
@@ -1,7 +1,7 @@
 use crate::{
-    computed::{ArcMemo, Memo},
+    computed::{ArcMemo, Memo, ScopedFuture},
     diagnostics::is_suppressing_resource_load,
-    owner::{ArcStoredValue, ArenaItem},
+    owner::{ArcStoredValue, ArenaItem, Owner},
     send_wrapper_ext::SendOption,
     signal::{ArcMappedSignal, ArcRwSignal, MappedSignal, RwSignal},
     traits::{DefinedAt, Dispose, Get, GetUntracked, GetValue, Update, Write},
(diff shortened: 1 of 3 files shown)


gbj:
I have traditionally not wanted to do this because it adds overhead for *every* use of actions/event listeners in exchange for a usability benefit for the subset of actions/event listeners that use context. In 0.6 and previous context wasn't intended to work in these locations but did in specific instances where it was assigned to a kind of "ambient null owner" (which itself has the potential for memory leaks/cleanup issues)

In all the cases in the example, context can be expected in the component body and moved into the action or event listener.

zakstucke:
> In all the cases in the example, context can be expected in the component body and moved into the action or event listener.

@gbj Totally, it's mainly for quality of life to have context everywhere and avoid the need to prop drill.
It usually arises as an issue when you want to start abstracting logic into reusable functions. You effectively can't do this inside actions, event listeners etc without some custom wrapper or prop drilling and is much more difficult than if you could expect some context by default. Just as importantly it's quite unintuitive and a tricky to debug quirk for inexperienced users.

In my opinion the ergonomics outweigh any performance impact here.

I've just run a quick benchmark for wasm+server and the overhead is coming in pretty negligible:

```
SERVER: Ran 21,585,000 times in 1000 ms total, 46.328 ns/run
BROWSER: Ran 32,702,000 times in 1000 ms total, 30.579 ns/run
```

 `30-45ns` (20-30 millionths of a second) of overhead per event listener call/`action.dispatch()`/future poll etc.

That seems reasonable to me for the benefit. 
If you'd still prefer not to have it, would you be open to adding an `owner_scope_extended` or similarly named feature  that opts into the performance/ergonomic tradeoff? I would want to make that tradeoff.

(The code being benchmarked is `Owner::current().unwrap().with(|| {});`)
```rust
    Effect::new_isomorphic(move || {
        #[cfg(feature = "ssr")]
        fn print(s: impl AsRef<str>) {
            println!("{}", s.as_ref());
        }
        #[cfg(not(feature = "ssr"))]
        fn print(s: impl AsRef<str>) {
            web_sys::console::log_1(&s.as_ref().into());
        }

        #[cfg(feature = "ssr")]
        let get_elapsed_ms = {
            let instant = std::time::Instant::now();
            move || instant.elapsed().as_millis() as f64
        };
        #[cfg(not(feature = "ssr"))]
        let get_elapsed_ms = {
            let perf = window().performance().unwrap();
            let start = perf.now();
            move || perf.now() - start
        };

        let mut count = 0u128;
        while !(count % 1_000 == 0 && get_elapsed_ms() >= 1000.0) {
            Owner::current().unwrap().with(|| {});
            count += 1;
        }

        let elapsed_ms = get_elapsed_ms();
        let ns_per_run = (elapsed_ms * 1_000_000.0) / count as f64;

        let result = format!(
            "Ran {} times in {:.3} ms total, {:.3} ns/run",
            count, elapsed_ms, ns_per_run,
        );
        if cfg!(feature = "ssr") {
            println!("SERVER: {}", result);
        } else {
            web_sys::console::log_1(&format!("BROWSER: {}", result).into());
        }
    });
```

gbj:
You're probably right. My performance concern was mostly about the creation cost for (say) 1000 event listeners at a time, when you render a big screen -- not about the cost of reestablishing it once on user interaction. I guess `Owner::current()` is just a single RefCell check and then upgrading a couple Weak pointers, so it's not actually that bad.

benwis:
It's funny we were just talking about this recently. I don't think this qualifies a breaking change, since I can't imagine anyone is depending on not having context in these scenarios. @gbj we can pull this out if we need to

PR #4269, opened 2025-08-29
Standardize ScopedFuture::new_untracked like untrack() and untrack_with_diagnostics()

zakstucke:
- Changes the behaviour of `ScopedFuture::new_untracked` to not emit diagnostics, in line with `subscriber::untrack`
- Adds new `ScopedFuture::new_untracked_with_diagnostics`, to match `subscriber::untrack_with_diagnostics`
- Replaces the call in `ArcAsyncDerived::new_with_manual_dependencies` to use `ScopedFuture::new_untracked_with_diagnostics`, as this one we want diagnostics. 
- The other usage of `ScopedFuture::new_untracked` is in `Action` usage, which is left unchanged, thereby disabling diagnostics, which are not needed in this context.
- Adds `untrack` to the sync part of `Action` s to disable diagnostics there too

In case there's some confusion about why this is useful: 

Having these diagnostics in action calls does not provide any value, a user is acutely aware that an action is something manually "dispatched". By disabling the warning, thereby it no longer being an error to use e.g. `.get()` inside an action, logic and getters that might contain signals and also be used in the `view!{}`/`Effects`, can be extracted into reusable functions, and not produce confusing warnings when used from actions.

### DIFF
--- a/reactive_graph/src/computed/async_derived/mod.rs
+++ b/reactive_graph/src/computed/async_derived/mod.rs
@@ -43,11 +43,25 @@ impl<Fut> ScopedFuture<Fut> {
         }
     }
 
-    #[doc(hidden)]
-    #[track_caller]
-    pub fn new_untracked_with_diagnostics(fut: Fut) -> Self {
+    /// Wraps the given `Future` by taking the current [`Owner`] re-setting it as the
+    /// active owner every time the inner `Future` is polled. Always untracks, i.e., clears
+    /// the active [`Observer`] when polled.
+    pub fn new_untracked(fut: Fut) -> Self {
         let owner = Owner::current().unwrap_or_default();
         Self {
+            owner,
+            observer: None,
+            fut,
+        }
+    }
+
+    #[doc(hidden)]
+    #[track_caller]
+    pub fn new_untracked_with_diagnostics(
+        fut: Fut,
+    ) -> ScopedFutureUntrackedWithDiagnostics<Fut> {
+        let owner = Owner::current().unwrap_or_default();
+        ScopedFutureUntrackedWithDiagnostics {
             owner,
             observer: None,
             fut,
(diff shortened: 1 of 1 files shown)


reactive_graph/src/computed/async_derived/mod.rs
```diff
@@ -25,6 +25,7 @@ pin_project! {
     pub struct ScopedFuture<Fut> {
         pub owner: Owner,
         pub observer: Option<AnySubscriber>,
+        diagnostics: bool,
```
gbj, line 28: Should these be marked `#[cfg(any(debug_assertions, leptos_debuginfo))]`? I think otherwise diagnostics are turned off anyway.

reactive_graph/src/computed/async_derived/mod.rs
```diff
@@ -61,8 +76,17 @@ impl<Fut: Future> Future for ScopedFuture<Fut> {
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.project();
-        this.owner
-            .with(|| this.observer.with_observer(|| this.fut.poll(cx)))
+        this.owner.with(|| {
+            this.observer.with_observer(|| {
+                #[cfg(debug_assertions)]
```
gbj, line 81: Should this be `#[cfg(any(debug_assertions, leptos_debuginfo))]`?

gbj (commented):
Sounds great! The explanation all makes sense to me. Couple small notes on debug-cfg'ing, otherwise I'm happy.

reactive_graph/src/computed/async_derived/mod.rs
zakstucke: I thought about this but I don't think so, `leptos_debuginfo` is there to help you debug panics in prod, so it's maintaining location information in all the errors, but non of the diagnostic stuff currently has `leptos_debuginfo` and I think that makes sense because: 
- It's a runtime thing so performance (`ReactiveZone.enter()`)
- They are just warnings

zakstucke:
I've had to restructure to avoid breaking semver, so we should merge this PR now, and https://github.com/leptos-rs/leptos/pull/4274 for 0.9. 

Edit: ~~hmm actually looking at the diff again, this might still theoretically break semver in the CI check, because the future type is different, but imo in practice it will not break any user code so will be fine.~~ should be fine now

gbj:
Ah! Yeah, fair enough. I'm fine with the new version. Thanks, Zak.

PR #4273, opened 2025-08-30
Into for Signal from reactive closure

zakstucke:


benwis:
So my first reaction was Woah and my second was How? so you've got a pretty interesting PR there. I'd love to not have to manually derive signals. Does the addition of generics on the generated into functions  affect compile times at all?

zakstucke:
> Does the addition of generics on the generated into functions affect compile times at all?

I haven't tested, I would assume non-zero, but also that it wouldn't be meaningful, I'm not actually asking the compiler for much here, it's an `A/B` generic, so there's not any type explosion or anything.

benwis:
> > Does the addition of generics on the generated into functions affect compile times at all?
> 
> I haven't tested, I would assume non-zero, but also that it wouldn't be meaningful, I'm not actually asking the compiler for much here, it's an `A/B` generic, so there's not any type explosion or anything.

I expect it to be minimal as well, just curious.

zakstucke:
I've made it robust and not specific to signals, I've added support for `Callback<T, _>` too, where before you'd be forced to do `Callback<(T,), _>` and use `(|(_,)| {}).into()`, now `Callback<T, _>` and `(|_| {}).into()` works too. 

This required me to move the callback module into `reactive_graph` though, the method only works with types defined inside the same crate.

zakstucke:
The semver CI failing is a false positive I believe, it doesn't understand the old `callback` module is still available, it's just sourced from `reactive_graph` now instead. The paths it complains are missing are all fine, `leptos::callback::*`, `leptos::prelude::Callback` etc.

I'm happy with this PR now I think, it's just blocked by the upstream PR:
https://github.com/idanarye/rust-typed-builder/pull/169

Cargo.toml
```diff
@@ -79,7 +79,10 @@ itertools = { default-features = false, version = "0.14.0" }
 convert_case = { default-features = false, version = "0.8.0" }
 serde_json = { default-features = false, version = "1.0.143" }
 trybuild = { default-features = false, version = "1.0.110" }
-typed-builder = { default-features = false, version = "0.21.2" }
+# typed-builder = { default-features = false, version = "0.21.2" }
+# typed-builder-macro = { default-features = false, version = "0.21.0" }
+typed-builder = { git = "https://github.com/zakstucke/rust-typed-builder.git", rev = "3bc55429a0036d0143ed96f9e48715def754f99d" }
```
gbj, line 84: Note: we should remember to revert these two when the upstream PR is merged. (We can't publish to crates.io with git dependencies)

reactive_graph/src/callback.rs
```diff
@@ -5,7 +5,7 @@
 //! # Usage
 //! Callbacks can be created manually from any function or closure, but the easiest way
 //! to create them is to use `#[prop(into)]]` when defining a component.
-//! ```
+//! ```ignore
 //! use leptos::prelude::*;
```
gbj, line 9: docs nit: granted this that is marked `ignore`, it's possible to use `reactive_graph` independently of `leptos` so if the module is being moved here the import should probably be updated

(I think it would be reasonable to update it to `# use reactive_graph::/* whatever */` and remove the `ignore` so that tests whether it compiles, but the reference to leptos doesn't appear in the docs)

EDIT: Ok I hadn't clicked to expand and see that yes, it's a whole Leptos example :smile: of course. Hm. Not sure the best way to handle this while migrating it.

reactive_graph/src/into_leptos_value.rs
```diff
@@ -0,0 +1,59 @@
+#[doc(hidden)]
+pub struct __IntoLeptosValueMarkerBaseCase;
+
+/// TODO docs
```
gbj, line 4: just flagging the TODO

gbj (commented):
... Wow! 😍 This is really great. Honestly kind of mindblowing.

I understand why using a marker generic lets us get around the conflicting trait implementations (they are no longer conflicting.)

I'm a little worried about the potential for this to break in subtle ways where the compiler can't infer the type of the marker. (I haven't thought it through that fully, or found cases where it does break -- I'm just used to the compiler enough to anticipate it! :smile:)

My only comment other than the individual nits is a bikeshedding one about the name: if we're putting it in the non-leptos-specific `reactive_graph` trait perhaps `IntoReactiveValue` or something rather than `IntoLeptosValue`?

In any case, awesome work and an ingenious solution.

reactive_graph/src/callback.rs
zakstucke: Yeah I'm not sure, I could make the example simpler and not reference leptos, or add a second example that doesn't have `ignore` and is local to `reactive_graph`

zakstucke:
> I'm a little worried about the potential for this to break in subtle ways where the compiler can't infer the type of the marker. (I haven't thought it through that fully, or found cases where it does break -- I'm just used to the compiler enough to anticipate it! 😄)

😅 I haven't been able to find a valid case it fails yet, as long as the `T` in `Signal<T>` is concretely known, the compiler seems to have no issue with it, which was already a requirement with our signal implementations. The fact it can handle `|| &'static str` -> `Signal<String>` without issues (the specific conversion is implemented mind), gives me more confidence. 

I'd encourage you to try and break it though!

> My only comment other than the individual nits is a bikeshedding one about the name: if we're putting it in the non-leptos-specific reactive_graph trait perhaps IntoReactiveValue or something rather than IntoLeptosValue?

Very happy to change it, but I don't think `IntoReactiveValue` works for `Callback`. It's also something a user might want to use directly, if they want a function that takes `foo<M>(val: impl IntoLeptosValue<Signal<usize>, M>)` or use `value.into_leptos_value()`.  Also, this `IntoLeptosValue` trait is being used instead of `.into()` for all `#[prop(into)]`, we might want to have implementations on all sorts of types down the line.

benwis:
Not sure I like IntoLeptosValue either, not really clear what a LeptosValue is 😛

maccesch:
> I'm a little worried about the potential for this to break in subtle ways where the compiler can't infer the type of the marker. (I haven't thought it through that fully, or found cases where it does break -- I'm just used to the compiler enough to anticipate it! 😄)

We're actually using the same kind of trick with `Effect` since 0.7 that accepts a function with and without parameters. And I haven't come across a case yet where the compiler couldn't infer the marker type. I suspect it's going to be same here.

gbj:
Thanks @maccesch. There were a few places I came across inference issues with the `Effect` changes, in places where it had trouble inferring the type of `T` in `prev: Option<T>` and I had to manually specify the type `Effect::new(move |prev: Option<()>|` or something similar. 

For example:
```rust
#[component]
fn App() -> impl IntoView {
    Effect::new(move |prev| {
        if prev.is_some() {
            log!("not the first run");
        }
        log!("foo bar {prev:?}");
    });
}
```
compiled on 0.6, but on 0.7/0.8 it yields the error
```
 --> src/main.rs:6:23
  |
6 |     Effect::new(move |prev| {
  |                       ^^^^
7 |         if prev.is_some() {
  |            ---- type must be known at this point
  |
help: consider giving this closure parameter an explicit type
  |
6 |     Effect::new(move |prev: /* Type */| {
```

(The changes to Effect are still a net win, in my opinion, to be clear!)

It's possible that this kind of inference error doesn't apply here because the component prop type needs to be provided. I haven't fully though through how the two implementations parallel each other... (and I'm going to log off for the evening instead!)

maccesch:
Understood. Excited to see where you land on this.

reactive_graph/src/nightly.rs
```diff
@@ -150,8 +147,8 @@ impl_get_fn_traits_get_arena![
     ReadSignal,
     RwSignal,
     ArcMemo,
-    ArcSignal,
-    Signal,
+    // ArcSignal,
```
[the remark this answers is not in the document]
zakstucke: Ah good catch, reminder I need to look at this, disabling them fixed something in the nightly CI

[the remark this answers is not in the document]
zakstucke: Yep exactly, so it's either keeping `Signal::derive()` on nightly or removing the `my_signal()` "getting" without `.get()` nightly feature. I'm assuming we should just disable this PR when the "nightly" feature is enabled, it's pretty easy to do, to prevent a breaking change.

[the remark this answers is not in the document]
zakstucke: I've disabled the trait impls for signal closures on nightly. Other parts, such as the `Callback` improvements still apply to nightly.

reactive_graph/src/callback.rs
zakstucke: Updated

zakstucke:
Should be good to go now other than any bikeshedding on `IntoLeptosValue`. 

As mentioned before the failed semver CI check is a false positive.

sabify:
I think the failed semver check is due to moving callback module from `leptos` to `reactive_gragh` and it shouldn't be a false positive...

zakstucke:
> I think the failed semver check is due to moving callback module from leptos to reactive_gragh and it shouldn't be a false positive...

The callback module is then re-exported from the leptos crate, and is accessible identically as before, cargo-semver-checks just doesn't seem to realise that. Unless you see something I'm missing?

image

leptos/src/lib.rs
```diff
@@ -204,7 +204,7 @@ pub mod prelude {
 pub mod form;
 
 /// A standard way to wrap functions and closures to pass them to components.
-pub mod callback;
+pub use reactive_graph::callback;
```
sabify, line 207: Just experimenting; can we trick the semver check by using the following?

```rust
pub use reactive_graph::callback as callback;
```

sabify: The failed semver error is clear that `A publicly-visible module cannot be imported by its prior path` (leptos/src/callback.rs). A dummy module in `leptos/src/callback.rs` with:

```rust
pub use reactive_graph::callback::*;
```

and in `lib.rs`:

```rust
pub mod callback;
```

may satisfy the checks but I don't know if that would be a viable option!

sabify: If that works, a TODO tag should be wrriten to remove the redundant module in next major release.

gbj:
I guess I'm okay with just disabling this on nightly, although it raises a moderately large issue: namely that it would prevent us from stabilizing the function-call syntax once the relevant nightly traits are stabilized without removing this implementation again (which would be a breaking change).

I may be having trouble remembering the history here.

Was the situation in 0.6 that `Signal<T>` supported `From<Fn() -> T>` but it did *not* implement `Fn()` on nightly? That seems plausible. In which case maybe removing the `Fn()` implementation for `Signal<_>` in 0.9 would allow us to solve this problem by unifying the two things. 

In any case I am sure that merging this is the right idea -- I will just wait for the CI to finish running.

gbj:
Oh and re: bikeshedding — are you okay with `IntoReactiveValue`/`.into_reactive_value()`?

zakstucke:
> Oh and re: bikeshedding — are you okay with IntoReactiveValue/.into_reactive_value()?

I'm okay with anything, although I'd hesitate that `IntoReactiveValue` doesn't make sense with things like `Callback`, which are already using and benefit from it. If it makes sense to you though I have no objection, for the main intended use case it's completely behind the scenes anyway. 

> I may be having trouble remembering the history here.
> Was the situation in 0.6 that Signal<T> supported From<Fn() -> T> but it did not implement Fn() on nightly? That seems plausible.

It's a good question, I'm unsure too and could check deeper next week but a quick check back in the git history seems like it might have been a 0.8 feature (https://github.com/leptos-rs/leptos/commit/61f5294f673baaf586e814b2aa103af4824495f9, 1f2b13a976e15fa2b07d48d096c4902b7e9d9aea)?  Either way I think the standardisation to this PR in a 0.9 for nightly also, rather than keeping the nightly `Fn()` feature which doesn't have a guarantee of stabilising, does make sense. Worth mentioning this PR still improves the `Callback` type on nightly today, the `Fn()` feature only blocks the signal improvements specifically, not the overall PR.

zakstucke:
> Was the situation in 0.6 that Signal<T> supported From<Fn() -> T> but it did not implement Fn() on nightly?

I checked this out, and it did support both by utilising `auto_traits` and `negative_impls`! I've been able to use the same trick here, so it actually now works on nightly too 🎉

gbj:
Hey @zakstucke — Sorry, I lost track of this for a while. I think we tested out the nightly solution and you were going to disable it due to the issues, right? Happy to merge after you've made that change.

zakstucke:
Hi @gbj all good! 

> I think we tested out the nightly solution and you were going to disable it due to the issues, right?

Yep, so as we discussed the last approach wouldn't work because the nightly feature is unsound in annoying ways...

But...

I was put onto some dark magic by @TethysSvensson dioxus are doing to make their signals work with function syntax even on stable, and it looks like we can use that to keep the existing nightly behaviour?

Updated PR to use said hack specifically for the `Signal`, `ArcSignal` case this effects and only on nightly, but theoretically could be used on stable, and apply to all the other Copy getters without args, too. Outside the scope of this PR.

The dioxus version:
https://bsky.app/profile/fasterthanli.me/post/3m2ckml34vs2x
https://docs.rs/dioxus-signals/0.6.3/src/dioxus_signals/signal.rs.html#485-494

gbj:
Working through PRs again. I'd love to merge this one relatively soon; as I noted via Discord, the deref-to-`dyn Fn()` approach doesn't quite work without breaking semver on `nightly` because "a type that Derefs to a `dyn Fn()`" does not `impl Fn()`, so components that took `impl Fn()` for a prop (for example, `Show`) would not longer accept `Signal` directly. 

Just writing that down publicly for anyone curious. If you remove the `dyn Fn()` piece for now and disable this on nightly again, I will merge it.

zakstucke:
Yep, I forgot to update it after our last discussion, just done so now, reverted back to the disabled on nightly version. 

I think on the next breaking change opportunity it would make sense to break nightly to "conform", as I think this is the overall right direction forward to some unified behaviour.

gbj:
Thanks, Zak! :tada:

PR #4278, opened 2025-09-02
Allow MultipartData ActionForms

DzenanJupic:
In [the chapter about `ActionForm`s](https://book.leptos.dev/progressive_enhancement/action_form.html), it says:
> Note: `<ActionForm/>` only works with the default URL-encoded `POST` encoding for server functions, to ensure graceful degradation/correct behaviour as an HTML form.

This, however, prevents file uploads from working without JavaScript being enabled. As multi part forms are also supported natively, I feel like it should be possible to allow using them as well.

This is a very naive - but functional - approach that implements support for this. There are currently a few issues though:

1. The `Clone for MultipartData` impl on the server side panics. I'm not sure how to best clone the server value. It's not required for this use case, but should obviously be fixed. I think I need some guidance here.
2. The `server_fn_macro` impl now refers to `leptos`, which is probably not desirable, as it is a standalone crate. I don't know the code base well enough to move this implementation somewhere else. How could I best solve this?
3. The `server_fn/multipart` feature is now enabled unconditionally. The easy solution would be to add another feature to `leptos`. But I feel like resolving issue `2` will also lead to a nicer solution here.

Also, while testing, I found what I believe is a bug. Before specifying the `enctype` on the `ActionForm`, the server-side code would panic with `couldn't parse boundary` when submitting the form. It seems like the `FromReq<MultipartFormData, Request, E>` impl doesn't handle encoding errors at all.

Here is some example code for testing:
```rust
use leptos::prelude::*;

#[component]
pub fn Upload() -> impl IntoView {
    let upload_file = ServerAction::<Upload>::new();

    view! {
        <ActionForm action=upload_file>
            <label>
                "File"
                <input
                    type="file"
                    name="file"
                    accept=".csv, text/*"
                />
            </label>
            <input type="submit" value="Upload" />
        </ActionForm>
    }
}

#[server(input = server_fn::codec::MultipartFormData)]
async fn upload(data: server_fn::codec::MultipartData) -> Result<(), ServerFnError> {
  let _ = data;
  Ok(())
}
```

### DIFF
--- a/server_fn/src/codec/multipart.rs
+++ b/server_fn/src/codec/multipart.rs
@@ -31,6 +31,20 @@ pub enum MultipartData {
     Server(multer::Multipart<'static>),
 }
 
+impl Clone for MultipartData {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Client(data) => {
+                let form_data = (*data.0).clone();
+                Self::from(form_data)
+            }
+            Self::Server(_) => {
+                panic!("Cannot clone server side multipart data")
+            }
+        }
+    }
+}
+
 impl MultipartData {
     /// Extracts the inner data to handle as a stream.
     ///
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -341,9 +341,10 @@ impl ServerFnCall {
                     Clone, #server_fn_path::rkyv::Archive, #server_fn_path::rkyv::Serialize, #server_fn_path::rkyv::Deserialize
                 },
             ),
-            Some("MultipartFormData")
-            | Some("Streaming")
-            | Some("StreamingText") => (PathInfo::None, quote! {}),
+            Some("MultipartFormData") => (PathInfo::None, quote! { Clone }),
+            Some("Streaming") | Some("StreamingText") => {
+                (PathInfo::None, quote! {})
+            }
             Some("SerdeLite") => (
                 PathInfo::Serde,
                 quote! {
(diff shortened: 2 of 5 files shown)


server_fn_macro/src/lib.rs
```diff
@@ -744,6 +745,31 @@ impl ServerFnCall {
...
+
+        let Some((field_name, _)) = self.single_field() else {
+            return quote! {};
+        };
+
+        let server_fn_path = self.server_fn_path();
+        let struct_name = self.struct_name();
+
+        quote! {
+            impl ::leptos::form::FromFormData for #struct_name {
```
gbj, line 761: I'd suggest that this should be handled in a similar way to the `server_fn_path`: have a `from_form_data` field in the `ServerFnCall` struct that takes `Option<Path>`, which allows leptos to provide `::leptos::form::FromFormData`, but would allow other crates to provide their own path here. If `from_form_data` is `None`, omit this implementation.

server_fn/src/codec/multipart.rs
```diff
@@ -31,6 +31,20 @@ pub enum MultipartData {
...
 
+impl Clone for MultipartData {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Client(data) => {
+                let form_data = (*data.0).clone();
+                Self::from(form_data)
+            }
+            Self::Server(_) => {
+                panic!("Cannot clone server side multipart data")
```
gbj, line 42: I agree with you that this is awkward. We should not implement `Clone` on a type that is not actually cloneable by panicking on the server. I think the piece of context I'm missing is — Why does this need to have a `Clone` implementation at all? (If you have a `MultipartData` and need to clone it you can call `.into_client_data().map(|data| data.take())` and clone that instead)

gbj (commented):
Thanks for the PR! Apologies for my delay in reviewing it. 

I think this is great. Making multipart data easier to work on is something that comes up occasionally, and this is a good step in that direction. I left some comments, which I hope are helpful.

PR #4294, opened 2025-09-12
fix: prevent double-rebuild and correctly navigate multiple times to same lazy route (closes #4285)

### LINKED ISSUE #4285: Lazy loading with split modules panicing on multiple calls
**Describe the bug**
Calling a lazy route multiple times while loading causes panics and halts the execution.

**Leptos Dependencies**

leptos [head](https://github.com/leptos-rs/leptos/commit/3eaabf85eac67d2271f33595d9e133e04bef1d1e).

**To Reproduce**

1. Go to `lazy_routes` example and run it.
2. Click on view 'C' or 'D' multiple times and quick.
3. You should see the panics in console and the execution halt.

**Next Steps**

- [ ] I will make a PR
- [x] I want someone else to take the time to fix this

gbj:


examples/regression/e2e/features/issue_4285.feature
```diff
@@ -0,0 +1,9 @@
+@check_issue_4005
```
jhoobergs, line 1: Small comment: I think you meant 4285 instead of 4005 on this line and the next.

PR #4295, opened 2025-09-14
Locking dependencies in cargo-leptos install example

agirorn:
This limits dependency errors on install


agirorn:
When I used `cargo install cargo-leptos` I got the error below that gets fixed by using --locked

```
error[E0599]: no variant or associated item named `deserialize` found for enum `serde_core::__private::content::Content` in the current scope
   --> /Users/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/swc_config-3.1.1/src/types/bool_or_data.rs:130:36
    |
130 |         let content = de::Content::deserialize(deserializer)?;
    |                                    ^^^^^^^^^^^ variant or associated item not found in `serde_core::__private::content::Content<'_>`
```

gbj:
Thanks. This is already in the `cargo-leptos` README and the book, must've just been missed here.

PR #4299, opened 2025-09-17
chore: add `referrerpolicy` attribute to `a` element

adoyle0:
When making a component that basically wraps an `<a>` I noticed the `referrerpolicy` attribute was missing.

It has [full browser support](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#browser_compatibility) and has for a while, it seems like it was simply overlooked?


gbj:
Yeah looks good to me. CI ran into a wasm-bindgen update so I've updated cargo-leptos and will try to re-run the ones that failed here -- but the PR itself is fine.

adoyle0:
I saw that. It's fine, take your time. It's not super important, I'm just supporting it in my component for completeness. Everyone has already gone this long without it and I just happened to stumble upon it lol

PR #4303, opened 2025-09-18
feat: Add support for #[prop(default)] syntax without value

nikessel:
# Add support for `#[prop(default)]` syntax without value

## Summary

This PR adds support for using `#[prop(default)]` without an explicit value, making it more ergonomic to use default values for component props. When no value is provided, the type's `Default::default()` implementation is used.

## Motivation

Previously, to use a default value for a component prop, users had to write:
- `#[prop(optional)]` - but this makes the prop an `Option<T>`
- `#[prop(default = Default::default())]` - verbose and repetitive
- `#[prop(default = false)]` for bools, `#[prop(default = 0)]` for numbers, etc.

With this change, users can simply write `#[prop(default)]` to use the type's default value:

```rust
#[component]
fn MyComponent(
    #[prop(default)] enabled: bool,        // defaults to false
    #[prop(default)] count: i32,          // defaults to 0
    #[prop(default)] name: String,        // defaults to empty string
    #[prop(default = 42)] custom: i32,    // explicit default still supported
) -> impl IntoView {
    // ...
}
```

## Implementation Details

The implementation works around a limitation of the `attribute-derive` crate, which expects `attribute = value` syntax. The solution:

1. **Preprocessing step**: Before parsing attributes, we transform `#[prop(default)]` into `#[prop(default = ())]` where the empty tuple serves as a marker.

2. **Recognition**: During typed builder generation, we recognize the empty tuple marker and treat it as "use `Default::default()`" rather than a literal value.

3. **Code generation**: The typed builder correctly generates code that uses `Default::default()` for props marked with standalone `#[prop(default)]`.

## Changes

- Added `preprocess_prop_attributes()` function to transform standalone `default` attributes
- Modified `TypedBuilderOpts::from_opts()` to recognize the empty tuple marker
- Added comprehensive documentation explaining the approach
- Added test coverage for the new functionality

## Testing

- ✅ Added new test `component_with_default` that verifies the functionality
- ✅ All existing tests pass
- ✅ Backward compatibility maintained - `#[prop(default = expr)]` continues to work

## Breaking Changes

None. This is a purely additive change that maintains full backward compatibility.

## Example Usage

```rust
#[component]
fn TodoItem(
    #[prop(default)] completed: bool,
    #[prop(default)] priority: i32,
    #[prop(into)] text: String,
) -> impl IntoView {
    view! {
        <li class:completed=completed>
            <span class="priority">{priority}</span>
            <span class="text">{text}</span>
        </li>
    }
}

// Usage:
view! {
    <TodoItem text="Learn Leptos"/>  // completed=false, priority=0 by default
}
```

## Future Improvements

This implementation uses a string replacement approach for simplicity. A future enhancement could use proper token parsing for more robust handling of edge cases, though the current implementation works well for all common use cases.

## Related Issues

This addresses a common ergonomics request from the community to make default props less verbose.

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -763,8 +763,8 @@ impl Prop {
         // that we'll recognize later in TypedBuilderOpts::from_opts.
         let modified_attrs = preprocess_prop_attributes(typed.attrs.clone());
 
-        let prop_opts =
-            PropOpt::from_attributes(&modified_attrs).unwrap_or_else(|e| {
+        let prop_opts = PropOpt::from_attributes(&modified_attrs)
+            .unwrap_or_else(|e| {
                 // TODO: replace with `.unwrap_or_abort()` once https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/17 is fixed
                 abort!(e.span(), e.to_string());
             });
@@ -1055,7 +1055,8 @@ impl TypedBuilderOpts {
         let (has_default, default_value) = match &opts.default {
             Some(expr) => {
                 // Check if it's our empty tuple marker for standalone #[prop(default)]
-                if matches!(expr, syn::Expr::Tuple(tuple) if tuple.elems.is_empty()) {
+                if matches!(expr, syn::Expr::Tuple(tuple) if tuple.elems.is_empty())
+                {
                     // This means the user wrote #[prop(default)] without a value
                     // We'll use Default::default() for this prop
                     (true, None)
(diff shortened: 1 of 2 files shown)


nikessel:
Ai generated code and pr. Hope it's not too messy or hacky of a solution, I'm not really that familiar with proc macros, so I didn't dig that deep

gbj:
> `#[prop(optional)]` - but this makes the prop an `Option<T>`

This is not actually true. `#[prop(optional)]` works for any type that implements `Default`. Your example works fine with `optional`. `default` is used if you want to pass a default value *other* than the one provided by the `Default` trait, or if there is no `Default` impl for the type.

```rust
#[component]
fn MyComponent(
    #[prop(optional)] enabled: bool,   // defaults to false
    #[prop(optional)] count: i32,      // defaults to 0
    #[prop(optional)] name: String,    // defaults to empty string
    #[prop(default = 42)] custom: i32, // explicit default still supported
) -> impl IntoView {
    // ...
}
```

nikessel:
Oh okay great I didn't know that!
Thanks, that's perfect

PR #4307, opened 2025-09-19
Add minimal support for `subsecond` and an example

gbj:


leptos_dom/Cargo.toml
```diff
@@ -18,6 +18,7 @@ tracing = { optional = true, workspace = true, default-features = true }
 wasm-bindgen = { workspace = true, default-features = true }
 serde_json = { optional = true, workspace = true, default-features = true }
 serde = { optional = true, workspace = true, default-features = true }
+subsecond.workspace = true
```
sabify, line 21: Please make it optional here and use it under a `subsecond` feature.

gbj: Thanks -- this was the initial experimental implementation just playing with `set_interval` and can be removed in any case.

PR #4308, opened 2025-09-19
chore: add missing attributes

adoyle0:
I stumbled upon another missing attribute. This time `dirname` for `<input>`. Interestingly it exists already for `<textarea>`.

It has [full support](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#browser_compatibility).


gbj:
Thanks!

PR #4309, opened 2025-09-19
Deprecate redundant `render_app_async_stream_with_context`

alexisfontaine:
I was inspecting the internals of the `leptos_axum` crate and noticed duplicate code in the logic for handling the async stream.
Furthermore, `render_app_async_stream_with_context` and `render_app_async_with_context` appear to have identical behavior. I looked for potential differences, but found none.

I propose to deprecate `render_app_async_with_context` in favor of `render_app_async_stream_with_context`, which is also exposed in the Actix integration.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -684,7 +684,7 @@ where
         additional_context.clone(),
         app_fn.clone(),
     );
-    let asyn = render_app_async_stream_with_context(
+    let asyn = render_app_async_with_context(
         additional_context.clone(),
         app_fn.clone(),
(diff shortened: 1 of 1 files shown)


gbj:
Thanks for pointing this out. I will tag this as 0.9 and simply remove the duplicate — marking something deprecated is not technically a breaking change but can be effectively break CI if people are linting, so it's easiest to deprecate in a minor release or simply remove, especially for something that's not likely to be widely used like this.

gbj:
Thanks! I've now created a `leptos_0.9` break so have merged this in there.

PR #4310, opened 2025-09-20
Speedup the `Free Disk Space` CI step

alexisfontaine:
The `Free Disk Space` job in CI is a significant bottleneck (~5min on hundreds of jobs per workflow). This change replaces `rm -rf` command with [`rmz`](https://crates.io/crates/rmz), which is apparently faster at deleting large directories.

If preferred, I can try the more conventional `find -delete` command instead of adding a dependency.

### DIFF
--- a/.github/workflows/run-cargo-make-task.yml
+++ b/.github/workflows/run-cargo-make-task.yml
@@ -21,22 +21,36 @@ jobs:
         toolchain: [stable, nightly-2025-07-16]
         erased_mode: [true, false]
     steps:
+      # Setup environment
+      - name: Install Glib
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libglib2.0-dev
+      - name: Setup Rust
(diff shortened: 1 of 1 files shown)


mohe2015:
How does the performance compare with `find`?

alexisfontaine:
You're right to question that claim – I hadn't actually benchmarked `rmz` or `find`. My apologies for stating it without verification.

I've been looking into how the Rust project currently handles this, and it seems they've built [their own script](https://github.com/rust-lang/rust/blob/9f2ef0f14d6028c5108643cafa6e2c617834594b/src/ci/scripts/free-disk-space-linux.sh) based on the [`free-disk-space`](https://github.com/jlumbroso/free-disk-space) action.

Interestingly, there's a PR on that action (jlumbroso/free-disk-space#26) with some numbers. It seems benchmarking the different commands within the Leptos CI environment is the best way to determine if it's actually an improvement.

PR #4325, opened 2025-09-24
Allow accessing a parent owner from a child

zakstucke:
This has come about from me trying to fix a really confusing bug in leptos-fetch. 

Scenario:
- I want from userland to clone and hold onto an `Owner`, to prevent it getting dropped. So that I can call `use_context()` as I please.
- A parent of this owner is dropped, and the context was actually stored on that owner.
- Panic

This PR allows me in userland (LF) to produce the chain of owners,`Vec<Owner>` by calling `.parent()` over and over again, to preserve the entire chain, not just the current owner.

### DIFF
--- a/reactive_graph/src/owner.rs
+++ b/reactive_graph/src/owner.rs
@@ -209,6 +209,25 @@ impl Owner {
         this
     }
 
+    /// Returns the parent of this `Owner`, if any.
+    ///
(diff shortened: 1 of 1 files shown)


gbj:
Definitely fine with me to expose this, thanks as always for the PR.

PR #4344, opened 2025-09-27
feat: `effect::immediate::batch`

QuartzLibrary:
This PR introduces a `batch` function that was present in an older version of `leptos`. It was removed because the new effects are deferred, so they are ~always batched.

Why is `batch` useful?
When using `ImmediateEffect`s, you might want to modify multiple signals atomically to ensure all subscribers always see a consistent view.
This can come up because of application invariants that are hard or inconvenient to encode in the type system, to avoid excessive updates, or under certain design patterns like applying multiple edit operations to a signal tree.

In general, normal effects should still be preferred.


gbj:
Awesome, thanks for reintroducing this and for the test.

PR #4357, opened 2025-10-01
Clean up window router events on unmount

Innominus:
Following on discussion from the Discord, this bug was found while using sub second hot patching as it re-runs all effects on a hot patch. Because of this, all window events from the router were being re-added without removing the previous events. They were not only polluting the window with unnecessary events, but these events would step on each other and it would break routing via a tags.

### DIFF
--- a/router/src/location/history.rs
+++ b/router/src/location/history.rs
@@ -3,7 +3,7 @@ use crate::{hooks::use_navigate, params::ParamsMap};
 use core::fmt;
 use futures::channel::oneshot;
 use js_sys::{try_iter, Array, JsString};
-use leptos::prelude::*;
+use leptos::{ev, prelude::*};
 use or_poisoned::OrPoisoned;
 use reactive_graph::{
     signal::ArcRwSignal,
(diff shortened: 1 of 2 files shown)


router/src/location/history.rs
```diff
@@ -206,14 +195,14 @@ impl LocationProvider for BrowserUrl {
...
-                "popstate",
-                closure.as_ref().unchecked_ref(),
-            )
-            .expect("couldn't add `popstate` listener to `window`");
+
+        let popstate_handle =
+            window_event_listener(ev::popstate, move |_| popstate_cb());
+
+        on_cleanup(|| {
+            click_handle.remove();
```
Innominus, line 203: I was originally going to just do window.remove_event_listener() to follow the above convention but the window object and send wrapper stuff was being annoying and this looked cleaner/reuse.

router/src/location/history.rs
```diff
@@ -159,27 +157,18 @@ impl LocationProvider for BrowserUrl {
 
         let handle_anchor_click =
             handle_anchor_click(base, Self::parse_with_base, navigate);
-        let closure = Closure::wrap(Box::new(move |ev: Event| {
+
+        let click_handle = window_event_listener(ev::click, move |ev| {
```
Innominus, line 161: `window_event_listener` doesn't return a result, so we lose the `expect(...)` context if this were to fail, but I don't really see that happening considering this is the `BrowserUrl` implementation, but if that's an issue I'm sure I can take another approach.

PR #4359, opened 2025-10-03
Allow more types to be converted to throw_error::Error

abusch:
## Context

At the moment it is quite difficult to get crates like `anyhow` to play well with leptos, in particular because it is _very_ difficult to convert an `anyhow::Error` type to a `leptos::Error` type. This is because the only way to construct a `leptos::Error` currently is via its `From` implementation, which exists for any type that implements the standard `Error` trait, but `anyhow::Error` does not implement `StdError` directly. It can however be converted to a boxed trait object via its [`.into_boxed_dyn_error()`][4] method, and you would think that `Box<dyn Error>` implements `Error`, but [that is sadly not the case][1].

## Solution

Change the blanket implementation of `From` for `throw_error::Error` from "any type that implements the Error trait" to "any type that can be converted to a boxed Error trait object".

This works because:
- A `Box` [can be converted][2] to an `Arc` for any type (including unsized types like trait objects),
- Any type that implements the standard `Error` trait [can be converted][3] to a `Box<dyn Error>`, so the new `From` blanket implementation covers a strict superset of what was previously allowed.

This change now allows types like `anyhow::Error`, but also `String`, to be easily converted to a leptos `Error`, therefore making them play well with things like `<ErrorBoundary>`.

[1]: https://stackoverflow.com/questions/65151237/why-doesnt-boxdyn-error-implement-error
[2]: https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#impl-From%3CBox%3CT,+A%3E%3E-for-Arc%3CT,+A%3E
[3]: https://doc.rust-lang.org/stable/std/error/trait.Error.html#impl-From%3CE%3E-for-Box%3Cdyn+Error%3E
[4]: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.into_boxed_dyn_error

### DIFF
--- a/any_error/src/lib.rs
+++ b/any_error/src/lib.rs
@@ -161,9 +161,8 @@ where
 
 #[cfg(test)]
 mod tests {
-    use std::error::Error as StdError;
-
     use super::*;
+    use std::error::Error as StdError;
 
     #[derive(Debug)]
     struct MyError;


gbj:
That's a really neat way to solve this problem! Thanks for the contribution.

PR #4362, opened 2025-10-04
Support `ToggleEvent` for `on:toggle`.

### LINKED ISSUE #4361: Support `ToggleEvent` for `on:toggle`
Heya, currently `on:toggle` only provides an [`Event`](https://github.com/leptos-rs/leptos/blob/v0.8.10/tachys/src/html/event.rs#L728), whereas `web_sys` supported `ToggleEvent` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/toggle_event)) since [`0.2.93`](https://github.com/wasm-bindgen/wasm-bindgen/blob/main/CHANGELOG.md#0293) (search for "Popover API", https://github.com/wasm-bindgen/wasm-bindgen/pull/3977 -- `gen_ToggleEvent.rs` got added here).

I'd like to have a `ToggleEvent` so that the [`new_state`](https://docs.rs/web-sys/latest/web_sys/struct.ToggleEvent.html#method.new_state) method can be called, so I can set a signal on a `bool`.

I'm using this on a `<details>` element, and based on the `open` state, I change the `<summary>` text.

Using an `<input>` element's `checked` state, and styling sibling elements as expanded or collapsed. But I'd rather use a built-in expandable element.

**Additional context**

I've tested my project with this code and it works: https://github.com/leptos-rs/leptos/pull/4362

It is a breaking change if there is code that explicitly uses `|event: Event|` as the parameter, but I think it's source compatible if they used an inferred type parameter and don't pass it somewhere else that is strict on the type.

azriel91:
* Closes #4361.

Allows you to do this:

```rust
#[component]
pub fn MyDetails() -> impl IntoView {
    let (expanded, expanded_set) = leptos::prelude::signal(true);
    view! {
        <details
            open={move || expanded.get()}
            on:toggle=move |event| *expanded_set.write() = event.new_state() == "open"
        >
            <summary><OneLineSummary expanded /></summary>
            <p>contents</p>
        </details>
    }
}

#[component]
#[component]
pub fn OneLineSummary(expanded: ReadSignal<bool>) -> impl IntoView {
    view! {
        <span>{move || if expanded.get() { "Long text" } else { "short" }}</span>
    }
}
```

### DIFF
--- a/tachys/Cargo.toml
+++ b/tachys/Cargo.toml
@@ -72,6 +72,7 @@ web-sys = { features = [
   "SecurityPolicyViolationEvent",
   "StorageEvent",
   "SubmitEvent",
+  "ToggleEvent",
   "TouchEvent",
   "TransitionEvent",
   "UiEvent",
--- a/tachys/src/html/event.rs
+++ b/tachys/src/html/event.rs
@@ -614,7 +614,7 @@ generate_event_types! {
   animation start: AnimationEvent,
   aux click: MouseEvent,
   before input: InputEvent,
-  before toggle: Event, // web_sys does not include `ToggleEvent`
+  before toggle: ToggleEvent,
   #[does_not_bubble]
   blur: FocusEvent,
   #[does_not_bubble]
@@ -725,7 +725,7 @@ generate_event_types! {
   #[does_not_bubble]
   time update: Event,
   #[does_not_bubble]
-  toggle: Event,
+  toggle: ToggleEvent,
   touch cancel: TouchEvent,
   touch end: TouchEvent,
   touch move: TouchEvent,


gbj:
Thanks! As you note, unless I'm mistaken this is a breaking change. Which is not a judgment of how large or small a change it is, simply the fact that including this in a patch release would cause previously-working code to stop compiling, which is bad. I'm fine making the change in 0.9.

In the meantime, you can use the `JsCast` trait from `wasm-bindgen` to convert to `ToggleEvent` pretty easily:
```rust
*expanded_set.write() = event.unchecked_ref::<ToggleEvent>().new_state() == "open"
```

azriel91:
ah! Thanks for the tip, I really appreciate both short and long term solutions 🙏

gbj:
I've now created a `leptos_0.9` branch so am merging this in there. Thanks for your patience!

PR #4369, opened 2025-10-07
Replace vendored wasm-split with out-of-repository version

WorldSEnder:
PR is for testing for now. The git dependency must be replaced by a proper release of the `wasm-split` crate before merging.


johansmitsnl:
My feedback on the test.

During the wasm compilation I see warnings on the trait:

```
warning: function `__ConsolePageSectionPreviewRoute_View` should have a snake case name
   --> src/app.rs:586:20
    |
586 | impl LazyRoute for ConsolePageSectionPreviewRoute {
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: `__console_page_section_preview_route_view`
 ```

The error I get in my project is:

```
Uncaught RuntimeError: indirect call to null
[app.wasm:8200590:1](http://127.0.0.1:8000/pkg/app.wasm)
    _ZN12wasm_bindgen4__rt8wbg_cast17h7e34a28c9f8f75fcE http://127.0.0.1:8000/pkg/app.wasm:8200590
    _ZN12wasm_bindgen7closure16Closure$LT$T$GT$4wrap17hb4d7d15805f372faE http://127.0.0.1:8000/pkg/app.wasm:6450621
    _ZN10leptos_use18use_event_listener31use_event_listener_with_options17h33e8f651187e3ed9E http://127.0.0.1:8000/pkg/app.wasm:877533
    _ZN10leptos_use18use_event_listener31use_event_listener_with_options17h33e8f651187e3ed9E http://127.0.0.1:8000/pkg/DDXqf-gPnDII5b5U8xfm7Q.wasm:4710614
```

Edit, in Chrome I get a better error:

```
Uncaught RuntimeError: null function or function signature mismatch
    at _ZN12wasm_bindgen4__rt8wbg_cast17h7e34a28c9f8f75fcE (app.wasm:0x7d218e)
    at _ZN12wasm_bindgen7closure16Closure$LT$T$GT$4wrap17hb4d7d15805f372faE (app.wasm:0x626dbd)
    at _ZN10leptos_use18use_event_listener31use_event_listener_with_options17h33e8f651187e3ed9E (app.wasm:0xd63dd)
    at _ZN10leptos_use18use_event_listener31use_event_listener_with_options17h33e8f651187e3ed9E (DDXqf-gPnDII5b5U8xfm7Q.wasm:0x47e0d6)
 ```

I'll share a zip of the output in Discord.

WorldSEnder:
@johansmitsnl if you have time, I'd value your test on the newest updated version of both this branch paired with the `cargo-leptos` PR (sorry if you got the earlier ping).
@gbj can we somehow disable or turn down CI while this is in draft mode? While not my wasted money and energy, I feel somewhat responsible :p

johansmitsnl:
@WorldSEnder I did a quick test and it is working with a debug build

WorldSEnder:
The test failures are caused by the fact that the released version *does* require an updated `cargo-leptos` and support from the cli due to making the module path that is used to import the link module (`__wasm_split.js`) a responsibility of the CLI. This would be a breaking change regardless. For future changes like this, the library now emits a (for the moment very small) custom section that the CLI can use to detect a specific `wasm_split_helpers`/`wasm_split_macros` version.

The option to emit a specific module path instead of a placeholder to be replaced by the CLI still exists, but is hidden and deprecated. Imo it makes most sense to just do this "breaking" change immediately, but let me know what you think, it's mostly a detail in how the `#[lazy]` macro uses `#[wasm_split]`.

EDIT: To be specific, `cargo-leptos` with the relevant PR compiling/splitting a binary depending on `leptos` with and without _this_ PR should work both ways. `cargo-leptos` without the relevant PR splitting `wasm` produces with this PR will fail to rewrite some module import paths.

gbj:
Thanks. I love when things like that cause CI to fail, as it reassures me that the CI will actually catch them breaking :smile: 

I'm going to try to trigger it again now that I've published a new `cargo-leptos`. Hopefully it will use the latest one.

PR #4376, opened 2025-10-08
Adding `bitcode` encoding/decoding to server fn IO

arpadav:
`bitcode` has been on the top of [rust ser/de benchmarks](https://github.com/djkoloski/rust_serialization_benchmark) and would be great to use within `leptos`'s `server_fn`.

~~On the implementation-side, one question: `Encodes` trait requires an error type, but `bitcode` can encode anything without error. I tried to do:~~

```rust
impl<T> Encodes<T> for BitcodeEncoding
where
    T: bitcode::Encode,
{
    type Error = ();
```

~~But am getting an error, since unit type does not impl Display. I just put a comment and made it `type Error = String` as a bandaid fix. Ideally it could be `()`, or perhaps if there was some blanket struct. Unsure what is preferred?~~

(I was told to modify the `Error` to be `std::convert::Infallible` here https://github.com/leptos-rs/leptos/pull/4376/commits/4f81773f6df511ffbc08069ad915495aed453db0 and this solves the issue above)

Tested on my end, but I'd like others to confirm. I did not see any particular places to put tests for server fn IO types, otherwise I can help populate those tests


arpadav:
@gbj Added!

And I noticed that `skip_feature_sets` has ser/de combinations between `serde`, `serde-lite`, `rkyv`, and `miniserde`. I assume these are for `Resource`'s? So there is no need to add `bitcode` here?

PR #4379, opened 2025-10-10
add `scrollend` event to view macro

adoyle0:
I was trying to use `on:scrollend` on an element in a view macro and got an error, upon investigation it seems like it just never got hooked up. Is there a reason for this? It seems like just an oversight. Anyway this appears to be all that is needed to hook it up? It works as expected now in my project.


gbj:
Thanks!

PR #4388, opened 2025-10-13
Remove `_with_handle` suffix from functions

### LINKED ISSUE #4387: `*_with_handle` functions are very confusing
I think that the next functions should be renamed in a breaking change, replacing their counterparts without handles:

- `request_animation_frame_with_handle` -> `request_animation_frame`
- `request_idle_callback_with_handle` -> `request_idle_callback`
- `set_interval_with_handle` -> `set_interval`
- `set_timeout_with_handle` -> `set_timeout`

Due to several reasons:

1. `with_handle` is very confusing. In Leptos the suffix `with_handle` means that the function returns a handle, but in web-sys it means that it accepts a handle. See [`clear_timeout_with_handle`](https://docs.rs/web-sys/latest/web_sys/struct.Window.html#method.clear_timeout_with_handle) and [`clear_interval_with_handle`](https://docs.rs/web-sys/latest/web_sys/struct.Window.html#method.clear_interval_with_handle). Note that `with_handle` in web-sys is used exclusively to differentiate the functions from their counterparts that don't accept a handle. As proof, [`cancel_animation_frame`](https://docs.rs/web-sys/latest/web_sys/struct.Window.html#method.cancel_animation_frame) accepts a handle but is not named `_with_handle` because it is implicit that cancelling needs something to be cancelled.
2. All the implementations of the above functions without handles are just syntactic sugar for `_ = *_with_handle`, so it would be more ergonomic just `_ = *` when the handle is discarded. IMO, it is preferable to call `_ = set_timeout` than needing to call another function when it should return a handle. That's more aligned with JavaScript behavior, as the handle is always returned.
3. Probably it is faster, at least in compile times, because it would remove an intermediate function when using functions without handles.
4. Traditionally in the Rust ecosystem, it is much more common to use `_with_*` for builders or for functions that accept an additional argument that differs from the original. I’ve never seen an API use `_with_*` to refer to a function that returns something different.

mondeja:
This is a breaking change. Resolves #4387

### DIFF
--- a/examples/axum_js_ssr/src/app.rs
+++ b/examples/axum_js_ssr/src/app.rs
@@ -674,7 +674,7 @@ fn CodeDemoWasm(mode: WasmDemo) -> impl IntoView {
                                 leptos::logging::log!("wasm csr_listener listener added");
 
                                 // Dispatch the event when this view is finally mounted onto the DOM.
-                                request_animation_frame(move || {
+                                _ = request_animation_frame(move || {
                                     let event = web_sys::Event::new("hljs_hook")
                                         .expect("error creating hljs_hook event");
                                     document.dispatch_event(&event)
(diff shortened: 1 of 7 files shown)


gbj:
Thanks! I have now created a `leptos_0.9` branch and will merge this in there.

PR #4393, opened 2025-10-15
Allow "infinite" timeouts in `set_timeout` helper

### LINKED ISSUE #4226: Change `set_timeout_with_handle()` and `set_interval_with_handle()` to not panic
We should change `set_timeout_with_handle()` and `set_interval_with_handle()` to not take a `Duration` because technically for wasm targets we can't take anything larger than `i32::MAX` and `Duration` takes values up to a `u128`. This will prevent panics if `Duration::as_millis() > i32::MAX`.

mondeja:
This is a crazy experimental solution for a crazy bug. It fixes #4226, I think in a backwards compatible way. If someone wants to schedule a timeout longer than 24.8551348032 days, now it's possible.

This is "better" than the JS implementation. When you pass a number greater than `2147483647` to the JS `setTimeout`, the browsers truncate it to 2147483647, so the solution not merged at #4212 is the JS behaviour.

This implements chunked timeouts. If the value passed to the timeout exceeds `i32::MAX`, schedule a timeout of `i32::MAX` and so on recursively until the remaining time is lower than `i32::MAX`, then execute the original callback.

Feel free to close this PR, just crafted for fun and it works. The only question here is if the memory, size and one conditional more  at runtime are worth.

If we want the same JS behaviour of silently truncate it to 24.8 days (probably not enough LOL), merge #4226. If we want to fix browsers implementations vote for this PR.


gbj:
The downside of this is that it adds a large amount of code in the default case in exchange for a small benefit in the really-wacky case. I am not sure the cost-benefit there works out.

mondeja:
I would merge #4212. It is fully consistent with the JavaScript implementation in browsers.

PR #4394, opened 2025-10-16
fix: Fixed to Add response headers for leptos_axum static files #4377

tqq1994516:
#4377


gbj:
Did you test this? When I try it out, your modified example panics on all static file requests.

tqq1994516:
> Did you test this? When I try it out, your modified example panics on all static file requests.

I verified it at the time. I don't know if it was a cache issue or something else, but now it does panic. I will continue to fix it.

gbj:
It still wasn't setting the headers as requested, because there was actual no ownership system/context set up -- easy enough fix so I thought I'd just modify it myself rather than bugging you!

Thanks for the PR.

tqq1994516:
> It still wasn't setting the headers as requested, because there was actual no ownership system/context set up -- easy enough fix so I thought I'd just modify it myself rather than bugging you!
> 
> Thanks for the PR.

That shouldn't be the case. When I ran that example according to my PR, I saw the headers I wanted.
image
image
image
image
Could you tell me the role of the newly added `Owner`? I think this can help me understand leptos better.

tqq1994516:
Oh, I found that the addition of this header in my PR is unstable; every time I refresh, it may not be added as expected.

PR #4408, opened 2025-10-25
Remove dashmap usage

xacrimon:
I'm the dashmap author and recently while working on a leptos project, I discovered dashmap in the deptree and got curious as to why it was there. Upon inspecting the usage of it in leptos, I've arrived the the conclusion that it's not appropriate in any of the places it's used.

The reason for this in all cases is that generally, you need to have *very* significant contention on the map across multiple cores for it to be worth it at all, after observing it's usage in the wild across a few years I like to say that most people that use it probably shouldn't. Dashmap is a very specialized tool and due to internal overheads necessary for a concurrent datastructure, it's generally doesn't provide any performance benefit (usually the opposite) unless you're hammering it with 1M+ lookups per second across cores, which is exceedingly unlikely in the places it's used.

Therefore, I've removed every usage of it across leptos that I found so far, and replaced them with the far simpler `Mutex<HashMap>` or `RwLock<HashMap>` depending on the spot.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1321,8 +1321,7 @@ where
                         .await;
                     (owner.with(use_context::<ResponseOptions>), html)
                 } else {
-                    let headers =
-                        STATIC_HEADERS.read().get(orig_path).cloned();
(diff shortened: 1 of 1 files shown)


gbj:
I did this separately in #4521.

PR #4417, opened 2025-10-29
add homepage to leptos cargo metadata

tswast:
This updates the links at Are We Web Yet..

See also https://doc.rust-lang.org/cargo/commands/cargo-metadata.html

Closes https://github.com/rust-lang/arewewebyet/issues/451


gbj:
Thanks!

PR #4422, opened 2025-10-31
ref: `drain_filter_polyfill` refactors

### LINKED ISSUE #4419: prefer std lib's `Vec::extract_if` over `drain_filter_polyfill` dep
Just something I noticed while scanning through the repo. Have not done any work yet to ensure the implementations are equivalent. Would be an easy deps to drop if so.

Currently, it brings in dependencies.

Use std lib APIs wherever applicable, unless the alternative is clearly better.

nil

**Additional context**
* [`Vec::extract_if`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extract_if)
* [Where this dependency is being used](https://github.com/leptos-rs/leptos/blob/2af6c6353c60822ac239d8243a284f6da35333ce/tachys/src/view/keyed.rs#L11)

chrisp60:
it is easy for me to get flipped around with binary logic, please confirm these changes are coherent.

I tried to keep these "filter" points very friendly to future contributors, but intentionally duplicated logic in certain places to *retain* a clear locality of behavior.

**Additionally** note the diffs contain source comments that stated `#[builder(_)]` was being discarded, but it was actually tossing `#[prop(_)]`s. I believe this was just a simple copy-paste mistake.

Resolves #4419

### DIFF
--- a/leptos_macro/src/slot.rs
+++ b/leptos_macro/src/slot.rs
@@ -5,8 +5,7 @@ use attribute_derive::FromAttr;
 use proc_macro2::{Ident, TokenStream};
 use quote::{quote, ToTokens, TokenStreamExt};
 use syn::{
-    parse::Parse, parse_quote, Field, ItemStruct, LitStr, Meta, Type,
-    Visibility,
+    parse::Parse, parse_quote, Field, ItemStruct, LitStr, Type, Visibility,
 };
 
 pub struct Model {


chrisp60:
#4425 should fix the CI failures

LeoniePhiline:
LGTM! I've reviewed all changes and can confirm they are coherent.

The binary logic was quite a lot easier to review thanks to your consistent semantic naming scheme! :)

LeoniePhiline:
(But this is ages old and needs a rebase. 😬 )

PR #4425, opened 2025-11-01
fix(ci): `clippy::derivable-impls`

chrisp60:
CI never fails.

### DIFF
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -277,18 +277,14 @@ impl TryFrom<String> for Env {
     }
 }
 
-/// An enum that can be used to define the websocket protocol Leptos uses for hotreloading
-/// Defaults to `ws`.
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


leptos_config/src/lib.rs
```diff
@@ -277,20 +277,16 @@ impl TryFrom<String> for Env {
     }
 }
 
-/// An enum that can be used to define the websocket protocol Leptos uses for hotreloading
-/// Defaults to `ws`.
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
+/// An enum that can be used to define the websocket protocol Leptos uses for hotreloading.
+#[derive(
+    Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq, Default,
```
LeoniePhiline, line 282: At this point, I would opt for importing:

```
```

The multi-line `#[derive]` formatting is rare and easily overlooked when grepping or performing regex-based replacements.

LeoniePhiline (commented):
I'd propose importing the `serde` traits. Looks good otherwise.

PR #4427, opened 2025-11-02
Force cleanup even if there are other references to the root owner

zakstucke:
This PR comes from me debugging a ssr memory leak. It seems that in some cases you can have a circular link back to the root owner, then the root won't cleanup on drop like intended at the end of the request, this PR forces cleanup server side during the `.unset()` at the end of the request, preventing some memory leak opportunities.


gbj:
Ah interesting. I guess in this case, if there is a circular reference it will still leak... *something*? Just the size of the `Owner` itself, rather than the size of all the owned data as well, I suppose.

zakstucke:
In my case it was worse because I was relying on the root owner being cleaned up to gc some global data (`on_cleanup`), but the root never clears so my global kept expanding!

PR #4452, opened 2025-11-18
Fix for InertElement hydration error

paul-hansen:
Seems like the hot-reload comments were tripping up hydration finding the correct element when using InertElement.
This tries to skip those comments similar to [other places](https://github.com/leptos-rs/leptos/blob/4f3a26ce88eb1b429d382a871c47e086be96d559/tachys/src/renderer/dom.rs#L163-L180) in the codebase.

Should fix https://github.com/carloskiki/leptos-icons/issues/64

### DIFF
--- a/tachys/src/svg/mod.rs
+++ b/tachys/src/svg/mod.rs
@@ -302,6 +302,22 @@ impl RenderHtml for InertElement {
         } else if curr_position != Position::Current {
             cursor.sibling();
         }
+
+        // if it's a comment node that starts with hot-reload, it's a marker that should be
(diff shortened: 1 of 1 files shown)


gbj:
Having looked a little at the issue, I don't think this fixes it — i.e., I can reproduce the same issue without any hot-reloading — but I will add a comment over on the issue.

paul-hansen:
Weird, I'm confused why this was fixing my test I had, but tbh your explanation does make a lot more sense. Thanks for looking and sorry for the noise.

PR #4466, opened 2025-11-26
Show unification

maccesch:
On the heels of `IntoReactiveValue` we can now change the `when` prop of `Show` to an `Into<Signal<bool>>>`.

Existing code should continue working without change but it is technically a breaking change.

But this unifies the API between `<Show>`, `<AnimatedShow>` and `<ShowLet>`.

Please note that I also tried to use the `Into<Signal<Option<T>>>>` approach for `<ShowLet>` but this creates an additional indirection for type inference which the compiler often can't do automatically. So I kept the custom approach from before which in effect yields the same result.

In addtion to that I wanted to update the examples to simplify the `Show` usage where applicable. That lead me to port the project `sso_auth_axum` to the latest Leptos version (was pre 0.6). I did not port other projects but it seems some of them would require it.


maccesch:
Am I supposed to increase the crate version directly in the PR?

gbj:
Thanks! The breaking change in this case is actually breaking when the `nightly` feature is activated, because the changes to add `IntoReactiveValue` don't (and can't) implement `From<Fn() -> T>` for `Signal<T>`. i.e., the code below currently compiles on main, but does not compile with this PR:
```rust
#[component]
pub fn SimpleCounter(
    /// The starting value for the counter
    initial_value: i32,
    /// The change that should be applied each time the button is clicked.
    step: i32,
) -> impl IntoView {
    let (value, set_value) = signal(initial_value);

    view! {
        <Show when=move || value.get() % 2 == 0>
            <p>"Even!"</p>
        </Show>
        <div>
            <button on:click=move |_| set_value.set(0)>"Clear"</button>
            <button on:click=move |_| *set_value.write() -= step>"-1"</button>
            <span>"Value: " {value} "!"</span>
            <button on:click=move |_| set_value.update(|value| *value += step)>"+1"</button>
        </div>
    }
}
```

I am planning to address this some time in early-mid January with a 0.9 release that will tweak the balance of these things slightly along the lines zak and I were discussing in #4273. My suggestion would be that we mark this PR with the 0.9 milestone and I will merge it into the `leptos_0.9` branch once I start working on that. (November-December is just a busy time for me in real life so I am planning to put the work in to actually make things make sense in January.)

maccesch:
Perfect. Thanks!

PR #4467, opened 2025-11-26
`ShowLet` now accepts static `Option` values. 

maccesch:
Also added `ShowLet` to examples where appropriate.


gbj:
Great, thanks for adding these!

PR #4473, opened 2025-12-05
feat: add support for variable keyed collections in reactive_stores

TiemenSch:
This PR adds reworks some of the internals of the reactive_stores to better support arbitrary collections. 

To achieve that I added a `KeyedAccess` trait that replaces the `IndexMut` from the standard library for arbitrary keys and values. I deliberately used different method names than `read` and `get` and friends to not confuse editors when certain trait implementations are missing (and it suggesting the wrong trait to implement).

I added an implementation for common indexed types and added built-in SlotMap support under a feature flag. Sadly, I couldn't make a blanket implementation since `IndexMut` has retracted implementations for some common collection types (BTreeMap etc) which would then become unusable. 

I made the assumption that requiring `Copy` on keys is good practice. This refrains users from using `Clone` keys rather than the cheaper options.

There is one part in particular I think should be thoroughly reviewed, though. It seems to be where the reactive path is retraced to see if all is still well, but the previous implementation still used the resulting `Option<id>` as the final collection key, rather than the `self.key` which was there all along. I'll link it below in a sec. These functions were hard to read with little documentation.

Also, working with the `stores` example was a bit frustrating. There seems to be a disconnect between the "users" in store doing effectively nothing and the actual "todo" items. Therefore, I kept the `stores` example unchanged (to see whether the reworked implementation broke anything or required any changes). I added a Spartan copy of that with some usability changes, but perhaps there could be a full rewrite at some point, where todo's are stored per user (and selected by a dropdown or something).

### DIFF
--- a/examples/stores/src/lib.rs
+++ b/examples/stores/src/lib.rs
@@ -107,7 +107,7 @@ pub fn App() -> impl IntoView {
     view! {
         <p>"Hello, " {move || store.user().name().get()}</p>
         <UserForm user=store.user() />
-        <UserAchievements store=store.clone() />
+        <UserAchievements store />
         <hr />
         <form on:submit=move |ev| {
             ev.prevent_default();


TiemenSch:
Context: https://discordapp.com/channels/1031524867910148188/1444434908490502286

The `serde` support under another feature flag was a byproduct of an earlier attempt and can clearly be removed if deemed an incorrect addition.

TiemenSch:
Just ran all commands from CONTRIBUTING and now they are in the green.

gbj:
Thanks for the PR!

Unfortunately I've never added any automated tests for keyed store fields, which is unhelpful. And as you note some of these implementation pieces are not very well documented, or documented at all.

At the bottom of this post, I've included a test that works on git main and fails on this PR, and is specifically related to the area you called out for further review -- which should hopefully help avoid regressions and make it easier for me to review things in the future.

The failing portion of the test:
1. creates a list of 3 todos with `id` field keys 10, 11, and 12, and `AtKeyed` fields for each
2. swaps the first and last items
3. updates the item with key 10

This then panics with the error `index out of bounds: the len is 3 but the index is 10`. 

This makes it reasonably clear right away what's going wrong.

The new implementation conflates two different senses of the word "key":
1) the "key" that allows you to index into a data structure (for a `Vec<_>`, this is a `usize` index; for a `HashMap<SomeKeyType, String>` this is a `SomeKeyType`)
2) the "key" that is a unique identifier for each entry in a keyed list like the keyed store

The keyed store field implementation is built around meaning #2 of "key." If you have a keyed field that is a `Vec<_>`, the key can be *any* type, not only a `usize`: and referring to that key refers to whichever entry in the `Vec<_>` has that key, *not* the nth entry in the `Vec<_>`. 

The implementation in the PR swaps from the second to the first meaning. It allows a variety of key types, but the key type *must* be the type that's used to access the underlying collection. So for example, for a `Vec<_>` the key now must be a `usize`. More importantly than that, however, it now understands "accessing the field with this key" to mean "indexing into the `Vec<_>` with this key as the index," which is not correct. If we have keys that are larger than the length of the vector, this will panic; if we have keys (like 0, 1, 2) that fit within the vector, it will read from/update the entry that happens to be at the index that has the same value as the key, whether the item with that key is at that index or not.

This actually works fine for map types, where the key *is* the index, and where items can't necessarily move around. But it doesn't work for slice types, where the key is *not* the index, and where items can move to different indices while keeping the same key.

That's what `let index = /* ... */;` was doing: it finds the location *in the slice* of the item with the current key, and uses that to index into the slice, rather than using the *key* to index into the slice.

So then the problem that the new implementation has to deal with is specifically the fact that what is returned here is a `usize` of the location, within the set of new keys, of this particular key.

I actually think this can be fixed without too much trouble, by changing `.keyed()` and `.keyed_mut()` so that they both take an `index: usize` as well. For slice-like types the implementation would use this index; for map-like types it would use the key. This should allow us to support multiple types of data structures while maintaining the correct behavior.

And I would suggest allowing `Clone` key types as well: of course it's not a good idea to clone big objects around, but there are plenty of types like `Arc<str>` that would be reasonable to use as keys and are cheap to clone for this purpose.

```rust
#[cfg(test)]
mod tests {
    use crate::{self as reactive_stores, tests::tick, AtKeyed, Store};
    use reactive_graph::{
        effect::Effect,
        traits::Track,
        traits::{GetUntracked, ReadUntracked, Set, Write},
    };
    use std::sync::atomic::{AtomicUsize, Ordering};
    use std::sync::Arc;

    #[derive(Debug, Store, Default)]
    struct Todos {
        #[store(key: usize = |todo| todo.id)]
        todos: Vec<Todo>,
    }

    #[derive(Debug, Store, Default, Clone, PartialEq, Eq)]
    struct Todo {
        id: usize,
        label: String,
    }

    impl Todo {
        pub fn new(id: usize, label: impl ToString) -> Self {
            Self {
                id,
                label: label.to_string(),
            }
        }
    }

    fn data() -> Todos {
        Todos {
            todos: vec![
                Todo {
                    id: 10,
                    label: "A".to_string(),
                },
                Todo {
                    id: 11,
                    label: "B".to_string(),
                },
                Todo {
                    id: 12,
                    label: "C".to_string(),
                },
            ],
        }
    }
     
    #[tokio::test]
    async fn keyed_fields_can_be_moved() {
        _ = any_spawner::Executor::init_tokio();

        let store = Store::new(data());
        assert_eq!(store.read_untracked().todos.len(), 3);

        // create an effect to read from each keyed field
        let a_count = Arc::new(AtomicUsize::new(0));
        let b_count = Arc::new(AtomicUsize::new(0));
        let c_count = Arc::new(AtomicUsize::new(0));

        let a = AtKeyed::new(store.todos(), 10);
        let b = AtKeyed::new(store.todos(), 11);
        let c = AtKeyed::new(store.todos(), 12);

        Effect::new_sync({
            let a_count = Arc::clone(&a_count);
            move || {
                a.track();
                a_count.fetch_add(1, Ordering::Relaxed);
            }
        });
        Effect::new_sync({
            let b_count = Arc::clone(&b_count);
            move || {
                b.track();
                b_count.fetch_add(1, Ordering::Relaxed);
            }
        });
        Effect::new_sync({
            let c_count = Arc::clone(&c_count);
            move || {
                c.track();
                c_count.fetch_add(1, Ordering::Relaxed);
            }
        });

        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 1);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);

        // writing at a key doesn't notify siblings
        *a.label().write() = "Foo".into();
        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 2);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);

        // the keys can be reorganized
        store.todos().write().swap(0, 2);
        let after = store.todos().get_untracked();
        assert_eq!(
            after,
            vec![Todo::new(12, "C"), Todo::new(11, "B"), Todo::new(10, "Foo")]
        );

        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 2);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);

        // and after we move the keys around, they still update the moved items
        a.label().set("Bar".into());
        let after = store.todos().get_untracked();
        assert_eq!(
            after,
            vec![Todo::new(12, "C"), Todo::new(11, "B"), Todo::new(10, "Bar")]
        );
        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 3);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);

        // we can remove a key and add a new one
        store.todos().write().pop();
        store.todos().write().push(Todo::new(13, "New"));
        let after = store.todos().get_untracked();
        assert_eq!(
            after,
            vec![Todo::new(12, "C"), Todo::new(11, "B"), Todo::new(13, "New")]
        );
        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 3);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);
    }
}
```

gbj:
I've added my test as a regression test in #4483. It would be worth adding a similar test using the alternate key types and keyed structures in your PR: for example, if you add the same test but use a HashMap or BTreeMap it should have the same reactive characteristics.

Re: the example — I would rather keep it to one example, as the slotmap addition duplicates most of the existing one, which means 2x the maintenance burden in order to show you can use keyed store fields with more than one kind of data structure. Happy to accept changes to the `stores` example that exists. To be clear, the `user` field is *intended* to be useless at the moment -- it only exists so there's an example of a sibling field, which is important because it shows that updating a field does not notify all its siblings. 

Here's a diff that shows what I mean about adding an index. I can confirm this fixes this issue in the test.

```diff
diff --git a/reactive_stores/src/keyed.rs b/reactive_stores/src/keyed.rs
index 24067e683..d212c33a4 100644
--- a/reactive_stores/src/keyed.rs
+++ b/reactive_stores/src/keyed.rs
@@ -29,57 +29,58 @@ pub trait KeyedAccess {
     /// Collection values.
     type Value;
     /// Acquire mutable access to a value.
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value;
+    fn keyed_mut(&mut self, index: usize, key: Self::Key) -> &mut Self::Value;
     /// Acquire read-only access to a value.
-    fn keyed(&self, key: Self::Key) -> &Self::Value;
+    fn keyed(&self, index: usize, key: Self::Key) -> &Self::Value;
 }
 impl<T> KeyedAccess for VecDeque<T> {
     type Key = usize;
     type Value = T;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
-        self.index(key)
+
+    fn keyed(&self, index: usize, _key: Self::Key) -> &Self::Value {
+        self.index(index)
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
-        self.index_mut(key)
+    fn keyed_mut(&mut self, index: usize, _key: Self::Key) -> &mut Self::Value {
+        self.index_mut(index)
     }
 }
 impl<T> KeyedAccess for Vec<T> {
     type Key = usize;
     type Value = T;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
-        self.index(key)
+    fn keyed(&self, index: usize, _key: Self::Key) -> &Self::Value {
+        self.index(index)
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
-        self.index_mut(key)
+    fn keyed_mut(&mut self, index: usize, _key: Self::Key) -> &mut Self::Value {
+        self.index_mut(index)
     }
 }
 impl<T> KeyedAccess for [T] {
     type Key = usize;
     type Value = T;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
-        self.index(key)
+    fn keyed(&self, index: usize, _key: Self::Key) -> &Self::Value {
+        self.index(index)
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
-        self.index_mut(key)
+    fn keyed_mut(&mut self, index: usize, _key: Self::Key) -> &mut Self::Value {
+        self.index_mut(index)
     }
 }
 impl<K: Ord, V> KeyedAccess for std::collections::BTreeMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(&key).expect("key does not exist")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(&key).expect("key does not exist")
     }
 }
 impl<K: Hash + Eq, V> KeyedAccess for std::collections::HashMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(&key).expect("key does not exist")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(&key).expect("key does not exist")
     }
 }
@@ -519,19 +520,20 @@ where
         let inner_path = self.inner.path().into_iter().collect();
         let keys = self.inner.keys()?;
 
-        // Check whether the key path still exists.
-        keys.with_field_keys(
-            inner_path,
-            |keys| (keys.get(&self.key), vec![]),
-            || self.inner.latest_keys(),
-        )
-        .flatten()?;
+        let index = keys
+            .with_field_keys(
+                inner_path,
+                |keys| (keys.get(&self.key), vec![]),
+                || self.inner.latest_keys(),
+            )
+            .flatten()
+            .map(|(_, idx)| idx)?;
 
         let key = self.key;
         Some(MappedMutArc::new(
             inner,
-            move |n| n.keyed(key),
-            move |n| n.keyed_mut(key),
+            move |n| n.keyed(index, key),
+            move |n| n.keyed_mut(index, key),
         ))
     }
 
@@ -544,13 +546,14 @@ where
             .keys()
             .expect("using keys on a store with no keys");
 
-        // Check whether the key path still exists.
-        keys.with_field_keys(
-            inner_path.clone(),
-            |keys| (keys.get(&self.key), vec![]),
-            || self.inner.latest_keys(),
-        )
-        .flatten()?;
+        let index = keys
+            .with_field_keys(
+                inner_path,
+                |keys| (keys.get(&self.key), vec![]),
+                || self.inner.latest_keys(),
+            )
+            .flatten()
+            .map(|(_, idx)| idx)?;
 
         let triggers = self.triggers_for_current_path();
         let key = self.key;
@@ -558,8 +561,8 @@ where
             triggers,
             MappedMutArc::new(
                 inner,
-                move |n| n.keyed(key),
-                move |n| n.keyed_mut(key),
+                move |n| n.keyed(index, key),
+                move |n| n.keyed_mut(index, key),
             ),
         ))
     }
diff --git a/reactive_stores/src/slotmap.rs b/reactive_stores/src/slotmap.rs
index f8e443120..1446b2513 100644
--- a/reactive_stores/src/slotmap.rs
+++ b/reactive_stores/src/slotmap.rs
@@ -4,50 +4,50 @@ use crate::KeyedAccess;
 impl<K: slotmap::Key, V> KeyedAccess for slotmap::SlotMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(key).expect("key does not exist.")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(key).expect("key does not exist")
     }
 }
 impl<K: slotmap::Key, V> KeyedAccess for slotmap::DenseSlotMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(key).expect("key does not exist.")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(key).expect("key does not exist")
     }
 }
 impl<K: slotmap::Key, V> KeyedAccess for slotmap::HopSlotMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(key).expect("key does not exist.")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(key).expect("key does not exist")
     }
 }
 impl<K: slotmap::Key, V> KeyedAccess for slotmap::SecondaryMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(key).expect("key does not exist.")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(key).expect("key does not exist")
     }
 }
 impl<K: slotmap::Key, V> KeyedAccess for slotmap::SparseSecondaryMap<K, V> {
     type Key = K;
     type Value = V;
-    fn keyed(&self, key: Self::Key) -> &Self::Value {
+    fn keyed(&self, _index: usize, key: Self::Key) -> &Self::Value {
         self.get(key).expect("key does not exist.")
     }
-    fn keyed_mut(&mut self, key: Self::Key) -> &mut Self::Value {
+    fn keyed_mut(&mut self, _index: usize, key: Self::Key) -> &mut Self::Value {
         self.get_mut(key).expect("key does not exist")
     }
 }
```

TiemenSch:
Thanks a bunch for the review, insight, test, and patch! I have some free time tomorrow to take a look at the example.

TiemenSch:
edit: just saw that you moved `reactive_stores` a bit on main as well.

Meanwhile I tried replicating the test for a slotmap example, but it seems the "moving keys around" doesn't really have an equivalent implementation for slotmaps or most other map types. 

How would you suggest we proceed?

gbj:
> it seems the "moving keys around" doesn't really have an equivalent implementation for slotmaps or most other map types.
> 
> How would you suggest we proceed?

My suggestion would be that you add an equivalent test for map-like types that removes the irrelevant parts of the test (moving things within a slice) but ensures that the behavior is correct in the relevant ways (for example, if you create a BTreeMap with items with keys 1, 3, 5, take keyed access to each of those fields, modify them, add keys 2 and 4, modify them, remove 3, does it all look the way it should? Basically come up with some scenario of a series of arbitrary changes and ensure that things notify when they should and that the state of the whole thing is correct. Whatever you think makes sense.)

TiemenSch:
Something like this?

TiemenSch:
Would you say this PR needs to update the stores example? What data should be kept as a `Vec`, and what as a `...Map`? An idea:

- I could add a some map that tracks the users/emails independently of the tasks that displays something if a user is "in store".
- I can also refrain from adding or changing any example for now, such that we cover this in a separate more well-thought out PR.

Alternatively or additionally, I can add some additional paragraphs to the documentation as well. I was thinking of:

- Snippet showcasing the use of arbitrarily keyed collection types.
- Use of `AtKeyed::new(...)` for specific object access in collection fields.

But those could be separate PRs.

gbj:
Copying thoughts on the examples from a comment above:

> Re: the example — I would rather keep it to one example, as the slotmap addition duplicates most of the existing one, which means 2x the maintenance burden in order to show you can use keyed store fields with more than one kind of data structure. Happy to accept changes to the stores example that exists. To be clear, the `user` field is intended to be useless at the moment -- it only exists so there's an example of a sibling field, which is important because it shows that updating a field does not notify all its siblings.

Just to expand a little: the `user` field is not intended to be a map between users and their todos, but to be "the currently logged-in user."

I would be fine with adding some other map-like field to the example as a demonstration.

And additions to the docs are always great, either as part of this PR or separately.

Overall: This PR looks like it's in great shape. Thanks so much for your work.

TiemenSch:
I tried to rework the example to include some "achievements" per user, which tracks their completed tasks. 

However, during some editing attempts, I found that working with `<For ...>` is somewhat unintuitive at times.

I.e., the syntax gets somewhat confusing/unwieldy and feels superfluous at times:

```rust
#[derive(Debug, Store, Serialize, Deserialize)]
struct Todos {
    user: User,
    #[store(key: usize = |todo| todo.id)]
    todos: Vec<Todo>,
    #[store(key: Arc<String> = |(name, _)| name.clone())]
    completed: BTreeMap<Arc<String>, (Arc<String>, usize)>,  
    // Why not `BTreeMap<Arc<String>, usize>`?
}

#[component]
fn SomeComponent(store: Store<Todos>) -> impl IntoView {
  view ! {
    /// For the key fn we get the value to work with, why not the (key, value) tuple?
    /// Or a fallback to the key function already in the KeyedSubfield?
    <For each=move||store.completed() key=|value|value.0.clone() let:value>
    ...
    </For>
  }
}
```

With SlotMaps, it's convenient enough to use their `insert_with_key` function to insert an entry for a key type that is `Copy`. For `Clone` types it quickly becomes unwieldy, though.

gbj:
> For the key fn we get the value to work with, why not the (key, value) tuple?

As I understand it, this is because `AtKeyed`'s value is the `KeyedAccess::Value` type, which you've defined as the value for maps. i.e., `.into_iter()` here is the equivalent of `.values()` on most map types, not `.into_iter()` on map types. i.e., this has to do with the `KeyedAccess` implementation, not with `For`.

TiemenSch:
Instead of exposing the key as part of the iterator, I tried to see how far I could get by exposing the key using an accessor function on the `AtKeyed` struct itself that is passed into the `key=...`  and `let:...` statement. The idea would be that the key function of the `<For>` would then become trivial such that you can `<For each=move||store.collection() key=|entry|entry.key() let:entry>`

However, I think I might be botching the example at this point, because the reactivity for the "user achievements" that I added is now broken. 

The overall serde JSON works, though, which is leaving me rather puzzled for the moment.

gbj:
Hm no you haven't messed up the example actually. This is a question/issue that's popped up in a number of different places , which is: Should writing to a parent field notify the children?

In this case, specifically, the issue is "Should writing to a `KeyedSubfield` notify its `AtKeyed` children?" Because you're tracking `store.completed().at_key(store.user().name().get().into())`, but you're writing to `store.completed()`.

I don't think there's a general solution to this other than data-diffing on all writes to a store field, which we currently do via `.patch()` but not in general. There are two possible approaches:

## Option 1
Solution: Track the `KeyedSubfield` in each `AtKeyed`, which will yield lots of false positives (effects that re-run when they don't need to) but avoid false negatives (the broken reactivity you see in the example)

The example can be fixed with the following patch, which borrows a `track_field` implementation used for several other types.
```diff
diff --git a/reactive_stores/src/keyed.rs b/reactive_stores/src/keyed.rs
index c2f180ed5..f402a7caf 100644
--- a/reactive_stores/src/keyed.rs
+++ b/reactive_stores/src/keyed.rs
@@ -641,6 +641,23 @@ where
     fn keys(&self) -> Option<KeyMap> {
         self.inner.keys()
     }
+
+    fn track_field(&self) {
+        let mut full_path = self.path().into_iter().collect::<StorePath>();
+        let trigger = self.get_trigger(self.path().into_iter().collect());
+        trigger.this.track();
+        trigger.children.track();
+
+        // tracks `this` for all ancestors: i.e., it will track any change that is made
+        // directly to one of its ancestors, but not a change made to a *child* of an ancestor
+        // (which would end up with every subfield tracking its own siblings, because they are
+        // children of its parent)
+        while !full_path.is_empty() {
+            full_path.pop();
+            let inner = self.get_trigger(full_path.clone());
+            inner.this.track();
+        }
+    }
 }

 impl<Inner, Prev, K, T> DefinedAt for AtKeyed<Inner, Prev, K, T>
 ```

However, this now leads to test failures:
```
failures:

---- keyed::tests::keyed_fields_can_be_moved stdout ----

thread 'keyed::tests::keyed_fields_can_be_moved' (4684942) panicked at reactive_stores/src/keyed.rs:1044:9:
assertion `left == right` failed
  left: 3
 right: 2
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
(and some other similar ones)

Looking at that particular test is a good illustration of the issue:
```rust
        // the keys can be reorganized
        store.todos().write().swap(0, 2);
        let after = store.todos().get_untracked();
        assert_eq!(
            after,
            vec![Todo::new(12, "C"), Todo::new(11, "B"), Todo::new(10, "Foo")]
        );

        tick().await;
        assert_eq!(a_count.load(Ordering::Relaxed), 2);
        assert_eq!(b_count.load(Ordering::Relaxed), 1);
        assert_eq!(c_count.load(Ordering::Relaxed), 1);
```
The patch above will modify these asserts so that they now require 3, 2, 2. i.e., if tracking a keyed field means you also track changes to its parent — *independent* of whether the `.into_iter()` iterator re-runs — then doing something like "rearranging three keyed items in a slice" will now notify each of those keyed items, even though none of their values has changed.

I think this is avoided in some other cases by the fact that the `.into_iter()` implementation does track the parent.

## Option 2

Require that if you're tracking the keyed field, you need to modify the record through that keyed field.

In other words, the following patch to the *example* code *also* fixes the issue, without changing the library code:

```diff
diff --git a/examples/stores/src/lib.rs b/examples/stores/src/lib.rs
index 9b93f2d59..c6bf7887e 100644
--- a/examples/stores/src/lib.rs
+++ b/examples/stores/src/lib.rs
@@ -213,9 +213,11 @@ fn TodoRow(
                 status.write().next_step();
                 if was_undone && status.done() {
                     let name = store.user().name().get();
-                    store
-                        .completed()
-                        .update(|completed| *completed.entry(name.into()).or_default() += 1)
+                    if store.completed().read().contains_key(&name) {
+                       *store.completed().at_key(name.into()).write() += 1;
+                    } else {
+                       *store.completed().write().entry(name.into()).or_default() += 1;
+                    }
                 }
             }>
                 {move || {
```

This prevents any "false positives" when modifying the parent, but it requires you to know that you need to update the keyed field.

TiemenSch:
Sorry for the lack in activity on this! Do you want to continue on the other PR or what is the best course of action? I might be able to spend some evenings on this the coming week.

Thanks for clarifying the reactivity mechanics. I think it's a delicate balance here. On the one hand, I would expect that fully setting a new value is allowed to trigger all children and it would be OK to have the user patch/update specific fields instead to avoid too many false positives. I think both paths (full replace and pinpoint field updates) absolutely have their merit and should have a clearly defined playbook/convention. 

I took a gander at your keyed patching PR, but need to dive in some further to fully understand the goings-on. It does seem however that the key-diffing could impact performance by quite a bit for large collections? If I'm not mistaken you would have to loop over the entire collection to check whether just one key might have been changed, while the user knew (or could have known) what keys were going to change during their update/patch. 

Would another option be implementing Option 1 (i.e. a parent's trigger notifies children) while making absolutely sure that there is a fine-grained alternative for users if they wish to do so such as Option 2. Option 2 looks like a good candidate to be reworked into a convenience method, though.

Intead of doing a key diff, wouldn't it be more intuitive to either falling back to notifying all children (but not siblings) when doing an update and providing an option for the user to limit the update scope to specific keys of which they know they just changed values)? 

For me as a user it would be intuitive that everything that is part of the mutable/replaced object will get notified, except for when the user supplied a specific key/field in a collection/store through the Leptos API. In that case, that key/those keys and their respective children should be triggered.

It could be an alternative to provide an extra option that does the somewhat heavier diffing in a centralized call for all listeners and does a better (though slightly more expensive) job of triggering for changes only, but that might involve a lot of key diffs and/or PartialEq's for large and/or heavily nested data structures.

gbj:
> Sorry for the lack in activity on this!

No worries at all! This has been a big PR and you've been very responsive and patient with me as we've worked through it.

> Do you want to continue on the other PR or what is the best course of action? I might be able to spend some evenings on this the coming week.

Yes, I think this is in great shape and will be happy to merge it. 

> I would expect that fully setting a new value is allowed to trigger all children and it would be OK to have the user patch/update specific fields instead to avoid too many false positives.

> For me as a user it would be intuitive that everything that is part of the mutable/replaced object will get notified

I think your intuition is correct that "false negatives" (broken reactivity) here are worse than "false positives" (notifications that are technically unnecessary). I think going ahead with my Option 1 (making the change in that diff and modifying tests accordingly) is probably the way to go. Make sure that fixes your example and feels right and I think it's fine.

Convenience methods like a keyed/fine-grained `.entry()` equivalent (so, my Option 2) would be great and can be done in separate PRs if desired.

> I took a gander at your keyed patching PR, but need to dive in some further to fully understand the goings-on. It does seem however that the key-diffing could impact performance by quite a bit for large collections? If I'm not mistaken you would have to loop over the entire collection to check whether just one key might have been changed, while the user knew (or could have known) what keys were going to change during their update/patch.

The important thing to know about patching is that it's primarily intended for cases where you do *not* have fine-grained control over the updates, but you want to have fine-grained notifications to the extent possible. So, think of it as "I have gotten new data from the server and am trying to patch the store with it." You don't know which fields may have changed on the server, so the two choices are 1) do no diffing and notify *every* entry in the collection that it has changed, or 2) do the keyed diff and notify *some* entries that they've changed.

The assumption here is that side effects are expensive relative to computation — so that if we have new data coming from the server, we actually *want* to loop over the whole collection in order to check whether just one key has changed, because then we only notify on that one key. (Unless I'm misunderstanding what you mean.)

But if, as you say, the user knew or could have known what keys were going to change, they should use the more direct update methods rather than `.patch()`.

******

tl:dr; let's go with Option 1; if you include that change, update the tests, and are happy with how the example feels at that point, I'll be happy to merge this.

TiemenSch:
Added option 1 and updated tests :)

From what I can see the example works as expected now

TiemenSch:
Forgot to run the tests with --all-features :nerd_face: 

It should now be in the green!

gbj:
Ok, looks like we're all set -- thanks for all your work on this!

PR #4484, opened 2025-12-12
Implement 'LeptosRoute' for actix_web::Scope

OhseyDev:
All this does is copy 'LeptosRoutes' implementation for actix_web::App and rename to actix_web::Scope as 'LeptosRoutes' trait implementation for actix_web::App uses functions that both actix_web::App and actix_web::Scope have.

Seems a little hideous, but actix-web doesn't provide a generic trait for the shared functionality.

### DIFF
--- a/leptos_macro/src/view/component_builder.rs
+++ b/leptos_macro/src/view/component_builder.rs
@@ -176,7 +176,9 @@ pub(crate) fn component_to_tokens(
     let spreads = (!(spreads.is_empty())).then(|| {
         if cfg!(feature = "__internal_erase_components") {
             quote! {
-                .add_any_attr(vec![#(#spreads.into_any_attr(),)*])
+                .add_any_attr({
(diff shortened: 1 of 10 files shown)


gbj:
I think copy-pasting the whole body is not the best solution here, because it essentially doubles the maintenance burden over time by requiring someone to keep the bodies of these two functions in sync, when in fact they do exactly the same thing.

I would recommend creating a local trait like `AddRoute` that you then implement for `App` and for `Scope` and that just dispatches to the appropriate `.route()` method for each type.

Then, you can implement `LeptosRoutes` for any type that implements `AddRoute`, and use the existing logic.

OhseyDev:
Good idea. I will get to working on that. Although I had already looked into having a trait. It would require contributing to actix-web. I am already looking into moving common functionality of the Scope, App and Resource struct's in actix-web into a trait.

OhseyDev:
The copy and paste was a little bit lazy/desperate on my part, as it is a needed feature for a project of mine.

OhseyDev:
Apologies for the delayed response. I had a football game that I had to stay focused for, and then a family gathering.

PR #4494, opened 2025-12-16
chore: update `wasm-bindgen`

LeoBorai:
Updates `wasm-bindgen-*` related crates. This allows to use recently exposed APIs in
`web-sys` such as [CookieStore](https://docs.rs/web-sys/latest/web_sys/struct.CookieStore.html).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,6 +616,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "276a59bf2b2c967788139340c9f0c5b12d7fd6630315c15c217e559de85d2609"
 
 [[package]]
+name = "cast"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
+
+[[package]]
 name = "cc"
 version = "1.2.40"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1904,9 +1910,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.80"
+version = "0.3.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "852f13bec5eba4ba9afbeb93fd7c13fe56147f055939ae21c43a29a0ecb2702e"
+checksum = "464a3709c7f55f1f721e5389aa6ea4e3bc6aba669353300af094b29ffbdde1d8"
 dependencies = [
  "once_cell",
  "wasm-bindgen",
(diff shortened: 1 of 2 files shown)


gbj:
Shouldn't access to new APIs already be available by using the latest versions in your crate?

LeoBorai:
> Shouldn't access to new APIs already be available by using the latest versions in your crate?

Hi @gbj!

Well the issue comes from `wasm-bindgen` version mismatches.

One issue is the supported versions for `web-sys` with regards to `wasm-bindgen` which is encountered when bumping `web-sys` to most recent version, the `web-sys`'s `wasm-bindgen` version turns out newer than then one used in Leptos causing a resolve dependency mismatch.

The other is `wasm-bindgen-cli`, which complains on compile time when theres 2 different versions of `wasm-bindgen` in two build artifacts.

gbj:
Shouldn't specifying the later version in your own `Cargo.toml` use the later version, which is what you want? We don't pin any of the versions here. I might be missing something obvious.

LeoBorai:
Hi @gbj, after taking a closer look I've found the issue.

So I had my `Cargo.toml` pointing to the desired `wasm-bindgen` version:

```toml
wasm-bindgen = "=0.2.106"
```

When I checked on packages using `0.2.100` (which was the one I was using before), I noticed
theres no crates pointing to such version anymore:

```bash
cargo tree --target aarch64-unknown-linux-gnu -i -p wasm-bindgen@0.2.100 -i
workspace root:
package:   ~/src/app/ui/Cargo.toml
workspace: ~/Cargo.toml
error: package ID specification `wasm-bindgen@0.2.100` did not match any packages
help: there are similar package ID specifications:

  wasm-bindgen@0.2.106
```

But even with this state I received the following message when building with Trunk.

```
it looks like the Rust project used to create this Wasm file was linked against
version of wasm-bindgen that uses a different bindgen format than this binary:

  rust Wasm file schema version: 0.2.106
     this binary schema version: 0.2.100

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating this binary or
the wasm-bindgen dependency in the Rust project.

You should be able to update the wasm-bindgen dependency with:

    cargo update -p wasm-bindgen --precise 0.2.100

don't forget to recompile your Wasm file! Alternatively, you can update the
binary with:

    cargo install -f wasm-bindgen-cli --version 0.2.106

if this warning fails to go away though and you're not sure what to do feel free
to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!
```

So I looked into my version of `wasm-bindgen-cli` and everything seems to be fine:

```bash
wasm-bindgen --version
wasm-bindgen 0.2.106
```

So far everythin with regards to `Cargo.toml` versions seems fine. But it wasn't compiling
so I looked into `Trunk.toml`:

```toml
[tools]
tailwindcss = "4.1.2"
wasm_bindgen = "0.2.100"
```

Trunk was performing WASM Bindgen tasks on version `0.2.100`, causing trouble
when linking due to version mismatches. Turns out Trunk is able to use its own
copies of tools instead of system's.

Based on this I think this PR is no longer relevant.

Thank you so much for taking a look at this!

PR #4495, opened 2025-12-17
fix: do not show Transition fallback on 2nd change if 1st change     resolved synchronously (closes #4492)

### LINKED ISSUE #3868: `Transition` component moves to fallback state long after initial load on newly subscribed resources (in edge case).
*EDIT:* There was some error in my reproduction, this is now fixed.

**Describe the bug**
It is expected that the `Transition` component will not switch to its fallback state once all the resources in it have loaded once.
However, there is an edge case that I have been encountering: If the initial load on the Transition doesn't subscribe to any resources, (i.e. it resolves without ever going into the fallback state), then it will not fully be marked as "resolved." If a signal then causes the Transition to subscribe to a new resource, then it will move back into fallback state. Note this only happens once - the Transition is still correctly ensuring that it never goes into fallback more than once. It also doesn't happen for SSR rendered transitions.

**Leptos Dependencies**

```toml
leptos = { version = "0.7.8", features = ["nightly"] }
leptos_meta = { version = "0.7.8" }
leptos_actix = { version = "0.7.8", optional = true }
leptos_router = { version = "0.7.8", features = ["nightly"] }
```

**To Reproduce**

Click the "Switch" button in order to render the Transition component on the client side, then click the "Switch to async" button. You will see that the Transition switches into fallback state the moment the button is being pressed.

```rs
#[component]
    let s_res = StoredValue::new(None::<ArcLocalResource<&'static str>>);

    let toggle = RwSignal::new(false);
    let is_async_mode = RwSignal::new(false);

    let dragon = move || {
        if !is_async_mode.get() {
            "Safi'jiiva"
        } else {
            let r = s_res.get_value();

            let res = r.unwrap_or_else(|| {
                let res = ArcLocalResource::new(|| async {
                    async_std::task::sleep(Duration::from_millis(1000)).await;
                    "The Ur-Dragon"
                });
                s_res.set_value(Some(res.clone()));
                res
            });

            res.get()
                .map(|s| *s)
                .unwrap_or("The Ur-Dragon is coming...")
        }
    };

    let contents = move || {
        if toggle.get() {
            let fallback = move || {
                logging::log!("Fallback rendered...");
                "Loading..."
            };

            let toggle_async_mode = move |_| {
                is_async_mode.update(|v| *v = !*v);
            };

            let async_toggle_label = move || match is_async_mode.get() {
                true => "Switch to const",
                false => "Switch to async",
            };

            view! {
                <button on:click=toggle_async_mode>{async_toggle_label}</button>
                <Transition fallback>
                    <h1>{dragon}</h1>
                </Transition>
            }
            .into_any()
        } else {
            view! {
                "Toggled off"
            }
            .into_any()
        }
    };

    view! {
        <div>
            <button on:click=move |_| toggle.update(|v| *v = !*v)>"Switch"</button>
            {contents}
        </div>
    }
}
```

**Next Steps**
[ ] I will make a PR
[ ] I would like to make a PR, but need help getting started
[x] I want someone else to take the time to fix this

In my app, I might work around this issue by simply not using LocalResource at all, which I can do because I am never doing anything with `Suspense` or fallback states (all my SSR stuff is blocking). However for other future apps I write (and possibly future features on this app, although unlikely) this might become a problem. Since it currently only results in a mildly annoying flicker, and I have other more important tasks to complete, I will wait a few weeks before implementing my workaround, in case someone else finds the time to resolve this.

### LINKED ISSUE #4492: Transition renders fallback after initial render of children
the logic of the `Transition` element currently means that it will render children initially _without_ awaiting a future, and then if triggered again _with_ awaiting a future it will render the fallback

this IMO is not how its behavior is described. here is the relevant code:

https://github.com/leptos-rs/leptos/blob/5af5fdeeedab7e37f768ab65ab1a8988eb1402e4/leptos/src/suspense_component.rs#L179-L184

the comment that says the logic will work `probably`  is key as this issue happens consistently when applying the logic in the description at the top of the issue

very similar logic is in the hydrate function:

https://github.com/leptos-rs/leptos/blob/5af5fdeeedab7e37f768ab65ab1a8988eb1402e4/leptos/src/suspense_component.rs#L526-L531

changing the `nth_run` check from 2 to 1 in the `build` function fixes this issue but that has the knock on effect of the fallback never being displayed when it should

the code below uses `leptos-rs/start-axum` as a base with the following extra dependencies

```
gloo-timers = { version = "0", features = ["futures"], optional = true }
send_wrapper = "0"
```

replace the `app.rs` with the following

```
use leptos::{logging::log, prelude::*};
use leptos_meta::{provide_meta_context, MetaTags, Stylesheet, Title};
use leptos_router::components::*;
use leptos_router::hooks::*;
use leptos_router::*;

    view! {
        <!DOCTYPE html>
        <html lang="en">
            <head>
                <meta charset="utf-8"/>
                <HydrationScripts options/>
                <MetaTags/>
            </head>
            <body>
                <App/>
            </body>
        </html>
    }
}

#[component]
    provide_meta_context();
    view! {
        <Stylesheet id="leptos" href="/pkg/bug6.css"/>
        <Router>
            <main>
                    <Route path=StaticSegment("p") view=OtherPage/>
                </Routes>
            </main>
        </Router>
    }
}

#[component]
    let query = use_query_map();
    let q = move || {
        query
            .get()
            .get("q")
            .unwrap_or("0".into())
            .parse::<u8>()
            .unwrap_or(0u8)
    };

    log!("initial");
    let initial = q();

    let r = Resource::new(
        move || q(),
        move |length| async move {
            let mut v = vec![];
            for i in 0..length {
                // if i >= initial {
                if i > 0 {
                    #[cfg(feature = "hydrate")]
                    log!("delay");
                    #[cfg(feature = "hydrate")]
                    send_wrapper::SendWrapper::new(gloo_timers::future::TimeoutFuture::new(2000))
                        .await;
                } else {
                    log!("cache");
                }
                v.push(i);
            }
            v
        },
    );

    view! {
        <h1>"Welcome to Leptos!"</h1>

        <Transition fallback=|| { view! { <p> "FALLBACK" </p> } }>
          <For each={move || r.get().unwrap_or(vec![])} key={|k| *k} let:k>
            <p><A href="/p"> "PATH" {k} </A></p>
          </For>
        </Transition>

        <p><A href=move || format!("/?q={}", q() + 1)> "Add " {q} </A></p>
        <p><A href="/p"> "PATH" </A></p>
        <p><A href="/"> "ROOT" </A></p>
    }
}

#[component]
fn OtherPage() -> impl IntoView {
    let count = RwSignal::new(0);
    let on_click = move |_| *count.write() += 1;

    view! {
        <h1>"Welcome to Leptos!"</h1>
        <button on:click=on_click>"Click Me: " {count}</button>
        <p><A href="/?q=1"> "Q 1" </A></p>
        <p><A href="/?q=2"> "Q 2" </A></p>
        <p><A href="/"> "ROOT" </A></p>

    }
}
```

the logic in the `Transition` element will load the first element without an awaited delay and the rest with a delay

steps to reproduce

1. click the `Add 0` link 
2. click the `PATH 0` link
3. click browser back
4. click the `Add 1` link
5. see the fallback triggered 

as you can probably tell i have looked around for an easy solution and failed to find one. i'm guessing any fix will also affect `Suspense` elements too 

please let me know if you need more information or a video

thanks for looking into it

gbj:
I'd like to close #3868 at the same time, I just haven't *quite* found the right combo yet.

### DIFF
--- a/examples/regression/e2e/features/issue_4492.feature
+++ b/examples/regression/e2e/features/issue_4492.feature
@@ -33,6 +33,6 @@ Feature: Regression test for issue #4492
 		When I click the button c-toggle
 		Then I see c-result has the text 0
 		When I click the button c-button
-		Then I see c-result has the text 0
+		Then I see c-result has the text 42
 		When I wait 100ms
 		Then I see c-result has the text 1
--- a/leptos/src/suspense_component.rs
+++ b/leptos/src/suspense_component.rs
@@ -134,7 +134,7 @@ where
             }
         });
         let has_tasks =
-            Arc::new(move || tasks.with_untracked(SlotMap::is_empty));
+            Arc::new(move || !tasks.with_untracked(SlotMap::is_empty));
 
         OwnedView::new(SuspenseBoundary::<false, _, _> {
             id,
--- a/leptos/src/transition.rs
+++ b/leptos/src/transition.rs
@@ -117,7 +117,7 @@ where
             }
         });
         let has_tasks =
-            Arc::new(move || tasks.with_untracked(SlotMap::is_empty));
+            Arc::new(move || !tasks.with_untracked(SlotMap::is_empty));
         if let Some(set_pending) = set_pending {
             Effect::new_isomorphic({
                 let none_pending = none_pending.clone();
--- a/leptos_server/src/once_resource.rs
+++ b/leptos_server/src/once_resource.rs
@@ -15,7 +15,7 @@ use codee::{
     Decoder, Encoder,
 };
 use core::{fmt::Debug, marker::PhantomData};
-use futures::Future;
+use futures::{Future, FutureExt};
 use or_poisoned::OrPoisoned;
 use reactive_graph::{
     computed::{
@@ -258,11 +258,17 @@ where
         if let Some(suspense_context) = use_context::<SuspenseContext>() {
             if self.value.read().or_poisoned().is_none() {
                 let handle = suspense_context.task_id();
-                let ready = SpecialNonReactiveFuture::new(self.ready());
-                reactive_graph::spawn(async move {
-                    ready.await;
-                    drop(handle);
-                });
+                let mut ready =
+                    Box::pin(SpecialNonReactiveFuture::new(self.ready()));
+                match ready.as_mut().now_or_never() {
+                    Some(_) => drop(handle),
+                    None => {
+                        reactive_graph::spawn(async move {
+                            ready.await;
+                            drop(handle);
+                        });
+                    }
+                }
                 self.suspenses.write().or_poisoned().push(suspense_context);
             }
         }
--- a/reactive_graph/src/computed/async_derived/arc_async_derived.rs
+++ b/reactive_graph/src/computed/async_derived/arc_async_derived.rs
@@ -632,12 +632,29 @@ impl<T: 'static> ReadUntracked for ArcAsyncDerived<T> {
 
     fn try_read_untracked(&self) -> Option<Self::Value> {
         if let Some(suspense_context) = use_context::<SuspenseContext>() {
+            // create a handle to register it with suspense
             let handle = suspense_context.task_id();
-            let ready = SpecialNonReactiveFuture::new(self.ready());
-            crate::spawn(async move {
-                ready.await;
-                drop(handle);
-            });
+
+            // check if the task is *already* ready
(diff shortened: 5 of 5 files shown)


jim-taylor-business:
this definitely fixes #4492

PR #4496, opened 2025-12-17
Allow argument names with suffixes in server functions

aisuzuki:
# Description
`#[server]` macro generates an argument struct from a server function's arguments. For this reason, developers have not way to control Clippy warnings even though the coding styles for function's argument is correct. 

Example: 
```rust
#[server]
async fn server_function(
  argument1_suffix: String,
  argument2_suffix: String,
  argument3_suffix: String,
) -> Result<(), ServerFnError> {
...
```

The code above causes `clippy::struct_field_names` warning
server_function_with_suffixed_arguments

# Environment
- Leptos version: 1.8.4

Related to: #4498

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -309,6 +309,16 @@ impl ServerFnCall {
             .collect::<TokenStream2>()
     }
 
+    /// Get the lint attributes for the server function.
+    pub fn lint_attrs(&self) -> TokenStream2 {
+        // pass through lint attributes from the function body
+        self.body
+            .lint_attrs
+            .iter()
+            .map(|lint_attr| quote!(#lint_attr))
+            .collect::<TokenStream2>()
+    }
+
     fn fn_name_as_str(&self) -> String {
         self.body.ident.to_string()
     }
@@ -388,6 +398,8 @@ impl ServerFnCall {
             PathInfo::None => quote! {},
         };
 
+        let lint_attrs = self.lint_attrs();
+
         let vis = &self.body.vis;
         let struct_name = self.struct_name();
         let fields = self
@@ -410,6 +422,7 @@ impl ServerFnCall {
             #docs
             #[derive(Debug, #derives)]
             #addl_path
+            #lint_attrs
             #vis struct #struct_name {
                 #(#fields),*
             }
@@ -1465,6 +1478,8 @@ pub struct ServerFnBody {
     pub block: TokenStream2,
     /// The documentation of the server function.
     pub docs: Vec<(String, Span)>,
+    /// The lint attributes of the server function.
+    pub lint_attrs: Vec<Attribute>,
     /// The middleware attributes applied to the server function.
     pub middlewares: Vec<Middleware>,
 }
@@ -1522,6 +1537,27 @@ impl Parse for ServerFnBody {
             };
             !attr.path.is_ident("doc")
         });
+
+        let lint_attrs = attrs
+            .iter()
+            .filter_map(|attr| {
+                if attr.path().is_ident("allow")
+                    || attr.path().is_ident("warn")
+                    || attr.path().is_ident("deny")
+                    || attr.path().is_ident("forbid")
+                {
+                    return Some(attr.clone());
+                }
+                None
+            })
+            .collect();
+        attrs.retain(|attr| {
+            !attr.path().is_ident("allow")
+                && !attr.path().is_ident("warn")
+                && !attr.path().is_ident("deny")
+                && !attr.path().is_ident("forbid")
+        });
+
         // extract all #[middleware] attributes, removing them from signature of dummy
         let mut middlewares: Vec<Middleware> = vec![];
         attrs.retain(|attr| {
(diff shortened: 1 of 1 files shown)


gbj:
Hm, I can't get clippy to generate this warning with the example. But that's mostly beside the point.

This PR blanket-allows, for all users of this library, a single, specific clippy warning, by embedding it into the macro's output. 

I'd suggest that a better approach would be to pass through `#[allow]` and other similar macros from the function definition to the struct definition. We already [check for a similar list](https://github.com/leptos-rs/leptos/blob/main/server_fn_macro/src/lib.rs#L1408-L1419)

Currently we only pass [docs](https://github.com/leptos-rs/leptos/blob/main/server_fn_macro/src/lib.rs#L410) through to the struct. Modifying this to pass docs and allows through would be a good start.

LeoniePhiline:
> This PR blanket-allows, for all users of this library, a single, specific clippy warning, by embedding it into the macro's output.

I find this to be the best possible solution. Here is my argumentation for why the blanket `#[allow]` attribute should be added:

From the perspective of the user, there is not even any struct. The struct only occurs in generated code, and generated code generally should not yield lint warnings or errors.

Since it is the macro, which generates the struct, the macro must also ensure that the resulting struct lints correctly, regardless of the user's linting configuration, and regardless of processed inputs, such as argument binding names.

The added `#[allow(clippy::struct_field_names)]`, attached to the generated struct, does exactly that.

Since `#[allow]` and not `#[expect]` is used, the attribute does not have negative side effects.

> I'd suggest that a better approach would be to pass through `#[allow]` and other similar macros from the function definition to the struct definition. 

I would politely disagree here, as struct-specific lints are semantically nonsensical when attached to function items:

Would it, from a user's perspective, be useful to annotate a `fn` item with `#[allow(clippy::struct_field_names)]`? Semantically, attached to a function, the attribute does not make sense, since it is for structs only.

The struct generation, however, is an implementation detail, and should be self-contained.

> I can't get clippy to generate this warning with the example.

The lint is not enabled by lenient default, but it is enabled by the popular `clippy::pedantic` lint group.

After creating a function as in @aisuzuki's example, for example in `examples/regression/src/app.rs`, ...:

```rust
#[server]
async fn server_function(
    argument1_suffix: String,
    argument2_suffix: String,
    argument3_suffix: String,
) -> Result<(), ServerFnError> {
    Ok(())
}
```

... run `clippy`, denying the entire `pedantic` lint group, or directly denying the triggering lint `clippy::struct_field_names`:

```bash
leptos/examples/regression $ cargo +nightly clippy -- --deny clippy::struct_field_names

warning: `auto` traits are unstable
[...]

error: all fields have the same postfix: `suffix`
  --> src/app.rs:63:1
   |
63 | #[server]
   | ^^^^^^^^^
   |
   = help: remove the postfixes
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names
   = note: requested on the command line with `-D clippy::struct-field-names`
   = note: this error originates in the attribute macro `server` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `regression` (lib) due to 1 previous error

```

gbj:
> `cargo +nightly clippy -- --deny clippy::struct_field_names`

Thanks — looks like this is a very recent addition? Even the nightly I had installed at present did not include this lint, but I've updated to the most recent nightly and it does.

> From the perspective of the user, there is not even any struct.

The fact that a struct is generated is [part of the documented behavior of the `#[server]` macro](https://docs.rs/leptos/latest/leptos/attr.server.html#named-arguments). The generated struct type's name is used as part of the API for things like `ActionForm` and form validation.

To the extent that this lint is useful at all (YMMV), it seems to me that it is equally useful for this generated struct.

> the attribute does not have negative side effects

Allowing `#[allow]` to be passed through to the generated struct gives the user control over which lints are shown. Applying a particular `#[allow]` to all generated structs takes away control from the user over which lints are shown. The `#[allow]` can always be added and passed through the macro; it can never be removed from the macro's output.

540655b090ffe9788b4cb13a5f92a20574c7a1b4 allows the user to do exactly what I'd expect to be able to do in this situation: see the lint warning, add an allow, and continue on with the day:
```rust
#[server]
#[allow(clippy::struct_field_names)]
async fn server_function(
    argument1_suffix: String,
    argument2_suffix: String,
    argument3_suffix: String,
) -> Result<(), ServerFnError> {
    Ok(())
}
```

aisuzuki:
Hi @gbj, thank you for your response. 
In the [list](https://github.com/leptos-rs/leptos/blob/main/server_fn_macro/src/lib.rs#L1408-L1419) of pass through attributes, there are still `deny` and `ignore` missing in my implementation. Should those items be also passed through to a generated struct?

gbj:
That would make sense to me, as they cover the same use cases.

aisuzuki:
`allow`, `warn`, `deny` and `forbid` are now pass through to a generated struct. `ignore` stays attribute to server function.

gbj:
Thanks very much!

PR #4497, opened 2025-12-17
chore: upgrade dependencies

sabify:
Supersedes: #4490.


sabify:
@gbj, can you please take this into effect as other dependents like `leptos-use` have already upgraded `actix-web` that causes compilation failure due to `server_fn`'s use of `actix-web` and `actix-ws`?

Please also look at my comment https://github.com/Baptistemontan/leptos_i18n/pull/239.

gbj:
Thanks for the ping; happy to merge.

Just so I understand, the Actix issue here is that in a minor release, Actix gated some of its websocket features behind a new `ws` feature, which is on by default (and therefore not a breaking change for most consumers of the library) — but because we deactivate all default features, we need to add it in?

sabify:
> Thanks for the ping; happy to merge.
> 
> Just so I understand, the Actix issue here is that in a minor release, Actix gated some of its websocket features behind a new `ws` feature, which is on by default (and therefore not a breaking change for most consumers of the library) — but because we deactivate all default features, we need to add it in?

Yes, this is exactly what happened.

sabify:
Can you please cut a patch release for `server_fn`?

gbj:
Done.

sabify:
Thanks!

PR #4505, opened 2025-12-22
chore: fix broken links in docs

fbdb:
Since the v0.7.0 release¹, the module structure has been reorganized: Some hardcoded links in the docs, readme and function definition were pointing to the previous (root) export instead of the new `prelude` module export.

¹: https://github.com/leptos-rs/leptos/releases/tag/v0.7.0

---

I was reading the book and stumbled upon some dead links (no such resource) on docs.rs. Then, I assumed that the broken links were due to v0.7.0 release, based on [what was said in the release note](https://github.com/leptos-rs/leptos/releases/tag/v0.7.0):
> Breaking Changes
>
> Imports
>
> I'm reorganizing the module structure to improve docs and discoverability. We will still have a prelude that can be used for glob imports of almost everything that's currently exported from the root.
> ```
> - use leptos::*;
> + use leptos::prelude::*;
> ```
> 
> (...)

I did a small grep of "docs.rs" links in the repo and tested all links, and stumbled upon some other links in the readme and a macro function documentation comment (for the latter, I did not take the time to wrap to 120 or whatever characters since the whole function doc comment has some wrapping inconsistencies).


gbj:
Thanks! The `COMMON_BUGS` example is just hopelessly out of date at this point and should probably be removed in any case, but I can do that separately.

PR #4514, opened 2025-12-29
Implement Params derive macro without nightly

luxalpa:
I was experimenting with trying to get my app rely less on nightly rust (as I am starting to get annoyed by the frequent ICE's).

I found a TODO that was asking to do the IntoParam implementation in a non-nightly way. Looking a bit around I found several methods of doing that. In the end, this one seemed to lead to the simplest implementation. It's based on the fact that the compiler will pick the inherent impl of a function over the trait impl.

Please note that my implementation also allows the user to override the default behavior by providing their own `IntoParam` implementations for their own types. I am unsure if that is desired or not, but we could change it in case it's not.

### DIFF
--- a/router/src/params.rs
+++ b/router/src/params.rs
@@ -202,8 +202,7 @@ where
 /// Helpers for the `Params` derive macro to allow specialization without nightly.
 pub mod macro_helpers {
     use crate::params::{IntoParam, ParamsError};
-    use std::str::FromStr;
-    use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


gbj:
Cool! Thanks. Yes, I think this implementation makes sense because it's being done inside a macro. I can confirm that this example code, which does not work on current stable, works with the PR
```rust
#[derive(Params)]
struct ExampleParams {
    id: String,
}
```

PR #4522, opened 2026-01-02
wip: Leptos 0.9

gbj:
Over time, a number of small changes have piled up that are semver-breaking but don't affect user code in significant ways. I'm opening a "Leptos 0.9" PR and `leptos_0.9` branch now for visibility. I don't anticipate significant migration needed for any of these, although publishing a new version always does create some ecosystem churn as dependencies need to update as well.


maruschin:
Saved me the day with the 0.9.0-alpha release, I couldn't figure out why the nested AtKeyed weren't working.

PR #4523, opened 2026-01-02
Signal API tweaks

gbj:
For 0.9, I'd like to adopt this suggestion from @zakstucke. This 
1) removes the `Fn()` implementation for `Signal<T>` (and `ArcSignal<T>`)
2) still allows the function-call syntax for them
3) allows both nightly and stable to have full support for the `IntoReactiveValue` trait, which allows converting any function type into a `Signal<_>` with `#[prop(into)]` (and no `Signal::derive()` wrapper)

Currently, this example code works on stable:
```rust
#[component]
fn App() -> impl IntoView {
    let (x, set_x) = signal(42);

    view! {
        <SomeComponent derived_value=move || x.get() * 2/>
    }
}

#[component]
fn SomeComponent(#[prop(into)] derived_value: Signal<i32>) {}
```
i.e., there is no longer a need for `derived_value=Signal::derive(move || x.get() * 2)`.

However, on nightly conflicting trait implementations made this impossible. This PR aligns the stable and nightly APIs here.

In exchange, it has to remove the `Fn()` implementation for the `Signal<_>` wrapper. It uses a deref trick to continue supporting `foo()` call syntax for `Signal<_>`, but it doesn't support the actual trait. This means a component like this can no longer take a `Signal<_>` directly. This can be worked around by manually dereferencing the `Signal<_>` wrapper.
```rust
#[component]
fn SomeComponentThatTakesFn(derived_value: impl Fn() -> i32) {}

// no longer works
<SomeComponentThatTakesFn derived_value=foo/>
// but we can deref the signal and it will work
<SomeComponentThatTakesFn derived_value=&*foo/>
```

### DIFF
--- a/reactive_graph/src/into_reactive_value.rs
+++ b/reactive_graph/src/into_reactive_value.rs
@@ -34,13 +34,27 @@ mod tests {
         let owner = Owner::new();
         owner.set();
 
-        #[cfg(not(feature = "nightly"))]
         let _: Signal<usize> = (|| 2).into_reactive_value();
         let _: Signal<usize, LocalStorage> = 2.into_reactive_value();
-        #[cfg(not(feature = "nightly"))]
         let _: Signal<usize, LocalStorage> = (|| 2).into_reactive_value();
         let _: Signal<String> = "str".into_reactive_value();
         let _: Signal<String, LocalStorage> = "str".into_reactive_value();
+
+        // Confirm doesn't affect nightly function syntax:
+        #[cfg(all(rustc_nightly, feature = "nightly"))]
+        {
+            let sig: Signal<usize> = Signal::stored(2).into_reactive_value();
+            assert_eq!(sig(), 2);
+        }
+
+        // Confirm can be used in more complex expressions:
+        {
+            use crate::traits::Get;
+            let a: Signal<usize> = (|| 2).into_reactive_value();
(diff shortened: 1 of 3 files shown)


zakstucke:
Awesome!

maccesch:
@gbj Does that mean that we will be able to use signals as functions on `stable`?

gbj:
@maccesch Yeah, Zak and I decided to add that *specifically* for Signal in order to allow `From<Fn() -> T>` for `Signal<T>` without conflicting implementations by removing the `Fn` implementation for `Signal` and adding the Deref-based approach instead. I'd initially avoided that for approach for signals in general, in the (naive?) hope that `fn_traits` would be merged sooner or later. It may be worth reconsidering now -- if we're using it for one type, perhaps we can use it for all of them? They should wouldn't `impl Fn` exactly, but could be called with `foo()` and deref to a `dyn Fn` with `*`.

Do you have thoughts? Would you like to have access to the function syntax on stable? If so it's certainly something we can move toward.

maccesch:
@gbj Thanks for the explanation!

Indeed I'd very much like to have the function-like API on stable. In production nightly can be a big headache so a while ago we switched all our internal stack to stable. 

And the function API is just nicer!

I guess since `impl Fn` can lead to ambiguities it might be preferable to have the Deref based approach even when the Fn trait is stabilized?

PR #4524, opened 2026-01-03
Fixed: Env inconsistent deserialization

### LINKED ISSUE #4511: Inconsistent deserialization of Leptos' Env enum
# Description

Deserialization of the `Env` enum (`leptos_config::Env`) is inconsistent depending on whether it is read from the Cargo.toml file or from the `LEPTOS_ENV` environment variable.

Normally, when `cargo-leptos` reads the `Cargo.toml` or if `main.rs` reads the same file via `get_configuration`, the `Env` is derived using [`env_from_str`](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L233), which transforms the input as lowercase before detecting whether it is `Env::PROD` (allowed strings: "prod", "production") or `Env::DEV` (allowed strings: "dev", "development").

However, when `get_configuration` is used in `main.rs` with a valid path, the environment variable are handled by [`get_config_from_str`](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L332) instead of [`get_config_from_env`](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L404), which is used when `get_configuration` is passed `None`.

Down the call stack, it uses `serde` to deserialize the values ([here in `get_config_from_str`](https://github.com/leptos-rs/leptos/blob/main/leptos_config/src/lib.rs#L373)) from the environment variables.

Serde is strict when it comes to deserialization: it only accepts same-case values as the variants from the targetted enum, therefore it will only allow the strings "PROD" and "DEV" as valid strings. If those are not provided, running a leptos app with `LEPTOS_ENV` set to something that would normally be valid (e.g. lowercase "prod") will result in the following panic message:

```
thread 'main' (180029) panicked at src/main.rs:11:54:
called `Result::unwrap()` on an `Err` value: ConfigError("enum Env does not have variant constructor prod for key `env`")
```

# Leptos Dependencies

```toml
leptos = { version = "0.8" }
leptos_axum = { version = "0.8", optional = true }
axum = { version = "0.8.0", optional = true }
```

# Bug reproduction steps

1. Clone `start-axum`
2. Change line 11 of `main.rs` to be `let conf = get_configuration(Some("Cargo.toml")).unwrap();`
3. Run `LEPTOS_ENV=prod cargo leptos serve`
4. Observe panic message in console

Also possible with `LEPTOS_ENV=production`, `LEPTOS_ENV=dev`, etc. as described in the description.

# Next steps

I would like to make a PR, but I also want to know which approach should be used. Should the implementation of `serde::Deserialize` be changed on the `Env` enum? Should the config sources be layered in a different way? (super unfamiliar with the crate `config`, so I'd appreciate if somebody knew if and how this could be done so we just replace the env layer with a call to `get_config_from_env()` or something of the kind).

I also don't mind if somebody wants to do the PR themselves, as this could be a good first issue for somebody looking to contribute to Leptos. (oc, i'm interested to contribute, but if somebody else wants to, I don't mind either)

maeldroem:
# Description

Fixes #4511

- Manually implemented `serde::Deserialize` to use `env_from_str()`
- Added test to prevent regression
- Extracted valid serialized values for `Env` into constants

### DIFF
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -226,9 +226,7 @@ pub const ENV_PROD_KEY_LONG: &str = "production";
 /// An enum that can be used to define the environment Leptos is running in.
 /// Setting this to the `PROD` variant will not include the WebSocket code for `cargo-leptos` watch mode.
 /// Defaults to `DEV`.
-#[derive(
-    Debug, Clone, serde::Serialize, PartialEq, Eq, Default,
-)]
+#[derive(Debug, Clone, serde::Serialize, PartialEq, Eq, Default)]
 pub enum Env {
     PROD,
     #[default]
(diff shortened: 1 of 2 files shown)


gbj:
Thanks!

PR #4526, opened 2026-01-05
Fix issue with missing documentation parts on `docs.rs`

ahlinc:
For example on the <https://docs.rs/leptos/0.8.15/leptos/mount/index.html> page there are no mentions of any `hydrate_*` functions because such functions require the `#cfg(feature = "hydrate")` definition to appear in docs.

Some lines in `Cargo.toml` files were reordered for uniformity and ability to quick review all docs.rs flags with a command like:
```
rg -A2 -g Cargo.toml package.metadata.docs.rs
```

### DIFF
--- a/integrations/actix/Cargo.toml
+++ b/integrations/actix/Cargo.toml
@@ -30,6 +30,7 @@ send_wrapper = { workspace = true, default-features = true }
 dashmap = { workspace = true, default-features = true }
 
 [package.metadata.docs.rs]
+all-features = true
 rustdoc-args = ["--generate-link-to-definition"]
 
 [features]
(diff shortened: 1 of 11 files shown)


gbj:
Thanks. As I understand it, we ideally want these `#[cfg_attr(docsrs, doc(cfg(feature = "...")))]` annotations on any conditionally-enabled function so that the nice little badges get generated in the docs; I don't think this gets done automatically so enabling all the features will show functions in docsrs without an indication that they need to be enabled special (I might be mistaken)

https://github.com/leptos-rs/leptos/blob/7bcaecfbc0e91b1f9ccae6f0fcd2522ec718826c/any_spawner/src/lib.rs#L177

[creates a nice badge](https://docs.rs/any_spawner/latest/any_spawner/struct.Executor.html#method.init_tokio)

I've done this in some of the smaller/newer utility crates but haven't gone through and done it everywhere.

I'm not saying you need to for this PR, but if anyone ever wants to it would be great!

ahlinc:
@gbj Thank you for the release with this PR!
Searching inside of the Leptos documentation became to be much comfortable now.

PR #4528, opened 2026-01-05
Break up `file_and_error_handler_with_context` into tower services

metatoaster:
The original implementation hasn't been removed to allow for this to be released as a patch release and also allow the decision on whether to deprecate the existing functions be made not here and now.

This implementation breaks up the existing implementation to permit the composition of the file fallback and the error handler with other tower services as they are now provided as such. My issue started from the inability to side-load additional static files using `tower_http::services::ServeDir` in a way that definitely won't override with the `.css`/`.js`/`.wasm` files from the application itself, and that even if I don't mind that, the existing handler cannot be used as a fallback for `ServeDir` as it's not a tower service.

This patch will effectively allow my specific use case and also other cases where composition with other tower services are required, illustrated with the example from my specific use case:

```rust
    let app = Router::new()
        })
        // ensure the site pkg routes are applied first before letting axum router's fallback logic
        .leptos_site_pkg_dir_route(&leptos_options)
        // search for assets located in some other filesystem location, and only when nothing is
        // found before the leptos shell is use to render some error message as the final fallback
        .fallback_service(
            ServeDir::new(media_root)
                .fallback(ErrorHandler::new(shell, leptos_options.clone()))
        )
        .with_state(leptos_options);
```

I have not ported over the logic for dealing with `ResponseOptions` introduced by #4394 as the logic only deals with responses, and given the use case are for headers this can be dealt with using tower's [`tower_http::set_header::SetResponseHeader`](https://docs.rs/tower-http/latest/tower_http/set_header/struct.SetResponseHeader.html) middleware, or well, the existing handlers.  The goal of this patch is to provide an alternative that plays nice/integrates well with the tower ecosystem at large.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,12 +1,53 @@
+experimental = ["setup-scripts"]
+
+# For tests that need to run one at a time to avoid resource contention from
+# parallel execution.
 [test-groups]
 serial-integration = { max-threads = 1 }
 
-# Given the inability to ensure all tests must run in the same process (like how `cargo test` does it),
-# the next best thing is to seralize it for the axum integration test.
+# The built-in `fn build_test_service` in the test runs a single instance of
+# `cargo leptos build` guaranteed with `std::sync::Once`.  However,
+# `cargo-nexttest` does not have the ability to group tests under the same
+# process to emulate what `cargo test` does, thus resulting in competing
+# `wasm-bindgen` executions that will clobber each other, breaking the tests.
+# While serializing the tests was done at one point as a quick workaround,
+# using setup-scripts is the better approach.
 #
 # References:
 # - https://github.com/nextest-rs/nextest/issues/27
 # - https://github.com/nextest-rs/nextest/issues/209
-[[profile.default.overrides]]
-filter = 'binary_id(leptos_axum::axum_integration)'
-test-group = 'serial-integration'
+#
+# [[profile.default.overrides]]
+# filter = "binary_id(leptos_axum::axum_integration)"
+# test-group = "serial-integration"
+
+# Ideally we want this to work to allow for as much as (relatively) cross-
+# platform compatibility via cargo, but there doesn't appear to be a way to
+# apply environment variables or otherwise set `LEPTOS_OUTPUT_NAME` within
+# this section.
+#
+# ```
+# [scripts.setup.axum_integration_service_mode]
+# command = ["cargo", "leptos", "--manifest-path", "integrations/axum/tests/service_mode/Cargo.toml", "build"]
+#
+# [[profile.default.scripts]]
+# filter = "binary_id(leptos_axum::axum_integration)"
+# setup = ["axum_integration_service_mode"]
+# ```
+#
+# Instead, we have to do the platform specific route...
+[scripts.setup.axum_integration_service_mode_sh]
+command = ["./scripts/axum_integration_service_mode.sh"]
+
+[scripts.setup.axum_integration_service_mode_bat]
+command = ["cmd", "/c", "scripts\\axum_integration_service_mode.bat"]
+
+[[profile.default.scripts]]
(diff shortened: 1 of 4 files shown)


metatoaster:
Ha, I forgot about possible downstream trait implementation of `leptos_axum::LeptosRoutes` that may happen.  Maybe it might be a good idea to [seal the trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) as part of the next semver release if we want to make it possible to add methods in the future without causing a major version bump, unless for whatever reason we want to allow downstream users to implement custom versions of it against newtypes of the `axum::Router`.  Let me know if you also want to seal the trait so I can do this, and/or I can change it so that I don't touch the trait for now to not break semver so this may be released faster.  Let me know.

metatoaster:
Thinking about the handling of logic for dealing with ResponseOptions, perhaps this could also be turned into a tower middleware, which then may be applied onto the router against desired routes and fallback, which may end up being more consistent and idiomatic with how tower is supposed to work.  Going down this path is a much more drastic change, and is probably more suitable in a completely separate patch.

gbj:
Thanks! I'd say: if it's going to be breaking either way, design what you think the best API is and we'll merge it into the `leptos_0.9` branch instead. I haven't had time at the moment to read through it. Let me know if you'd like to do more (and possibly breaking) work on it right now, or if you want me to review as is.

metatoaster:
Adding a new method would be the best way and had the trait been sealed in the first place this wouldn't have been a breaking change, and so I also include that here.

In any case it is ready for review.

metatoaster:
Pulling back a bit, it should be possible to merge this to main without f7d1762 and 61d71b7 and merge f7d1762 and 61d71b7 into `leptos_0.9`.  Basically the new helpers are just additional non-breaking changes with only the new trait method being breaking but not critical to the working of this.

I will be updating this pull request and create a new one to reflect this.

metatoaster:
Hopefully this all works now without breaking changes in this pull request, see #4535 for the breaking changes, as I shuffled and remerged all the patches to where they are relevant.

metatoaster:
Now that I have more time I am noticing that the tests for integrations/axum does not appear to test against the default features, which means the one doc test I wrote for service did not appear to show up.  I manually added `default` as `extra-features` to test via `cargo-all-features` and it did the job locally so I am pushing this to see if it works for CI.

I also want to add a test for fetching the actual artifacts to make sure this all really works, and I might want to add integration tests directly to the `leptos_axum` package, unless you want me to stuff this back into `examples`.  Let me know what you think.

Oh hmm, looking at what I've done again, maybe it doesn't have to be gated behind `default` feature, as the original error handler only got that simply due to it being coupled with `ServeDir` which deals with filesystem calls but with it decoupled it might be possible to provide this error service?  I am not familiar with how wasm on the server is set up so also please do let me know if I should remove the feature gate.

metatoaster:
This PR is definitely ready for a proper review now that I got all the lower level integration tests in and I am finally happy with what's been done, ~~aside from the configuration added to `cargo-nextest` which is more of a punt rather than being thorough, though I am trying to keep it focused so cleaning up how this setup is done may be done later.~~ I RTFM'd and ensured what I've done and the usage of `cargo-nextest` is well integrated and correct as per the recommendations suggested by the documentation from that excellent test runner.

gbj:
This looks great, thanks! I am not sure what's up with the failing router example tests but that is a CSR example so I can't imagine it's related to this.

metatoaster:
Yeah that is strange though I would assume it might have something to do with contention with test runner resource with parallel browser runs, assuming playwright does that. Moreover, that same failing test has passed previously and subsequent changes only affected testing for `leptos_axum` so yeah the failure is unrelated.

Thanks for merging the PR! I'll rebase the breaking changes later.

PR #4535, opened 2026-01-10
Full change to #4528

metatoaster:
See #4528 for the discussion, but the intent here is to split out the final two commits which include the breaking changes.

The new method will benefit from at least a test or two to verify that the route providing the assets remains working in the future, I'll get this in at some point.

Making this a draft request until the other non-breaking changes are merged, unless it was decided to do a full breaking change.


metatoaster:
This change now provides the following shortcut methods for doing the `site_pkg_dir` and the error fallback:

```rust
Router::new()
    })
    // this:
    .leptos_route_site_pkg_dir(&leptos_options, shell)
    .leptos_route_fallback(&leptos_options, shell)
    // instead of this:
    // .fallback(leptos_axum::file_and_error_handler(shell))
    .with_state(leptos_options),
```

Naturally, `leptos_route_fallback` isn't a strictly necessary addition as it's a shortcut for

```rust
    .fallback_service(ErrorHandler::new(shell, leptos_options))
```

Naturally we could go one step further to offer another method that combine the two to effectively replicate `file_and_error_handler` in one step, but maybe that's at the point where adopting the service builder pattern to customize each and every aspect of all the `.leptos_routes*` methods while also avoid repeating `leptos_options` and `shell` everywhere.  I am open to doing this if that's desirable.

metatoaster:
I've been thinking about the axum integration (and actix to a much smaller extent), I think given how patterns have started to emerge I think I have a better idea to formalize everything into a configuration builder to get the `axum::Router` all set up in much less boilerplate for the default case with a configuration builder.

Moreover, now that I've stepped back out to take a grander view of this integration, I found that the typical `shell` (from the `leptos-rs/start-axum` template) and the `file_and_error_handler` are actually somewhat coupled when they really shouldn't be, as the `shell` that the template provides expects a `LeptosOptions` which feeds into `file_and_error_handler` as that also expects this, while the entire time the `LeptosRoutes::leptos_routes` and friends allow all kinds of `S` as long as it satisfies the bound `LeptosOptions: FromRef<S>`. Hence I pushed a change that cleaned up `ErrorHandler` so that it also doesn't require `LeptosOptions` at all.

Backtracking to the configuration builder, perhaps it may be beneficial to formalize that and provide documentation to show that by using the existing state can serve as an `AppState` that tracks additional common configuration values (e.g. analytics/firebase tokens), how to make that `AppState` usable (ensuring `LeptosOptions: FromRef<S>`) and that avoids having to use `leptos_routes_with_context`.

To expand a bit with an example:

In `app.rs`:
```rust
#[derive(Clone)]
#[cfg_attr(feature = "ssr", derive(axum::extract::FromRef))]
pub struct MyState { 
    pub leptos_options: LeptosOptions,
    // doesn't matter what this is, these are additional States for the router
    pub token: Token,
} 
        
pub fn shell(state: MyState) -> impl IntoView {
    todo!()
}
```

In `main.rs`:
```rust
async fn main() {
    let my_state = MyState { leptos_options, .. };
    let routes = generate_route_list(App);

    let app = Router::new()
        .leptos_routes(&my_state, routes, {
            let my_state = my_state.clone();
            move || shell(my_state.clone())
        })
        .fallback_service(ErrorHandler::new(shell, my_state));
        .with_state(my_state);
    todo!()
}
```

The configuration helper I am thinking of could simply configure the `App`, `S` (default to `LeptosOptions`), the `shell`, and then an optional extra context, and it might end up looking like (name pending):

```rust
let config = LeptosAxumConfig::default()  // or `LeptosAxumConfig::new()` to disable defaults
    .app(App)
    .shell(shell);
    // this could even be optional as `get_configuration` could be called automatically
    // .state(leptos_options);
let app = Router::new()
    .leptos_config(config);
```

With that we don't need to worry about calling `generate_route_list(App)` and setting up the default file handler and error handler, though naturally that should be possible to be toggled.  The idea of this helper is to have less typing going forward, it will also basically use the existing configuration methods as they are working fine still.

metatoaster:
I went ahead and implement the configuration. Since the configuration uses generics inside optional fields it gets slightly tricky with the trait bounds, but thankfully it isn't an insurmountable barrier.  What I got so far:

```rust
let app = Router::new().leptos_route_configure(
    leptos_axum::RouterConfiguration::new()
        .app(App)
        .shell(shell)
        .state(leptos_options)
        .with_context(|| {})  // optional
);
```

And that's it.  The additional context provided using `.with_context` will apply to all the routes _and_ the `ErrorHandler` without having to repeat oneself. The application of that to the site pkg file service will require an additional piece of middleware which I think should be the approach, and this middleware will be generic enough that it should be usable with other tower/axum services (even though they will have a more straightforward way for the intended use case, this additional context is too good of a hammer while lowering cognitive load for Leptos developers who are only familiar with the Leptos tooling).

What I had also hoped to add was the ability to have `.state` be optional if it's a `LeptosOptions`, but this proved difficult and complicated with duplicate generic constraints so I left this out, and nevermind that `LeptosOptions` typically will have other fields that are needed alongside with the Router setup so it's typically instantiated by the developer anyway.

metatoaster:
The additional middleware `AdditionalContext` is now provided and that also is applied to the configuration for the `ServeDir` handler for site pkg for the application of configured headers, though not for setting it for the underlying service itself as there isn't a currently reported requirement to do so, but I think this at least serves as a start to potentially achieve that when that's required.

This whole pull request is definitely ready for review now.

I also note that the documentation right now for the new types are rather sparse, this will be amended in due time.

gbj:
I don't understand the CI failure here -- I don't think it's happening on main. Could you try merging main in and we'll rerun the CI?

gbj:
(Oh I meant to say but lost the comment in a separate tab: Thanks so much for all your work! This looks like a great API and it all makes sense to me.)

metatoaster:
I ended up rebasing the new set of changes onto main as I anticipate a bit further work on documentation and possible doctests, but hopefully whatever broke should work.  If some other request end up merging some potentially later main in again I can rebase as necessary.

Also, thanks for looking into this!

metatoaster:
Looks like it did work for real this time after #4591 provided the fix.

That said I do want to look into providing the reactive context for real for actual tower/axum services, rather than just calling the day by patching the headers like what that file and error handler needed.  Also the documentation still need updating. Though by and large the intent of what I originally set out to do, which is providing feature parity replacement for the `file_and_error_handler_with_context` function but in a fully decoupled manner through some underlying tower service, so a thorough code review may be done in that particular light if desired.

Thinking about this some more, I think it really can be implemented to function as the middleware for all server-function-like handlers. At a casual glance it looks to be possible, basically it would be a `Service` with `type Future = ScopedFuture`, though I will have to see how this might be constructed.

metatoaster:
Okay, I did not end up changing the `type Future` bit as that particular optimization isn't necessary as it will require constructing a specific typed future, so for now I just followed the approach taken by `handle_server_fns_inner` but only deal with providing the contexts that should be available. It may benefit from not being copy/pasted but it will suffice for now, as that can be optimized later when necessary.

With this change, generic tower services can leverage `LeptosContextLayer` separately to set up the middleware so that it can access the same contexts available to server functions and any other additional contexts that are provided, though minus the `Location` header handling as that's too specific to the actual server functions (also that the user can just use the `ResponseOptions` context to do what they need).

I am happy for this to be reviewed completely and am happy if it's merged as is ~~Edit: pending cleaning up of documentation for new items~~.  Though if we're feeling more adventurous (well, elimination of effectively duplicate feature) we can try replacing the existing `file_and_error_handler_with_context` with a one-shot call to this new service stack, but I am as equally happy to leave that alone for now, too.

gbj:
Looks great -- thank you!

PR #4536, opened 2026-01-11
feat: Allow customizing the status code of server fn error responses

spencewenski:
Currently, all server function error responses have status code 500. This is not ideal, especially when it comes to observability. If every error response has status 500, then it's difficult to see which errors are client errors (404, unauthorized access attempts, etc) vs which ones are actual internal server errors that need to be investigated.

This PR builds on https://github.com/leptos-rs/leptos/pull/4249 to allow customizing the http status code for server function error responses.


gbj:
Looks good, thanks!

PR #4549, opened 2026-01-24
Add closedby attribute to HtmlDialogElement

bngo92:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy


gbj:
Thanks!

Issue #4403, opened 2025-10-22
`to_map` version of `Params`


kyrias:
**Is your feature request related to a problem? Please describe.**
It would be nice if there was a way to easily generate the query string for a given route based on the `Params` struct that view uses to deserialize the query string, to prevent both typos and accidentally using keys that might not exist, especially while refactoring.

**Describe the solution you'd like**
Either on `Params` or in a new trait matching it I'd like to have a `to_map` method returning a `ParamsMap`, the opposite of the `Params::from_map` method, which would then let the user call `ParamsMap::to_query_string` to build an appropriate query string for linking the user to.

**Describe alternatives you've considered**
Manually constructing the `ParamsMap` which is what I'm doing now, but it easily leads to mistakes that are only caught at runtime when exercising that code path.

PR #4553, opened 2026-01-29
Adding feature for easy conversion from Params to ParamsMap

cerdelen:
Feature requested in #4403. This new to_map function easily lets you convert from a Params struct to a ParamsMap struct which decreases typo errors and is a QOL improvement.

Edit: This assumes that all fields of a struct are Options<T> where T: ToString. If this is not sufficient one would have to extend the ::leptos_router::params::macro_helpers::Wrapper type to also provide __from_param().

If this is wanted i can look into that one to, if this only working for structs with Fields Option<T: ToString> is suficient no further work required.

Fix: #4403

### DIFF
--- a/leptos_macro/src/params.rs
+++ b/leptos_macro/src/params.rs
@@ -56,7 +56,7 @@ pub fn params_impl(ast: &syn::DeriveInput) -> proc_macro::TokenStream {
 				let ty = &field.ty;
 				let span = field.span();
 				quote_spanned! {
-                    span=> if let Some(val) = ::leptos_router::params::macro_helpers::Wrapper::<#ty>::__from_param(&self.#ident, #field_name_string) {
+                    span=> if let Some(val) = ::leptos_router::params::macro_helpers::Wrapper::<#ty>::__param_to_string(&self.#ident, #field_name_string) {
                         map.insert(#field_name_string, val);
                     }
 				}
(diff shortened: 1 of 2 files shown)


gbj (commented):
Thanks for the PR! I left some nit comments as I went through. However, I think there's a slightly bigger issue: params are not required to be `Option<_>`. For example this currently compiles on the main branch but fails with this PR:
```rust
#[derive(Params)]
struct MyParams {
    foo: i32,
    bar: usize,
    baz: String,
}
```

cerdelen:
Oh ... interesting ... will look into why T doesn't work and only works with Option T.

Will also clean up the docu placeholders.

gbj:
Thanks! This looks good. `cargo-semver-checks` is rightly warning us that it is a breaking change because it adds a public method to a trait (`Params::to_map()`) without a default implementation. I think there's probably not a sensible default implementation here (I wouldn't necessarily want to default to an empty map...) but I am happy to merge this into the `leptos_0.9` branch where we are collecting breaking changes. Does that work for you?

cerdelen:
Yes i was expecting this to be a breaking change for anyone who implements Params manually and not by the derive macro.

For people that use the macro nothing will change.

gbj:
Sounds good! I will merge it to the 0.9 branch.

PR #4554, opened 2026-01-30
feat: Implement InnerHtmlValue for Oco<'static, str>

pheki:
Hi!

I have a type that implements `IntoRender<Output=Signal<Oco<'static, str>>`, but I can't use it directly in `inner_html` as `InnerHtmlValue` is not implemented for `Oco`.

I mostly copied `Arc<str>` / `&str`'s implementation, adapting it to Oco based on how it was done for the other traits (mostly by calling `upgrade_inplace()`.

Originally I implemented it for `'_`, but I realized we'd then need to allocate for CloneableOwned even for `&'static` references instead of just copying it, so I kept it only for `'static` like the other traits.

### DIFF
--- a/tachys/src/oco.rs
+++ b/tachys/src/oco.rs
@@ -2,6 +2,7 @@ use crate::{
     html::{
         attribute::{any_attribute::AnyAttribute, AttributeValue},
         class::IntoClass,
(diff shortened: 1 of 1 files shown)


gbj:
Thank you!

PR #4563, opened 2026-02-06
Remove `linear-map` dependency

alexisfontaine:


gbj:
These both look totally reasonable to me. Thanks!

PR #4565, opened 2026-02-06
add IntoIterator implementantion for &ParamsMap

jonnyso:
When manually implementing `Params` the trait only gives access to a reference of `ParamsMap`, for convenience this PR adds an `IntoIterator` implementation for `&ParamsMap` as well as a helper function `get_iter` that returns an iterator over the most-recently-added values for every param on the map.

### DIFF
--- a/router/src/params.rs
+++ b/router/src/params.rs
@@ -66,7 +66,7 @@ impl ParamsMap {
     }
 
     /// Gets an iterator for all the most-recently-added values on the map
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! This looks good overall. I am wondering about the name `.get_iter()` though. The other `.get()` methods take a key, and return the values for that key. This returns the (latest) values for all keys. I wonder whether something like `.latest_values()` may make it clearer what it does.

PR #4567, opened 2026-02-07
Warn when trying to update a signal with an active read guard

MarkusG:
Using `reactive_stores`, I had the following code
```rs
let value = t.value().read();

other_t.value().set(value.clone());
```
where `t` is a value of a custom struct type with store field `value`. This would silently fail to update the value of `value`.

This PR adds code to log a warning on a call to `set()` when the the update fails and the signal is not disposed.

### DIFF
--- a/reactive_graph/src/traits.rs
+++ b/reactive_graph/src/traits.rs
@@ -545,11 +545,13 @@ where
 
     #[track_caller]
     fn set(&self, value: Self::Value) {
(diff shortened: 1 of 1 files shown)


gbj:
Thanks! Could you gate this so that it's not included in release builds? That's what we do for similar debug warnings elsewhere. They're marked with `#[cfg(any(debug_assertions, leptos_debuginfo))]`

MarkusG:
@gbj done

gbj:
Thanks!

Issue #4548, opened 2026-01-24
Adopt policy to avoid bumping patch versions?


gbj:
Currently, all our dependencies are specified in a [workspace Cargo.toml](https://github.com/leptos-rs/leptos/blob/main/Cargo.toml).

As of #3987, these are specified precisely, to the patch version, in order to be able to use `cargo-minimal-versions` to ensure that we specify the correct minimal version for dependencies within the set of leptos packages.

One downside of this is that it leads to both dependabot and human contributors bumping patch versions in order to "update" our dependencies. In the context of a library like this, however, the primary effect of these updates is to constrain users unnecessarily.

For example, currently all Leptos packages [specify a dependency on `wasm-bindgen 0.2.106`](https://github.com/leptos-rs/leptos/blob/main/Cargo.toml#L82). This is the result of a version bumped group in with other, necessary updates in #4497. For users with `wasm-bindgen` pinned to an earlier version for any reason, this meant that their builds no longer worked, and added the time-consuming work of hunting down the correct `leptos` patch version that supported the necessary version of `wasm-bindgen`.

After a conversation with them I'm wondering whether a better policy would be to *avoid* updating our own workspace dependencies, but to keep them at their *minimal* versions. Users can still, of course, consume the latest patches in their own applications.

@sabify, you know much more about both `cargo-minimal-versions` and dependabot than I do, I think. I'm wondering if you could share thoughts on what effects this change might have that I'm not thinking of.

luxalpa:
For my own libraries, I try to keep the dependency specification as minimal as possible and I consider updating them always a breaking change. Especially when building for the web, binary size matters, so it's important I think to give the developer the maximum ability to deduplicate their dependencies. Also stability matters too - upgrading a dependency even on patch level usually requires some level of testing in a large project (my current one has 26,000 lines of code).

PR #4568, opened 2026-02-07
feat: rely on dependencies major versions with minor bounds

sabify:
Closes: #4548.

What's changed:

- Relaxing patch segment in our dependency versions. By doing this, we are still respect the minor bound and will never break the major bound, giving users flexibility in choosing and pinning the right versions. 
   For example: **`dep = "0.1"`** has minimum version 0.1.0, accepts `>=0.1.0, <0.2.0`; And **`dep = "1.2"`** has minimum version 1.2.0, accepts `>=1.2.0, <2.0.0`.
- We still keep `Cargo.lock` to make our CI catch new releases, even patches, and test against them. This will be done by utilizing `dependabot` and `cargo-minimal-versions`. Note that `Cargo.lock` is not part of releases/publishes.

### DIFF
--- a/examples/lazy_routes/e2e/Makefile.toml
+++ b/examples/lazy_routes/e2e/Makefile.toml
@@ -15,6 +15,8 @@ args = [
   "--",
   "--retry",
   "2",
+  "--retry-after",
+  "5sec",
   "--fail-fast",
   "${@}",
 ]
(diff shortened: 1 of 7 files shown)


sabify:
Is the failed CI related to this PR?

@gbj  Can you please re-run the failed CI?

gbj:
Thank you very much! I'm really grateful for your work.

QuartzLibrary:
Small note: this introduced a breaking change by bumping `reqwest` from `0.12` to `0.13` and `wasm-streams` from `0.4` to `0.5`.
See #4651 for more info.

PR #4569, opened 2026-02-09
fix: Strip trailing `_` to fix inert SVG `use_` elements

### LINKED ISSUE #4555: SVG `use_` does not work for inert elements anymore (0.8.3 regression)
**Describe the bug**

SVG `use_` will insert a `use_` element instead of a `use` element. It used to work fine up to 0.8.2. I found that after updating from 0.8.2 to 0.8.15 one of my SVGs stopped showing correctly.

**Leptos Dependencies**

```toml
leptos = { version = "0.8.15", features = ["csr"] }
```

**To Reproduce**
1. Create an example app following the [Getting Started](https://book.leptos.dev/getting_started/index.html) tutorial from the book
2. Replace `src/main.rs` with the following:

```rust
use leptos::prelude::*;

fn main() {
    leptos::mount::mount_to_body(|| view! { <MyOwnSvg /> })
}

#[component]
fn MyOwnSvg() -> impl IntoView {
    view! {
        <svg
            width="200"
            viewBox="0 0 200 400"
            id="lamp"
            xmlns="http://www.w3.org/2000/svg"
            xmlns:xlink="http://www.w3.org/1999/xlink"
        >
            <rect id="example" x="12" y="10" width="140" height="32" stroke="black" fill="transparent" />
            // An inert `<use_ />` component is rendered as `<use_ />` instead of `<use>`, so this won't show up.
            <use_ y="100" href="#example" />
            // This is not inert, so it works
            {
                move || view! {
                    <use_ y="200" href="#example" />
                }
            }
            // This is not inert, so it works
            <use_ y=300 href="#example" />
        </svg>
    }
}
```

4. Open http://localhost.:8080/ in your browser (tested on both Firefox and Chromium)
5. Check that the HTML includes `<use_>` instead of `<use>`

**Next Steps**

- [ ] I will make a PR
- [X] TBH I'm not sure, I do think this is important as a regression that silently breaks SVGs, but I don't know the complexity of it to make a decision. I think this should be at least known before discussing next steps.

**Additional context**

Bisection result:

```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
50a4c3b0d918680d5cacce11ac992389acfba5ff
30b0a579cad0765e9cc5c62739785cb155fe6154
c76649d77b1be71e84fd04821d0edf06f2d90294
df09d4a7f6de4e8a2279bb19dac8df4403629225
We cannot bisect more!
```

I skipped commits that failed to compile.

PRs that those commits are from: https://github.com/leptos-rs/leptos/pull/4085, https://github.com/leptos-rs/leptos/pull/4100, https://github.com/leptos-rs/leptos/pull/4051 (improbable), https://github.com/leptos-rs/leptos/pull/4072 (very improbable).

Edit: updated reproduction and title, clarifying that the issue is specific to inert elements

pheki:
Fixes #4555

I also updated the issue reproduction to highlight the difference between inert vs non-inert elements

### DIFF
--- a/leptos_macro/src/view/mod.rs
+++ b/leptos_macro/src/view/mod.rs
@@ -444,6 +444,12 @@ fn inert_svg_element_to_tokens(
                     Node::Element(node) => {
                         let self_closing = is_self_closing(node);
                         let el_name = node.name().to_string();
+                        // strip trailing underscores, for identifiers such as SVG use_
+                        let el_name = el_name
+                            .strip_suffix('_')
+                            .map(str::to_string)
+                            .unwrap_or(el_name);
+
                         let escape = el_name != "script"
                             && el_name != "style"
                             && el_name != "textarea";


gbj:
Thanks!

PR #4579, opened 2026-02-16
Fix SSR Suspense livelock when multiple boundaries share Resources via .get()

alilee:
## Summary

One-line fix for an SSR livelock where 2+ `Suspense`/`Transition` boundaries sharing 2+ Resources via `.get()` causes the Suspense Effect to endlessly re-run `dry_resolve()`, consuming 100% CPU (~297k closure invocations per boundary per 5s) and never completing SSR.

Manifests on AWS Lambda (128MB, 2 cores → `worker_threads=2`) as intermittent 500 errors.

## Root Cause

The Effect's isomorphic closure returns a `bool` that becomes `double_checking: Option<bool>`. After `dry_resolve()` returns `true` (setting `double_checking = Some(true)`), the Effect is woken by individual monitor completions. If tasks aren't empty yet, the catch-all `false` at the end resets `double_checking` to `Some(false)`. When tasks finally drain, the Effect sees `double_checking == Some(false)` and calls `dry_resolve()` again — infinite cycle.

```
dry_resolve → true → double_checking = Some(true)
  monitor1 completes → Effect wakes → tasks NOT empty → returns false (RESETS!)
  monitor2 completes → Effect wakes → tasks empty + double_checking == Some(false) → dry_resolve again!
```

## The Fix

`leptos/src/suspense_component.rs`, line 409:

```diff
-                false
+                double_checking.unwrap_or(false)
```

This preserves `Some(true)` through intermediate wakeups so the completion branch is reached:

```
dry_resolve → true → double_checking = Some(true)
  monitor1 completes → Effect wakes → tasks NOT empty → returns true (PRESERVED!)
  monitor2 completes → Effect wakes → tasks empty + double_checking == Some(true) → COMPLETION
```

Only the `Some(true)` → non-empty-tasks path changes behavior. All other paths return the same value as before.

## Reproduction

Minimal repro at https://github.com/alilee/feedback/tree/main/bugs/ssr-race (Cargo.toml + src/main.rs).

```bash
cargo run --features ssr  # uses worker_threads = 2
curl -s -m 5 http://localhost:3000/broken | grep -c RESOLVED_RESOURCES  # 0 (livelock)
curl -s -m 5 http://localhost:3000/works  | grep -c RESOLVED_RESOURCES  # 1 (Suspend::new workaround)
```

Thread count dependency:
| `worker_threads` | Result |
|---|---|
| 1 | Works |
| **2-3** | **Livelocks** |
| 4+ | Works |

## Testing

- New regression test: `leptos/tests/ssr_suspense_livelock.rs` — 2 Suspense boundaries, 2 shared Resources, `worker_threads=2`, 2s timeout
- All existing `leptos` crate tests pass (72 tests: 58 doc + 11 ssr + 2 pr_4061 + 1 new)
- Test fails without the fix, passes with it

### DIFF
--- a/leptos/src/suspense_component.rs
+++ b/leptos/src/suspense_component.rs
@@ -406,7 +406,7 @@ where
                         tasks.track();
                     }
                 }
-                false
+                double_checking.unwrap_or(false)
             }
         });
 
--- a/leptos/tests/ssr_suspense_livelock.rs
+++ b/leptos/tests/ssr_suspense_livelock.rs
@@ -1,0 +1,62 @@
+//! Regression test for SSR livelock when multiple Suspense boundaries share
+//! Resources via `.get()`.
+//!
+//! Root cause: The Suspense Effect's state machine (in suspense_component.rs)
+//! returned `false` as a catch-all, resetting `double_checking` from `Some(true)`
+//! to `Some(false)` when the Effect saw non-empty tasks between individual monitor
+//! completions. This caused `dry_resolve()` to be called again when tasks drained,
+//! spawning new monitors in an infinite cycle.
+//!
+//! The fix: replace the catch-all `false` with `double_checking.unwrap_or(false)`,
+//! preserving the `Some(true)` state through intermediate Effect wakeups.
+
+#[cfg(feature = "ssr")]
+mod imports {
+    pub use any_spawner::Executor;
+    pub use futures::StreamExt;
+    pub use leptos::prelude::*;
+}
+
+/// Two Suspense boundaries sharing two Resources via `.get()` must resolve
+/// within a reasonable time. Previously this livelocked with worker_threads=2-3
+/// due to the Suspense Effect's dry_resolve re-registration cycle.
+#[cfg(feature = "ssr")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn shared_resources_across_suspense_boundaries_no_livelock() {
+    use imports::*;
+
+    _ = Executor::init_tokio();
+    let owner = Owner::new();
+    owner.set();
+
+    let r_a = Resource::new(|| (), |_| async { "a".to_string() });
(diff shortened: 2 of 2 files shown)


alilee:
I've tried to run the regression test suite, but I'm not familiar with the ci so I'm not confident I've fully leveraged the available testing - apologies in advance. Happy to do more work as you direct. 

I think you might want to get ready for an avalanche of Claude Code originated PR's and set ultra-convenient standards for yourself, and maybe even provide prompts or agent commands to do what you need so that you can sit on the top - surgeon to all us nurses. You should feel completely justified in asking for people to invest their tokens on tasks! I really hope the shift is net beneficial for you and leptos.

gbj:
I'm having trouble reproducing the bug, either against the main branch with the reproduction you provided in the issue or in the PR branch by reverting the change and running the test. The `alilee/feedback` repo linked in this PR appears to be private (or at least it 404s for me). Could you share that with me, and I'll see if I can reproduce it given that?

gbj:
Ah ok, interesting. I am able to reproduce it against the latest published version, but not again git main. Could you confirm whether the bug occurs against the current state of the main branch? I am about to publish a new set of patch releases in any case.

alilee:
Yep can do. Thanks!

alilee:
Fixed in v0.8.16. Thanks!

All 4 routes tested against leptos-rs/leptos main (commit 4eb4f27f, leptos 0.8.16):

  - /broken — works (was deadlocking on v0.8.15)
  - /hyp2 — works, only 3 closure invocations per boundary (was ~297,000)
  - /hyp3 — works (3 resources, was deadlocking)
  - /works — works (control, always worked)

Tested /broken 10 consecutive times — 10/10 success.

This means the fix was incorporated into upstream changes after v0.8.15, and we should be able to drop our Leptos fork once we upgrade to a release that includes commit 4eb4f27f.

PR #4583, opened 2026-02-17
fix: Remove the dependency on `::std` in `either_of`

DanikVitek:
Since the MSRV is 1.88, the `Error` trait is now accessible in `::core`, therefore the dependency on `::std` can be removed completely

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1079,7 +1079,7 @@ checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
 name = "either_of"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "paste",
  "pin-project-lite",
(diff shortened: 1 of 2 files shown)


DanikVitek:
@gbj Hi. This one is technically a breaking change due to the `no_std` feature being removed, but as the crate is v0.1 it could either be bumped to v0.2, or v1.0 (as API-stable)

gbj:
Fair enough; some of its types are re-exported in other packages, though, so bumping its minor version would (I believe) requiring bumping their versions as well. We could also simply leave the `no_std` feature in place. It would be a no-op right now, but would allow for future `std` use without further breaking changes.

gbj:
Awesome, thanks!

DanikVitek:
> some of its types are re-exported in other packages, though, so bumping its minor version would (I believe) requiring bumping their versions as well

@gbj Should then there be a PR opened against the `leptos_0.9` branch which removes the `no_std` feature, or leave it be?

gbj:
I'd suggest leaving it be -- unless there's a reason you can think of that a `no_std` feature is bad, this leaves the flexibility to add other things that require `std` in the future and already have them gated.

PR #4587, opened 2026-02-20
fix: always create client-side values in tachys constructors regardless of ssr feature

DrudgeDance:
## What this fixes

When deploying a CSR WASM app in a workspace where another crate enables `leptos/ssr`, Cargo feature unification activates `tachys/ssr` globally. The `on()`, `directive()`, and `prop()` constructors then create `None` values via `(!cfg!(feature = "ssr")).then(...)`, causing every event handler, directive, and property to panic at runtime with opaque messages like `"callback removed before attaching"`.

This PR makes the constructors always create `Some(...)` values. The SSR path (`to_html()`) never accesses them — it writes nothing for events, directives, and properties — so the values are harmlessly dropped.

Related issue: Fixes #4586

### Always create client-side values (core fix)

```diff
 // tachys/src/html/event.rs
 pub fn on<E, F>(event: E, cb: F) -> On<E, F> {
     On {
         event,
-        cb: (!cfg!(feature = "ssr")).then(|| SendWrapper::new(cb)),
+        cb: Some(SendWrapper::new(cb)),
     }
 }
```

Same one-line change in [`directive.rs`](https://github.com/DrudgeDance/leptos/commit/2d535dd) and [`property.rs`](https://github.com/DrudgeDance/leptos/commit/2d535dd).

### Compile-time feature conflict guards

Added `compile_error!` to `leptos/src/lib.rs` for `csr`+`ssr`, `csr`+`hydrate`, and `ssr`+`hydrate` — with an actionable message pointing the user to `cargo tree -e features -i leptos`.

### Actionable diagnostic messages

Replaced opaque `.expect("callback removed before attaching")` with a shared diagnostic constant explaining the SSR feature unification cause.

### Regression tests

15 `wasm_bindgen_test` tests covering all 3 types × 4 code paths (build, hydrate, rebuild, to_html):

```
running 15 tests ... test result: ok. 15 passed; 0 failed
```

### CI workflow

`feature-safety.yml` — runs WASM tests with `ssr` feature + verifies all `compile_error!` guards fire.

## Reproduction

Three-variant standalone reproduction: **[DrudgeDance/contributions/leptos/fix/always-create-client-values](https://github.com/DrudgeDance/contributions/tree/main/leptos/fix/always-create-client-values)**

```bash
git clone https://github.com/DrudgeDance/contributions.git && cd contributions/leptos/fix/always-create-client-values
cd issues/clean  && trunk serve --port 8100  # ✅ works
[...]

PR #4588, opened 2026-02-20
fix(tachys): replace opaque panic messages with actionable SSR feature diagnostic

### LINKED ISSUE #4586: Cargo feature unification of tachys/ssr silently panics all event handlers, directives, and properties in CSR apps
**Describe the bug**

When the `ssr` feature is activated on `tachys` — even in a CSR-only WASM app — all `on:click`, `use:*`, and `prop:*` handlers silently become `None` and panic at runtime. The `on()`, `directive()`, and `prop()` constructors conditionally skip value creation based on `cfg!(feature = "ssr")`, so when the feature is active via Cargo feature unification, client-side values are never created.

The practical trigger is Cargo feature unification: any workspace dependency that enables `leptos/ssr` (e.g., `radix-leptos v0.9.0`) poisons the entire workspace, activating `tachys/ssr` for CSR apps too.

Runtime panics:

```
panicked at tachys/src/html/event.rs:190:30: callback removed before attaching
panicked at tachys/src/html/directive.rs:122:30: directive removed early
panicked at tachys/src/html/property.rs:81:30: property removed early
```

**Leptos Dependencies**

```toml
leptos = { version = "0.8.16", features = ["csr"] }
console_error_panic_hook = "0.1"

# THIS SINGLE LINE CAUSES THE BUG:
# Simulates what happens when any workspace dependency enables ssr
tachys = { version = "0.2", features = ["ssr"] }
```

**To Reproduce**

1. Create a CSR Leptos app with any `on:click` handler
2. Add `tachys = { version = "0.2", features = ["ssr"] }` to `Cargo.toml` (simulating feature unification from another workspace crate)
3. Run `trunk serve`
4. Page is blank; console shows `panicked at tachys/src/html/event.rs:190:30: callback removed before attaching`

Standalone reproduction:

```bash
git clone https://github.com/DrudgeDance/contributions.git && cd contributions/leptos/fix/always-create-client-values
cd issues/broken && trunk serve   # → blank page + panic
```

Repo: [DrudgeDance/contributions/leptos/fix/always-create-client-values](https://github.com/DrudgeDance/contributions/tree/main/leptos/fix/always-create-client-values)

**Root cause**

Three constructor functions use `cfg!(feature = "ssr")` to decide whether to create values:

`tachys/src/html/event.rs:116` — `on()`:
```rust
cb: (!cfg!(feature = "ssr")).then(|| SendWrapper::new(cb)),
//   ^^^ when ssr feature is active: (!true) → false → .then() = None
```

Same pattern in `directive.rs:50` and `property.rs:25`.

When these values are `None`, the subsequent `.build()` → `.attach()` / `.expect()` calls panic. The SSR rendering path (`to_html()`) never accesses these values — it writes nothing for events, directives, and properties — so the conditional creation is an unnecessary optimization that becomes dangerous under feature unification.

**Why this affects real apps**

[Cargo feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) means if any dependency in a workspace enables `leptos/ssr`, the `ssr` feature is activated for all crates. Known real-world triggers:

- `radix-leptos v0.9.0` — unconditionally enables `leptos/ssr`
- Any library that depends on `leptos` with `features = ["ssr"]` in a mixed workspace

The developer experience is particularly poor because there are multiple layers of silence:

1. **No compile-time error or warning** — the app builds cleanly
2. **No runtime warning** — the panic is the first indication
3. **Opaque error message** — "callback removed before attaching" gives no hint that the `ssr` feature is the cause
4. **No obvious diagnostic path** — `cargo tree -e features -i tachys` reveals the issue, but nothing suggests running it

**Next Steps**

- [x] I will make a PR

**Additional context**

- Leptos: 0.8.16, Tachys: 0.2.12
- Target: `wasm32-unknown-unknown` (CSR/WASM via Trunk)
- Rust: stable
- Originally discovered when `radix-leptos v0.9.0` activated `leptos/ssr` via feature unification in a CSR workspace
- The `issues/clean` variant in the reproduction repo confirms the app works correctly when `tachys/ssr` is not active — identical `src/main.rs`, only `Cargo.toml` differs

DrudgeDance:
When the `ssr` feature gets activated in a CSR or hydrate build (typically via Cargo feature unification from another workspace dependency), tachys constructors set event handlers, directives, and properties to `None`. The `.expect()` calls then panic with messages like `"callback removed before attaching"`, which tells you nothing about the cause.

This replaces all 8 of those messages with a shared constant that tells you what's wrong and what to run to fix it.

Fixes #4586

### Before / after

**Before:**
```
panicked at tachys/src/html/event.rs:190:30: callback removed before attaching
```

**After:**
```
panicked at tachys/src/html/event.rs:190:30: Value is None because the `ssr`
feature is active. When `ssr` is enabled, tachys skips creating client-side
values (event handlers, directives, properties) to avoid cross-thread panics
on multithreaded servers. If you are building a client-side (CSR or hydrate)
target, this means the `ssr` feature is being activated unintentionally
via Cargo feature unification; another dependency in your workspace is
enabling it. Run `cargo tree -e features -i tachys` to identify the source.
```

### Testing

```bash
git clone https://github.com/DrudgeDance/contributions.git
cd contributions/leptos/diagnostic/ssr-panic-message
trunk serve --port 8103
# → page panics with the improved diagnostic
```

### Context

Follow-up to [#4587](https://github.com/leptos-rs/leptos/pull/4587) (closed), incorporating feedback:
- No constructor changes (keeps `SendWrapper` thread-safety on servers)
- No `compile_error!` (doesn't break rust-analyzer `--all-features`)
- Text covers both CSR and hydrate, not just CSR

### DIFF
--- a/tachys/src/html/mod.rs
+++ b/tachys/src/html/mod.rs
@@ -20,13 +20,13 @@ use std::borrow::Cow;
 /// the `ssr` feature was activated unintentionally via Cargo feature
 /// unification in a client-side (CSR or hydrate) build.
 pub(crate) const FEATURE_CONFLICT_DIAGNOSTIC: &str =
-    "Value is None because the `ssr` feature is active. When `ssr` is enabled, \
-     tachys skips creating client-side values (event handlers, directives, \
-     properties) to avoid cross-thread panics on multithreaded servers. If \
-     you are building a client-side (CSR or hydrate) target, this means the \
-     `ssr` feature is being activated unintentionally via Cargo feature \
-     unification; another dependency in your workspace is enabling it. \
-     Run `cargo tree -e features -i tachys` to identify the source.";
+    "Value is None because the `ssr` feature is active. When `ssr` is \
+     enabled, tachys skips creating client-side values (event handlers, \
+     directives, properties) to avoid cross-thread panics on multithreaded \
+     servers. If you are building a client-side (CSR or hydrate) target, this \
+     means the `ssr` feature is being activated unintentionally via Cargo \
+     feature unification; another dependency in your workspace is enabling \
+     it. Run `cargo tree -e features -i tachys` to identify the source.";
 
 /// Types for HTML attributes.
 pub mod attribute;


gbj:
Thanks very much!

PR #4592, opened 2026-02-21
feat(`leptos_config`): Implement `LeptosOptions::css_path`

### LINKED ISSUE #3513: `LeptosOptions`: provide the CSS path
As in the title - it'd be great if `LeptosOptions` provided a function (or member?) with the path to the CSS file generated.

From what I can see, the generated CSS file is put in `/pkg/{output_name}.css`. It could be a simple helper function, currently I have:

```rust
fn css_path(leptos_options: &LeptosOptions) -> String {
    let mut path = String::with_capacity(200);
    path.push('/');
    path.push_str(&*leptos_options.site_pkg_dir);
    path.push('/');
    path.push_str(&*leptos_options.output_name);
    path.push_str(".css");

    path
}
```

DanikVitek:
Closes #3513

### DIFF
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -126,6 +126,20 @@ pub struct LeptosOptions {
 }
 
 impl LeptosOptions {
+    /// Returns the path to the generated CSS file.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use leptos_config::LeptosOptions;
+    /// use std::path::{PathBuf, MAIN_SEPARATOR};
+    ///
+    /// let options = LeptosOptions::builder().output_name("test").build();
+    /// let path: PathBuf = options.css_path();
+    ///
+    /// assert_eq!(path, PathBuf::from(format!("pkg{MAIN_SEPARATOR}test.css")));
+    /// ```
+    #[must_use = "allocates"]
     pub fn css_path(&self) -> PathBuf {
         use std::path::MAIN_SEPARATOR_STR;
 


gbj:
Thanks!

metatoaster:
Wait but this does something entirely different vs the original issue when used under Windows, as it's not the same separator used on the web.

gbj:
Oh thanks @metatoaster, you're completely right. This method (I think correctly?) generates the local filesystem path to the CSS file in the target directory, as distinct from the originally-requested URL for the CSS file.

metatoaster:
> Oh thanks @metatoaster, you're completely right. This method (I think correctly?) generates the local filesystem path to the CSS file in the target directory, as distinct from the originally-requested URL for the CSS file.

Yeah, but given the intent of the original issue is to use `site_pkg_dir` only to easily generate the value to fill into the view (or route in actix/axum), for the filesystem version to be useful it will likely need the `site_root` prefix also to be more useful for filesystem usage purpose. I'm not opposed to having the helper to return the filesystem path but having this `.css_path` shorthand for the more useful purpose be done like what the other issue likely needed.

For the filesystem counterpart I feel would benefit from using a new type wrapper around the `LeptosOptions` could serve as a useful disambiguation if we want to maintain the exact method names. I'm not opposed to leaving this as-is though, but I would note that the utility of this method as-is now is rather limited.

gbj:
I agree with you. I'd welcome a PR that adds the other one and renames this one to `css_file_path` or something similar, including the rest of the needed path, etc. Or can fix it myself, I'm just limited on time.

metatoaster:
Working on this in #4615.

johansmitsnl:
This doesn't work if hashes are enabled, right?

PR #4594, opened 2026-02-23
Allow `server_fn_macro` to specify a default client type

darakreqo:
Useful for custom usage of `server_fn` and `server_fn_macro`. Usage outside Leptos often requires customizing the `static CLIENT: LazyLock<Client>` and this PR helps with that.


gbj:
Thanks! This looks like a completely sensible PR. I'll just let the CI run and then will be happy to merge it.

darakreqo:
CI isn't failing because of my commits. Since it's my first PR here, can I ask how should I deal with that? It doesn't seem logical to change CI in this PR, right?

gbj:
I re-ran the one that was flaky, and it's fine.

However, CI now fails on `cargo-semver-checks`, for good reason -- this adds a new parameter to the public function `server_fn_macro::server_macro_impl`, which means that any callers need to update their calls.

There are two ways we can handle this:
1. Make it a breaking release
2. Non-breaking: Provide a `server_macro_impl_with_preset_client` function takes that the `preset_client: Option<_>` parameter, and have `server_macro_impl` call it with `None` by default. (And then change the signature in the next breaking release.)

2 is straightforward and could be released soon. 1 is also fine, but might not be released right away.

darakreqo:
> 2 is straightforward and could be released soon. 1 is also fine, but might not be released right away.

I don't mind whichever option is chosen, my project will continue to use my fork in the meantime anyway.

But I personally prefer making it a breaking change because there would be a bit of "asymmetry" with the fact that `server_macro_impl` already accepts `preset_server: Option<_>` without needing a wrapper fn like `server_macro_impl_with_preset_server`.

Just my thoughts.

PR #4598, opened 2026-02-24
fix: apply server function lint attributes to original and generated fn

LeoniePhiline:
## Relations

Fixes a regression introduced in https://github.com/leptos-rs/leptos/pull/4496.

Closes https://github.com/leptos-rs/leptos/issues/4597.

## Changeset

Given the following server function:

```rust
#[server(input = Postcard, output = Postcard)]
#[allow(clippy::too_many_lines)]
async fn get_article_content(
    article_slug: String,
) -> Result<Option<ArticleContent>, ServerFnError> {
    use axum_login::AuthSession;
    use leptos_axum::extract;
    // ...
    Ok(None)
}
```

Pre-https://github.com/leptos-rs/leptos/pull/4496 output:

```rust
    #[serde(crate = "leptos::server_fn::serde")]
    struct GetArticleContent {
        article_slug: String,
    }
    // ...
    #[allow(clippy::too_many_lines)]
    async fn get_article_content(
        article_slug: String,
    ) -> Result<Option<ArticleContent>, ServerFnError> {
        __server_get_article_content(article_slug).await
    }
    #[doc(hidden)]
    #[allow(clippy::too_many_lines)]
    async fn __server_get_article_content(
        article_slug: String,
    ) -> Result<Option<ArticleContent>, ServerFnError> {
        use axum_login::AuthSession;
        use leptos_axum::extract;
        // ...
        Ok(None)
    }
```

Post-https://github.com/leptos-rs/leptos/pull/4496 output (regression in function lint attributes):

```diff
    #[serde(crate = "leptos::server_fn::serde")]
+   #[allow(clippy::too_many_lines)]
    struct GetArticleContent {
        article_slug: String,
    }
    // ...
-   #[allow(clippy::too_many_lines)]
    async fn get_article_content(
        article_slug: String,
    ) -> Result<Option<ArticleContent>, ServerFnError> {
        __server_get_article_content(article_slug).await
    }
    #[doc(hidden)]
-   #[allow(clippy::too_many_lines)]
    async fn __server_get_article_content(
        article_slug: String,
    ) -> Result<Option<ArticleContent>, ServerFnError> {
        use axum_login::AuthSession;
        use leptos_axum::extract;
        // ...
        Ok(None)
}
```

With this changeset:

```diff
    #[serde(crate = "leptos::server_fn::serde")]
    #[allow(clippy::too_many_lines)]
    struct GetArticleContent {
        article_slug: String,
    }
    // ...
+   #[allow(clippy::too_many_lines)]
[...]

PR #4599, opened 2026-02-25
chore: upgrade deps

sabify:


sabify:
The failed SemVer check is false negative here.

sabify:
The SemVer breaking happened in the merged PR #4598.

gbj:
Thanks!

I'm aware of the semver piece, and am just waiting for docsrs to get around to building docs for the latest released version of `server_fn_macro`. 

Thanks for the updated nightly and associated changes. 

My understanding is that we should try to avoid updating minor or patch versions of our dependencies, unless Leptos itself *requires* that later version, so that end users can use earlier-than-latest versions if they need to for any reason. No? Most users will end up with the latest versions due to cargo version resolution in any case.

(I also notice that in the case of `serde_qs` the change downgrades it from 0.15 to 0.5 while all other version changes are updates.)

sabify:
Sorry for the long delay!

I see you've managed the nightly changes and merged them which is good.

You are right. Some of the upgrades (like actix_ws, glib, digest, serde_qs and rand) will be considered breaking changes based on our new policy. We may have to find a middle ground for upgrading deps without waiting for major leptos releases and breaking the current policy.

PR #4607, opened 2026-03-02
Systemic Fixes for Tachys Panics, WebView2 Stability, and E0283 Ambiguity

ordokr:
## Resolution for Critical Tachys Panics & Systemic Ambiguity

This PR introduces optimizations and resilience fixes for the Tachys rendering engine and Reactive Graph, specifically targeting unrecoverable panics in production and embedded environments (WebView2).

#### 1. Resilient Hydration & Rendering (Tachys)
- **Zero-Crash Fallbacks:** Replaced brittle `unwrap()` calls in `dom.rs` and `hydration.rs`. Mismatches now fallback to safe dummy nodes (e.g., `<div>` or empty `Comment`) in production instead of crashing the WASM thread.
- **Throttled Diagnostics:** Introduced throttled logging (`HYDRATION_ERROR_COUNT`). Logs high-fidelity data for the first 10 errors, then silences to maintain performance. Zero-cost in production.
- **Interning Optimization:** Optimized fallback creation using thread-local static interned strings to minimize JS-bridge overhead.

#### 2. WebView2 & Embedded Resilience
- **Resilient Attribute Attachment:** Refactored `event.rs`, `directive.rs`, and `property.rs` to handle missing callbacks/values gracefully. Instead of the common `"callback removed before attaching"` panic, it now logs an error and returns a no-op handler. This is critical for race conditions common in WebView2.

#### 3. Reactive Graph & `ImmediateEffect`
- **SOTA Recursion Loop:** Refactored the update loop to minimize lock contention and ensure absolute consistency during recursive triggers.
- **Dependency Integrity:** Fixed logic ensuring only the latest active run tracks sources, preventing memory leaks and stale state.

#### 4. Systemic Trait Ambiguity Fix (`IntoReactiveValue`)
- **Marker-Based Specialization:** Resolved the widespread `E0283` ("type annotations needed") errors by implementing a marker-based strategy in `reactive_graph` and `leptos_macro`.
- **Macro Update:** `#[prop(into)]` now uniquely resolves for all reactive types (Signals, Memos, Closures) without manual disambiguation.

###  Verification Results
- **Full Workspace Tests:** All crates (`tachys`, `reactive_graph`, `leptos`, etc.) pass unit and integration tests in both debug and release modes.
- **Real-World Integration:** Verified against the complex `Ordo LMS` project; confirmed that previously required manual event workarounds are now redundant and standard Leptos syntax is fully resilient.

---
**This PR addresses several long-standing 'papercut' panics and makes Leptos 0.9 significantly more viable for mission-critical and embedded applications.**

### DIFF
--- a/leptos/src/lib.rs
+++ b/leptos/src/lib.rs
@@ -187,7 +187,8 @@ pub mod prelude {
             signal::*,
             wrappers::{read::*, write::*},
             IntoReactiveValueTrait, IntoSignal,
-            __IntoReactiveValueMarkerBaseCase, __IntoReactiveValueMarkerIdentity,
+            __IntoReactiveValueMarkerBaseCase,
+            __IntoReactiveValueMarkerIdentity,
         };
         pub use server_fn::{
             self,
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -1072,8 +1072,10 @@ impl ToTokens for TypedBuilderOpts<'_> {
 
         let into = if self.into {
             let ty_str = self.ty.to_token_stream().to_string();
-            let is_signal = ty_str.contains("Signal") || ty_str.contains("Memo") || ty_str.contains("RwSignal");
-            
+            let is_signal = ty_str.contains("Signal")
+                || ty_str.contains("Memo")
+                || ty_str.contains("RwSignal");
+
             if !self.strip_option {
                 let ty = &self.ty;
                 if is_signal {
--- a/leptos_macro/src/slot.rs
+++ b/leptos_macro/src/slot.rs
@@ -199,8 +199,10 @@ impl ToTokens for TypedBuilderOpts<'_> {
 
         let into = if self.into {
             let ty_str = self.ty.to_token_stream().to_string();
-            let is_signal = ty_str.contains("Signal") || ty_str.contains("Memo") || ty_str.contains("RwSignal");
-            
+            let is_signal = ty_str.contains("Signal")
+                || ty_str.contains("Memo")
+                || ty_str.contains("RwSignal");
+
             if !self.strip_option {
                 let ty = &self.ty;
                 if is_signal {
--- a/reactive_graph/src/effect/immediate.rs
+++ b/reactive_graph/src/effect/immediate.rs
@@ -300,7 +300,7 @@ mod inner {
 
         fn update_if_necessary(&self) -> bool {
             let mut did_update = false;
-            
+
             loop {
                 let state = {
                     let guard = self.read().or_poisoned();
@@ -351,7 +351,12 @@ mod inner {
                     }
 
                     guard.state = ReactiveNodeState::Clean;
-                    (guard.owner.clone(), any_subscriber, guard.fun.clone(), recursion_count)
+                    (
+                        guard.owner.clone(),
+                        any_subscriber,
+                        guard.fun.clone(),
+                        recursion_count,
+                    )
                 };
 
                 owner.with_cleanup(|| any_subscriber.with_observer(|| fun()));
@@ -360,7 +365,8 @@ mod inner {
                 let needs_re_run = {
                     let mut guard = self.write().or_poisoned();
                     guard.run_done_count += 1;
-                    guard.run_done_max = Ord::max(recursion_count, guard.run_done_max);
+                    guard.run_done_max =
+                        Ord::max(recursion_count, guard.run_done_max);
 
                     let needs_re_run = guard.state == ReactiveNodeState::Dirty;
 
--- a/reactive_graph/src/into_reactive_value.rs
+++ b/reactive_graph/src/into_reactive_value.rs
@@ -42,7 +42,9 @@ where
     }
 }
 
-impl<T, S, I> IntoReactiveValueTrait<Signal<T, S>, __IntoReactiveValueMarkerIdentity> for I
+impl<T, S, I>
+    IntoReactiveValueTrait<Signal<T, S>, __IntoReactiveValueMarkerIdentity>
+    for I
 where
     I: IntoReactiveValue<Signal<T, S>, __IntoReactiveValueMarkerIdentity>,
     S: Storage<T>,
@@ -52,7 +54,9 @@ where
     }
 }
 
-impl<T, S, I> IntoReactiveValueTrait<ArcSignal<T, S>, __IntoReactiveValueMarkerIdentity> for I
+impl<T, S, I>
+    IntoReactiveValueTrait<ArcSignal<T, S>, __IntoReactiveValueMarkerIdentity>
+    for I
 where
     I: IntoReactiveValue<ArcSignal<T, S>, __IntoReactiveValueMarkerIdentity>,
     S: Storage<T>,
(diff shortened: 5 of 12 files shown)


ordokr:
@gbj  This PR provides a systemic resolution for the IntoReactiveValue ambiguity (E0283) that has been blocking complex component hierarchies in 0.9.  It's been a pain in my repo for awhile now, so I decided to take it upon myself to fix.  It also implements resilient attribute attachment to fix the callback removed before attaching panics frequently seen in WebView2/embedded environments. Verified against a production LMS codebase that I own.

ordokr:
Hi @gbj, thanks for your feedback!  As requested and to make review easier, here are three small repros that map 1:1 to the changes in this PR. All three were observed on the 0.9 branch prior to this PR; with this PR applied they either compile cleanly or stop panicking / behave consistently.

1) Systemic #[prop(into)] trait ambiguity (E0283) for reactive values

Problem (before): #[prop(into)] + reactive inputs like RwSignal<T> could produce E0283 because RwSignal participates in multiple conversion paths (e.g. callable / reactive conversion), and the compiler can’t choose a unique IntoReactiveValue route.

Repro (compile-fail before / compiles after):

use leptos::prelude::*;

#[component]
fn AmbiguityRepro(#[prop(into)] sig: Signal<String>) -> impl IntoView {
    move || sig.get()
}

#[test]
fn ambiguity_e0283() {
    let s = RwSignal::new("Hello".to_string());

    // BEFORE: fails with E0283 ("type annotations needed"/"multiple applicable items in scope")
    // AFTER: compiles (macro-generated marker path removes ambiguity)
    let _ = view! { <AmbiguityRepro sig=s /> };
}
2) ImmediateEffect recursive update consistency (dirty-but-not-rerun)

Problem (before): If an ImmediateEffect updates a signal it depends on, it can be marked dirty but not re-run until a later tick, which breaks “run-to-stability” expectations and leads to inconsistent intermediate state in more complex graphs.

Repro (behavior differs before / stabilizes after):

use leptos::prelude::*;
use std::{cell::RefCell, rc::Rc};

#[test]
fn immediate_effect_recursive_updates_should_stabilize() {
    let owner = Owner::new();
    owner.set();

    let s = RwSignal::new(0);
    let log = Rc::new(RefCell::new(Vec::new()));

    let _guard = ImmediateEffect::new({
        let log = log.clone();
        move || {
            let v = s.get();
            log.borrow_mut().push(v);

            // Self-triggered update: effect depends on `s`, then mutates `s`.
            if v < 2 {
                s.set(v + 1);
            }
        }
    });

    // EXPECTED with this PR: it runs until stable: 0 -> 1 -> 2
    // BEFORE: commonly observed as only the first run (e.g. log == [0], s == 1),
    //         with the rerun deferred to a future tick.
    assert_eq!(&*log.borrow(), &[0, 1, 2]);
    assert_eq!(s.get_untracked(), 2);
}
3) WebView2 / embedded panic: “callback removed before attaching” during attribute/event attachment

Problem (before): In WebView2-embedded environments (wry/tauri-style), rapidly mounting/unmounting nodes with event handlers / dynamic attributes could intermittently panic with “callback removed before attaching”. This PR makes attribute attachment resilient to that timing window.

Repro (panic before in WebView2 / no panic after):

use leptos::prelude::*;

#[component]
fn WebView2AttachRepro() -> impl IntoView {
    let (show, set_show) = create_signal(true);

    view! {
        <button on:click=move |_| set_show.update(|v| *v = !*v)>
            "toggle fast"
        </button>

        <Show when=move || show()>
            // Rapidly mount/unmount this element in a WebView2 host.
            // BEFORE (WebView2): intermittent panic "callback removed before attaching"
            // AFTER: no panic
            <div
                class=("hot", move || show())
                on:click=move |_| {}
            >
                "click / toggle"
            </div>
        </Show>
    }
}

Notes: I wasn’t able to make this 100% deterministic in all desktop browsers, but it reproduced reliably in a production WebView2 host where rapid toggling/mount churn happens during navigation and reactive updates.

If you'd like, I can also add these as actual tests (where applicable) under the relevant crate/module layout for 0.9, but I didn’t want to bloat the PR until you confirm the preferred location/shape.

gbj:

I'm going to preface my review by saying: the great thing about open source is that it doesn't force anyone to do any work they don't want to. So, if you want to take the steps I describe below, great! If not, you can keep using the fork/modified version that fixes the issues in a way you're satisfied with.

Here are a few general notes on what constitutes a good PR. I have no problem with AI-generated PRs per se; all of these apply to any PR, however the problem was analyzed and however the code was generated:
- Please remember when you submit a PR that you're communicating with other people. Language like "CRITICAL RESILIENCE" and "🚨 Urgent" is not a great way to engage with a project for your first contribution.
- PRs should do 1 thing. i.e., rather than 1 PR with 1 commit that claims to fix 4 different, unrelated issues, submit 4 PRs that each fix one issue, in separate commits
- PRs should close issues. When you encounter a bug, open an issue using the issue template in this repo. (i.e., not as a comment on this PR) If you want to submit a PR to fix it, that is awesome; please open an issue and submit the PR at the same time.
- Issues should be opened in the Issues tab of the repo and follow the template there, including a reproducible example.
- When feasible, please include a regression test. For example, include a test that would fail before the `ImmediateEffect` changes but passes after them. We have all the infrastructure set up for both unit tests and integration tests in this repo.
- If additional description is needed, please include it as part of the description text of the PR or the issues, not as an extraneous `.md` file to commit to the repo (i.e., please remove files like `TACHYS_FIXES_WALKTHROUGH.md` rather than including them in a PR)

I'd be happy to review individual PRs for any of these issues, given a reproduction as an issue.

ordokr:
Appreciate the review. Can I sanity-check something: are your notes mainly about contribution format (tone/scope/issues/tests), rather than feedback on the actual repros/fixes because that can't be done as is? If yes, is it okay if I leave this up for cherry-picking since I can’t rework it further right now?  Like I said, it's AI summaries of the work so thus the emojis, sorry about that!  To be honest this is my first PR on an open source repo, so still learning!

gbj:
My assumption based on similar previous PRs is that the actual issues and fixes contain a mix of incorrect configuration, actual bugs, correct fixes, and bad decisions/incorrect fixes. It's very hard to tell without having a 1:1 "this change is supposed to fix this issue," which is why I asked for that.

gbj:
> To be honest this is my first PR on an open source repo, so still learning!

Oh yeah, no problem. We all get started somewhere.

The notes I gave above about how to make a good PR are true for this or any open source project.

My advice would be: start small, with clear and focused issues, and then opened PRs with the fixes for them. My experience has been that AI coding agents do better with this kind of task anyway.

I'll take a look at the issues you opened! Happy to leave this open for the moment.

ordokr:
Sounds good, thank you @gbj , and I can see that after thinking about it a bit, and it is better practice than I have been going at it.  ( I tend to solve a bunch of issues in one commit and as a downside have to untangle things later when targeting a specific issue).  If I can later on this week will separate out the code to match the issues, but would like to have it verified that these issues are reproducible on other people's repos as well.  I really like Leptos, its the foundation for my LMS project along with Tauri, and so rather than having a local dependency in my cargo.toml I'd rather just pull from the official one that includes my fixes, so that was my thinking for this PR.  Thanks again for making Leptos possible.

PR #4613, opened 2026-03-03
feat: Extend either macro

domwst:
Unfortunately, no AI was used in creation of these changes.


domwst:
Can we have a macro for generating these?

gbj:
One of the reasons that these are limited to certain numbers is because the compile time balloons as they become larger. Could these changes be gated behind a feature to enable the larger sizes?

domwst:
Well, as far as I know, currently Rust doesn't allow to feature-gate certain arms of a declarative macro. We can however generate compile error in these new macro arms instead of generating the match expression/have separate macro. But is it really that big of a deal compared to already existing definitions of `EitherOf9`...`EitherOf16`? I could imagine, just having these macro arms doesn't hurt compilation times nearly as badly as having definitions and method/trait implementations for these types.

gbj:
Well, I just ran `cargo build --timings` for both versions and there's absolutely no difference :smile: I'm happy to merge this. CI was broken for unrelated reasons.

domwst:
Yeah, I probably should have run it myself. Thanks!

PR #4615, opened 2026-03-04
Better implementation of `css_path` with that moved to `css_file_path` to better reflect/compliment underlying use cases

metatoaster:
Provide both `css_path` and `css_file_path` to address both use cases in a way that is more generally useful for the respective context they will be used. Also provide `site_pkg_dir_route_base` as that has all the slashes added without duplication (it doesn't account for duplicate slashes added to `site_pkg_dir`) which `css_path` uses, and is in fact lifted from [`leptos_axum::site_pkg_dir_service_route_path`](https://docs.rs/leptos_axum/latest/leptos_axum/fn.site_pkg_dir_service_route_path.html), with the axum specific tail removed.

Additional integration tests are added alongside to the ones already at `leptos_axum` to show how these new methods may be used.  For brevity, the following may now be done to provide a singular service to route that one file (lifted/modified from the integration test case for axum that was added):

```rust
let router = Router::new()
    .nest(
        &opts.css_path(),
        Router::new()
            .route_service("/", ServeFile::new(&opts.css_file_path())),
    );
```

Though in practice under `leptos_axum`, `leptos_axum::site_pkg_dir_service` may be used for individual targeted paths, typically it can be cloned to serve multiple paths under `site_root`.

Fixes #4592 to actually do what #3513 wanted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2060,6 +2060,7 @@ dependencies = [
  "reqwest",
  "server_fn",
  "tachys",
+ "tempfile",
  "tokio",
  "tower",
  "tower-http",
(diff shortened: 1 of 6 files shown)


gbj:
Awesome! Thank you so much.

Issue #2142, opened 2023-12-29
`callback::NotRawCallback` is not implemented for <closure>


slyons:
**Describe the bug**
When using `Callback`s as demonstrated in the `login_with_token_csr_only` example I get the following error:

```rust
100 | |  ...                       <Login
101 | |  ...                           api=unauthorized_api
102 | |  ...                           on_success=move |api| {
    | |                                ---------- ----------
    | |                                |          |
    | | _______________________________|__________within this `{closure@lift_frontend/src/lib.rs:102:48: 102:58}`
    | ||                               |
    | ||                               required by a bound introduced by this call
103 | || ...                              log::info!("Successfully logged in");
104 | || ...                               authorized_api.update(|v| *v = Some(api));
105 | || ...                               let navigate = use_navigate();
106 | || ...                               navigate(Page::MeasureList.path(), Default::default());
107 | || ...                               fetch_user_info.dispatch(());
108 | || ...                           }
    | ||_______________________________- this tail expression is of type `{closure@lib.rs:102:48}`
109 | |  ...                       />
110 | |  ...                   }
    | |________________________^ within `{closure@lift_frontend/src/lib.rs:102:48: 102:58}`, the trait `callback::NotRawCallback` is not implemented for `(dyn for<'a> std::ops::Fn(&'a ()) -> Pin<Box<(dyn Future<Output = ()> + 'static)>> + 'static)`
```

And my build is full of messages like:
```rust
warning: a method with this name may be added to the standard library in the future
  --> frontend/src/pages/login.rs:31:36
   |
31 |                         on_success.call(res);
   |                                    ^^^^
   |
   = warning: once this associated item is added to the standard library, the ambiguity may cause an error or change in behavior!
   = note: for more information, see issue #48919 <https://github.com/rust-lang/rust/issues/48919>
   = help: call with fully qualified syntax `leptos::Callable::call(...)` to keep using the current method
   = help: add `#![feature(fn_traits)]` to the crate attributes to enable `std::ops::Fn::call`
```

**Leptos Dependencies**

For example:
```toml
leptos = { version = "0.5.4", features = ["csr", "nightly"] }
leptos_meta = { version = "0.5.4", features = ["csr"] }
leptos_router = { version = "0.5.4", features = ["csr"] }
```

**To Reproduce**
1. Create new crate
2. Copy code from example into crate
3. `trunk build`

gbj:
Is this a duplicate of #2041?

slyons:
[...]

Issue #2041, opened 2023-11-18
Derived signals can't be converted Into() Callbacks


pscott31:
**Describe the bug**
I have a component that takes a callback prop using e.g. `#[prop(into)] on_click: Callback<()>`

When I pass in a closure that makes use of standard signal types, like `ReadSignal`, it works - however if my closure access a Derived signal (or indeed a signal made by create_slice), it complains with 

```
 within `{closure@src/app.rs:106:42: 106:50}`, the trait `callback::NotRawCallback` is not implemented for `(dyn std::ops::Fn() -> i32 + 'static)`
```

**Leptos Dependencies**

```toml
 leptos = { version = "0.5.2", features = ["csr", "nightly"] }
```

**To Reproduce**

Here's a little snippet that reproduces the problem

```
use leptos::logging::*;
use leptos::*;

fn main() {
    // This is fine
    let (signal, _) = create_signal::<i32>(1);
    let callback1 = Callback::<()>::from(move |_| log!("{}", signal()));

    // But this doesn't work:
    //  trait bound `(dyn std::ops::Fn() -> i32 + 'static): callback::NotRawCallback` is not satisfied
    let derived = Signal::derive(move || signal() + 1);
    let callback2 = Callback::<()>::from(move |_| log!("{}", derived()));
}
```

**Expected behavior**

The program above to compile! And hopefully if that compiles, my actual program will compile 

**Additional context**
I don't quite understand what the NotRawCallback trait is trying prevent me from doing; presumably it's to prevent me shooting myself in the foot and maybe what I'm trying to do I shouldn't be doing for some reason. I had a look at the code and it seems like the point of NotRawCallback is to prevent wrapping a Callback in another Callback but I might be mistaken.

It might also well be this is a bug in rust's negative trait implementation as it's not stable yet :)

gbj:
> I don't quite understand what the NotRawCallback trait is trying prevent me from doing; presumably it's to prevent me shooting myself in the foot and maybe what I'm trying to do I shouldn't be doing for some reason. I had a look at the code and it seems like the point of NotRawCallback is to prevent wrapping a Callback in another Callback but I might be mistaken.

It allows implement `From<Fn(In) -> Out>` for `Callback<In, Out>` while also implementing `Fn(In) -> Out` for `Callback<In, Out>`, which would otherwise conflict with the global `From<T> for T` implementation. In other word, it isn't there to allow you to do anything or prevent you from doing anything.

[...]

PR #4617, opened 2026-03-04
fix: `.into_reactive_value` conversion for signal-function types on `nightly` (closes #4608)

### LINKED ISSUE #4608: #[prop(into)] + RwSignal<T> causes E0283 ambiguity on 0.9
On the 0.9 branch, using #[prop(into)] on a component prop typed as Signal<T> (and observed similarly with Memo<T>) can fail to compile when the caller passes an RwSignal<T>. The compiler reports E0283 (“type annotations needed” / “multiple applicable items in scope”). This appears to be caused by RwSignal<T> satisfying multiple valid conversion routes (e.g., it can behave like a callable reactive value and also participates in direct Into<Signal<_>>-style conversions), leaving the macro-generated conversion path unable to disambiguate.

To Reproduce

use leptos::prelude::*;

#[component]
fn AmbiguityRepro(#[prop(into)] sig: Signal<String>) -> impl IntoView {
    move || sig.get()
}

#[test]
fn test_compile_fail() {
    let s = RwSignal::new("test".to_string());

    // Fails to compile on 0.9 with E0283:
    // "type annotations needed" / "multiple applicable items in scope"
    let _ = view! { <AmbiguityRepro sig=s /> };
}

Additional context
A robust fix likely requires disambiguation in the IntoReactiveValue conversion strategy (e.g., a marker-based specialization approach that ensures “identity” conversions for reactive types are preferred over competing conversion paths).

gbj:
I *think* this fixes #4608 in an appropriate way (using type system changes rather than changes to the macro) without breaking other things. But at a point this level of marker types for traits, nightly features, etc. starts to boggle the mind. @zakstucke and @maccesch, this intersects with your work on IntoReactiveValue and Show/ShowLet, so I'm hoping you can help me confirm it doesn't break either of those.

### DIFF
--- a/reactive_graph/src/nightly.rs
+++ b/reactive_graph/src/nightly.rs
@@ -219,8 +219,13 @@ impl<T, S> !NotASignalType for Memo<T, S> {}
 impl<T, S> !NotASignalType for MaybeProp<T, S> {}
 
 // Types that do NOT need negative impls:
-// - Signal, ArcSignal: use Deref to dyn Fn() -> T, don't directly impl Fn traits
+// - Signal: wraps ArenaItem (an index), so auto trait propagation preserves NotASignalType
 // - WriteSignal, ArcWriteSignal, SignalSetter: only impl Fn(T) -> (), not Fn() -> T
+
+// ArcSignal contains SignalTypes directly (which holds ArcReadSignal, ArcMemo, etc.),
+// so auto trait propagation would remove NotASignalType. We add an explicit positive
+// impl because ArcSignal uses Deref (not direct Fn impls), so there's no ambiguity.
+impl<T, S> NotASignalType for ArcSignal<T, S> where S: Storage<T> {}
 
 mod readable_deref_impl {
     /// Derived from the implementation and original creaters at dioxus


maccesch:
Yep, certainly mind boggling to me 😀. From what I can see this should be good to go.

- `Show` and `AnimatedShow` use standard `#[prop(into)] when: Signal<bool>` so this fix should work nicely there.
- `ShowLet` uses it's own custom tailored implementation because the compiler was not able to infer types otherwise. This means it should not be affected by this at all.

zakstucke:
Hey! IIRC I wanted to use `auto_traits` and `negative_impls` to get around the nightly issues in my original PR ([commit](https://github.com/leptos-rs/leptos/commit/f617c7fe1c08544a6119aba3699f34221b0df337#diff-43dbcf3c056de18e08f5239beb35b006ff1be070f37938c391b45846c655aa9a)) but we ended up having to give up on it because those nightly features have fundamental issues? 

This was what you originally warned me about it 😅 

> IIRC we have been burned by auto_traits 
https://github.com/leptos-rs/leptos/issues/2142
https://github.com/leptos-rs/leptos/issues/2041
Not sure if the same sorts of issues would apply here but that's one of the things I'd want to look at.
(i.e., the issue in the past was that a closure that captures a negative auto trait then acquires that same auto trait. In other words, a closure that captures a Signal would itself be !NotSignalMarker because the Signal is !NotSignalMarker)

And I think it still applies here, I can break this PR on nightly. 

```rust
    1 use reactive_graph::{wrappers::read::ArcSignal, IntoReactiveValue};
    2 let a: ArcSignal<usize> = (|| 2).into_reactive_value();
    3 let b: ArcSignal<usize> = ArcSignal::stored(2).into_reactive_value();
    4 let _: ArcSignal<usize> = (move || a.get() + b.get()).into_reactive_value();
```

Stable works fine, on master nightly this fails on line `2` and `4` with:
```
error[E0277]: the trait bound `ArcSignal<usize>: From<{closure@reactive_graph/tests/async_derived.rs:13:32: 13:34}>` is not satisfied
  --> reactive_graph/tests/async_derived.rs:13:38
   |
13 |     let a: ArcSignal<usize> = (|| 2).into_reactive_value();
   |                                      ^^^^^^^^^^^^^^^^^^^ the trait `From<{closure@reactive_graph/tests/async_derived.rs:13:32: 13:34}>` is not implemented for `ArcSignal<usize>`
   |
   = help: the following other types implement trait `From<T>`:
             `ArcSignal<String, S>` implements `From<&str>`
             `ArcSignal<T, S>` implements `From<ArcMemo<T, S>>`
             `ArcSignal<T, S>` implements `From<RwSignal<T, S>>`
             `ArcSignal<T, S>` implements `From<Signal<T, S>>`
             `ArcSignal<T, S>` implements `From<reactive_graph::computed::Memo<T, S>>`
             `ArcSignal<T, S>` implements `From<reactive_graph::signal::ReadSignal<T, S>>`
             `ArcSignal<T>` implements `From<ArcReadSignal<T>>`
             `ArcSignal<T>` implements `From<ArcRwSignal<T>>`
             `ArcSignal<T>` implements `From<T>`
   = note: required for `{closure@reactive_graph/tests/async_derived.rs:13:32: 13:34}` to implement `Into<ArcSignal<usize>>`
   = note: required for `{closure@reactive_graph/tests/async_derived.rs:13:32: 13:34}` to implement `IntoReactiveValue<ArcSignal<usize>, reactive_graph::__IntoReactiveValueMarkerBaseCase>`

error[E0277]: the trait bound `ArcSignal<usize>: From<{closure@reactive_graph/tests/async_derived.rs:15:32: 15:39}>` is not satisfied
  --> reactive_graph/tests/async_derived.rs:15:59
   |
15 |     let _: ArcSignal<usize> = (move || a.get() + b.get()).into_reactive_value();
   |                                                           ^^^^^^^^^^^^^^^^^^^ the trait `From<{closure@reactive_graph/tests/async_derived.rs:15:32: 15:39}>` is not implemented for `ArcSignal<usize>`
   |
   = help: the following other types implement trait `From<T>`:
             `ArcSignal<String, S>` implements `From<&str>`
             `ArcSignal<T, S>` implements `From<ArcMemo<T, S>>`
             `ArcSignal<T, S>` implements `From<RwSignal<T, S>>`
             `ArcSignal<T, S>` implements `From<Signal<T, S>>`
             `ArcSignal<T, S>` implements `From<reactive_graph::computed::Memo<T, S>>`
             `ArcSignal<T, S>` implements `From<reactive_graph::signal::ReadSignal<T, S>>`
             `ArcSignal<T>` implements `From<ArcReadSignal<T>>`
             `ArcSignal<T>` implements `From<ArcRwSignal<T>>`
             `ArcSignal<T>` implements `From<T>`
   = note: required for `{closure@reactive_graph/tests/async_derived.rs:15:32: 15:39}` to implement `Into<ArcSignal<usize>>`
   = note: required for `{closure@reactive_graph/tests/async_derived.rs:15:32: 15:39}` to implement `IntoReactiveValue<ArcSignal<usize>, reactive_graph::__IntoReactiveValueMarkerBaseCase>`

For more information about this error, try `rustc --explain E0277`.
```

On this branch it fails on line `3` with:
```
error[E0277]: the trait bound `&dyn Fn() -> {integer}: IntoReactiveValue<ArcSignal<usize>, _>` is not satisfied
  --> reactive_graph/tests/async_derived.rs:14:52
   |
14 |     let b: ArcSignal<usize> = ArcSignal::stored(2).into_reactive_value();
   |                                                    ^^^^^^^^^^^^^^^^^^^ the trait `IntoReactiveValue<ArcSignal<usize>, _>` is not implemented for `&dyn Fn() -> {integer}`

For more information about this error, try `rustc --explain E0277`.
```

So I think it does the classic nightly conundrum with this feature of fixing one but breaking the other :(

gbj:
Thanks Zak. You're right to question the negative impls approach and you're 100% correct that I pointed out that problem in the past :smile: I was a little hesitant to use it here, but it *seems* not to break things in the same way when it's used in this marker trait context.

I added another test to clarify this. If I recall correctly, in the past our problem has been that auto trait are infectious: i.e., `!NotASignalType` works like `!Send`, so a closure that captures a `!NotASignalType` also becomes `!NotASignalType`. With the approaches we've taken in the past, the problem here has been that this means that (for example) `Signal` can't capture a `Signal` (and so on). I've added a couple tests in the last two commits to clarify that this isn't an issue. 

I think the differences between your original PR and this one is that in your original `!NotASignal` was implemented for `Signal` and `ArcSignal`, whereas in this one it's the other way around -- it's implemented for all the signal types *other than* `Signal` and `ArcSignal`.

It looks like all your failing versions are with `ArcSignal` (as opposed to `Signal`). I noticed I had `!NotASignalType` implemented for `ArcSignal` but removing it doesn't seem to fix. I'll need to wrap my head around this one a little more... for now I've added it as a failing test case :smile: 

Any help welcome!

PR #4620, opened 2026-03-05
Fixes deriving `Patch` for `Store` where `store` attribute is being used.

mskorkowski:
I've made a few things public, so the code generated by the `rective_stores_macro` will compile.
When adding/updating comments, whenever I couldn't understand the meaning of something I've left `??`

**Questions left open**

1. Why are we always calling `initialize` in the `with_field_keys`? Shouldn't we move the call
    
    ```rust
        .or_insert_with(move || Box::new(FieldKeys::new(initialize())));
    ```
    This will cause memory allocation only in case we use the `initial_keys`, on the other hand if initialize has any side effects, which FnOnce can, they might not be triggered.

2. There is an example where `store` attribute is used but it derives only a `Store` trait. 

   - Should there be an example Store + Patch?
   - Should just add the deriving of the `Patch` to the existing example?


gbj:
Thanks for the PR!

Feel free to add the deriving of `Patch` to the current example.

Your suggestion to use `.or_insert_with(move || Box::new(FieldKeys::new(initialize())));` certainly *sounds* right to me. I've looked and I don't think we use effectful `initialize` anywhere. If it sounds correct and the current tests pass with it, I'm sure this is fine.

My suggestion would be that these should be exported `#[doc(hidden)] pub`. They really are internal implementation details. 

Additional explanatory comments are always helpful for people reading the code, of course. The `reactive_stores` implementations are fairly tricky to follow along with, as I'm sure you've noticed. I'll add some one-by-one comments on the doc comments you've added.

reactive_stores/src/lib.rs
```diff
@@ -424,12 +425,48 @@ type Map<K, V> = Arc<std::sync::RwLock<std::collections::HashMap<K, V>>>;
...
+    ///     
+    ///     
+    ///     ### Entries
+    ///
+    ///     Entry in the vector is a tuple (idx, segment) where
+    ///
+    ///     - **idx** - index of the element in the collection
+    ///     - **segment** - key of the element in the collection
+    ///     
+    /// - **initialize** - ??
```
gbj: `initialize` is the set of keys with which to initialize the `KeyMap` for this field, if there aren't keys listed yet. In all cases inside the library this is `|| self.latest_keys()` or `|| self.inner.latest_keys()`

reactive_stores/src/patch.rs
```diff
@@ -134,6 +141,14 @@ where
     ///
     /// Returns `true` if the structure of the collection changed (items added, removed,
     /// or reordered). Individual item changes are notified via the `notify` callback.
+    ///
+    /// # Arguments
+    ///
+    /// - **new** - updated values
+    /// - **notify** - callback to notify about the update
+    /// - **keys** - ??
```
gbj: see above

mskorkowski:
@gbj PR is ready for review

gbj:
Thanks very much -- this looks good!

gbj:
Note to self for the future: the `initialize()` change here breaks initialization for nested keyed stores:

```rust
    #[test]
    fn nested_keyed_subfields_dont_deadlock_on_first_iter() {
        // Regression test for a panic/deadlock that occurred when a `For`
        // over a keyed subfield was nested inside another `For` over a
        // keyed subfield. Iterating the inner subfield's `KeyedSubfield`
        // would call `update_keys`, which, while holding a write lock on
        // the global `KeyMap`, would initialize its entry by calling
        // `latest_keys()`. That traversed back through the parent
        // `AtKeyed::reader` -> `resolve_index` -> `with_field_keys`,
        // attempting to re-acquire the same write lock and aborting.
        use crate::Field;

        #[derive(Debug, Store, Clone)]
        struct Part {
            id: usize,
            #[store(key: usize = |c| c.id)]
            chapters: Vec<Chapter>,
        }

        #[derive(Debug, Store, Clone)]
        struct Chapter {
            id: usize,
        }

        #[derive(Debug, Store)]
        struct Structure {
            #[store(key: usize = |p| p.id)]
            parts: Vec<Part>,
        }

        let store = Store::new(Structure {
            parts: vec![Part {
                id: 1,
                chapters: vec![Chapter { id: 10 }],
            }],
        });

        // Mimic the behavior of nested `<For>` components: iterate the
        // outer keyed subfield, take a `Field<Part>` for each item, and
        // iterate its inner keyed subfield. Doing so on a freshly
        // constructed store ensures both the outer and the inner entries
        // need to be initialized, which is the case that triggered the
        // re-entrant write lock.
        for at_part in store.parts() {
            let part: Field<Part> = at_part.into();
            let _: Vec<_> = part.chapters().into_iter().collect();
        }
    }
```

PR #4627, opened 2026-03-09
feat(`either_of`): Add `Either!` convenience macro

DanikVitek:
This PR adds a convenience `Either!` macro for referring to different `EitherOf{N}` types.

```rs
let a: Either!(A, B, C) = todo!();
let _: EitherOf3<A, B, C> = a;
```

Specifying a single type parameter just aliases it:

```rs
let a: Either!(&str);
let _: &str = a;
```

> [!NOTE]
> TODO: Add an empty branch that aliases the `!` (never) type when it gets stabilized,
> or possibly add a nightly-only (e.g. `nightly_never_type`) feature that does this.

### DIFF
--- a/either_of/src/lib.rs
+++ b/either_of/src/lib.rs
@@ -942,36 +942,110 @@ macro_rules! either {
     };
 }
 
+/// Convenience macro for refering to `Either` types by listing their variants.
+///
+/// # Example
+///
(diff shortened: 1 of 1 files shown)


gbj:
Looks nice -- thanks!

Issue #4630, opened 2026-03-09
Property key autocomplete in view! macro is not working


lpotthast:
Users are unable to discover the required (and optional) props of a component they want to use. They cannot simply begin to write `<SomeComp |>` and hit the IDE autocomplete shortcut at cursor position '|' in a view! macro. No meaningful thing will be shown. Neither actually defined props, nor inherently available things like `on` handler attachment, `class`, `style`, etc.

This inability to discover goes against common user expectations. It may slow down development, as it may require more jumps to external (to the view! macro call) component definitions and/or external resources (documentation, websites, ....) or roundtrips with the compiler.

Due to https://github.com/leptos-rs/leptos/issues/4628, roundtrips with the compiler are also inefficient.

gbj:
All components have docs for the props auto-generated so a simple LSP hover on the component name should show the optional/required props.

I have no idea how to achieve autocomplete support in a different way — do you?

Genuine question, not passive aggression. I know very little about the topic. I'd be happy for any help you can provide.

lpotthast:
I think the most relevant change in the PR (that enabled automcomplete by making it easier for the compiler to see thorugh and not bail out) was the move towards individual let bindings when building the props builder, as in
```
let __helper = component_helper(&SomeName);

let __checked_prop1 = __helper.check_and_wrap_prop1(value).extract_value();

let __props_builder = __helper.builder();
let __props_builder = __props_builder.prop1(__checked_prop1);
let __props_builder = __props_builder.prop2(__checked_prop2);
let __props_builder = __props_builder.prop3(__checked_prop3);
let __props_builder = __props_builder.children(__checked_children);

let __props_builder = __presence.check_missing(__props_builder);
let props = __props_builder.build();
```
I will try to address this in more detail in the PR.

Issue #4628, opened 2026-03-09
Errors in view! macros are not properly scoped and lead to a bad developer experience


lpotthast:
Certain errors made in a view! macro invocation are reported with too broad scoping. These errors typically point to, explicitely or accidentally using `Span::call_site()`, the entire view! macro invocation.

These include, for example:
- providing a prop value of incorrect type (in the case of a generic prop)
- missing to provide a required prop
- missing to provide children (special case of the above)
- providing children which don't implement the required Fn* trait (e.g. providing `FnOnce` children, where `Fn` children would be required)

This makes it very hard (especially for newcomers to both Leptos and Rust) to see quickly what exactly went wrong, why it went wrong, and how this issue can be remedied.

Errors spanning the entire macro are most likely rendered in an IDE by squiggly-underlining the entire view! macro invocation in red, which is very off-putting. Even for experienced devs, this error reporting is cumbersome to work with, as the errors must be actively read by either hovering anywhere in the view! macro in your IDE or by specifically looking at the errors in the console. But even then, the root cause, the thing that’s actually incorrect/missing might not be discoverable from the first line of the error, requiring users to look deeper into the message.

In a tiny view! call, an error coped to the entire invocation is easy to correlate. When a view! macro spans many more lines and reuses the same components multiple times, it might not be obvious what exact callsite of <SomeComp> this error `SomeCompPropsBuilder_Error_Missing_required_field_concrete_i32` correlates with.

All types of error-sources described above could happen from manually working on a view!, in which case an error is easily correlatable to the last changes a user made. But such errors could simply occur because of bad merges, updates of libs providing components which are in use and now take additional props, etc…

In some cases (multiple required props missing) only errors for one missing prop are reported, swallowing/hiding the other issue until the first one is resolved.

An additional observation: Sometimes, there are actually clearly scoped errors, which then get shadowed by multiple Span::call_site() scoped downstream errors.

To summarize the problem: There is no concept/framework in leptos_macro that allows for properly scoped errors. There is no control of the amount of errors emitted for each type of error. There are no tests that would catch regressions in error-locality.

gbj:
Could you please provide some examples? It sounds to me like ~all of these are related to the component props builder implementation -- does that sound right?

gbj:
Ah okay so I saw and replied to these notifications in order — but did not know you had a big PR incoming :smile: Feel free to ignore my request here.

lpotthast:
[...]

PR #4631, opened 2026-03-09
Localized view macro errors + Prop autocomplete

lpotthast:
Fixes https://github.com/leptos-rs/leptos/issues/4628
Fixes https://github.com/leptos-rs/leptos/issues/4630

Hey @gbj,

in this branch, I tried to make errors be reported where they actually happen, so that they do not disturb a user, but instead steer them in the right direction for quickly fixing them. No error should ever span the entire view! macro. Custom messages should be emitted where possible. When multiple things are erroneous, no error should be hidden until an earlier error is resolved. The changes should be as non-breaking as possible.

(Though this could be done with a few quick  "replace `quote!` with `quote_spanned!`" and a few span re-calculations here and there... But oh boy, was I wrong. I do think there could have been a smaller/more focused change that might accomplish the same, but I had to refactor some parts to make sense of it. Sorry for the inconveniences of such large changesets! :/)

Please see [leptos_macro/LOCALIZED_ERRORS.md](https://github.com/leptos-rs/leptos/compare/main...lpotthast:leptos:localized-view-macro-errors?expand=1#diff-26bcf4396588163449b3a100c6e2f469d5be3af61774351996621ae8eddb0dec) for an overview of what changed and how.

As a summary:
- Both components and slots now generate a companion module that can be used to store an arbitrary amount of additional type-machinery
- Props no longer take the full generics, following the late-bound approach also followed in most other leptos areas
- A new helper struct is used to track presence of required props and to validate each props type
- As a result:
- view! errors are now properly scoped
- Custom on_unimplemented messages provide nicer feedback for users
- A new suite of trybuild tests can catch regressions in error locality 
- Prop autocompletion now works! (totally by accident ^^, was kinda surprised when I discovered this)

These macro code-gen changes add purely compile-time validation. Sadly, this comes at a slight cost of around 6.5% WASM binary size increase in debug builds. In release builds, there is no difference with main/latest on crates.io, as all new code  is indeed optimized away. I don't have any more ideas on how the impact on debug builds could be reduced. But I also do think that this is an acceptable price for the gain in developer experience.

Examples:

Missing required prop:
image

Component requires children:
image

Wrong concrete prop type:
image

Wrong generic prop type:
image

Wrong children:
image

Autocomplete:
image

### DIFF
--- a/leptos/Cargo.toml
+++ b/leptos/Cargo.toml
@@ -173,8 +173,6 @@ skip_feature_sets = [
   ["miniserde", "rkyv"],
   ["serde-lite", "rkyv"],
   ["default-tls", "rustls"],
-  # do not test against nightly feature alone
-  ["nightly"],
 ]
 max_combination_size = 2
 
--- a/leptos/src/show_let.rs
+++ b/leptos/src/show_let.rs
@@ -1,5 +1,6 @@
 use crate::{children::ViewFn, IntoView};
 use leptos_macro::component;
+#[cfg(not(all(feature = "nightly", rustc_nightly)))]
 use reactive_graph::traits::Get;
 use std::{marker::PhantomData, sync::Arc};
 use tachys::either::Either;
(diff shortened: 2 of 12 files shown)


gbj:
Wow!

Well, the results in the screenshots certainly look great, along with the other results you've reported.

> Sadly, this comes at a slight cost of around 6.5% WASM binary size increase in debug builds.

Debug build WASM size is not an issue at all. Release build WASM size would be a concern if it was affected. It sounds like you're saying it is not?

Compile time in larger codebases is the other question that matters.

> Sorry for the inconveniences of such large changesets!

No worries -- sometimes it takes what it takes to get something done. I hope you'll understand it may take me a while to find time to review.

Thanks so much, this looks like it could be a big improvement.

For anyone else reading: I'd really appreciate help reviewing some of this. The next six weeks or so is the time of year when I have the least time, so any help would be welcome.

PR #4633, opened 2026-03-11
4610 webview2 callback panic repro

ordokr:
Adds a minimal reproduction for the WebView2/Tauri panic caused by a callback being removed before attach.

Branch: 4610-webview2-callback-panic-repro
Location: projects/webview2-callback-panic-repro/

To reproduce, install Rust, Trunk, and Tauri CLI. If your ~/.cargo/config includes -fuse-ld=lld, clear RUSTFLAGS for the wasm build, then build the frontend:

$env:RUSTFLAGS=''
cd projects/webview2-callback-panic-repro/frontend
cargo build --release --target wasm32-unknown-unknown --config 'target.wasm32-unknown-unknown.rustflags=[]'
trunk build --release
cd ../..

Run trunk serve in projects/webview2-callback-panic-repro/frontend and cargo tauri dev in projects/webview2-callback-panic-repro.

The app auto-triggers 10 rapid toggle cycles on load, or the issue can be triggered by repeatedly clicking Toggle rapidly (30x @ 5ms). The panic shows up in the WebView DevTools console.

This reproduces the issue on Leptos CSR with Tauri 2.x + Wry/WebView2 on Windows, using a <Show> boundary with rapid mount/unmount via staggered gloo_timers::Timeout callbacks and child on:click handlers. It appears to hit the tachys race where self.cb is None during attach.

### DIFF
--- a/projects/webview2-callback-panic-repro/ISSUE_COMMENT.txt
+++ b/projects/webview2-callback-panic-repro/ISSUE_COMMENT.txt
@@ -1,10 +1,6 @@
-# Comment to post on https://github.com/leptos-rs/leptos/issues/4610
-
----
-
 I've added a minimal reproduction for the "callback removed before attaching" panic in WebView2/Tauri.
 
-**Branch:** [`4610-webview2-callback-panic-repro`](https://github.com/ordokr/leptos/tree/4610-webview2-callback-panic-repro)  
+**Branch:** [4610-webview2-callback-panic-repro](https://github.com/ordokr/leptos/tree/4610-webview2-callback-panic-repro)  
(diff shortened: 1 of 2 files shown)


gbj:
I'm not able to reproduce this using this example, running `trunk serve` and `cargo tauri dev` (on macos, so Safari as the webview), or running `trunk serve` alone (in Chrome or Firefox) after checking out this PR branch.

I suppose it *might* be a Windows-specific issue but that would be surprising.

I don't think Tauri should be necessary at all to the reproduction. Does the same problem occur for you with `trunk serve` alone?

PR #4634, opened 2026-03-11
fix(reactive_graph): ImmediateEffect re-runs to stability on recursive updates (closes #4609)

### LINKED ISSUE #4609: ImmediateEffect fails to re-run to stability during recursive signal updates
### Describe the bug
In the `0.9` reactive graph, an `ImmediateEffect` that updates a signal it depends on (a recursive update) does not always re-run to reach a stable state within the same synchronous cycle. While the node is marked `Dirty`, the execution can be lost or deferred, leading to an inconsistent state where the effect has run, but the underlying signal has changed again without triggering a follow-up execution.

### Leptos Dependencies
```toml
leptos = { git = "https://github.com/leptos-rs/leptos", branch = "0.9", features = ["csr"] }
```

### To Reproduce
```rust
use leptos::prelude::*;
use std::{cell::RefCell, rc::Rc};

#[test]
fn immediate_effect_should_stabilize() {
    let owner = Owner::new();
    owner.set();

    let s = RwSignal::new(0);
    let log = Rc::new(RefCell::new(Vec::new()));

    // The effect should run until the condition (v < 2) is no longer met.
    let _guard = ImmediateEffect::new({
        let log = log.clone();
        move || {
            let v = s.get();
            log.borrow_mut().push(v);

            // Self-triggered update: effect depends on `s`, then mutates `s`.
            if v < 2 {
                s.set(v + 1);
            }
        }
    });

    // EXPECTED: [0, 1, 2]
    // ACTUAL on 0.9: [0] (the rerun is often deferred or lost)
    assert_eq!(&*log.borrow(), &[0, 1, 2]);
    assert_eq!(s.get_untracked(), 2);
}
```

### Expected behavior
The effect should re-run synchronously until it reaches a stable state (no further self-triggered updates), producing:
- `log == [0, 1, 2]`
- `s == 2`

### Actual behavior
On `0.9`, the follow-up runs are often deferred or lost (commonly only `[0]`), even though the effect logically dirtied itself.

### Additional context
This looks like a missed “run-to-fixpoint” behavior for `ImmediateEffect` under recursive updates: the effect can mark itself dirty while running, but the follow-up execution does not reliably happen in the same synchronous cycle.

ordokr:
Changes
reactive_graph/src/effect/immediate.rs – ImmediateEffect loop fix:

Wraps update_if_necessary in a loop so it re-runs until the effect is no longer Dirty
When an effect updates a signal it depends on, it can mark itself Dirty; the loop ensures it runs again in the same cycle
Adjusts mark_check and mark_dirty to avoid redundant state transitions
reactive_graph/tests/effect_immediate.rs – Regression test:

immediate_effect_recursive_updates_should_stabilize – effect that updates its own signal runs until stable (0 → 1 → 2)

### DIFF
--- a/reactive_graph/src/effect/immediate.rs
+++ b/reactive_graph/src/effect/immediate.rs
@@ -299,100 +299,110 @@ mod inner {
         fn mark_subscribers_check(&self) {}
 
         fn update_if_necessary(&self) -> bool {
-            let state = {
-                let guard = self.read().or_poisoned();
+            let mut did_update = false;
 
-                if guard.owner.paused() {
-                    return false;
+            loop {
+                let state = {
+                    let guard = self.read().or_poisoned();
+                    if guard.owner.paused() {
+                        return did_update;
+                    }
+                    guard.state
+                };
+
+                let needs_update = match state {
+                    ReactiveNodeState::Clean => false,
+                    ReactiveNodeState::Check => {
+                        let sources = self.read().or_poisoned().sources.clone();
+                        sources
+                            .into_iter()
+                            .any(|source| source.update_if_necessary())
+                    }
+                    ReactiveNodeState::Dirty => true,
+                };
+
+                if !needs_update {
+                    break;
                 }
 
-                guard.state
-            };
+                {
+                    if let Some(batch) = &*BATCH.read().or_poisoned() {
+                        let mut batch = batch.write().or_poisoned();
+                        let subscriber =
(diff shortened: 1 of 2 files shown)


gbj:
Please either target this PR against the `leptos_0.9` branch or include only the relevant commit ([92213b0](https://github.com/leptos-rs/leptos/pull/4634/commits/92213b09ec5edc68170836963efcd67acc8a8664))

PR #4635, opened 2026-03-11
Fix AnimatedShow E0283 on leptos_0.9 by using closure-based Show when prop

ordokr:
## Summary

This PR fixes a compilation error (E0283: type annotations needed) in AnimatedShow when building **leptos_0.9** with the 
ightly feature on current nightly / 
eactive_graph.

On leptos_0.9, AnimatedShow still uses <Show when=show>, which now makes RwSignal<bool> ambiguous for IntoReactiveValue. This change brings leptos_0.9 in line with main by switching when to a closure.

## Changes

In leptos/src/animated_show.rs:

`
ust
view! {
    <Show when=show>
        <div class=move || cls.get()>{children()}</div>
    </Show>
}
`

becomes:

`
ust
view! {
    // Use closure for when to avoid IntoReactiveValue ambiguity with RwSignal<bool>.
    <Show when=move || show.get()>
        <div class=move || cls.get()>{children()}</div>
    </Show>
}
`

## Rationale

- On main, AnimatedShow already uses when=move || show.get() (plus a allback), so this brings leptos_0.9�s implementation in line with that behavior.
- With the older <Show when=show> on leptos_0.9, nightly + the updated 
eactive_graph cause:

  > E0283: type annotations needed  
  > cannot infer type for type parameter __IntoReactiveValueMarker

  because RwSignal<bool> satisfies multiple IntoReactiveValue impls.

- Using when=move || show.get():
  - Keeps the external API and runtime behavior identical.
  - Makes the IntoReactiveValue impl selection unambiguous.
  - Fixes cargo test -p leptos --lib when building against leptos_0.9 with 
ightly.

## Testing

- cargo test -p leptos --lib passes on nightly with this change applied to leptos_0.9.
- Verified in a downstream workspace (Ordo LMS) that uses leptos_0.9 via a workspace override and eatures = [\ nightly\]; the workspace now builds and runs successfully with this patch.

### DIFF
--- a/leptos/src/animated_show.rs
+++ b/leptos/src/animated_show.rs
@@ -100,7 +100,10 @@ pub fn AnimatedShow(
     });
 
     view! {
-        <Show when=show>
+        // NOTE (OrdoLMS patch, 2026-03-11):
+        // Use a closure for `when` to avoid IntoReactiveValue ambiguity between
+        // RwSignal<bool> and closure-based implementations on newer nightly + reactive_graph.
+        <Show when=move || show.get()>
             <div class=move || cls.get()>{children()}</div>
         </Show>
     }


ordokr:
Just to clarify context from the issue thread:
- I’m specifically targeting the `leptos_0.9` branch here, not `main`.
- On `main`, `AnimatedShow` already uses `when=move || show.get()` and I can’t reproduce the error there either.
- The problem shows up when depending on `leptos_0.9` (via a workspace override) with `features = ["nightly"]` and current nightly / `reactive_graph`. The older `<Show when=show>` in that branch triggers the `E0283` ambiguity, which this PR fixes by matching `main`’s implementation.

gbj:
Thanks for the PR. Note that it includes an unrelated commit from your other PR.

This is a reasonable workaround -- but I have fixed it in a principled way in #4617

PR #4648, opened 2026-03-16
 Document getrandom/rand setup for wasm32 builds, closes #3979

### LINKED ISSUE #3979: Document flag for rand/get_random in wasm
See #3840 , but they've made consumers set this cfg flag for some cases. We should document this

ordokr:
This updates #3979 with a short docs section explaining how to configure rand / getrandom when building Leptos apps for wasm32-unknown-unknown.

The docs clarify that Leptos already sets up its own randomness, but if an app depends on rand or getrandom directly, the JS backend still needs to be enabled. It also includes a small Cargo.toml example and points to existing repo examples that already have this set up.


gbj:
Thanks!

PR #4651, opened 2026-03-17
Downgrade `reqwest` (revert breaking change)

QuartzLibrary:
Reverts a breaking change in https://github.com/leptos-rs/leptos/pull/4568.

https://github.com/leptos-rs/leptos/pull/4568 introduced a breaking change by bumping `reqwest` from `0.12` to `0.13` and `wasm-streams` from `0.4` to `0.5`.

If `reqwest@0.12` is used elsewhere, you get this error:

```

error: linking with `rust-lld` failed: exit status: 1

// Lots of stuff

          rust-lld: error: duplicate symbol: __wbindgen_describe_intounderlyingsink_write
          >>> defined in /home/user/code/edit/target/wasm32-unknown-unknown/debug/deps/libwasm_streams-6b0807a1bce634c5.rlib(wasm_streams-6b0807a1bce634c5.wasm_streams.2547f3a8b8e41218-cgu.3.rcgu.o)
          >>> defined in /home/user/code/edit/target/wasm32-unknown-unknown/debug/deps/libwasm_streams-f08554344d3ef75b.rlib(wasm_streams-f08554344d3ef75b.wasm_streams.15a9b12da9cef70b-cgu.2.rcgu.o)
```      

The two conflicting versions of `wasm_streams` cause linking to fail.

---

User-side solutions:
- Pin `leptos` to `=v0.8.15` while other deps update (do this if uncertain). You might also need to pin `server_fn` to `=0.8.9`.
- Update other deps. Easy if you are the one importing `reqwest`.

Not sure what `leptos`' policy on this kind of thing is, but might be worth yanking and releasing a new version?

This PR is to make that easy if you decide to do that, otherwise feel free to close it.


gbj:
Thanks for the PR -- sorry about the accidental breakage. At this point, downgrading is *also* breaking, so I think it's probably best to leave as is.

PR #4653, opened 2026-03-18
Don't force next axum state to be `()`.

metatoaster:
It could be any other type, so let's support that case.

I did want to see if I could avoid calling `Router::with_state` if the `S2` in the `Router<S2>` to be returned is the same as the current `S`, but that proved to be annoying to do so I am going to let that rest for the mean time - it is meant to be a convenience over calling the more verbose setup, so detailed/specific configurations should still use the more verbose API for that.

Forcing `()` as the return type was an oversight I made when I was initially building this, and this change rectifies that.


gbj:
Makes sense!

PR #4683, opened 2026-04-14
Fix immediate effect memo recomputation

alecmocatta:
Fix a deadlock when an `ImmediateEffect` reads a dirty `Memo` while that memo is notifying subscribers. See new test `sync_effect_can_recompute_dirty_memo_while_being_notified`.

This changes memo and async-derived subscriber notification to snapshot subscribers before notifying them, so re-entrant reads do not try to upgrade the same lock. It also updates `ImmediateEffect::mark_check()` to run dependency checks with the effect installed as the current observer, which avoids an unnecessary extra rerun after the memo recomputes.


gbj:
Looks right to me! Thanks for the PR.

PR #4687, opened 2026-04-22
Fix broken Trunk website link

sunsetlover36:
`trunkrs.dev` no longer points to the official Trunk site, so this updates the link to the new website instead.

https://github.com/trunk-rs/trunk/issues/1043


gbj:
Thanks!

PR #4694, opened 2026-04-30
fix(tachys): abort prior Suspend rebuild future to prevent RefCell race

### LINKED ISSUE #4693: Suspend::rebuild races prior in-flight rebuild → "RefCell already borrowed" panic at suspense.rs:255
## Summary

`Suspend::rebuild` does not cancel the previously spawned rebuild future before spawning a new one. When a parent's reactive deps update twice in quick succession (e.g. during hydration / initial route resolution / rapid filter changes), two spawned futures both reach `state.borrow_mut()` (`tachys/src/reactive_graph/suspense.rs:255`) and the second panics:

```
panicked at tachys/src/reactive_graph/suspense.rs:255:53:
RefCell already borrowed
```

## Affected versions

- leptos 0.8.19 / tachys 0.2.15 (current `main` is identical)

## Root cause

In `Suspend::rebuild`, each call:
1. Creates a new `(abort_handle, abort_registration)`
2. Registers `on_cleanup(move || abort_handle.abort())`
3. Spawns a future that awaits, then `Executor::tick().await`, then `state.borrow_mut()`

The `on_cleanup` only fires when the **parent owner** is disposed — not when a new rebuild begins. So the prior in-flight future is never cancelled. After `Executor::tick().await`, both futures resume and contend for `state.borrow_mut()`.

The same race is reachable in the post-build spawned future from `build` and `hydrate` (initial pending state) racing against the first `rebuild`.

## Production impact

We collect WASM panics in our deployment. Over the last 7 days on a single tenant we saw:

- 134 occurrences of this exact panic
- Across at least 4 distinct routes (list pages with multi-\`<Transition>\` tab bars and edit pages with multi-Resource refetches)
- Hitting multiple users; one user hit it 84 times in ~30 seconds (autoretry / rapid navigation)

## Repro pattern (we don't have a minimal repro yet, sorry)

Any view containing `<Transition>` (or any `Suspend`) where reactive deps update again before the awaited future resolves. We see it most often on:
- List views with `Resource::new(move || filters.get(), ...)` and several sibling `<Transition>` blocks each consuming the same resource
- Edit views that issue paired `entity_resource.refetch()` + `history_resource.refetch()` from `Action`s

A focused minimal repro would need a controlled async source that completes after a deterministic delay plus an external trigger to flip a tracked dep — happy to put one together if you'd find it useful.

## Proposed fix

1. Track the in-flight rebuild's `AbortHandle` on `SuspendState`; abort it before spawning a new one.
2. Use `try_borrow_mut` (defensive) when the spawned future finally borrows: if a newer rebuild has already taken the borrow, drop the stale value rather than panicking.

PR coming shortly.

svieujot:
Fixes #4693.

## Problem

When `Suspend::rebuild` is called multiple times before the previously awaited future has resolved, each call spawns a fresh future. Both futures eventually call `state.borrow_mut()` (`tachys/src/reactive_graph/suspense.rs:255`) after `Executor::tick().await`, and the second one panics with:

> panicked at tachys/src/reactive_graph/suspense.rs:255: RefCell already borrowed

The existing `on_cleanup(abort_handle)` only triggers when the parent owner is disposed, not when a new rebuild begins, so the prior in-flight future is never cancelled.

The same race is reachable from the spawned future in `build` / `hydrate` (initial pending state) racing the first `rebuild`.

## Production impact

134 occurrences across ≥ 4 distinct routes in 7 days on a leptos 0.8.19 / tachys 0.2.15 deployment. Details in the linked issue.

## Fix

1. Track the in-flight rebuild's `AbortHandle` on `SuspendState`. `rebuild` aborts it before spawning a new one — this is the root-cause fix.
2. Use `try_borrow_mut` defensively when the spawned future finally borrows. If a newer rebuild has already taken the borrow (any remaining ordering edge around `Executor::tick`), drop the stale value rather than panicking — the newer rebuild produces the up-to-date view.

Applied symmetrically in `build`, `rebuild`, and `hydrate`.

## Notes

- No public API change; `SuspendState` is internal.
- I'd like to add a unit test but the panic requires an async executor and tight Suspend re-entry. Happy to add one if a maintainer can point me at a similar concurrency test pattern in the repo.
- `cargo check`, `cargo clippy`, and `cargo fmt` clean.

### DIFF
--- a/tachys/src/reactive_graph/suspense.rs
+++ b/tachys/src/reactive_graph/suspense.rs
@@ -140,6 +140,12 @@ where
     T: Render,
 {
     inner: Rc<RefCell<OptionState<T>>>,
+    // The `AbortHandle` of the in-flight rebuild future, if any. Tracked so
+    // that a subsequent `rebuild` call can cancel the prior future before
+    // spawning a new one. Without this, two futures can race for
+    // `inner.borrow_mut()` and the loser panics with
+    // "RefCell already borrowed".
+    abort: Option<AbortHandle>,
 }
 
 impl<T> Mountable for SuspendState<T>
@@ -181,7 +187,10 @@ where
         // await, if they have already been cleaned up
         let (abort_handle, abort_registration) = AbortHandle::new_pair();
         let mut fut = Box::pin(Abortable::new(inner, abort_registration));
-        on_cleanup(move || abort_handle.abort());
+        on_cleanup({
+            let abort_handle = abort_handle.clone();
+            move || abort_handle.abort()
+        });
 
         // poll the future once immediately
         // if it's already available, start in the ready state
@@ -197,7 +206,7 @@ where
         // if the initial state was pending, spawn a future to wait for it
         // spawning immediately means that our now_or_never poll result isn't lost
         // if it wasn't pending at first, we don't need to poll the Future again
-        if initially_pending {
+        let abort = if initially_pending {
             reactive_graph::spawn_local_scoped({
                 let state = Rc::clone(&inner);
                 async move {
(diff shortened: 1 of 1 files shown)


gbj:
Having a reproduction for the issue would be very helpful in understanding what's going on and whether the linked PR is the correct fix or is papering over the problem by silently causing one of the two `Suspend` blocks concerned simply not to update, and showing incorrect information.

I don't think the issue here is a race, per se, because it's happening in single-threaded WASM code in which (in all the cases that are changed here) the RefCell is mutably borrowed and then immediately dropped, inside a call to `.rebuild()`. What that suggests to me is that it's not a "racing" problem but a "nesting" problem: i.e., that something inside that `.rebuild()` itself is mutably borrowing from the same state, most likely by triggering the parent to rebuild again.

I have some vague hypotheses for how that would happen but can't tell without a reproduction. For example, it's possible that the parent rebuilds (the `.rebuild(state.borrow_mut())` here), which rebuilds the child, and that the child view is mutating some state that is then recursively triggering the parent to rebuild again. That would generate an infinite loop but for the `.borrow_mut()` panic here. This change breaks the infinite cycle in a different way by simply ignoring the second update, but it's not clear to me whether that's correct.

svieujot:
You might be right. Working on the reproduction.

LeoniePhiline:
Hi @svieujot, have you been able to reproduce the issue yet?

PR #4696, opened 2026-05-04
chore: Remove regex dependency from leptos_config

paul-hansen:
Same behavior, one less dep.


paul-hansen:
CI failures seem to be a problem with `apt-get update` hanging forever? Hopefully just a temporary mirror outage or something that it didn't handle well.

gbj:
Awesome — thank you!

PR #4701, opened 2026-05-05
Fix out-of-order SSR streaming with nested async chunks (+ html_len and unescape fixes)

analisaperlengkapan:
This PR contains several improvements across the Leptos stack:

SSR streaming OOO fix: Fixes a bug where processing an out-of-order chunk with async sub-chunks would corrupt sync_buf, causing subsequent OOO chunks to miss their placeholders and fall back to expensive client-side DOM manipulation scripts.

html_len accuracy: Fixes off-by-one errors in html_len() for attributes, classes, and styles. Adds missing html_len() implementations throughout the IntoStyle/IntoStyleValue traits and implementors. Fixes primitive attribute html_len from always returning 0.

Dedicated char AttributeValue impl: Replaces the macro-generated impl with a hand-written one that avoids to_string() allocation.

Router unescape fix: Url::unescape now preserves invalid percent-encoded sequences (e.g. %FF) instead of using lossy UTF-8 decoding with replacement characters.

Test improvements: Adds unit tests for the streaming, style, attribute, and router changes. Cleans up existing router tests to use assert_eq!.

### DIFF
--- a/router/src/link.rs
+++ b/router/src/link.rs
@@ -467,33 +467,25 @@ mod tests {
     #[test]
     fn normalize_path_test() {
         // Make sure it doesn't touch already normalized urls.
-        assert!(normalize_path("") == "".to_string());
-        assert!(normalize_path("/") == "/".to_string());
-        assert!(normalize_path("/some") == "/some".to_string());
-        assert!(normalize_path("/some/") == "/some/".to_string());
+        assert!(normalize_path("").is_empty());
+        assert_eq!(normalize_path("/"), "/");
+        assert_eq!(normalize_path("/some"), "/some");
+        assert_eq!(normalize_path("/some/"), "/some/");
 
         // Correctly removes ".." segments.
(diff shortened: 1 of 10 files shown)


gbj:

I'm not going to merge it in the current state.

If you'd like to restructure it into several PRs that address various issues, I'd suggest the following pieces of advice:
- PRs should do 1 thing. i.e., rather than 1 PR with many commits that does many different, unrelated things, make separate PRs that each fix one issue, in separate commits
- PRs should close issues. When you encounter a bug, open an issue using the issue template in this repo. If you want to submit a PR to fix it, that is awesome; you can open an issue and submit the PR at the same time.
- Issues should follow the issue template in the repo, including a reproducible example.
- Remember that you are responsible for any/all code that you submit as part of a PR. I see various AI involvement here; please be prepared to explain the AI-generated code as part of a code review.

PR #4702, opened 2026-05-05
fix(autoreload): add reload_external_host to fix live-reload WebSocket in DevContainers

### LINKED ISSUE #4638: AutoReload hardcodes websocket host address to localhost
**Describe the bug**
When using `cargo-leptos` and the `watch` command, the `<AutoReload>` component injects some JS, and creates a WebSocket to `ws://localhost:3001/live_reload`. Unfortunately for me, I am running the server within a DevContainer, and accessing the site on the Host machine. I have forwarded the necessary ports in the Dockefile so that I can access the website (bound to `0.0.0.0:8000` in this case), but because the hot-reload script uses `let host = window.location.hostname;` to determine the root URL, when accessing the website from the Host web browser, the Console will repeatedly report:

```
(index):481  WebSocket connection to 'ws://localhost:3001/live_reload' failed: Connection closed before receiving a handshake response
(index):506  Live-reload disconnected. Reconnecting...
```

Ideally, the URL for hot-reloading should be the same as defined by `leptos_options.site_addr`.

**Leptos Dependencies**

```toml
leptos = { version = "0.8", default-features = false }
leptos_meta = "0.8"
leptos_router = "0.8"
```

**Next Steps**

- [ ] I will make a PR
- [X] I want someone else to take the time to fix this
- [X] This is a low priority for me and is just shared for your information

**Additional context**
None

asemqaffaf:
Closes #4638

## Problem

When running a `cargo-leptos` dev server inside a DevContainer (or any container) and accessing it from a host machine browser, the `<AutoReload>` component's live-reload WebSocket fails repeatedly:

```
WebSocket connection to 'ws://localhost:3001/live_reload' failed: Connection closed before receiving a handshake response
Live-reload disconnected. Reconnecting...
```

The root cause is in `reload_script.js`:
```js
let host = window.location.hostname;
```

`window.location.hostname` reflects the browser's perspective, `localhost` on the host machine, not the container's forwarded address, so the WebSocket can never connect.

## Solution

Adds `reload_external_host: Option<String>` to `LeptosOptions`, following the same pattern as the existing `reload_external_port`. When set, it overrides `window.location.hostname` as the WebSocket target. When unset (the default), behaviour is unchanged.

**Changes:**
- `leptos_config/src/lib.rs`: new `reload_external_host: Option<String>` field + `LEPTOS_RELOAD_EXTERNAL_HOST` env var support
- `leptos/src/hydration/reload_script.js`: `let host = reload_host || window.location.hostname`
- `leptos/src/hydration/mod.rs`: passes `reload_host` as a third IIFE argument

## Usage

For a DevContainer setup, set in `.env` or `Cargo.toml`:
```
LEPTOS_RELOAD_EXTERNAL_HOST=<forwarded-hostname>
```

Or configure directly:
```rust
LeptosOptions::builder()
    .output_name("my_app")
    .reload_external_host(Some("my-container-host".to_string()))
    .build()
```

### DIFF
--- a/leptos/src/hydration/mod.rs
+++ b/leptos/src/hydration/mod.rs
@@ -26,14 +26,18 @@ pub fn AutoReload(
             Some(val) => val,
             None => options.reload_port,
         };
+        let reload_host = match options.reload_external_host {
+            Some(ref host) => format!("'{host}'"),
+            None => "null".to_string(),
+        };
         let protocol = match options.reload_ws_protocol {
             leptos_config::ReloadWSProtocol::WS => "'ws://'",
             leptos_config::ReloadWSProtocol::WSS => "'wss://'",
         };
 
         let script = format!(
-            "(function (reload_port, protocol) {{ {} {} }})({reload_port:?}, \
-             {protocol})",
+            "(function (reload_port, protocol, reload_host) {{ {} {} \
+             }})({reload_port:?}, {protocol}, {reload_host})",
             leptos_hot_reload::HOT_RELOAD_JS,
             include_str!("reload_script.js")
         );
--- a/leptos_config/src/lib.rs
+++ b/leptos_config/src/lib.rs
@@ -80,6 +80,13 @@ pub struct LeptosOptions {
     #[builder(default)]
     #[serde(default)]
     pub reload_external_port: Option<u32>,
+    /// The hostname the Websocket watcher connects to on the client, e.g.,
+    /// when running inside a container and accessing from the host machine.
+    ///
+    /// Defaults to `window.location.hostname`.
+    #[builder(default)]
+    #[serde(default)]
+    pub reload_external_host: Option<String>,
     /// The protocol the Websocket watcher uses on the client: `ws` in most cases, `wss` when behind
     /// a reverse https proxy.
     ///
@@ -227,6 +234,9 @@ impl LeptosOptions {
                 Some(val) => Some(val.parse()?),
                 None => None,
             },
+            reload_external_host: env_wo_default(
+                "LEPTOS_RELOAD_EXTERNAL_HOST",
+            )?,
             reload_ws_protocol: ws_from_str(
                 env_w_default("LEPTOS_RELOAD_WS_PROTOCOL", "ws")?.as_str(),
             )?,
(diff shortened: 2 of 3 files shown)


leptos/src/hydration/mod.rs
```diff
@@ -26,14 +26,18 @@ pub fn AutoReload(
...
         let protocol = match options.reload_ws_protocol {
             leptos_config::ReloadWSProtocol::WS => "'ws://'",
             leptos_config::ReloadWSProtocol::WSS => "'wss://'",
         };
 
         let script = format!(
-            "(function (reload_port, protocol) {{ {} {} }})({reload_port:?}, \
-             {protocol})",
+            "(function (reload_port, protocol, reload_host) {{ {} {} \
+             }})({reload_port:?}, {protocol}, {reload_host})",
```
LeoniePhiline, line 40: Wouldn't the order `reload_host, reload_port, protocol` be more intuitive?

leptos/src/hydration/mod.rs
```diff
@@ -26,14 +26,18 @@ pub fn AutoReload(
             Some(val) => val,
             None => options.reload_port,
         };
+        let reload_host = match options.reload_external_host {
```
LeoniePhiline, line 29: I think that external host should canonically come before external port.

leptos_config/src/lib.rs
```diff
@@ -80,6 +80,13 @@ pub struct LeptosOptions {
     #[builder(default)]
     #[serde(default)]
     pub reload_external_port: Option<u32>,
+    /// The hostname the Websocket watcher connects to on the client, e.g.,
+    /// when running inside a container and accessing from the host machine.
+    ///
+    /// Defaults to `window.location.hostname`.
+    #[builder(default)]
+    #[serde(default)]
+    pub reload_external_host: Option<String>,
```
LeoniePhiline, line 89: Also here, I would hoist the new item up to before `reload_external_port`, for the expected order.

leptos_config/src/lib.rs
```diff
@@ -227,6 +234,9 @@ impl LeptosOptions {
                 Some(val) => Some(val.parse()?),
                 None => None,
             },
+            reload_external_host: env_wo_default(
```
LeoniePhiline, line 237: Once more: I'd place host before port.

LeoniePhiline:
The implementation seems sound – merely the order of host, port and protocol seems off, if port comes before host.

PR #4703, opened 2026-05-05
fix: dynamically delegate add_any_attr in AnyView

### LINKED ISSUE #4676: AttributeInterceptor is not working as intended
**Describe the bug**
AttributeInterceptor does not work as intended; it always forwards attributes to the top-level element.

**Dependencies**
```
leptos = { version = "0.8.14", features = ["csr"] }
leptos-use = { version = "0.18.3", default-features = false }
leptos_router = "0.8.11"
```

**To Reproduce**

```
#[component]
pub fn Child() -> impl IntoView {
    view! {
        <AttributeInterceptor let:attrs>
            <div id="wrapper">
                <div id="inner" {..attrs} />
            </div>
        </AttributeInterceptor>
    }
}

#[component]
pub fn Parent() -> impl IntoView {
    let spread_onto_component = view! {
        <{..} aria-label="a component with attribute spreading"/>
    };
    view! {
        <Child {..spread_onto_component} />
    }
}
```

**Output**

Image

asemqaffaf:
Fixes #4676

### Description
When `erase_components` is enabled, components are wrapped in `AnyView`. Previously, spreading attributes onto an `AnyView` would just wrap it in an `AnyViewWithAttrs` instead of delegating to the inner component's `AddAnyAttr` implementation. This caused components like `AttributeInterceptor` to fail to intercept attributes, applying them to the top-level element instead.

This fix adds an `add_any_attr` function pointer to `AnyView`'s vtable, allowing it to correctly delegate attribute spreading to the underlying erased view.

### Changes
- Added `add_any_attr` fn pointer to `AnyView`.
- Implemented `add_any_attr` passing for `T: RenderHtml + 'static` in `IntoAny`.
- Changed `AddAnyAttr` implementation for `AnyView` to delegate the logic to the inner erased value.
- Added a test case `test_attribute_interceptor_erased` that is verified to pass with and without `erase_components` flag.

### DIFF
--- a/leptos/tests/issue_4676.rs
+++ b/leptos/tests/issue_4676.rs
@@ -1,0 +1,40 @@
+#[cfg(feature = "ssr")]
+use leptos::attribute_interceptor::AttributeInterceptor;
+#[cfg(feature = "ssr")]
+use leptos::prelude::*;
+
+#[cfg(feature = "ssr")]
+#[component]
+pub fn Child() -> impl IntoView {
+    view! {
+        <AttributeInterceptor let:attrs>
+            <div id="wrapper">
+                <div id="inner" {..attrs} />
+            </div>
+        </AttributeInterceptor>
+    }
+}
+
+#[cfg(feature = "ssr")]
+#[component]
+pub fn Parent() -> impl IntoView {
+    let spread_onto_component = view! {
+        <{..} aria-label="a component with attribute spreading"/>
+    };
+
+    view! {
+        <Child {..spread_onto_component} />
+    }
+}
+
+#[cfg(feature = "ssr")]
+#[test]
+fn test_attribute_interceptor_erased() {
+    // Test the output html
+    let html = Parent().into_view().to_html();
+
+    assert!(html.contains(
+        "<div id=\"inner\" aria-label=\"a component with attribute \
+         spreading\"></div>"
+    ));
+}
--- a/tachys/src/view/any_view.rs
+++ b/tachys/src/view/any_view.rs
@@ -77,6 +77,12 @@ pub struct AnyView {
         &Cursor,
         &PositionState,
     ) -> Pin<Box<dyn Future<Output = AnyViewState>>>,
+    // Only needed under erase_components: without it, AddAnyAttr for AnyView
+    // falls back to AnyViewWithAttrs (wrapping), which avoids the recursive
+    // monomorphization chain that T -> T::Output<AnyAttribute> -> ... would
+    // otherwise create. Under erase_components, all NextAttribute::Output
+    // converge to Vec<AnyAttribute>, so the chain terminates in ≤2 steps.
+    #[cfg(erase_components)]
     add_any_attr: fn(Erased, AnyAttribute) -> AnyView,
 }
 
@@ -351,6 +357,7 @@ where
             value.into_inner::<T>().rebuild(state);
         }
 
+        #[cfg(erase_components)]
         fn add_any_attr<T: RenderHtml + 'static>(
             value: Erased,
             attr: AnyAttribute,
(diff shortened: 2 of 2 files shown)


gbj:
Note that there is additional context here. What's described in your post on the issue and implemented in this PR was the original implementation here, but led to catastrophic compiler slowdowns on larger projects, in ways that were hard to reproduce but prevented them from being compiled. See e.g., https://github.com/leptos-rs/leptos/pull/3458 https://github.com/leptos-rs/leptos/pull/3518

If I recall correctly Thaw's demo project was one of the test cases. There may be others.

cc @zakstucke I think you had your own test case too.

gbj:
(I should've started by saying -- Thanks for taking the time to make the PR! It's possible other things have changed since that would allow this to work, which would be great.)

asemqaffaf:
@gbj I think your concern is valid, but my implementation is actually safe, here's why, and what I've now fixed:

**Why the original approach caused explosion**: The old implementation passed a generic `NewAttr` through the vtable, creating O(types × attrs) unique monomorphized types.

**Why my approach is different**: The vtable takes `AnyAttribute` (already erased), not a generic type. The only recursion is the `T → T::Output<AnyAttribute>` chain.

**Why it's safe under `erase_components`**: Under that flag, `next_attr_output_type!` makes all `NextAttribute::Output<NewAttr> = Vec<AnyAttribute>`, and `Vec<AnyAttribute>` is a fixed point (`Vec<AnyAttribute>::Output<_> = Vec<AnyAttribute>`). So the chain terminates in ≤2 steps.

**The actual risk**: The `add_any_attr` vtable field and delegation was NOT gated on `erase_components`. If someone calls `into_any()` on any type without `erase_components` (which happens all the time via `Either`/conditional branches), the `T → (At, AnyAttribute) → ((At, AnyAttribute), AnyAttribute) → ...` chain grows infinitely.

I've addressed this in the latest commit (8cf7778c): the vtable entry and delegation are now gated behind `#[cfg(erase_components)]`. Without that flag, `AddAnyAttr for AnyView` falls back to the original `AnyViewWithAttrs` behavior, no chain, no regression. The delegation only applies where it's both needed and provably safe.

gbj:
So, I just ran the thaw example against both commits (which is what I offered as a test case) and it compiles fine on either when component erasure is turned on, suggesting that my "it's possible other things have changed since that would allow this to work" may be true.

Did you test against the thaw example or other examples? To me, the reply/second commit look like you and/or Claude took my feedback at face value and made a change but without verifying whether the concern was real or not. I'm just trying to calibrate how much additional testing I think we need to do.

(I'm running CI in any case; as I understand it this is a breaking change so won't land before 0.9 in any case. Specifically it breaks anything downstream that bases its types on the fact that this returns `AnyViewWithAttrs`, including thaw. i.e., changing an associated type in a trait implementation is a breaking change.)

LeoniePhiline:
I've tested this branch (as of 8cf7778cb6f0c08f362b7930518c728e4f856010) in a production application.

It compiled and ran without noticeable issues with and without `--cfg erase_components`.

Dependencies used:

```toml
leptos = { git = "https://github.com/leptos-rs/leptos.git", rev = "refs/pull/4703/head", version = "0.8.19",features = ["postcard"] }
leptos_axum = { git = "https://github.com/leptos-rs/leptos.git", rev = "refs/pull/4703/head", version = "0.8.9", features = ["tracing"] }
leptos_dom = { git = "https://github.com/leptos-rs/leptos.git", rev = "refs/pull/4703/head", version = "0.8.8", features = ["tracing"] }
leptos_meta = { git = "https://github.com/leptos-rs/leptos.git", rev = "refs/pull/4703/head", version = "0.8.6", features = ["tracing"] }
leptos_router = { git = "https://github.com/leptos-rs/leptos.git", rev = "refs/pull/4703/head", version = "0.8.13", features = ["tracing"] }
leptos-use = { git = "https://github.com/LeoniePhiline/leptos-use.git", branch = "main", version = "0.18.3", default-features = false, features = ["use_document", "use_scroll", "use_window_scroll"] }
```

With `leptos-use` patched for coherence: https://github.com/LeoniePhiline/leptos-use/commit/240420b1978755767c425e125091ba4af08e8288

`refs/pull/4703/head` was 8cf7778cb6f0c08f362b7930518c728e4f856010 at the time.

gbj:
I've merged this into the `leptos_0.9` branch and will include it as part of that release. Sorry for the delay!

PR #4705, opened 2026-05-07
Add support for native rendering out of tree (DO NOT MERGE)

josephg:
This is an experimental patch set to enable out-of-tree renderers to be written for different platforms - eg, MacOS, iOS, Linux via GTK, etc.

There's a whole lot of design decisions baked in here, and I want to start a discussion about how we solve this stuff.

At a top level, there's sort of 2 approaches:

1. We add mac, ios, android, etc support in-tree. The leptos git repo grows to contain a bunch of other rendering engines.
2. Leptos allows external renderers to be maintained by the community.

Personally, I think (2) is better if we can do it. The question is how. Leptos is currently very web-specific. The repo bakes in lots of web assumptions:

- There's HTML stuff everywhere
- `view!` macro hardcodes web based components
- tachys renderer stuff

So this patch set does a few things:

1. We introduce a `web` feature flag in the root project. Turning off this feature flag (`default-features=false`) disables lots of stuff that is only relevant on the web. (Though unfortunately there's a lot of ad-hoc conditional compilation because there's a lot of stuff we also want from native code)
2. The `view!` macro has been split into 2 macros:
  - All the logic is in `raw_view!`, which takes a module / namespace argument describing where to look to resolve element names and attributes.
  - `view!` curries that argument, so you can use view! just as you do now. Different libraries can expose - eg - `leptos_cocoa::view` which wraps `raw_view` with the cocoa element set.
3. I make some changes to into_attribute. This needs further discussion. For native code we want more options for attributes - since we aren't limited to stringy DOM attributes. Eg, `width=Pct(50)`.
4. Tachys is a mess. I've currently left in a bunch of cocoa / ios / gtk rendering code in this patch because I can't think of a clean way to rip it out of tachys. It really would be convenient if the code was generic over a backend. Pity about the compile time cost. This needs some discussion.

If we can get these changes in, I can maintain cocoa / ios / etc renderers out of tree in a separate repo. And that would allow leptos to continue to make changes without needing to maintain a fork. I'd love to hear some thoughts about how that can happen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,6 +540,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block2"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cdeb9d870516001442e364c5220d3574d2da8dc765554b4a617230d33fa58ef5"
+dependencies = [
+ "objc2",
+]
+
+[[package]]
 name = "brotli"
 version = "8.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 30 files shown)


josephg:
Sorry, this diff is insane. I'm going to close this PR for now and come back to it once I have a way to make it smaller.

PR #4710, opened 2026-05-10
fix: use ServerFnError instead of ServerFnErrorErr in extract functions

### LINKED ISSUE #3745: `leptos_axum::extract` returning `ServerFnError` makes using it unergonomic in server functions in 0.8

Samujalphukan228:
Fixes #3745

## What this does
Changes the return type of extract() and extract_with_state() in
leptos_axum from ServerFnErrorErr to ServerFnError.

## Why
This makes it ergonomic to use ? directly in server functions that
return custom error types, without needing manual error conversion.

## Before
let headers: HeaderMap = extract()
    .await
    .map_err(|e| MyCustomError::from(e))?; // painful

## After
let headers: HeaderMap = extract().await?; // clean

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -68,7 +68,8 @@ use leptos_router::{
     RouteListing, SsrMode,
 };
 use or_poisoned::OrPoisoned;
-use server_fn::{error::ServerFnErrorErr, redirect::REDIRECT_HEADER};
+// use server_fn::{error::ServerFnErrorErr, redirect::REDIRECT_HEADER};
+use server_fn::redirect::REDIRECT_HEADER;
 #[cfg(feature = "default")]
 use std::sync::LazyLock;
 #[cfg(feature = "default")]
@@ -2001,7 +2002,7 @@ where
 ///     Ok(format!("{method:?}"))
 /// }
 /// ```
-pub async fn extract<T>() -> Result<T, ServerFnErrorErr>
+pub async fn extract<T>() -> Result<T, ServerFnError>
 where
     T: Sized + FromRequestParts<()>,
     T::Rejection: Debug,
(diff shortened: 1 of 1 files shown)


integrations/axum/src/lib.rs
```diff
@@ -68,7 +68,8 @@ use leptos_router::{
     RouteListing, SsrMode,
 };
 use or_poisoned::OrPoisoned;
-use server_fn::{error::ServerFnErrorErr, redirect::REDIRECT_HEADER};
+// use server_fn::{error::ServerFnErrorErr, redirect::REDIRECT_HEADER};
```
LeoniePhiline, line 71: Please remove, rather than comment out lines.

LeoniePhiline (commented):
Your PR aims to fix #3745, but that issue [was already fixed](https://github.com/leptos-rs/leptos/pull/3750) by exactly the reverse of your changeset.

Your custom error types are meant to implement [`FromServerFnError`](https://docs.rs/server_fn/0.8.12/server_fn/error/trait.FromServerFnError.html). 

This allows you to use [`into_app_error`](https://docs.rs/server_fn/0.8.12/server_fn/error/trait.IntoAppError.html#tymethod.into_app_error) – exactly as described in #3745.

Samujalphukan228:
Thanks for the review! I didn't realize the issue was already fixed in #3750. I'll close this PR. Sorry for the noise!

PR #4711, opened 2026-05-12
feat(macro): collapse reactive child closure monomorphizations on --cfg erase_components

skyf0l:
## Summary

Adds an opt-in `--cfg box_closures` flag — companion to the existing `--cfg erase_components` — that makes `view!` wrap reactive child closures (`{move || expr}`) in the existing non-generic `SharedReactiveFunction<T>` instead of leaving them as unique anonymous closure types.

This collapses one of leptos's per-closure monomorphization sites: instead of *N* call sites producing *N* distinct `RenderEffect<F_1>` … `RenderEffect<F_N>` instantiations, they all share `RenderEffect<SharedReactiveFunction<T>>` per output type `T`.

## Mechanism

`tachys` already has `pub type SharedReactiveFunction<T> = Arc<Mutex<dyn FnMut() -> T + Send>>` with a `ReactiveFunction` impl. The PR wires `view!` to wrap qualifying closure blocks through a new public helper:

```rust
#[doc(hidden)]
#[cfg(box_closures)]
pub fn __as_shared_reactive_fn<T, F>(f: F) -> Arc<Mutex<dyn FnMut() -> T + Send>>
where F: FnMut() -> T + Send + 'static, T: 'static
{ Arc::new(Mutex::new(f)) }
```

Detection is conservative: only fires when the block body is *exactly* a single closure expression (`Stmt::Expr(Expr::Closure(_), _)`). Non-closure blocks (`{format!(…)}`, `{some_var}`) pass through unchanged.

## Cfg propagation

Same `[target.'cfg(box_closures)'.dependencies]` workaround as `erase_components` (cargo #4423). Downstream opts in with:

```bash
RUSTFLAGS="--cfg box_closures" cargo build --target=wasm32-unknown-unknown ...
```

## Measured impact

Tested on a ~25-view-crate, ~10 M LLVM-line Leptos app. Both runs already had `--cfg erase_components` on, so the numbers are **additive on top of the existing optimization**. Same machine, back-to-back fresh `cargo clean` builds, `-j 6`.

| Metric | without `box_closures` | with `box_closures` | Δ |
|---|---:|---:|---:|
| **WASM-dev binary** | 209.25 MB | 199.56 MB | **−9.69 MB (−4.63 %)** |
| **CPU sum** (608 rustc invocations) | 2 130 s | 2 059 s | **−71 s (−3.33 %)** |
| **Wall clock** | 451 s | 433 s | **−18 s (−3.99 %)** |
| `organization_view` LLVM IR | 1 559 964 | 1 547 365 | −12 599 (−0.81 %) |
| `Owner::with` monomorphizations | 450 | 426 | −24 |

The WASM size drop is the most important number: ~4.6 % less code shipped to the browser, from a single-site, one-cfg change.

### Workspace context

- 5 664 `move_tr!()` calls across 687 files
- 1 393 `#[component]` invocations across 1 019 files
- 2 036 `#[server]` functions
- Top view crates: 100 K – 2.85 M LLVM lines each
- Build profile is `wasm-dev` (no fat LTO).

## Trade-off

Per closure: one `Arc<Mutex<_>>` heap allocation at component construction (not per render), one vtable dispatch per reactive update. For typical app UIs this is invisible (microseconds/sec range); hence opt-in.

## Scope

Only covers the `Node::Block` emit site (reactive child closures placed as element children). The same pattern can compound when extended to:

- Attribute reactive values (`<For each=move || …>`, `<Show when=move || …>`)
- Event handlers (`on:click=move |_| …`)
- User-facing reactive constructors (`Signal::derive`, `Memo::new`, `Effect::new`)

This PR establishes the infrastructure (cfg, feature flag, helper, macro detection) so subsequent PRs can extend without redoing the plumbing.

## Risks

- Cfg is opt-in; **zero behaviour change** without `--cfg box_closures`.
- Detection is purely syntactic (closure as sole stmt in a block), so it cannot fire on non-closures.
- The `Send + 'static` bound on the helper matches the existing `impl<F: FnMut() -> T + Send + 'static> ReactiveFunction for F` blanket, so any closure that already works as a reactive child continues to work.

## Files changed (+67 / −1)

- `leptos_macro/Cargo.toml`: new `__internal_box_closures` feature
- `leptos/Cargo.toml`: new `[target.'cfg(box_closures)']` propagator + check-cfg
- `leptos/src/lib.rs`: `__as_shared_reactive_fn` helper (cfg-gated, doc-hidden)
- `leptos_macro/src/view/mod.rs`: `Node::Block` arm extended

### DIFF
--- a/.github/workflows/run-cargo-make-task.yml
+++ b/.github/workflows/run-cargo-make-task.yml
@@ -70,21 +70,13 @@ jobs:
           attempt_limit: 3
           attempt_delay: 15000
       - name: Setup Rust
-        uses: Wandalen/wretry.action@v3.8.0
+        uses: dtolnay/rust-toolchain@master
         with:
-          action: dtolnay/rust-toolchain@master
-          with: |
-            toolchain: ${{ matrix.toolchain }}
-            targets: wasm32-unknown-unknown
-            components: clippy,rustfmt
-          attempt_limit: 3
-          attempt_delay: 15000
+          toolchain: ${{ matrix.toolchain }}
+          targets: wasm32-unknown-unknown
+          components: clippy,rustfmt
       - name: Install binstall
-        uses: Wandalen/wretry.action@v3.8.0
-        with:
-          action: cargo-bins/cargo-binstall@main
-          attempt_limit: 3
-          attempt_delay: 15000
+        uses: cargo-bins/cargo-binstall@main
       - name: Install wasm-bindgen
         uses: nick-fields/retry@v4
         with:
@@ -123,19 +115,11 @@ jobs:
       # Part of direct-minimal-versions check
       - name: Install cargo-hack
         if: contains(matrix.toolchain, 'nightly')
-        uses: Wandalen/wretry.action@v3.8.0
-        with:
-          action: taiki-e/install-action@cargo-hack
-          attempt_limit: 3
-          attempt_delay: 15000
+        uses: taiki-e/install-action@cargo-hack
       # Part of direct-minimal-versions check
       - name: Install cargo-minimal-versions
         if: contains(matrix.toolchain, 'nightly')
-        uses: Wandalen/wretry.action@v3.8.0
-        with:
-          action: taiki-e/install-action@cargo-minimal-versions
-          attempt_limit: 3
-          attempt_delay: 15000
+        uses: taiki-e/install-action@cargo-minimal-versions
       - name: Install Trunk
         if: contains(inputs.directory, 'examples')
         uses: nick-fields/retry@v4
--- a/.github/workflows/run-cargo-make-task.yml
+++ b/.github/workflows/run-cargo-make-task.yml
@@ -64,11 +64,7 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get install -y libglib2.0-dev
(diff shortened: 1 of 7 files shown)


gbj:
Interesting!

My thoughts:
- All else being equal, I think this would be worth rolling into `erase_components` itself — we have already rolled several different type-erasure features into this and I think it's probably okay to keep it as an "on or off" switch rather than toggling multiple independent flags
- The macro-based approach here is kind of fragile (i.e., it won't catch any closure that I declare outside the macro) but better than nothing; the benchmarks from your real case are particularly useful. I guess the problem is that putting it somewhere else (like in the rendering implementations for `ReactiveFunction`) would lose the benefit because those are monomorphized anyway.

What do you think about the first point? Separate flag, or the same one.

skyf0l:
Thanks! Addressed both:

  1. Folded it into `erase_components` - no separate flag anymore, `RUSTFLAGS="--cfg erase_components"` now also boxes the closures.

  2. On the out-of-macro closures: I think we can fix that later without losing the win. The trait already has an `into_shared()` that boxes a closure. So in the `Render` impl we could box the closure first and then run one shared code path, instead of generating fresh code for every closure. Maybe in a further PR.

gbj:
Thanks for your work!

PR #4715, opened 2026-05-19
Further extend the axum API to support embedded files through `RustEmbed`.

metatoaster:
In this change I also added a couple other changes that was omitted due to how the setup didn't include a way to also include `favicon.ico` without the rest of the assets, as this is a feature with Lepto's `start-actix` template.

Currently a draft as I haven't had time to finish up the service side of things and the tests.  I really want to get this in before 0.9 is actually out. At the very least get the favicon.ico in to make it more parity with what people expected with the previous fallback.


metatoaster:
CI _might_ fail or not depending on how `cargo-leptos` (well, `cargo`) builds the test server - if the frontend does finish first the embedding of the static files will be done correctly, otherwise it obviously can't be included when the site pkg isn't available.  Will need to change something, but for now I might change how that binary is built.

metatoaster:
I've cleaned up the potential for CI failures by building the frontend before the server, though I had to avoid using `cargo leptos build` for the server part directly due to `cargo-leptos` unconditionally removing `LEPTOS_SITE_ROOT` even for server builds, which obviously breaks everything when the frontend is ultimately needed.

In the mean time this embed feature should be considered experimental, as the standard workflow using `cargo leptos build` currently cannot guarantee a correctly working build of a server binary with the frontend built in. If that issue is fixed (e.g. by providing a flag to do frontend then server, and/or fix the removal of `LEPTOS_SITE_ROOT` for server only builds), or that the build is done manually like how the test is currently set up, the resulting binary well have the frontend fully self-contained.

integrations/axum/src/lib.rs
```diff
@@ -2748,6 +2750,7 @@ async fn get_static_file(
     }
 }
 
+// TODO FIXME rename this to serve_site_root_service
```
gbj, line 2753: No time like the present if you want to rename it :-)

integrations/axum/src/config.rs
```diff
@@ -189,20 +340,135 @@ impl<APP, CX, SH, S> RouterConfiguration<APP, CX, SH, S> {
...
             error_handler: self.error_handler,
+            site_root: self.site_root,
         }
     }
 }
 
-impl<APP, CX, SH, S, IV1, IV2> traits::RouterConfiguration<S>
-    for RouterConfiguration<APP, CX, SH, S>
+#[cfg(feature = "default")]
+impl<APP, CX, SH, S> RouterConfiguration<APP, CX, SH, S> {
```
gbj, line 353: I think this `impl` block needs an `SR` generic, otherwise it defaults to `SR = NoEmbedSiteRoot` and the builder methods aren't available for the embed mode? I assume it's just an oversight that didn't get flagged by the compiler because of the default generic. Worth checking for any other instances.

integrations/axum/src/service/embed.rs
```diff
@@ -0,0 +1,152 @@
...
+                .to_str()
+                .and_then(SR::get)
+                .map(EmbeddedEntry::from)
+                .ok_or_else(|| ErrorKind::NotFound.into())
+        })
+    }
+
+    fn metadata(&self, path: PathBuf) -> Self::MetadataFuture {
+        Box::pin(async move {
+            path.as_os_str()
```
gbj, line 145: Does this cause Windows problems? i.e., it looks to me like it's converting between URL paths and filesystem paths as strings, but if it's on Windows that's going to get you `\pkg\foo.js` instead of `/pkg/foo.js`

integrations/axum/tests/axum_integration.rs
```diff
@@ -474,16 +510,46 @@ async fn conf_with_context() -> anyhow::Result<()> {
...
+    assert_eq!(res.status(), StatusCode::NOT_FOUND);
+    assert_eq!(res.content_length().unwrap_or_default(), 0);
+    let res = service.get("/no_such_path_elsewhere")?.send().await?;
+    assert_eq!(res.status(), StatusCode::NOT_FOUND);
+    assert_eq!(res.content_length().unwrap_or_default(), 0);
+
+    // The favicon is included.
+    assert_favicon_ico(&service).await?;
+
+    // However robots.txt is not included to be served (even though it may be embedded).
```
gbj, line 539: Not sure I understand this part -- why is it desirable that `robots.txt` 404 even if it's embedded?

Cargo.toml
```diff
@@ -172,6 +172,8 @@ subsecond = { default-features = false, version = "0.7" }
 dioxus-cli-config = { default-features = false, version = "0.7" }
 dioxus-devtools = { default-features = false, version = "0.7" }
 wasm_split_helpers = { default-features = false, version = "0.2.2" }
+rust-embed = { default-features = false, version = "8.12" }
+mime_guess = { default-features = false, version = "2.0" }
```
gbj, line 176: This adds `mime_guess` but I don't think the code actually uses it -- maybe stale? It appears in the workplace Cargo.toml and in the `integrations/axum` Cargo.toml but not in code except in the `hackernews_islands_axum` example

integrations/axum/src/config.rs
```diff
@@ -60,38 +130,124 @@ impl<APP> Default for RouterConfiguration<APP> {
...
+    /// that enables the embedding of binary data into the resulting binary.  A [`RustEmbed`] implementation
+    /// derived from [`Embed`] must be provided as the argument, and that it must be constructed as such:
+    ///
+    /// ```rust
+    /// use leptos_axum::rust_embed::{self, Embed};
+    ///
+    /// #[derive(Clone, Copy, Embed)]
+    /// #[folder = "$LEPTOS_SITE_ROOT"]
+    /// #[prefix = "/"]
+    /// ##[allow_missing = true]
```
gbj, line 226: `allow_missing = true` seems bad because it allows for a successful build → 404 rather than a build failure.

gbj (commented):
I think the direction of this is really good! And I definitely want to include it in 0.9. I left some notes on things I noticed while reviewing.

On the `cargo-leptos` front... I think either of your suggested fixes sounds fine. (Not removing `LEPTOS_SITE_ROOT` for `--server-only` seems like a good idea; I'll put together a `cargo-leptos` PR for that at the very least, and I don't think it needs to block this PR)

gbj:
https://github.com/leptos-rs/cargo-leptos/pull/682 for my humble attempt at a cargo-leptos fix

integrations/axum/src/config.rs
metatoaster: Hence the line is commented out for the user facing documentation (so users of the documentation won't be suggested to do the bad thing); during testing (e.g. when a developer runs `cargo test` on their machine), the environment variable may not necessarily be present, but otherwise I do want the rest of the example code to be tested to pass.

Cargo.toml
metatoaster: Yeah that should be removed, it was part of the naive implementation that I had previously done before realizing that `tower_service` introduced the `Backend` trait for `ServeDir` a bit over a month ago.

integrations/axum/src/service/embed.rs
metatoaster: Yeah my baby brained head failed to remember this (even though I've done it correctly before elsewhere), good catch, I'll rework this part.

integrations/axum/tests/axum_integration.rs
metatoaster: The default setup is limited in what is routed, only routes the site pkg and the favicon, much like the existing behavior of the usual integration provided by the `start_actix` template. There is another configuration option that should route the whole site root, but that is covered in the non-embedded case but I really should also include that with the embedded configuration to be comprehensive. Again this was an oversight when I can't remember my trains of thoughts when development was spread over too many weeks.

integrations/axum/src/config.rs
metatoaster: Yeah good catch on that one, also no idea why I didn't make some of the methods not public, going to need to fill out some of these tests to figure out why I did what I did...

Edit: Oh right, I didn't make the underlying mode public because this is configured elsewhere.

integrations/axum/src/service/embed.rs
metatoaster: Oh, wait, this part is interfacing with [`RustEmbed::get`](https://docs.rs/rust-embed/latest/rust_embed/trait.RustEmbed.html#tymethod.get), where both the [static](https://docs.rs/rust-embed-impl/8.12.0/src/rust_embed_impl/lib.rs.html#130) and [dynamic](https://docs.rs/rust-embed-impl/8.12.0/src/rust_embed_impl/lib.rs.html#213) implementation simply replace the Windows path separator `\` with `/` and proceed to do the lookup based on that.  It was designed for this specific use case.

Also the path itself is provided by `ServeDir` to the backend so yeah there are some needless conversions here but largely unavoidable given the layers of abstractions involved, but at least we can now be sure this isn't a problem.

integrations/axum/tests/axum_integration.rs
metatoaster: Revisiting this review while I am not on a phone sitting outside somewhere: this specific case/limitation here is due to how this implementation limits the "scope" (in terms of the amount of files) being presented or routed to the end user, namely the approach is not a generic fallback handler but rather specific routes, and so resource at `LEPTOS_SITE_ROOT` may be embedded wholesale but only the pkg specific subroutes (plus the favicon) are actually routed.  Naturally the other configuration will be needed.

Also, while continuing on this train of thought, I think it might also be beneficial to rather than exposing `RustEmbed` directly (albeit through a pub reexport), it might be good to introduce two new derive macro to cover both these cases and have the setup to automatically pick up on it.  This one I might have to think a bit more about how to best approach it.  Basically, rather than:

```rust
#[derive(Clone, Copy, Embed)]
#[folder = "$LEPTOS_SITE_ROOT"]
#[prefix = "/"]
struct SiteRoot;
```

Which has the end user having to remember to include all of those very specific set of derive macros and attributes, this probably is much more preferable (actual name tbd):

```rust
// Embed just the site pkg.
#[derive(EmbedLeptosSitePkg)]
struct LeptosSitePkg;

// Embed the whole site root.
#[derive(EmbedLeptosSiteRoot)]
struct LeptosSiteRoot;
```

Both will have other different private traits from which the config helper method may be implemented to do the correct thing for the relevant derive so that the correct sets of routes and/or fallback service may be set up on the underlying `axum::Router`.

Naturally, this can come later, I don't want this additional work to block the release of 0.9.0 needlessly (e.g. if I can't find time again because the infant/toddler catches influenza again). Since the traits have been sealed we should have the flexibility to introduce additional setup method without causing a semver breaking change to include this addition, though I think I might need to better present the API for the generic `RustEmbed`, perhaps be a bit clear with the documentation as I do want to simply have the config builder call `.embed(...)` and the correct thing just happens.  Though I am not opposed to having `.embed_site_root()` and `.embed_site_pkg()`, too, as this also would help make it clear at a glance. Need to think about this more.

PR #4716, opened 2026-05-20
fix: preserve sync chunk order in out-of-order SSR template

tysen:
## Repro

https://github.com/tysen/leptos-out-of-order-bug

(`cargo run`)

## Bug

In `StreamBuilder`'s `Stream::poll_next` ooo handler, both branches iterated `chunks.into_iter().rev()` and dispatched sync vs. non-sync in the same pass. `.rev()` + `push_front` is correct for non-sync chunks (places them at the head of the main queue in original order), but for `StreamChunk::Sync` it `push_str`s into the buffer in reverse, so a sequence `[Sync(A), OutOfOrder, Sync(B)]` rendered as `"BA"` in the emitted `<template>`.

This shape arises when a resolved `<Suspense>` contains `<ErrorBoundary>` wrapping a nested `<Suspense>`: ErrorBoundary's `buf.append(side_buffer)` flushes the parent's `sync_buf` as a standalone `StreamChunk::Sync`, the nested Suspense pushes an `OutOfOrder` chunk, and `StreamBuilder::finish` can't re-merge the trailing sync into the leading one. The reversed concatenation then shows up as an unrecoverable hydration mismatch on the client.

## Fix

Iterate forward, append sync chunks in order, collect non-sync into a local `VecDeque` via `push_back`, then drain `pop_back` + `push_front` onto `this.chunks` — same head-of-queue invariant, correct order for both kinds. Apply to both branches (placeholder-present and emit-as-template).

## Notes

As might be obvious, I played ~zero part in identifying the root cause nor the proposed fix (thanks claude), but this did happen in my project and this really does fix it, and as far as I can tell the change is sensible and minimal. I've verified the tests pass and the examples build with this change. I've been using this in my project for a couple months with no issues. Seeing your post about the 0.9 RC prompted me to finally cut a repro/PR!

### DIFF
--- a/tachys/src/ssr/mod.rs
+++ b/tachys/src/ssr/mod.rs
@@ -403,8 +403,6 @@ impl Stream for StreamBuilder {
                                             .sync_buf
                                             .find(&closing)
                                             .unwrap();
-                                        let chunks_iter =
-                                            chunks.into_iter().rev();
 
                                         // TODO can probably make this more efficient
                                         let (before, replaced) =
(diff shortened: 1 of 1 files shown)


gbj:
Nice. Thanks for sharing this! Fix certainly looks correct to me — the reproduction is a real reproduction and the fix fixes it. Assuming it passes the relevant CI checks I'm happy to merge it.

~~I think the PR itself should be targeted against `main`, as the bug also exists there (I can retarget it) and I will merge it into the `leptos_0.9` branch as well.~~

Oops no I can't but I guess I can cherry-pick this commit from 0.9 into main later, all good.

PR #4717, opened 2026-05-20
leptos_0.9: rust edition 2024 and upgrade deps

sabify:
It's time to upgrade the breaking changes. Migrating to edition 2024 won't change MSRV as we are on v1.88 already. (edition 2024 requires v1.85)


sabify:
We have conflict between nightly-only features like `format_strings = true` and stable rustfmt flags. I will limit cargo-fmt checks to nightly-only.

gbj:
Thanks for the edition update and formatting!

My understanding is that there is no reason for us to bump minor versions of our dependencies (i.e., from `tokio` 1.49 to 1.52) unless we specifically require new features introduced after 1.49; listing 1.52 prevents our users from using anything earlier than 1.52, but listing 1.49 doesn't prevent them from using the latest version (which will happen automatically). As a result, we should keep these low unless we specifically need later features. Is that different from your understanding?

lukexor:
I'd agree with that, barring security or bug related issues that may affect code paths called by leptos crates

sabify:
You are right. But, should we consider that for new major leptos version release which contains breaking changes? If so, please let me know to revert those numbers. We have many other upgrades that will be considered breaking changes and we can make these minor version bump as minimum standard for the new major release.

gbj:
Yes, please revert them. It's true that there are other breaking changes because it's a breaking release, but (as I understand it) there is *no* benefit to increasing these minor version numbers, unless I'm missing something.

sabify:
Done!

gbj:
Thanks very much!

PR #4718, opened 2026-05-20
fix(server_fn): avoid panic on malformed Content-Type in multipart codec

sabify:
server_fn/src/codec/multipart.rs::from_req used `.expect("couldn't parse boundary")` on the result of parsing the request's `Content-Type` header. The header is fully attacker controlled, so any unauthenticated request to a `MultipartFormData` endpoint with a missing, non-multipart, or boundary-less Content-Type panicked the request future. Under runtimes that abort on panic this is a trivial remote-DoS primitive; under runtimes that catch it this is still an expensive panic-per-request.


gbj:
Thanks!

PR #4719, opened 2026-05-20
fix(server_fn): make FormatType::into_encoded_string lossy instead of panicking

sabify:
server_fn/src/lib.rs::FormatType::into_encoded_string used `String::from_utf8(bytes.into()).expect(...)` for the `Format::Text` branch. The bytes passed in originate from an `Encodes` implementation (potentially user-defined). Any text-format encoder that produced non-UTF-8 bytes panicked the process.

Worse, this same default is reached from `ServerFnErrorWrapper::Display::fmt` (server_fn/src/error.rs:539), so a malformed error message could panic *during* error reporting - a fail-during-fail bug.

Replace `.expect(...)` with an explicit fallback that converts the bytes via `String::from_utf8_lossy` when the UTF-8 check fails. The happy path is unchanged (a single allocation, same as before); the failure path now substitutes U+FFFD for invalid bytes and returns a best-effort string. This preserves the public API (still returns `String`) and avoids a remote-input-reachable panic.

### DIFF
--- a/server_fn/src/lib.rs
+++ b/server_fn/src/lib.rs
@@ -822,11 +822,22 @@ pub trait FormatType {
     const FORMAT_TYPE: Format;
 
     /// Encodes data into a string.
+    ///
+    /// For [`Format::Text`], any bytes that are not valid UTF-8 are replaced
+    /// with the Unicode replacement character (`U+FFFD`) rather than panicking.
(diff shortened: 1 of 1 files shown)


gbj:
Fair enough -- thanks!

PR #4720, opened 2026-05-21
fix(server_fn): browser and reqwest client prepend server_url for form_data and streaming requests

sabify:
In server_fn/src/request/browser.rs, the WASM `BrowserRequest` client implements 6 request constructors. Four of them (`try_new_req_query`, `try_new_req_text`, `try_new_req_bytes`, `try_new_req_multipart`) prepend `client::get_server_url()` to the caller-supplied `path` before building the `Request`. Two of them silently did not:

  - `try_new_req_form_data` passed the bare `path` to
    `Request::post/put/patch`.
  - `try_new_req_streaming` passed the bare `path` to the private
    `streaming_request` helper, which then handed it to
    `web_sys::Request::new_with_str_and_init`.

For cross-origin clients (which call `server_fn::client::set_server_url("https://api.example.com")` so the WASM client can talk to a separate API host), URL-encoded form submissions and streaming requests were dispatched against the page's own origin instead of the configured API host. This produced 404/CORS failures that were very hard to diagnose because the other request types in the same struct continued to work.

Apply the exact pattern already used by the four sibling constructors in the same file:

    let url = format!("{}{}", get_server_url(), path);

Then build the gloo/`web_sys` request from `&url` rather than `path`.

It is the same for the `reqwest` implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3597,6 +3597,7 @@ dependencies = [
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "wasm-bindgen-test",
  "wasm-streams",
  "web-sys",
  "xxhash-rust",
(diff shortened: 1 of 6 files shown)


gbj:
Thank you -- and thanks for the tests!

PR #4721, opened 2026-05-21
fix(server_fn): use StreamingText::CONTENT_TYPE for StreamingText codec

sabify:


sabify:
The failed CI is not related to this PR. I ran it locally and it succeed after 3 times. The `body_runs_with_conditional_nested_resource` test is flaky. (Is it part of our expectation that the test is flaky?)

gbj:
Thanks!

PR #4722, opened 2026-05-21
fix(server_fn): use correct redirect header in reqwest client

sabify:
The browser implementation correctly checks `REDIRECT_HEADER` (`server_fn::redirect::REDIRECT_HEADER`, `"serverfnredirect"`); the `reqwest` implementation checks the standard HTTP `Location` header instead!


sabify:

PR #4724, opened 2026-05-21
fix(server_fn): avoid panic on invalid path bytes in error_response

sabify:
Both `http::Response::error_response` and
`generic::Response::error_response` set the `serverfnerror` header directly from the request `path` and then called `.unwrap()` on the builder:

    Response::builder()
        .status(err.status_code)
        .header(SERVER_FN_ERROR_HEADER, path)
        .header(header::CONTENT_TYPE, err.content_type)
        .body(err.body.into())
        .unwrap()

`http::Response::builder` accumulates header errors and surfaces them at `.body(...)`. Any caller that passes a string with bytes that `HeaderValue` rejects (control bytes, etc.) crashes the worker on the error path. The public function is also reachable from custom code that does not pre-validate the path through hyper's URI parser.

Fix: validate `path` via `HeaderValue::from_str` and drop the diagnostic header on failure; keep an `unwrap_or_else` backstop that preserves the status code and body, so the client still sees a meaningful error response instead of a panic.

Add regression tests in both modules that call `error_response` with a path containing `\r\n` and assert the response is well-formed with the malformed header omitted.

### DIFF
--- a/server_fn/src/response/generic.rs
+++ b/server_fn/src/response/generic.rs
@@ -93,12 +93,24 @@ where
 
 impl Res for Response<Body> {
     fn error_response(path: &str, err: ServerFnErrorResponseParts) -> Self {
-        Response::builder()
-            .status(err.status_code)
-            .header(SERVER_FN_ERROR_HEADER, path)
-            .header(header::CONTENT_TYPE, err.content_type)
-            .body(err.body.into())
-            .unwrap()
+        let status_code = err.status_code;
+        let content_type = err.content_type;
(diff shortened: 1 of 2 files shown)


sabify:
@gbj, can you please re-run the `todo_app_sqlite_csr` test? It seems it timed out.

sabify:
@gbj, it seems the github actions has been having tough times these days! please re-run the failed workflows...

gbj:
Thanks, this looks good!

PR #4725, opened 2026-05-22
fix(server_fn): apply back-pressure in websocket forwarding loop

sabify:
Both the axum and actix websocket adapters create a bounded `futures::channel::mpsc` channel (capacity 2048) to forward messages from the underlying session into the server-function output stream, then called `start_send` directly on the sender inside the forwarding loop. Per the `Sink` contract, `start_send` requires a prior `poll_ready` returning `Ready`; calling it on a full bounded sender returns `Err(TrySendError)`. The previous code swallowed that error with `_ = ...`, so once the consumer of the output stream fell behind, incoming WebSocket frames were silently dropped with no error surfaced to either side and no back-pressure on the producer.

Switch each forwarding call to `SinkExt::send(...).await`, which awaits capacity via `poll_ready` before placing the item, pausing the forwarding loop when the consumer is slow instead of dropping frames. When `send` reports the receiver has been dropped, break the loop and close the session cleanly rather than spinning on a dead channel.

### DIFF
--- a/server_fn/src/request/actix.rs
+++ b/server_fn/src/request/actix.rs
@@ -6,7 +6,7 @@ use crate::{
 use actix_web::{web::Payload, HttpRequest};
 use actix_ws::Message;
 use bytes::Bytes;
-use futures::{FutureExt, Stream, StreamExt};
+use futures::{FutureExt, SinkExt, Stream, StreamExt};
 use send_wrapper::SendWrapper;
 use std::{borrow::Cow, future::Future};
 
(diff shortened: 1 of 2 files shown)


lukexor:
Thank you for finding and fixing this! I think I've been hitting this constantly and not realizing it streaming FFTs at a high rate for DSP processing of an IQ stream.

Not sure if it's related but hot reloading or refreshing the page will re-connect to the websocket after refresh and get pending items from the websocket buffer, when it should have closed and disconnected. I'll have to test this PR and see if it resolves that.

sabify:

PR #4726, opened 2026-05-22
fix(server_fn_macro): quality of life fixes

sabify:
In summary:

  - Proc-macro panics on malformed types: output_type, err_type, err_ws_in_type, and err_ws_out_type indexed path segments and generic args without bounds checks, so empty paths or types like Result<>/BoxedStream<> crashed the compiler.
  Replaced unchecked indexing with safe accessors; removed a dead panic! branch.
  - Fully-qualified paths: these helpers now match the last path segment, so std::result::Result, core::result::Result,
  and server_fn::BoxedStream are recognized instead of silently misdetected.
  - Build-path-dependent URLs: the server-fn URL hash defaulted to CARGO_MANIFEST_DIR, causing client/server URL
  mismatches (silent 404s) in split/CI builds. Now defaults to CARGO_PKG_NAME@CARGO_PKG_VERSION for reproducible URLs.

Please refer to commits' message for detailed changes.

### DIFF
--- a/server_fn_macro/src/lib.rs
+++ b/server_fn_macro/src/lib.rs
@@ -932,20 +932,20 @@ impl Parse for Middleware {
 }
 
 fn output_type(return_ty: &Type) -> Option<&Type> {
-    if let syn::Type::Path(pat) = &return_ty
-        && pat.path.segments[0].ident == "Result"
-    {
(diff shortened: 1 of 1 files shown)


sabify:

gbj:
Thanks!

PR #4729, opened 2026-05-23
feat(leptos_macro): Adding a `lazy_preload` macro

### LINKED ISSUE #4727: Explicit lazy preload
Hi!

I've ran into a situation with `lazy` functions in SSR split mode. My lazy function is multiple MB worth of wasm and is intended to run only on the client side as its doing rendering in a `<canvas>`.
I wanted to modify my view to display a loading message while the split wasm is being fetched.

I think that something along those lines would be nice:
```rust
#[lazy]
fn my_lazy() {}

#[component]
fn MyView() -> impl IntoView {
    let is_loaded = lazy_preload!(my_lazy);

    view!{<p>{if is_loaded() { "Loaded" } else { "Loading..." }}</p>}
}
```
Where  `lazy_preload!` would return a `Fn() -> bool` signal that would turn to true once the lazy wasm is loaded.

I made a prototype that works for my use case:
```rust
macro_rules! lazy_preload {
    (my_lazy) => {
        cfg_select! {
            feature = "hydrate" => {
                {
                    let loaded = LocalResource::new(|| __preload_my_lazy());
                    move || loaded.get().is_some()
                }
            },
            _ => || false // Would never load outside of client
        }
    };
}
```
This would probably need to turn into a `proc_macro` to generate the `__preload_{}` function identifier (and ideally reuse a common function with the `lazy` macro).

I tried using `Resource` but it would evaluate on the server. `Suspense` ran into similar issues with an added layer of hydration bugs when I tried to limit actual execution to the client only.
I also tried declaring a `let (loaded, set_loaded) = signal(false)`, giving `set_loaded` to my lazy function and setting it to `true`, but then in my use case the lazy function would re-actively be called multiple times and would need some internal book-keeping to manage the different states it might be in (first call to notify that it's ready and second call once the view is rendered and ready to be manipulated).
Would there be another pattern I've missed that would solve the issue?

If what I proposed for a `lazy_preload` macro sound reasonable I'd be down to work on the implementation.

edmondj:
Closes #4727 

Adds a `lazy_preload` convenience macro to explicitly download lazy function's split WASM and react to it's loading status.

### DIFF
--- a/leptos_macro/src/lazy.rs
+++ b/leptos_macro/src/lazy.rs
@@ -1,13 +1,20 @@
 use convert_case::{Case, Casing};
 use proc_macro::TokenStream;
 use proc_macro2::Ident;
-use proc_macro_error2::abort;
+use proc_macro_error2::{OptionExt, ResultExt, abort};
 use quote::{format_ident, quote};
 use std::{
     hash::{DefaultHasher, Hash, Hasher},
     mem,
 };
-use syn::{parse_macro_input, parse_quote, ItemFn, ReturnType, Stmt};
+use syn::{
+    parse::Parse, parse_macro_input, parse_quote, ItemFn, Path, ReturnType,
+    Stmt,
+};
+
+fn preload_name(ident: &Ident) -> Ident {
+    format_ident!("__preload_{}", ident)
+}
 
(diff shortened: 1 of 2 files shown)


gbj:
Looks good -- thanks!

PR #4730, opened 2026-05-23
Fix HTML-injection (XSS) sinks in tachys and leptos_macro

sabify:
This PR closes cases where untrusted data could reach the HTML body without escaping, across both the runtime renderer and the `view!` macro's static-HTML path.

### 1. `char` rendered without escaping (`tachys`)
`RenderHtml::to_html_with_buf` for primitive types ignored the `escape` flag and wrote values directly. For numeric/`bool`/`IpAddr`/`NonZero*` types this is harmless, their `Display` output is a subset of HTML text. But a `char` can be any Unicode scalar value, including `<`, `>`, `&`, `"`, `'`, so an attacker-controlled `char` rendered into an element body was an injection vector.

The `render_primitive!` macro is now parametrized with an escaping flag: `char` is escaped via `html_escape::encode_text` when the enclosing element escapes its children, and written verbatim only in raw-text contexts (`escape == false`). All other primitives keep their original allocation-free direct write, output is byte-identical.

### 2. `<textarea>` children not escaped (`tachys`)
`<textarea>` was declared with `ESCAPE_CHILDREN: false`, grouping it with the true raw-text elements `<script>`/`<style>`/`<noscript>`. But per the HTML spec, textarea content is *escapable* raw text, any `</textarea>` closes the element, so input like `</textarea><script>alert('xss')</script>` broke out and injected live markup. The flag is now `true`. This path handles all **dynamic** textarea content.

### 3. `<textarea>` not escaped in the macro's inert path (`leptos_macro`)
The `view!` macro has a separate compile-time static-HTML generator that *also* lumped `textarea` with `script`/`style` as non-escaping. This only affects fully-static ("inert") textarea content (not untrusted input), but it diverged from the runtime path and is incorrect per spec. Dropped `textarea` from the raw-text exemption at all three inert sites (element, SVG element, node dispatcher), so static textarea content is now escaped consistently with the runtime path.

### DIFF
--- a/leptos/tests/ssr.rs
+++ b/leptos/tests/ssr.rs
@@ -205,3 +205,39 @@ fn ssr_option() {
 
     assert_eq!(rendered.to_html(), "<option></option>");
 }
+
+#[cfg(feature = "ssr")]
+#[test]
+fn ssr_textarea_escapes_static_content() {
+    use leptos::prelude::*;
+
+    // Nested (non-top-level) static textarea exercises the macro's inert
+    // HTML path; its content must be HTML-escaped.
+    let rendered: View<HtmlElement<_, _, _>> = view! {
+        <div><textarea>"a < b & c"</textarea></div>
+    };
+
+    assert_eq!(
+        rendered.to_html(),
+        "<div><textarea>a &lt; b &amp; c</textarea></div>"
+    );
+}
+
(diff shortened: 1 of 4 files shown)


gbj:
Interesting! Good catch on both chars and textareas.

PR #4731, opened 2026-05-23
Fix five bugs in the `router`

sabify:

### 1. `<Form>` no longer lowercases the action URL
`extract_form_attributes` applied `.to_lowercase()` to the form `action` (and the `<input>`/`<button>` `formaction` fallbacks). Unlike `method`/`enctype`, the action is a URL, only scheme and host are case-insensitive (RFC 3986), so this silently corrupted case-sensitive paths and query values (e.g. `/API/v1/Users/AbC123` → `/api/v1/users/abc123`, broken HMAC/JWT signatures). Removed `.to_lowercase()` from the three action sites; kept it for `method`/`enctype`.

### 2. Query parameters are now decoded exactly once
Both URL sources hand back already-decoded values, `url::query_pairs()` (server) and `URLSearchParams` (browser), but each was re-decoded a second time via `ParamsMap::insert` → `Url::unescape`. This corrupted any value containing a literal `%xx` (e.g. `id=foo%252Fbar` → `foo/bar` instead of `foo%2Fbar`, truncated base64 padding, a parameter-smuggling surface). Added a crate-internal `ParamsMap::insert_decoded` and used it for query pairs on **both** server and client (fixing only one side would desync hydration). Path params are unchanged, they're extracted raw and correctly rely on `insert`'s single decode.

### 3. `RoutingProgress` no longer produces NaN width
`max_time` is optional and defaults to `Duration::ZERO`, making the per-tick step `100.0 / 0.0` = `inf`/`NaN` and rendering `width: NaN%`. Guarded the division to fill the bar in a single increment when `max_time` isn't positive.

### 4. `has_scheme` detects opaque and compound schemes
`resolve_path` treated absolute URLs with opaque schemes (`data:`, `javascript:`, `view-source:`, `blob:`) and compound schemes (`git+ssh`) as relative paths and mangled them. Rewrote `has_scheme` to follow RFC 3986 §3.1 (split on first `:`, validate the scheme prefix), keeping the `//` protocol-relative shortcut.

### 5. Route tuples (arity 3–16) preserve the inner `RouteMatchId`
The `tuples!` macro substituted a route's positional index for its real `RouteMatchId`, so routes at the same position in different sub-trees compared equal (stale-view risk for id-based cache/transition logic). The 2-tuple impl was already correct; aligned the macro with it and removed the now-unused `$count`.

### DIFF
--- a/router/src/components.rs
+++ b/router/src/components.rs
@@ -649,7 +649,15 @@ pub fn RoutingProgress(
     const INCREMENT_EVERY_MS: f32 = 5.0;
     let expected_increments =
         max_time.as_secs_f32() / (INCREMENT_EVERY_MS / 1000.0);
-    let percent_per_increment = 100.0 / expected_increments;
+    // `max_time` is optional and defaults to `Duration::ZERO`, which would make
+    // `expected_increments` zero and `100.0 / 0.0` evaluate to `inf` (and then
+    // `NaN`), producing `width: NaN%`. Fill the bar in a single increment when
+    // no positive `max_time` was provided.
+    let percent_per_increment = if expected_increments > 0.0 {
+        100.0 / expected_increments
+    } else {
+        100.0
+    };
 
     let (is_showing, set_is_showing) = signal(false);
     let (progress, set_progress) = signal(0.0);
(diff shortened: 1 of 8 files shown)


sabify:
Github Actions is truly getting worse!

gbj:
Thanks! These all look good.

PR #4732, opened 2026-05-24
Quality of life fixes for `router_macro`

sabify:

### `path!` macro

**Panic on non-literal input → diagnostic**
```rust
const ROUTE: &str = "/foo";
path!(ROUTE);   // before: proc-macro panic "not implemented"; now: clean compile error
```

**Multiple literals silently concatenated → rejected**
```rust
path!("/foo" "/bar");  // before: (StaticSegment("foo"), StaticSegment("bar")); now: error
```

**Raw strings mishandled → parsed by value**
```rust
path!(r"/home");     // before: error `Invalid static segment: r"home`; now: (StaticSegment("home"),)
path!(r"/");         // before: (StaticSegment("/"),); now: ()
```

**Errors pointed at whole invocation → at the literal**
```rust
path!("/foo/^/");    // caret now highlights the path string, not the entire path!(...) call
```

**Empty segments accepted → rejected**
```rust
path!("/:");         // before: ParamSegment(""); now: error "Invalid param segment:"
```

**RFC 3986 `pchar` set incomplete → completed**
```rust
path!("/v1:beta/items;color=red/$5/me!/(a+b)");  // before: "Invalid static segment"; now: OK
```
(`*` stays disallowed inside a segment — it's the wildcard sigil.)

### `lazy_route` macro

**Argument pattern spliced into call position → forwarded by value**
```rust
fn view(mut this: Self) -> AnyView { ... }          // before: broken `__View(mut this)`; now: OK
fn view(Self { id, body }: Self) -> AnyView { ... }  // before: broken; now: OK
```

**Missing `view` silently dropped work → aborts**
```rust
#[lazy_route]
impl LazyRoute for R { fn data() -> Self { Self } }  // now: "requires a `view` method"
```

**Extra `view` params silently dropped → rejected**
```rust
fn view(this: Self, extra: u32) -> AnyView { ... }   // now: "view must take exactly one argument"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2280,6 +2280,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
+ "trybuild",
 ]
 
 [[package]]
--- a/router_macro/Cargo.toml
+++ b/router_macro/Cargo.toml
@@ -22,6 +22,7 @@ syn = { features = ["full"], workspace = true, default-features = true }
 leptos = { path = "../leptos" }
 leptos_router = { path = "../router" }
 leptos_macro = { path = "../leptos_macro" }
+trybuild = { workspace = true, default-features = true }
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(leptos_debuginfo)'] }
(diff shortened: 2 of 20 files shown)


gbj:
Thanks!

PR #4734, opened 2026-05-25
Fix several bugs in `reactive_graph`

sabify:

### Fixes

- **Hash/Eq contract on pointer-identity handles**: `ArcRwSignal`, `ArcReadSignal`, `ArcWriteSignal`, `ArcMemo`, and `ArcStoredValue` compare equal via `Arc::ptr_eq`, but their `Hash` impls hashed the address of a stack temporary instead of the shared heap pointer. Equal handles could hash differently, breaking `HashMap`/`HashSet` lookups. Now hashes the `Arc` pointer.

- **`ArcAction` out-of-order dispatch guard**: the dispatch counter was never incremented, so the last-write-wins check never fired and a slower earlier dispatch could clobber the result of a newer one. The counter is now bumped per dispatch and the guard compares versions exactly.

- **`AsyncTransition` scoping**: transition state was held in a process global, so concurrent transitions on different tasks interfered with each other. It's now scoped to the running task via a thread-local restored around each poll, keeping it executor-agnostic.

- **`Sandboxed` arena restoration**: polling a `Sandboxed` future installed its arena but never restored the caller's, leaking the sandboxed arena into whatever ran next on the thread and breaking per-request isolation. The caller's arena is now restored on drop (across `await` points and panics).

- **Off-thread `LocalStorage` access**: `try_read_value`/`try_update_value`/`try_write_value` on a `StoredValue<T, LocalStorage>` panicked in the `SendWrapper` when called from another thread, instead of returning `None` as documented. They now check thread validity and return `None`.

- **Owner `NodeId` leak**: disposed nodes' ids accumulated in their owner indefinitely. Dead ids are now pruned at the `Vec` reallocation boundary (amortized O(1)).

- **Owner child `Weak` leak**: dropped child owners left dangling `Weak` references that accumulated in a long-lived owner. These are pruned at the same boundary.

### DIFF
--- a/reactive_graph/src/actions/action.rs
+++ b/reactive_graph/src/actions/action.rs
@@ -5,7 +5,10 @@ use crate::{
     owner::{ArcStoredValue, ArenaItem, Owner},
     send_wrapper_ext::SendOption,
     signal::{ArcMappedSignal, ArcRwSignal, MappedSignal, RwSignal},
-    traits::{DefinedAt, Dispose, Get, GetUntracked, GetValue, Update, Write},
+    traits::{
+        DefinedAt, Dispose, Get, GetUntracked, GetValue, Update, UpdateValue,
+        Write,
+    },
     unwrap_signal,
 };
 use any_spawner::Executor;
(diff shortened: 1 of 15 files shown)


sabify:
@gbj, the test is passing successfully on my machine and it seems it's flacky. Please re-run the workflow.

gbj:
Thanks for these -- I've reviewed them all and they look good!

PR #4735, opened 2026-05-25
Fix bugs in `reactive_stores` and `reactive_stores_macro`

sabify:

### `reactive_store`

- **`ArcField::is_disposed()` always returned `false`**: It now reports the real disposal state via a captured callback, so subscribers and read-paths that gate on `is_disposed` short-circuit correctly instead of reading a dead field.

- **`AtIndex::reader`/`writer` panicked on out-of-bounds**: A stale or never-valid index produced a guard that panicked on first deref. They now bounds-check under the held lock and return `None`, matching the "`None` means projection invalid" contract.

- **`Field<T, S>` only implemented `Write` for the default `SyncStorage`**: The `Write` impl dropped the `S` generic, so writing through e.g. `Field<T, LocalStorage>` failed to compile. The impl is now generic over `S` like the crate's other trait impls.

- **`PatchField` for `f32`/`f64` notified spuriously**: Using `!=` meant `NaN`→`NaN` fired a notification while `+0.0`→`-0.0` did not. Comparison now uses `total_cmp`, giving the correct "notify only on change" behavior.

- **`OptionStoreExt::unwrap` panicked via raw `Option::unwrap` (TOCTOU)**: Its projection closures unwrapped the inner `Option` lazily on deref, so clearing the option between capture and read panicked. A new `OptionSubfield` type performs the `is_some` check inside `reader()`/`writer()` (under the held lock) and returns `None` when the option is currently `None`; `unwrap`/`invert`/`map`/`map_untracked` now return it. Trigger/notify/track semantics mirror `Subfield` exactly, so notification counts are unchanged.

--- 

### `reactive_store_macro`

- Fix enum-variant path segmentation in the `Store` derive: all fields of a variant collapsed onto segment `0`, so writing one field spuriously woke subscribers of its siblings. Each field now gets a distinct segment from its index within the variant.
- Implement the `Patch` derive for enums (previously a compile error despite `Store` supporting enums): same-variant patches update fields in place and notify only what changed; cross-variant patches replace the value and notify the enum's path. Uses the same field-index segment numbering as `Store`, so patches target exactly the triggers the accessors subscribe to.
- Honor `#[store(skip)]` in the `Patch` derive and emit clear diagnostics for malformed `#[patch]` attributes.

### DIFF
--- a/reactive_stores/src/arc_field.rs
+++ b/reactive_stores/src/arc_field.rs
@@ -40,6 +40,7 @@ where
     keys: Arc<dyn Fn() -> Option<KeyMap> + Send + Sync>,
     track_field: Arc<dyn Fn() + Send + Sync>,
     notify: Arc<dyn Fn() + Send + Sync>,
+    is_disposed: Arc<dyn Fn() -> bool + Send + Sync>,
 }
 
 impl<T> Debug for ArcField<T>
@@ -155,6 +156,7 @@ where
             keys: Arc::new(move || value.keys()),
             track_field: Arc::new(move || value.track_field()),
             notify: Arc::new(move || value.notify()),
+            is_disposed: Arc::new(move || value.is_disposed()),
         }
     }
 }
@@ -204,6 +206,10 @@ where
                 let value = value.clone();
                 move || value.notify()
             }),
+            is_disposed: Arc::new({
+                let value = value.clone();
+                move || value.is_disposed()
+            }),
         }
     }
 }
@@ -212,7 +218,7 @@ impl<Inner, Prev, T> From<Subfield<Inner, Prev, T>> for ArcField<T>
 where
     T: Send + Sync,
     Subfield<Inner, Prev, T>: Clone,
-    Inner: StoreField<Value = Prev> + Send + Sync + 'static,
+    Inner: StoreField<Value = Prev> + IsDisposed + Send + Sync + 'static,
     Prev: 'static,
 {
     #[track_caller]
@@ -256,13 +262,17 @@ where
                 let value = value.clone();
                 move || value.notify()
             }),
+            is_disposed: Arc::new({
+                let value = value.clone();
+                move || value.is_disposed()
+            }),
         }
     }
 }
 
 impl<Inner, T> From<DerefedField<Inner>> for ArcField<T>
 where
-    Inner: Clone + StoreField + Send + Sync + 'static,
+    Inner: Clone + StoreField + IsDisposed + Send + Sync + 'static,
     Inner::Value: Deref<Target = T> + DerefMut,
     T: Sized + 'static,
 {
@@ -307,6 +317,10 @@ where
                 let value = value.clone();
                 move || value.notify()
             }),
+            is_disposed: Arc::new({
+                let value = value.clone();
+                move || value.is_disposed()
+            }),
         }
     }
 }
@@ -314,7 +328,7 @@ where
 impl<Inner, Prev> From<AtIndex<Inner, Prev>> for ArcField<Prev::Output>
 where
     AtIndex<Inner, Prev>: Clone,
-    Inner: StoreField<Value = Prev> + Send + Sync + 'static,
+    Inner: StoreField<Value = Prev> + IsDisposed + Send + Sync + 'static,
     Prev: IndexMut<usize> + Send + Sync + 'static,
     Prev::Output: Sized + Send + Sync,
 {
@@ -359,6 +373,10 @@ where
                 let value = value.clone();
                 move || value.notify()
             }),
+            is_disposed: Arc::new({
+                let value = value.clone();
+                move || value.is_disposed()
+            }),
         }
     }
 }
@@ -370,7 +388,7 @@ where
     KeyedSubfield<Inner, Prev, K, T>: Clone,
     T: KeyedIterable,
     for<'a> &'a T: IntoIterator<Item = <T as KeyedIterable>::IterItem<'a>>,
-    Inner: StoreField<Value = Prev> + Send + Sync + 'static,
+    Inner: StoreField<Value = Prev> + IsDisposed + Send + Sync + 'static,
     Prev: 'static,
     T: KeyedAccess<K> + 'static,
     T::Value: Sized,
@@ -416,6 +434,10 @@ where
                 let value = value.clone();
                 move || value.notify()
             }),
+            is_disposed: Arc::new({
+                let value = value.clone();
+                move || value.is_disposed()
+            }),
         }
     }
 }
@@ -434,6 +456,7 @@ impl<T> Clone for ArcField<T> {
             keys: Arc::clone(&self.keys),
             track_field: Arc::clone(&self.track_field),
             notify: Arc::clone(&self.notify),
+            is_disposed: Arc::clone(&self.is_disposed),
         }
     }
 }
@@ -489,6 +512,35 @@ impl<T> Write for ArcField<T> {
 
 impl<T> IsDisposed for ArcField<T> {
     fn is_disposed(&self) -> bool {
-        false
+        (self.is_disposed)()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{self as reactive_stores, ArcField, Store};
+    use reactive_graph::{owner::Owner, traits::IsDisposed};
+
+    #[derive(Default, reactive_stores_macro::Store)]
+    struct State {
+        value: i32,
+    }
+
+    #[test]
+    fn arc_field_reports_disposal_of_underlying_store() {
+        let owner = Owner::new();
+        let field: ArcField<i32> = owner.with(|| {
+            let store = Store::new(State { value: 42 });
+            store.value().into()
+        });
+
+        // while the owner is alive, the field is not disposed
+        assert!(!field.is_disposed());
+
+        // once the owner is disposed, the underlying store is gone and the
+        // field must report it, instead of always returning `false`
+        owner.cleanup();
+        drop(owner);
+        assert!(field.is_disposed());
     }
 }
--- a/reactive_stores/src/arc_field.rs
+++ b/reactive_stores/src/arc_field.rs
@@ -329,7 +329,7 @@ impl<Inner, Prev> From<AtIndex<Inner, Prev>> for ArcField<Prev::Output>
 where
     AtIndex<Inner, Prev>: Clone,
     Inner: StoreField<Value = Prev> + IsDisposed + Send + Sync + 'static,
-    Prev: IndexMut<usize> + Send + Sync + 'static,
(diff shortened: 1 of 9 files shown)


gbj:
Are these PRs are AI-generated? I don't have a problem with that per se and have been happy to merge the changes that are obviously wrong or very straightforward fixes (like the `server_fn` ones that had copy-paste issues with the `Content-Type`, replacing panics with `Err` returns, etc.)

It just helps me calibrate how much *human* thought has gone into any particular change, as part of the review process.

For example [this change](https://github.com/leptos-rs/leptos/pull/4735/changes/37d47715f98d43f4108bebb6d910fc9358dd4398#diff-71367434fac7a37510db15206ff63f7f7eb361bbb3d5e4f079c463be2ea2dd8cR646-R650) is documented by a comment as falling back to the inner path rather than panicking. [This one](https://github.com/leptos-rs/leptos/pull/4735/changes/37d47715f98d43f4108bebb6d910fc9358dd4398#diff-71367434fac7a37510db15206ff63f7f7eb361bbb3d5e4f079c463be2ea2dd8cR664-R667) skips over a keyed segment rather than panicking. 

If that's an intentional choice by you then we can discuss it. If it's an LLM helpfully replacing a panic by returning wrong results from a function, I'd rather just revert that commit.

If these are AI-generated, I will probably prioritize approving and merging obviously-correct changes, and/or any changes that close issues someone has been reported. (Rather than trying to understand and reproduce some of the more obscure ones to see if they're hallucinations or to try to figure out how they'd be triggered.)

sabify:
Does it matter what tools we use to improve the codebase? You are right, the [37d4771](https://github.com/leptos-rs/leptos/pull/4735/commits/37d47715f98d43f4108bebb6d910fc9358dd4398) was behavioral change and off topic. It was not intentional and I will revert it. I do my best to carefully curate and review these bugs.

About the reproducible example, almost all the reported bugs have their reproducible example. You can just look at the commit's message. Tests are also provided for each fix and cannot pass without it.

reactive_stores/src/arc_field.rs
```diff
@@ -329,7 +329,7 @@ impl<Inner, Prev> From<AtIndex<Inner, Prev>> for ArcField<Prev::Output>
 where
     AtIndex<Inner, Prev>: Clone,
     Inner: StoreField<Value = Prev> + IsDisposed + Send + Sync + 'static,
-    Prev: IndexMut<usize> + Send + Sync + 'static,
+    Prev: IndexMut<usize> + crate::len::Len + Send + Sync + 'static,
```
gbj: nit: `use crate::len::Len` rather than the fully qualified name here

gbj:
Oh, let me clarify: the choice of tool in this case *only* makes a difference in three specific way: 1) it finds some obscure bugs that nobody has noticed and reported (really good!) and 2) it makes some fixes that are hacks (like the path one) — this is just where it requires extra attentiveness, because it's a plausible-looking change when a person reviews it but not one that a person sat down and said "yeah this is the right fix" before writing it; and 3) it has the potential to hallucinate bugs.

Identifying whether things are category 3 in non-obvious cases is all that I mean about prioritizing the obvious bugs. In this case all of the commits except 6ae9060d9acb9f547b9d424fe01c0253a0481eeb look straightforwardly good to me. For 6ae9060d9acb9f547b9d424fe01c0253a0481eeb, I needed to think a little more about how the paths should actually work out here (see comments). I had a little time to do that this morning so I've looked it over. I seem to have lost where my comments went in the GitHub review UI (lol) but https://github.com/leptos-rs/leptos/commit/6ae9060d9acb9f547b9d424fe01c0253a0481eeb#diff-f0882aa9d1bcd7ab0bfd3dfa0bb28c31a0ce24fd791ba39b938859bf9914e48e


Thanks very much for putting all of these together!

sabify:
Just added another test that reflects the changes in https://github.com/leptos-rs/leptos/commit/6ae9060d9acb9f547b9d424fe01c0253a0481eeb.

Here is the full analysis:

### Summary

Both review comments are individually correct, but each addresses a different half of the change. The `field_idx`-based segment resolves the reported bug — collisions between sibling fields *within* a single variant. The per-variant `base_segment` offset resolves a distinct, second class of collision — between fields of *different* variants. The reason removing `base_segment` leaves the test suite passing is that the committed regression test never transitions between variants, and therefore exercises only the first half. A new test closing that gap has been added.

### Background: path-based trigger semantics

Each reactive field is identified by a path, `parent_path ++ [segment]` (`subfield.rs:80`), and every path owns a pair of triggers, `this` and `children`. Subscription and notification are intentionally asymmetric:

- **Tracking** a field at `[form, k]` subscribes to `this@[form,k]`, `children@[form,k]`, and the `this` trigger of every ancestor (`this@[form]`, `this@[]`) — but not the ancestors' `children` (`subfield.rs:126`). This asymmetry is what isolates sibling fields from one another.
- **Writing** `[form, k]` notifies `this@[form,k]`, `children@[form,k]`, and the `children` trigger of every ancestor (`store_field.rs:75`).

A write wakes a subscriber only where their trigger sets intersect, which occurs precisely when two fields are assigned the same `segment`. Every defect in this area therefore reduces to two logically independent fields receiving an identical segment.

### Comment at line 335 — path stability across variants

The concern that a field's path should not depend on the active variant is valid, and the implementation satisfies it. `base_segment` does not alter the path of the enum field itself: `foo` remains `[form]` in every variant, determined solely by its position in the parent struct. The offset applies only to the segments of each variant's *inner* fields, and those assignments are static — `loaded_name()` resolves to `[form, 0]` and `loaded_age()` to `[form, 1]` unconditionally. No field's path varies at runtime with the active variant.

The substantive question this raises is whether two fields residing in different variants "at the same position" should share a segment. They should not: they represent distinct data, and a shared trigger slot permits a write to one to wake a subscriber of the other, as demonstrated below.

### Comment at line 539 — sufficiency of `field_idx`

`field_idx` is indeed the operative change for the intra-variant defect, and the committed test passes without `base_segment` because it remains within the `Loaded` variant throughout and validates only that case. `base_segment` addresses cross-variant collisions, which that test does not reach.

Consider:

```rust
#[derive(Store, Default)]
enum Multi { #[default] Empty, A { a: u32 }, B { b: u32 } }
```

- With `field_idx` alone: `A::a` and `B::b` both receive segment `0` and collide at `[multi, 0]`.
- With the base offset: `A::a` receives `0` and `B::b` receives `1`.

The behaviour was confirmed empirically by toggling the offset in the macro and running an identical scenario: track (rather than read) the `A::a` subfield, transition `A → B`, then write `B::b` twice.

| Macro behaviour              | Intra-variant test | Cross-variant test                |
| ---------------------------- | ------------------ | --------------------------------- |
| `field_idx` only             | Passes             | Fails — subscriber wakes twice    |
| `field_idx` + `base_segment` | Passes             | Passes — subscriber does not wake |

Two aspects make this scenario representative rather than contrived:

- **Tracking rather than reading.** Reading a field belonging to a non-active variant panics on the `"accessed an enum field that is no longer matched"` guard. Tracking is therefore the supported means by which a subscription legitimately outlives a variant transition.
- **The transition does not absolve the collision.** The `A → B` write wakes the subscriber once, through the ancestor `this@[multi]` trigger; this is expected and occurs identically with or without the offset. A tracking subscriber re-establishes the same subscription on re-run rather than dropping it, so subsequent `B::b` writes still collide and still produce spurious wakeups whenever the segments alias.

### Rationale for retaining the base offset

The offset establishes a bijection between fields and segments across the entire enum, ensuring that no two distinct data locations alias onto a single trigger slot. It is strictly non-regressive: within a variant the segments remain consecutive and distinct, and across variants they become distinct as well, so the change can only eliminate spurious wakeups and cannot introduce a missed one. The cost is a single `usize` of compile-time bookkeeping per variant.

### Change in response to the review

The gap identified by the review was in test coverage rather than in the implementation: the suite justified only the `field_idx` half of the fix. A regression test, `writing_one_variant_field_does_not_wake_a_different_variants_sibling`, has been added to `enum_subfield_paths.rs`. It fails if and only if the base offset is removed, while the original test passes in either configuration. Both tests pass against the committed macro, which is otherwise unchanged.

gbj:
I have two comments:
1) I don't think that the reasoning in the test introduced by 8c5ddbde6fed8672a5ecd920f3f4c17a3e655285 is correct: `Subfield for `A::a`. We only *track* it (never read), so its subscription can outlive a switch to a different variant -- reading a non-matching variant would panic on the "accessed an enum field that is no longer matched" guard.` I don't think this implementation distinguishes between tracking and reading. If we switch from `A::a` to `B::b`, the old reference to `A::a` is no longer valid, period. The text of your reply adds, "Tracking is therefore the supported means by which a subscription legitimately outlives a variant transition." What does this mean in an actual application?
2) I am not comfortable with a new normal in which people submit AI-generated code and then respond to review comments with 750 words of AI-generated text.

Would you please explain to me, person to person, why this change is a good idea? Do you have an example of application code where you've run into this?

gbj:
For example, here's an application that uses the "broken" store example from the test, and is just fine. What's supposed to be wrong here?
```rust
#[component]
    let store = Store::new(MultiState {
        multi: Multi::Empty,
    });

    view! {
        <button on:click=move |_| store.multi().set(Multi::A { a: 1 })>"Switch to A"</button>
        <button on:click=move |_| store.multi().set(Multi::B { b: 42 })>"Switch to B"</button>
        <Show when=move || store.multi().a()>
            <pre>{move || format!("a_a: {:?}", store.multi().a_a().unwrap().get())}</pre>
        </Show>
        <Show when=move || store.multi().b()>
            <pre>{move || format!("b_b: {:?}", store.multi().b_b().unwrap().get())}</pre>
            <button on:click=move |_| store.multi().b_b().unwrap().set(13)>"Set b_b to 13"</button>
        </Show>
    }
}
```

sabify:
@gbj Sorry for bothering you and pushing this further with the changes. I think my purpose in providing the analysis is to provide you the reasoning behind the changes and it's certainly not proof. I've already done like this in https://github.com/leptos-rs/leptos/pull/4755#issuecomment-4636036739 which was **helpful** in **our** reasoning.

I just applied your suggestion to `field_idx` only, removed the excessive `base_segment` (with never happen/unsupported cross-variant usage) and reverted the added test. Hope it's helpful.

sabify:
I just merged #4752 into this branch with applied changes, as these two needed to be in sync.

gbj:
No need to apologize, and thanks again for your work. I'd suggest what is most helpful in the future is for *you* to read that analysis, and then to distill it into the 1-2 points that are relevant — rather than offering it unfiltered. The main chokepoint for me at this point (and I think this is true across open source projects) is my capacity to read and review both code and comments/analysis/discussion.

sabify:
I totally understand your position and thank you for maintaining this awesome project.

PR #4738, opened 2026-05-26
Quality of life fixes for `leptos_macro`

sabify:

### `#[derive(Params)]`

**Generics/lifetimes dropped from impl → threaded through**
```rust
#[derive(Params)]
struct PagedParams<const N: usize> { page: Option<i32> }
// before: impl Params for PagedParams { ... }  → E0107 missing generics
// now:    impl<const N: usize> Params for PagedParams<N> { ... }
```
(Same fix covers `<T>`, lifetimes, and where-clauses.)

### `#[lazy]` / `#[island]`

**Identifiers hashed with `DefaultHasher` → deterministic FNV-1a + fmix64**
```rust
// DefaultHasher is not stable across Rust releases and is native-endian,
// so a server/wasm toolchain skew silently breaks prefetch/hydrate lookup.
// before: let mut h = DefaultHasher::new(); location.hash(&mut h); h.finish();
// now:    fnv1a_64(location.as_bytes())  → byte-identical across builds/targets
```

### `include_view!`

**Edits to the included file triggered no rebuild → dependency registered**
```rust
include_view!("views/hello.rs");
// before: edit views/hello.rs alone → cargo build recompiles nothing
// now:    emits const _: &[u8] = include_bytes!(<path>) → Cargo rebuilds
```

**Opaque read error → path + OS cause**
```rust
include_view!("views/does_not_exist.rs");
// before: "could not open file"
// now:    "could not open file `views/does_not_exist.rs`: No such file or directory (os error 2)"
```

### `view!` — inert elements

**`bind` prefix matched any attribute → matches `bind:` exactly**
```rust
view! { <custom-element binding="foo" /> }
// before: "binding" treated as a directive → forced onto slow dynamic path
// now:    inert fast path (only real bind: directives are excluded)
```

### `view!` — event modifiers

**Modifiers matched by substring → matched exactly**
```rust
view! { <button on:click:targeted=move |_| {}>"click"</button> }
// before: contains(":target") → activates :target on event "clicked", fails deep in codegen
// now:    error: unknown event modifier `:targeted`; expected `:undelegated`, `:target`, or `:capture`
```
(Real usage like `on:input:target` is unaffected.)

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -1,3 +1,4 @@
+use crate::stable_hash::fnv1a_64;
 use attribute_derive::FromAttr;
 use convert_case::{
     Case::{Pascal, Snake},
@@ -9,7 +10,6 @@ use leptos_hot_reload::parsing::value_to_string;
 use proc_macro_error2::abort;
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{ToTokens, TokenStreamExt, format_ident, quote, quote_spanned};
-use std::hash::DefaultHasher;
 use syn::{
     AngleBracketedGenericArguments, Attribute, FnArg, GenericArgument,
     GenericParam, Item, ItemFn, LitStr, Meta, Pat, PatIdent, Path,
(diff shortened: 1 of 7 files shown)


gbj:
Thanks!

PR #4739, opened 2026-05-27
Harden Leptos server integrations (axum): panics, info leaks, path traversal & resource leaks

sabify:

This is all about axum integration with a shared fix for actix.

### Changes

| Area                    | Fix                                                                                                                                                                                                                         |
| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Path traversal**      | `static_file_path` now validates centrally, returning `Option` (`None` rejects `..`, `%2e`/`%2f`/`%5c`, backslash). Axum + Actix read handlers answer `404`; generation refuses to write.                                   |
| **Info disclosure**     | Static-route serving no longer renders raw `std::io::Error` (server paths) into the response body — logs at `warn`, returns generic 500.                                                                                    |
| **Unbounded memory**    | Per-path static-header cache switched from an unbounded `HashMap` to a bounded `lru::LruCache` (capacity via `LEPTOS_STATIC_HEADERS_CACHE_SIZE`, default 1024).                                                             |
| **Resource leak**       | Reactive `Owner` cleanup moved from a trailing stream item to a `Drop` guard, so it runs on client disconnect, not only on full stream drain.                                                                               |
| **DoS via panic**       | `redirect()` no longer panics on an invalid `Location` value (CRLF/control chars).                                                                                                                                          |
| **Panics → graceful**   | Missing path-and-query treated as `/`; `render_route` returns 500 instead of panicking on a missing matched-path; server-fn / static-file / request-build / response-builder paths return 500 instead of `unwrap`/`expect`. |
| **Content negotiation** | `Accept` header parsed with the `mime` crate (honors `q=0`) instead of substring-matching `"text/html"`, fixing wrong 302/HTML decisions.                                                                                   |
| **Correctness**         | Static-route 404 status set explicitly rather than relying on `ResponseOptions` sequencing.                                                                                                                                 |

### Threat model

- **Path traversal (`f3eb2621`)** — *Attacker:* a client whose request path (or a slug feeding static generation) reaches a static route. *Vector:* `..` segments or percent-encoded separators in the raw, undecoded URI. *Scope note:* `static_file_path` always appends `.html` and there is no NUL-truncation vector, so this is **not** arbitrary file read — the reachable set is limited to `.html` files. *Impact:* (1) **write-side** — generation does `create_dir_all` + `fs::write`, so an attacker-influenced path can create dirs and write/overwrite `.html` files outside `site_root` (defacement, **stored XSS** via the rendered body, clobbering another app's files); (2) **read-side** — `ServeFile` can return `.html` files outside the intended subtree (cross-tenant or not-yet-public pre-rendered pages). *Mitigation:* central rejection at the single helper that builds the on-disk path; percent-encoding rejection is defense-in-depth for any layer that decodes.
- **Info disclosure (`8fea649e`)** — *Attacker:* client triggering a TOCTOU file-removal race on a static route. *Impact:* raw filesystem error (absolute server paths) leaked in the response body. *Mitigation:* generic 500, details only in server logs.
- **Memory exhaustion / DoS (`8bb803a3`)** — *Attacker:* driving many unique static paths (e.g. attacker-chosen slugs). *Impact:* one cache entry leaked per path for the process lifetime → unbounded RSS growth. *Mitigation:* bounded LRU eviction.
- **Resource leak / DoS (`145c4ad7`)** — *Attacker (or routine condition):* disconnecting mid-response (slow client, browser cancel, proxy timeout). *Impact:* the `Owner` and everything it transitively holds leaks per aborted request → unbounded RSS growth on the main render path. *Mitigation:* `Drop`-based cleanup runs regardless of how the stream ends.
- **Panic-based DoS (`2c85b74e` and others)** — *Attacker:* crafting inputs (invalid header values, authority-form URIs) reaching `unwrap`/`expect`/`panic!`. *Impact:* request-handler panic. *Mitigation:* graceful error propagation / generic 500.

The remaining commits (Accept-header parsing, 404 status) are correctness fixes without a direct attacker path.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -2528,7 +2528,21 @@ where
                 let options = LeptosOptions::from_ref(&state);
                 let res =
                     get_static_file(uri, &options.site_root, req.headers());
-                let res = res.await.unwrap();
+                // `get_static_file` returns `Err` if the underlying `ServeDir`
+                // fails. This handler is the documented "reasonable default"
+                // fallback, so it must not panic: log and serve a generic 500.
+                let res = match res.await {
+                    Ok(res) => res,
+                    Err((status, err)) => {
+                        #[cfg(feature = "tracing")]
+                        tracing::warn!(
+                            "static file handler failed: {status} {err}"
+                        );
+                        #[cfg(not(feature = "tracing"))]
+                        let _ = (status, err);
+                        return internal_server_error();
+                    }
+                };
 
                 if res.status() == StatusCode::OK {
                     let owner = Owner::new();
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -181,6 +181,13 @@ pub(crate) fn extend_response<ResBody>(
         .extend(std::mem::take(&mut res_options.headers));
 }
 
+/// A generic `500 Internal Server Error` response with no body details, used
+/// when an integration handler hits an unrecoverable but non-fatal condition
+/// that previously panicked.
+fn internal_server_error() -> Response<Body> {
+    (StatusCode::INTERNAL_SERVER_ERROR, "Internal Server Error").into_response()
+}
+
 struct AxumResponse(Response<Body>);
 
 impl ExtendResponse for AxumResponse {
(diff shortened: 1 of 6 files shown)


integrations/axum/src/lib.rs
```diff
@@ -245,18 +279,33 @@ pub fn redirect(path: &str) {
...
+        // parameter), so it may contain bytes that are illegal in a header
+        // value (CR, LF, NUL, ...). `HeaderValue::from_str` rejects those, and
+        // turning that recoverable error into a panic would let any client take
+        // down the worker handling the request. Skip the redirect instead.
+        let location = match header::HeaderValue::from_str(path) {
+            Ok(location) => location,
+            Err(_) => {
+                #[cfg(feature = "tracing")]
+                tracing::warn!(
+                    "redirect() ignored: target is not a valid header value"
```
LeoniePhiline, line 292: For better diagnostics, `path` could be included in its debug representation (escaping non printable characters).

integrations/axum/src/lib.rs
```diff
@@ -717,22 +776,49 @@ where
...
+                 handler must be mounted through the Axum router."
+            );
+            #[cfg(not(feature = "tracing"))]
+            eprintln!(
+                "render_route handler invoked without a MatchedPath; the \
+                 handler must be mounted through the Axum router."
+            );
+            return Box::pin(async { internal_server_error() });
+        };
         // 2. Find RouteListing in paths. This should probably be optimized, we probably don't want to
```
LeoniePhiline, line 804: A `phf` map could help optimize this.

integrations/utils/src/lib.rs
```diff
@@ -17,6 +17,51 @@ pub type PinnedStream<T> = Pin<Box<dyn Stream<Item = T> + Send>>;
...
+/// Wraps a response body stream and ties cleanup of the reactive [`Owner`] to
+/// the stream's `Drop`, rather than to a chained terminal future.
+///
+/// The previous approach appended an `owner.unset_with_forced_cleanup()` future
+/// as the last item of the stream. That only runs if the stream is polled to
+/// completion, so a client disconnecting mid-response (slow client, browser
+/// cancel, proxy timeout) would drop the body before the terminal future ran,
+/// leaking the `Owner` and everything it transitively keeps alive. Cleaning up
+/// on `Drop` runs whether the stream finishes or is cancelled.
+struct OwnerCleanupStream {
```
LeoniePhiline, line 29: Great solution!

PR #4740, opened 2026-05-27
fix(CI): do not fail fast

sabify:
Github Actions is getting worse and unstable. So, I just applied a retry strategy for necessary steps to prevent failing fast.


gbj:
Thanks! We'll cross our fingers... it's gotten pretty grim :smile:

PR #4741, opened 2026-05-27
Remove unused unstable FromWasmAbi bound

ealmloff:
Tachys relies on FromWasmAbi which is an unstable api from wasm bindgen under the convert module. It appears to be completely unused. This PR removes it so tachys doesn't break if it updates in a minor version in the future


gbj:
Thanks!

PR #4742, opened 2026-05-27
fix(CI): use official actions for toolchain and binary installs, #4740 follow up

sabify:
Wandalen/wretry.action only supports node and docker actions and errors on composite actions. Use dtolnay/rust-toolchain, cargo-bins/cargo-binstall and taiki-e/install-action directly; they already retry their downloads internally.


gbj:
Thanks!

sabify:
Now we get different error on pnpm step. I'll provide a fix for it.

PR #4747, opened 2026-05-29
Fix HTML injection (XSS) and other bugs in `hydration_context`

sabify:

## Changes

- **Escape `<` in serialized error messages (XSS).** Errors emitted into `__SERIALIZED_ERRORS` were formatted with `{:?}`, which does not escape `<`. Any error message containing `</script>` could close the surrounding streaming `<script>` tag and inject markup. The error path now applies the same `<` → `\u003c` escape the resolved-resource path already uses.

- **Drop the `async_buf` write lock before polling resource futures.** `AsyncDataStream::poll_next` held a write lock across user futures; any polled future that transitively called `write_async` on the same context tried to re-acquire the lock and deadlocked. The lock is now taken only for short critical sections that drain and reinsert the buffer, with polling happening in between.

- **Make `SsrSharedContext::default()` consistent with `new()`.** The derived `Default` produced `is_hydrating = false` and `non_hydration_id = 0`, immediately minting colliding IDs against a freshly hydrating ID counter. `Default` now delegates to `new()`, and both constructors spell out every field so they cannot drift apart.

- **Declare `__INCOMPLETE_CHUNKS` up front and push markers live.** The marker array was previously emitted only as the final chunk of the response, after every async resource had resolved — too late for client-side `get_incomplete_chunk` probes during streaming. The initial chunk now declares `__INCOMPLETE_CHUNKS=[…]` (snapshotting markers known at that moment) and additional markers are streamed as `.push(id)` statements as `set_incomplete_chunk` fires.

- **Announce late-registered async resources via `__PENDING_RESOURCES.push(...)`.** The initial `__PENDING_RESOURCES=[…]` literal only contained resources known at snapshot time; any resource registered afterwards was resolved on the client without ever appearing in `__PENDING_RESOURCES`. Any resource whose id was not in the initial snapshot now gets a `__PENDING_RESOURCES.push(id);` written before its `__RESOLVED_RESOURCES[id] = …` write.

- **Skip malformed entries instead of panicking hydration.** The hydrate-side parsers for `__SERIALIZED_ERRORS` and `__INCOMPLETE_CHUNKS` used `.unwrap()` / `.expect(...)` on every JsValue cast, so any malformed entry (truncated streams, version mismatches, proxy rewrites) panicked the entire hydration step. Both parsers now thread `?` through every fallible step and drop bad entries via `filter_map`.

### DIFF
--- a/hydration_context/src/hydrate.rs
+++ b/hydration_context/src/hydrate.rs
@@ -30,34 +30,36 @@ extern "C" {
 }
 
 fn serialized_errors() -> Vec<(SerializedDataId, ErrorId, Error)> {
+    // Treat `__SERIALIZED_ERRORS` as untrusted input: a truncated stream,
+    // a version mismatch between server and client, a proxy that mangles
+    // the script, or any other source of malformed entries would
+    // otherwise panic the entire hydration step. Skip-and-continue is
+    // safer than unwrap-and-die.
     __SERIALIZED_ERRORS.with(|s| {
         s.iter()
-            .flat_map(|value| {
-                value.dyn_ref::<Array>().map(|value| {
-                    let error_boundary_id =
-                        value.get(0).as_f64().unwrap() as usize;
-                    let error_id = value.get(1).as_f64().unwrap() as usize;
-                    let value = value
-                        .get(2)
-                        .as_string()
-                        .expect("Expected a [number, string] tuple");
-                    (
-                        SerializedDataId(error_boundary_id),
-                        ErrorId::from(error_id),
-                        Error::from(SerializedError(value)),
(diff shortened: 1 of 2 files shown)


sabify:
@gbj, you may want to merge branch main into leptos_0.9 to make the retries (#4740) work here too. Please re-run the failed workflow.

gbj:
Sure, I've merged `main` into `leptos_0.9`

sabify:
The merge couldn't bring those 4 commits to `leptos_0.9`. Maybe your local `main` branch is out of sync.

gbj:
These look good, thanks!

PR #4750, opened 2026-05-31
fix: conditionalize server_fn browser feature on WASI to support WASIp2 compilation

codeitlikemiley:
### Context & Problem
This PR resolves the WASI build regression reported in [leptos_rs/leptos_wasi#21](https://github.com/leptos-rs/leptos_wasi/issues/21).

When compiling Leptos applications for WASI targets (such as `wasm32-wasip2` or `wasm32-wasip3` using Spin/Wasmtime), the build process encountered unresolved `__wbindgen_placeholder__` linker errors. This occurred because:
1. `leptos` unconditionally enabled the `browser` feature of `server_fn` inside its `Cargo.toml`.
2. Upgraded dependencies pulled in `wasm-streams 0.5.0` which forces `wasm-bindgen >= 0.2.108` and browser-specific dependencies.
3. Linker failures on WASI occurred because we do not run the `wasm-bindgen` post-processing CLI on WASI targets, leaving placeholder imports unresolved.

### Solution
This PR updates the dependency structure to conditionally compile browser features and dependencies on WASI:
1. **Target-gated dependencies & features**: Gated `wasm-bindgen`, `web-sys`, `wasm-bindgen-futures`, and `any_spawner`'s `wasm-bindgen` feature under `target.'cfg(not(target_os = "wasi"))'`.
2. **Conditional `server_fn`**: Enabled the `browser` feature of `server_fn` only on non-WASI targets (`target.'cfg(not(target_os = "wasi"))'`). When targeting WASI, `server_fn` is compiled with only `form-redirects`.
3. **Internal Mocks in `tachys`**: Because the core view-tree logic in `tachys`/`leptos` still references DOM/JS types (like `Element`, `JsValue`, etc.), we added self-contained internal mock modules inside `tachys` under `#[cfg(target_os = "wasi")]` to allow compiling without external stub crates.
4. **Dummy Macros**: Added a no-op `#[wasm_bindgen]` attribute macro to `leptos_macro` so components compile cleanly on WASI.

All E2E tests and guest apps target builds verify cleanly with zero regressions on standard browser features.


codeitlikemiley:
@benwis here is the patch that would fix the issue https://github.com/leptos-rs/leptos_wasi/issues/21 
I  would like you to review this, and here feedback from you...

You can also checkout [leptos_wasi](https://github.com/leptos-rs/leptos_wasi/pull/22) with using the git hash commit on my fork of [leptos](https://github.com/codeitlikemiley/leptos/tree/servernfnwasi)

gbj:
I just want to say at the top that I'm not going to merge this using this approach: i.e., a change to target-specific conditional compilation that touches 57 files with a +3k -1.3k diff is probably not the right solution to this problem.

Almost all of the wasm-bindgen-related code being disabled here has existed in exactly this configuration for almost 2 years, i.e., since leptos 0.7. 

What's the change that you think is responsible for the breakage between the leptos patch versions you identified as working and being broken? Let's address that specific change.

benwis:
@codeitlikemiley I'd have to agree, this seems like AI may have completely missed the ball.

codeitlikemiley:
@gbj you're probably right, I do have a working solution with 4 file changes , but that requires me to make PR to  [wasm-bindgen](https://github.com/wasm-bindgen/wasm-bindgen/pull/5175)



@benwis Its me who missed the context lol 

Thanks for replies, 

I will keep you guys posted once I fully tested it...

codeitlikemiley:
The compilation failure on version `0.2.115` is caused by a change in how `wasm-bindgen` gates its target-specific code generation. 

Specifically, they simplified target detection from a strict arch/OS check to a broad `target_family = "wasm"` check. Because **WASI targets belong to the `wasm` target family**, this forces WASI builds down the browser/JS codegen path, emitting unresolved JavaScript imports.

Here are the exact technical details:

### 1. The Gate in Version `0.2.114` (Working)
In `0.2.114`, the procedural macro backend [wasm-bindgen-macro-support](file:///Users/uriah/.gemini/antigravity/brain/80a36730-0fc8-4b51-ab43-d73a27a64871/scratch/diff_wasm-bindgen-macro-support.txt#L98) and the core library [wasm-bindgen](file:///Users/uriah/.gemini/antigravity/brain/80a36730-0fc8-4b51-ab43-d73a27a64871/scratch/diff_wasm-bindgen.txt#L277) restricted Web-specific code emission to targets where the target OS is `unknown` or `none`:

```rust
#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
```

* **For `wasm32-unknown-unknown`:** Evaluates to `true` $\rightarrow$ emits the actual JavaScript bindings and placeholders.
* **For `wasm32-wasip2` / `wasm32-wasi`:** Evaluates to `false` (since `target_os = "wasi"`) $\rightarrow$ compiles the target-independent fallback stubs, which are just Rust functions that panic at runtime if called, and **do not** generate external imports.

---

### 2. The Change in Version `0.2.115` (Breaking)
In `0.2.115`, these guards were simplified to:

```rust
#[cfg(target_family = "wasm")]
```

* **For `wasm32-wasip2`:** Evaluates to `true` because the target family is `wasm`.
* Consequently, `wasm-bindgen` evaluates the build as a browser target and emits the real JS imports linked to the `__wbindgen_placeholder__` module rather than the dummy stubs:

```rust
#[link(wasm_import_module = "__wbindgen_placeholder__")]
#[cfg(target_family = "wasm")]
extern "C" {
    fn __wbindgen_object_drop_ref(ptr: u32) -> u32;
}
```

---

### 3. Why This Fails at the Component Linker

Because `__wbindgen_object_drop_ref` is declared inside a `#[link(wasm_import_module = "__wbindgen_placeholder__")]` block, the Rust compiler leaves `__wbindgen_placeholder__::__wbindgen_object_drop_ref` as an unresolved import in the compiled `.wasm` file.

When `spin build` compiles the code and tries to translate the core WebAssembly module into a WASIp2 Component, `wasm-component-ld` (or `wit-component`) looks for a WIT interface to resolve the `__wbindgen_placeholder__` import. Since no such WIT interface exists for browser bindings in a WASI execution environment, the linker fails:

```
failed to resolve import __wbindgen_placeholder__::__wbindgen_object_drop_ref
```

codeitlikemiley:
I already created a PR here:

https://github.com/wasm-bindgen/wasm-bindgen/pull/5175

If they merged it  and tag 0.2.123  then it. would be okay to use.

meaning we dont have to deal with changing anything on leptos at all...

@gbj  @benwis

PR #4752, opened 2026-06-01
Quality of life fixes for `reactive_stores_macro`

sabify:

### 1. Enums are now supported
|||
| ---------- | ---------------------------------------------------------------------------------- |
| **Before** | `#[derive(Patch)]` on an enum aborted: *"only structs can be used with `Patch`"*   |
| **After**  | The derive generates a `match` over `(old, new)` and patches each variant's fields |

- Variant fields are numbered with path segments **unique across the whole enum** (mirrors the `Store` derive), so `notify` hits the same triggers the field accessors subscribe to.
- Same-variant unit case → no-op. Mismatched variants → full replace.

### 2. `#[store(skip)]` is now honored
|||
| ---------- | ------------------------------------------------------------------------------------------------- |
| **Before** | Skipped fields were still patched → forced their type to implement `PatchField`                   |
| **After**  | Skipped fields are left untouched; only the path segment is advanced to keep later fields aligned |

### 3. No more panic on malformed `#[patch(...)]`
|||
| ---------- | ----------------------------------------------------------------------------------------------------------------------- |
| **Before** | `expect_or_abort` **panicked the proc-macro** on missing/duplicate closures                                             |
| **After**  | Explicit, span-targeted errors pointing at the offending tokens, naming the expected form `#[patch(\|this, new\| ...)]` |

### DIFF
--- a/reactive_stores/tests/enum_patch.rs
+++ b/reactive_stores/tests/enum_patch.rs
@@ -1,0 +1,154 @@
+//! Regression test: the `Patch` derive supports enums.
+//! - patching within the same variant updates fields in place and only
+//!   notifies the fields that changed,
+//! - patching across variants replaces the value and notifies subscribers.
+
+use reactive_graph::{
+    effect::Effect,
+    owner::Owner,
+    traits::{Get, Read, ReadUntracked},
+};
+use reactive_stores::{Patch, Store};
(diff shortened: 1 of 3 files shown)


sabify:
Will get merged into #4735 to be in sync.

PR #4754, opened 2026-06-01
Fix `leptos_dom` correctness bugs

sabify:

## Changes

- **Clamp timer delays instead of panicking on overflow** — `set_timeout`/`set_interval` converted the delay via `as_millis().try_into().unwrap_throw()`, aborting for any `Duration` over ~24.85 days (`i32::MAX` ms). Now clamps to `i32::MAX`, matching the browser's spec-defined behavior.

- **Guard `location_pathname()` with `is_server()`** — it called `window().location()` unconditionally and panicked off-browser, unlike its sibling `location_hash()`. Now returns `None` on the server.

- **`event_target_checked` uses `unwrap_throw()`** — was the lone helper using std `.unwrap()`; now consistent with the rest of the module (smaller wasm panic path).

- **`request_idle_callback` accepts `FnOnce`** — it required `Fn` while `request_animation_frame` takes `FnOnce`, despite both firing once. Relaxed to `FnOnce` (backward compatible) so callers can consume captures without `Rc<RefCell<…>>` ceremony.

- **Fix `cancelIdleCallback` doc link text** — `IdleCallbackHandle::cancel` rendered `cancelAnimationFrame()` while linking/behaving as `cancelIdleCallback`.

- **Free scheduled closures on cancel instead of leaking** — timer/RAF/idle callbacks were `into_js_value()`'d and the handle kept only the integer id, so `clear()`/`cancel()` stopped the timer but never freed the Rust-side closure (a soft leak, hard under GC starvation). Handles now own their `Closure`; `clear()`/`cancel()` cancel the timer and drop it deterministically, while dropping a handle without cancelling (fire-and-forget) still hands it to the JS engine — no behavior change for `_ = set_timeout(..)` / `_ = request_animation_frame(..)`.

## Breaking change (last item)

`TimeoutHandle`, `IntervalHandle`, `AnimationFrameRequestHandle`, and `IdleCallbackHandle`:
- lose `Copy`/`Clone`/`PartialEq`/`Eq`/`Hash` (a `Closure` is neither `Copy` nor `Clone`); `Debug` is retained.
- `clear()`/`cancel()` now consume `self` (`self` instead of `&self`).

`AnimatedShow` is updated to take-and-clear its stored handle via `update_value`/`try_update_value`. All other workspace call sites (`router`, `debounce`, fire-and-forget sites) needed no changes.


gbj:
8d7f6fed51be6314f4afdeb62810b682424379ef is a genuine breaking change for users of these APIs. The others look good!

sabify:
I believe the correctness changes with API breakage can be placed on major version releases.

As an example, the migration in `leptos-use` is straightforward with minimal changes: 

```diff
diff --git a/src/use_timeout_fn.rs b/src/use_timeout_fn.rs
index c905c31..e336402 100644
--- a/src/use_timeout_fn.rs
+++ b/src/use_timeout_fn.rs
@@ -76,8 +76,7 @@ where
             let timer = Arc::clone(&timer);
 
             move || {
-                let timer = timer.lock().unwrap();
-                if let Some(timer) = *timer {
+                if let Some(timer) = timer.lock().unwrap().take() {
                     timer.clear();
                 }
             }
diff --git a/src/use_websocket.rs b/src/use_websocket.rs
index 4535e7f..33555f9 100644
--- a/src/use_websocket.rs
+++ b/src/use_websocket.rs
@@ -448,7 +448,7 @@ where
                     && ws
                         .get_value()
                         .is_some_and(|ws: WebSocket| ws.ready_state() != WebSocket::OPEN)
-                    && reconnect_timer_ref.get_value().is_none()
+                    && reconnect_timer_ref.with_value(|timer| timer.is_none())
                 {
                     reconnect_timer_ref.set_value(
                         set_timeout(
@@ -474,10 +474,11 @@ where
             let on_error = Arc::clone(&on_error);
 
             Some(Arc::new(move || {
-                if let Some(reconnect_timer) = reconnect_timer_ref.get_value() {
-                    reconnect_timer.clear();
-                    reconnect_timer_ref.set_value(None);
-                }
+                reconnect_timer_ref.update_value(|timer| {
+                    if let Some(reconnect_timer) = timer.take() {
+                        reconnect_timer.clear();
+                    }
+                });
 
                 if let Some(web_socket) = ws.get_value() {
                     let _ = web_socket.close();
diff --git a/src/utils/filters/debounce.rs b/src/utils/filters/debounce.rs
index 20a50ea..04b64eb 100644
--- a/src/utils/filters/debounce.rs
+++ b/src/utils/filters/debounce.rs
@@ -29,9 +29,8 @@ where
 
     let clear_timeout = move |timer: &Arc<Mutex<Option<TimeoutHandle>>>| {
         let mut timer = timer.lock().unwrap();
-        if let Some(handle) = *timer {
+        if let Some(handle) = timer.take() {
             handle.clear();
-            *timer = None;
         }
     };
 
diff --git a/src/utils/filters/throttle.rs b/src/utils/filters/throttle.rs
index fde6aac..fa4b0f5 100644
--- a/src/utils/filters/throttle.rs
+++ b/src/utils/filters/throttle.rs
@@ -42,9 +42,8 @@ where
     let t = Arc::clone(&timer);
     let clear = move || {
         let mut t = t.lock().unwrap();
-        if let Some(handle) = *t {
+        if let Some(handle) = t.take() {
             handle.clear();
-            *t = None;
         }
     };
 ```

Or `thaw-ui` already uses the right API by using `take()` and `clear()`.

**Comment by gbj** [comment]
Two-part response — feel free to skip to the bold at the end.

Obviously breaking changes are allowed in major version releases. My point about breaking user code is that this kind of change (removing both `Copy` and `Clone` from a struct's trait implementations) is the kind of thing that tends to drive people nuts.

Here's a trivial example prior to this PR:
```rust
#[component]
    let handle = set_interval(
        || leptos::logging::log!("tick"),
        Duration::from_millis(1000),
    )
    .unwrap();

    view! {
        // pointless conditional for the sake of the example
        <Show when=|| true>
            <button
                on:click=move |_| { handle.clear(); }
            >
                "Cancel the interval"
            </button>
        </Show>
    }
}
```

Here's the result after this PR
```
error[E0525]: expected a closure that implements the `Fn` trait, but this closure only implements `FnOnce`
  --> src/lib.rs:12:5
   |
12 | /     view! {
13 | |         // pointless conditional for the sake of the example
14 | |         <Show when=|| true>
15 | |             <button
16 | |                 on:click=move |_| { handle.clear(); }
   | |                                     ------ closure is `FnOnce` because it moves the variable `handle` out of its environment
...  |
20 | |         </Show>
21 | |     }
   | |     ^
   | |     |
   | |_____this closure implements `FnOnce`, not `Fn`
   |       the requirement to implement `Fn` derives from here
   |
   = note: required for `TypedChildrenFn<HtmlElement<Button, (...,), ...>>` to implement `ToChildren<{closure@src/lib.rs:12:5: 21:6}>`
```

A user who's a relative beginner is probably going to be baffled. (I have helped many people in this situation over the years). This is the point at which many people give up and stop trying to migrate to the new version.

(The problem is amplified by the fact that `.clear()` takes `self` not `&mut self` or `&self`, so initial attempts to fix it by sticking it in an Arc etc. won't work)

A user who's pretty good at reading Rust error messages is going to hack their way into something like this
```rust
use leptos::prelude::*;
use std::time::Duration;

#[component]
    let handle = StoredValue::new(Some(
        set_interval(
            || leptos::logging::log!("tick"),
            Duration::from_millis(1000),
        )
        .unwrap(),
    ));

    view! {
        // pointless conditional for the sake of the example
        <Show when=|| true>
            <button
                on:click=move |_| { handle.write_value().take().unwrap().clear(); }
            >
                "Cancel the interval"
            </button>
        </Show>
    }
}
```

In exchange, they've received: a closed memory leak on systems that don't support weak refs. Which is fair enough, but to be perfectly honest, this should be our problem to solve, not the user's.

**My suggestion**: Use the existing resource ownership and cleanup mechanisms built into the framework. i.e., rather than handing ownership to the JS runtime with `.into_js_value()`, hand ownership to the reactive graph with `Owner::on_cleanup(move || scheduled_closure.free())` or whatever.

sabify:
Thanks for the clarification. I now understand what you mean by "breaking change". You are right, it really matters. 

Your suggestion is genuine but it has the original pitfall when using a timer in a timer's callback where ownership is not known there and we have to fall back into `into_js_value()`. (Please correct me if I'm wrong)
Also, your suggestion is not an immediate-free approach and is bound to the owner disposal/cleanup. This is less concern and only happens if your really want to run many timers in one owner scope without dropping the owner immediately.

I just provided a cloneable approach in https://github.com/leptos-rs/leptos/pull/4754/commits/0751418a7fe11b5357e19ec3212e30e383b49945, if `Clone` is an acceptable approach.

The usage would be like:

```rust
#[component]
    let handle = set_interval(
        || leptos::logging::log!("tick"),
        Duration::from_millis(1000),
    )
    .unwrap();

    view! {
        // pointless conditional for the sake of the example
        <Show when=|| true>
            <button
                id="cancel-interval"
                on:click={
                    let handle = handle.clone();
                    move |_| { handle.clear(); }
                }
            >
                "Cancel the interval"
            </button>
        </Show>
    }
}
```

Please let me know if this approach is desirable or you need me to apply your suggestion.

sabify:
I'm reading it again and again and the ergonomics really matters here! The emitted error on the "cloneable" approach is not helpful here and is the same as you've posted. It really confuses the users that they have to clone it. I will implement your suggestion.

sabify:
Done!

https://github.com/leptos-rs/leptos/pull/4754/commits/0751418a7fe11b5357e19ec3212e30e383b49945

gbj:
Looks good, thanks!

PR #4755, opened 2026-06-02
Harden leptos actix server integrations

sabify:

These fixes  are mostly identical to the Axum counterpart.

### Panic fixes (worker DoS)
- **`redirect()`** — no longer `expect()`s on `HeaderValue::from_str`. User-influenced redirect targets containing CR/LF/NUL (header injection attempts) returned an `InvalidHeaderValue` that panicked the worker; now logs a warning and skips the redirect.
- **`set_default_content_type()`** — replaces `from_str(..).unwrap()` with a match; an invalid `&str` (e.g. containing a NUL) now skips the header instead of crashing.
- **Unsupported `SsrMode`** — both `LeptosRoutes` impls ended their match with `_ => unreachable!()`. A future enum variant would compile fine then panic at runtime; replaced with a helper that logs and serves `500`.

### Correctness fixes
- **Accept header parsing** — `redirect()` and `handle_server_fns_with_context` used `accept.contains("text/html")`. Now parses media ranges with the `mime` crate, so `text/html;q=0` (explicit refusal) and substring false-positives like `application/x-text/html-fake` are handled correctly.
- **HEAD requests** — registered HEAD handlers returned a bare `200 OK` (wrong status/headers per RFC 9110), and the `ServiceConfig` impl didn't register HEAD at all. HEAD now routes to the GET render handler via `guard::Get().or(guard::Head())`, mirroring GET's real status/headers across both mount paths.

### Static route fixes
- **TOCTOU error leak** — a failed `NamedFile::open` after `try_exists` rendered the raw `io::Error` into the 500 body, disclosing server paths. Now logs server-side and returns a generic 500.
- **Unbounded cache** — `STATIC_HEADERS` (a `HashMap`) grew one entry per distinct path forever. Replaced with an `lru::LruCache` (default 1024, override via `LEPTOS_STATIC_HEADERS_CACHE_SIZE`).
- **Atomic writes** — `tokio::fs::write` truncates then writes; a crash mid-write left empty/partial files that were later served. Now writes to a temp file and atomically `rename`s it over the target.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -204,6 +204,33 @@ impl ExtendResponse for ActixResponse {
     }
 }
 
+/// Returns whether an `Accept` header value indicates the client will accept
+/// an HTML response — i.e. an ordinary browser navigation or a plain `<form>`
+/// submission, as opposed to a programmatic client expecting structured data.
+///
+/// Unlike a naive `contains("text/html")` check, each comma-separated media
+/// range is parsed with the [`mime`] crate and an explicit `q=0` refusal is
+/// honoured. So `text/html;q=0` (the client refusing HTML) and
+/// `application/x-text/html-fake` (an unrelated, unparseable range) are both
+/// correctly treated as *not* accepting HTML.
+fn accept_header_includes_html(accept: &str) -> bool {
+    accept.split(',').any(|range| {
+        let Ok(media) = range.trim().parse::<mime::Mime>() else {
+            return false;
+        };
+        if media.type_() != mime::TEXT || media.subtype() != mime::HTML {
+            return false;
+        }
+        // honour an explicit `q=0`, which means the client refuses HTML
+        match media.get_param("q") {
+            Some(q) => {
+                q.as_str().parse::<f32>().map(|w| w > 0.0).unwrap_or(true)
+            }
+            None => true,
+        }
+    })
+}
+
 /// Provides an easy way to redirect the user from within a server function.
 ///
 /// Calling `redirect` in a server function will redirect the browser in three
@@ -266,7 +293,7 @@ pub fn redirect(path: &str) {
             .headers()
             .get(ACCEPT)
             .and_then(|v| v.to_str().ok())
-            .map(|v| v.contains("text/html"))
+            .map(accept_header_includes_html)
             .unwrap_or(false);
         if accepts_html {
             // if the request accepts text/html, it's a plain form request and needs
@@ -388,7 +415,7 @@ pub fn handle_server_fns_with_context(
                                 .headers()
                                 .get(ACCEPT)
                                 .and_then(|v| v.to_str().ok())
-                                .map(|v| v.contains("text/html"))
+                                .map(accept_header_includes_html)
                                 .unwrap_or(false);
                             let referrer = req.headers().get(REFERER).cloned();
 
@@ -1708,7 +1735,8 @@ mod tests {
     // `actix_web::test`, which would shadow the `#[test]` attribute macro.
     use super::{
         ActixResponse, ExtendResponse, HttpResponse, LOCATION, OrPoisoned,
-        Owner, Request, ResponseOptions, header, provide_context, redirect,
+        Owner, Request, ResponseOptions, accept_header_includes_html, header,
+        provide_context, redirect,
     };
     use actix_web::test::TestRequest;
 
@@ -1774,5 +1802,33 @@ mod tests {
                 .map(|v| v.as_bytes()),
             Some(&b"text/html; charset=utf-8"[..])
         );
+    }
+
+    #[test]
+    fn accept_header_plain_navigation_is_html() {
+        // typical browser navigation
+        assert!(accept_header_includes_html(
+            "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+        ));
+        assert!(accept_header_includes_html("text/html"));
+        assert!(accept_header_includes_html("text/html; charset=utf-8"));
+        assert!(accept_header_includes_html("text/html;q=0.1"));
+    }
+
+    #[test]
+    fn accept_header_explicit_refusal_is_not_html() {
+        // `q=0` means the client explicitly does not want HTML
+        assert!(!accept_header_includes_html(
+            "text/html;q=0, application/json"
+        ));
+        assert!(!accept_header_includes_html("text/html;q=0.0"));
+    }
+
+    #[test]
+    fn accept_header_substring_is_not_html() {
+        // these contain the literal substring "text/html" but are not it
+        assert!(!accept_header_includes_html("application/x-text/html-fake"));
+        assert!(!accept_header_includes_html("application/json"));
+        assert!(!accept_header_includes_html("*/*"));
     }
 }
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1341,6 +1341,8 @@ async fn write_static_route(
     path: &str,
     html: &str,
 ) -> Result<(), std::io::Error> {
+    use leptos_integration_utils::write_file_atomic;
+
     // Reject anything that would escape the site root before caching headers
     // or touching the filesystem.
     let Some(file_path) = static_path(options, path) else {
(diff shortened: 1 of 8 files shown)


sabify:
Even retries fail; good for Github Actions!

@gbj Please re-run the failed CI.

integrations/actix/src/lib.rs
```diff
@@ -1301,9 +1301,35 @@ impl StaticRouteGenerator {
...
+/// Default upper bound on the number of per-path [`ResponseOptions`] entries
+/// cached for static routes. Without a bound the cache grew for the life of the
+/// process, one entry per unique static path served (e.g. attacker-driven slugs
+/// on a regenerated `/posts/{slug}` route).
+///
+/// Eviction is graceful: the static file is still served from disk, the cache
+/// only drops the custom headers/status captured at generation time for the
+/// evicted path (re-populated on the next regeneration). 1024 covers a typical
+/// static site's working set for a worst case on the order of ~1 MB.
+const STATIC_HEADERS_DEFAULT_CAPACITY: std::num::NonZeroUsize =
```
gbj: prefer `use` to the fully-qualified path here

integrations/actix/src/lib.rs
```diff
@@ -1301,9 +1301,35 @@ impl StaticRouteGenerator {
...
+    };
+
+/// Environment variable that overrides [`STATIC_HEADERS_DEFAULT_CAPACITY`].
+/// A missing, unparseable, or zero value falls back to the default.
+const STATIC_HEADERS_CAPACITY_ENV: &str = "LEPTOS_STATIC_HEADERS_CACHE_SIZE";
+
 static STATIC_HEADERS: LazyLock<
-    std::sync::RwLock<HashMap<String, ResponseOptions>>,
-> = LazyLock::new(Default::default);
+    std::sync::RwLock<lru::LruCache<String, ResponseOptions>>,
```
gbj: prefer `use` to the fully-qualified path here

integrations/actix/src/lib.rs
```diff
@@ -1301,9 +1301,35 @@ impl StaticRouteGenerator {
...
+> = LazyLock::new(|| {
+    let capacity = std::env::var(STATIC_HEADERS_CAPACITY_ENV)
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+        .and_then(std::num::NonZeroUsize::new)
+        .unwrap_or(STATIC_HEADERS_DEFAULT_CAPACITY);
+    std::sync::RwLock::new(lru::LruCache::new(capacity))
```
gbj: prefer `use` to the fully-qualified path here

gbj:
a18d7e7d54ab869917cbf5c5e333a57553c1d05d duplicates code added to `leptos_axum` in e9e34ca99fcdbf2cc0a597835594a9d380b6a2ee — this should probably be refactored into a shared helper in `leptos_integration_utils`

d7db40fa66c964e9ac5d3087aa42eedfba925a66 seems... extreme. 

The rest seem good.

sabify:
Done.

    d7db40f seems... extreme.

Edge cases in production would seem extreme too! (lol)

gbj:
To expand a little on what I mean by "extreme," here's my understanding of how this bug is triggered. I think it's been a couple years since I wrote the SSG code, so I might be missing something, but it looks to me like this commit just modifies the body of `write_static_route`, which is called during `StaticRouteGenerator::new()`, which is called during `generate_route_list_with_exclusions_and_ssg_and_context`, which the user will call during the startup of the application or during the runtime of the static site generator. `write_static_route` itself does not check whether the file exists; it will overwrite the contents of the file when it's called. The file-serving handler *does* check whether the file exists.

So the mechanism for this bug to be triggered is:
- `generate_route_list_with_exclusions_and_ssg_and_context` runs
- it spawns a Tokio task per route, each of which calls `write_static_route`
- while one of these is running, the application is killed catastrophically, i.e., in such a way that `tokio::fs::write` never returns an `Err(_)` but the process is just killed (SIGKILL, OOM, etc.). Due to the joys of parallelism this happens in such a way that it reveals the non-atomic nature of `std::fs::write`, i.e., it leaves a truncated file.
- The user then *actually deploys* the generated static files, including the truncated one, even though the process was killed halfway through?

It's that last step that confuses me. For the truncated file to be served, either the user needs to manually deploy the built static files (after a failure during the build process and a non-success exit code), or the user needs to restart the server process itself. The commit description says
```
# after restart, requests to /foo may return an empty 200 response
```
But `generate_route_list_with_ssg` (etc.) *runs during restart*, which means that the truncated file will be *overwritten* by `write_static_route`.

The commit text handwaves about "a panic in a sibling future" as a possible cause of killing the process. I might be wrong, but I don't *think* this is how Tokio works? i.e., I do think that the only way to end up with a truncated file here is to kill the whole process.

I could be wrong!

sabify:
Please let me provide you the full analysis:

## Where your mental model is correct

- `write_static_route` does **no existence check** — it unconditionally truncates and overwrites (`:1338`). ✓
- The serving handler decides what to serve with `tokio::fs::try_exists` (`:1422`) and, if present, serves it straight off disk via `NamedFile::open` (`:1476`) — with **no "is this file complete?" check**. ✓
- A process death between the truncate and write completion persists a truncated/empty file. ✓ The truncate is an immediate inode-metadata operation, so a *process* kill (not power loss) leaves the kernel page cache/inode intact — the empty file is genuinely there on the next open, no fsync required.

## The gap in your model: `write_static_route` has *two* callers

Your analysis followed only the **ahead-of-time (AOT)** path:

> `StaticRouteGenerator::generate()` → `generate_static_files` → `path.build()` → `write_static_route` (`:1264`)

But the same writer is also invoked from the **live request handler**:

> `handle_static_route` → on cache miss → `ResolvedStaticPath::build()` → `write_static_route` (`:1443`)

`handle_static_route` is what's actually wired into the running server for `SsrMode::Static` routes (`:1606-1614`, `:1715-1723`). On a request: if the file exists, serve from disk; if not, render, write via `write_static_route`, serve. **Files land in `site_root` and are served directly — no manual redeploy of build artifacts is involved.** That deployment step, which confused you, isn't required for the bug.

## Why "regeneration on restart overwrites it" doesn't save you

Your claim holds *only* for the canonical AOT setup, and the example shows why: in `examples/static_routing/src/main.rs:18`, `static_routes.generate(&opts).await` is awaited to completion **before** the server starts serving. In that exact configuration a prior truncated file is overwritten before any request sees it. So for **a param-free `Static` route, AOT-awaited-before-serve, no ISR (Incremental Static Regeneration)**, you're right — the bug is masked. That's the one safe case, and your reasoning nails it.

It breaks down because that's not the general case. The decisive fact, verified in code:

**Dynamic-param `Static` routes are on-demand only.** `RouteListing::into_static_paths` (`router/src/generate_route_list.rs:79-81`):

```rust
let params = self.static_route()?.to_prerendered_params().await;
Some(StaticPath::new(self.path).into_paths(params))
```

- `StaticRoute::new()` defaults `prerender_params: None` (`router/src/static_routes.rs:53,59-61`), so `to_prerendered_params()` returns `None`.
- In `into_paths`, a `Param`/`Splat` segment with `params == None` produces an **empty** vec — `new_paths` is never filled (`router/src/static_routes.rs:235-261`).

So a route like `/blog/{slug}` without `prerender_params` contributes **zero** paths to AOT generation. `static_routes.generate()` never touches it — at first start or on restart. It is generated purely on demand by `handle_static_route`. (Secondarily, `.generate()` is itself optional: it's handed back to the caller, and a pure lazy app never calls it. But you don't even need that loophole — dynamic routes are on-demand regardless.)

## The verified concrete paths to the bug

**Path A — crash, survives restart (strongest; no ISR needed; holds even if you call `.generate()` faithfully):**

1. App has `SsrMode::Static` route `/blog/{slug}`, no `prerender_params`.
2. `static_routes.generate(&opts).await` runs at startup and produces **nothing** for it.
3. First request `/blog/hello` → file missing → render → `write_static_route` → `tokio::fs::write` **creates `blog/hello.html`, truncates to 0**, begins `write_all`.
4. Process is killed (SIGTERM/OOM/SIGKILL) after the create/truncate, before `write_all` finishes → 0-byte (or partial) file persists.
5. Restart → `.generate()` again produces nothing for `/blog/{slug}`.
6. Request `/blog/hello` → `try_exists == true` → `NamedFile::open` serves the 0-byte file → **empty `200`**.

For a never-before-generated route this is even starker: "create empty file → die → serve empty file."

**Path B — no crash at all (requires ISR):**

1. `/foo.html` exists; a `.regenerate(...)` stream fires (`router/src/static_routes.rs:77-88`).
2. The long-lived regenerate loop (`:361-373`) calls `write_static_route` on the **existing** file → `tokio::fs::write` truncates `/foo.html` to 0 in place.
3. A concurrent request hits `handle_static_route`, sees `try_exists == true`, `NamedFile::open`s the 0-byte window → **empty `200`**. No process death, no restart.

## On "a panic in a sibling future"

You were right to be skeptical of this phrasing in the commit message.

- Each path is `spawn`ed as an **independent** task (`router/src/static_routes.rs:315`). Under Tokio's default unwinding, a panic in one spawned task is caught at the `JoinHandle` boundary; it does **not** abort sibling tasks or the process. So a sibling task's panic does not truncate another route's file.
- It becomes true only if the panic actually *kills the process* — i.e., the crate is built with `panic = "abort"` (panic → `SIGABRT` → process death). At that point it's really just "process abort," not Tokio propagating a panic across futures. The commit message conflates the two by listing it next to SIGTERM/OOM.
- The honest non-crash variant the message *should* have cited: **dropping/cancelling the `write` future after the truncate but before `write_all` completes** — e.g., runtime shutdown dropping pending tasks, or task cancellation. The `create` commits the truncate; if the `write_all` future is then dropped, the empty file remains, no SIGKILL required.

## "Only a whole-process kill can truncate" — too strong

Path B serves a truncated response with no death at all (concurrent reader during in-place regeneration). And future-drop on shutdown/cancellation truncates without SIGKILL. So the truncation is not gated on a hard process kill.

## Bottom line

- **Mechanics:** correct.
- **"User must manually deploy artifacts":** not the real path — on-demand/ISR writes into `site_root` and serves directly.
- **"Restart regenerates and overwrites it":** true only for param-free `Static` routes with `.generate().await` completed before serving and no ISR. False for dynamic-param routes (on-demand only, never in the AOT set) and lazy-only apps — which is where the bug actually lives.
- **"Only a process kill can truncate":** too strong — the ISR concurrent-reader window and future-drop on shutdown both truncate without a hard kill.
- **"Panic in a sibling future":** misleading under default unwinding; relevant only when a panic aborts the whole process.

The fix is sound for every one of these triggers (crash-survives-restart, ISR concurrent read, future-drop on shutdown). The commit *message* simply over-attributes the cause to crashes when the cleaner justification is the dynamic-route on-demand path plus the concurrent-reader window during regeneration. A regression test (`write_static_route_is_atomic`) asserts the target is written in full with no temp leftovers.

gbj:
That contains two useful points I had missed:
- I missed the other call site of `write_static_route`, which happens dynamically and not on startup
- There's a TOCTOU issue in which a request that happens mid-write can potentially return the empty file, not the written file
  - handler A begins writing the file
  - handler B sees that the file exists and returns a response with the file
  - handler A finishes writing the file

Fair enough. It's a worthwhile change.

Could you please refactor the atomic file write helper into `leptos_integration_utils` and share it between the Axum and Actix integrations?

sabify:
> Could you please refactor the atomic file write helper into leptos_integration_utils and share it between the Axum and Actix integrations?

Good call! I totally forgot about the axum part!

Done.

PR #4757, opened 2026-06-02
Fix `any_spawner` correctness bugs

sabify:

## Changes

**Make `init_local_custom_executor` truly thread-local**
It was documented as a per-thread override but stored its dispatch table in the *global* `EXECUTOR_FNS` while its function pointers read a thread-local instance. This panicked on any spawn from another thread (`unwrap()` on `None`) and locked out every other `init_*`. Added a per-thread `LOCAL_EXECUTOR_FNS` table with a `current_executor_fns()` resolver that prefers the thread-local executor and falls back to the global one; the global table is now left untouched.

**Reject local re-init without residue**
`init_local_custom_executor` could return `AlreadySet` while leaving the thread-local instance populated, permanently pinning the executor and breaking retries. It now checks the per-thread "already configured" flag up front and rejects before allocating, dropping the rejected executor.

**Claim the global slot before storing a custom executor**
`init_custom_executor` stored the executor in a `'static` slot *before* claiming `EXECUTOR_FNS`; losing the race left the executor leaked in static memory forever. It now claims the global slot first and only stores the instance on success, so a rejected executor is dropped.

**Surface a dropped `tick()` instead of returning silently**
`Executor::tick()` discarded the channel result, so if the executor dropped the synchronization task it returned as though a cycle had elapsed. It now panics with a clear message when the task is dropped. Real executors are unaffected; the signature is unchanged.

**Emit the uninitialized-spawn diagnostic in release builds**
The `tracing::error!` for spawning before initialization was gated on `all(debug_assertions, feature = "tracing")`, so it never fired in release builds — `spawn` dropped silently and `spawn_local` panicked, both with no log. The diagnostic now fires in debug and release; non-tracing paths are unchanged. Also removes the unused-`caller` release warning.

### DIFF
--- a/any_spawner/Cargo.toml
+++ b/any_spawner/Cargo.toml
@@ -28,6 +28,7 @@ tokio = { default-features = false, features = [
 ] , workspace = true }
 wasm-bindgen-test = { workspace = true, default-features = true }
 serial_test = { workspace = true, default-features = true }
+tracing = { workspace = true, default-features = true }
 
 [features]
 async-executor = ["dep:async-executor"]
--- a/any_spawner/src/lib.rs
+++ b/any_spawner/src/lib.rs
@@ -420,6 +420,16 @@ impl Executor {
             > = const { OnceLock::new() };
         };
 
+        // Reject re-initialization up front, before allocating the box or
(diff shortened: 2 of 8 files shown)


sabify:
@gbj Please re-run the failed CI.

PR #4760, opened 2026-06-04
fix+perf: server `Url::origin()` returns the real request origin

### LINKED ISSUE #4153: bug: url `.origin()` returns `leptos.dev`
**Describe the bug**

**Leptos Dependencies**

For example:
```toml

leptos = { version = "=0.8.2" }
leptos_meta = { version = "=0.8.2" }
leptos_router = { version = "=0.8.2" }
leptos_axum = { version = "=0.8.2", optional = true }
```

**To Reproduce**

Using the axum-ssr template, create a component that uses the `use_url()` router hook. My component looks similar to this: 

```rust
    let url = leptos_router::hooks::use_url();
    view! {
            <label
                for="slug-input"
            >
                {move || format!("{}/", url.get().origin().to_string())}
            </label>
            <input
                id="slug-input"
                type="text"
            />
}
```

Then navigate to the page that has this input, observe that the label says `http://localhost:{your-port}`. 

Refresh the page observe that the label is now `http://leptos.dev`. I believe this based on the `const BASE` defined at https://github.com/leptos-rs/leptos/blob/131251b3612e40e1934b550a34351f280c4ebaa4/router/src/location/mod.rs#L25

**Screenshots**

What i see when navigating to the page via hyperlink: 

Image

what I see when I navigate to this page via url or refresh the page: 

Image

**Next Steps**

I'm happy to make a PR, I just don't know where the problem stems from and what the cause is

- [ ] I will make a PR

**Additional context**

sabify:
Server-side request URLs were parsed by resolving the path against a hardcoded `BASE = "https://leptos.dev"` (a constant that `url::Url` needs because it can't decompose a relative reference without a base). That dummy host leaked into `Url::origin()`, so `use_url().origin()` returned a fake constant server-side — never the real host — and mismatched the client after hydration. This makes the server derive the **real** origin from the request and removes the hardcoded base.

### Bug fix
- `Url::origin()` (via `use_url()`) now returns the actual request origin server-side (e.g. `http://127.0.0.1:3000`) instead of the hardcoded dummy (`https://leptos.dev` on axum, `http://leptos` on actix).
- Server origin now **matches the client** origin after hydration, fixing a latent SSR/client mismatch.
- Origin is sourced correctly per integration: axum from the URI authority / `Host` header with `X-Forwarded-Proto` (defaults to `http`); actix from `connection_info` (honors reverse-proxy forwarding headers).
- A leading `//` stays literal in the path instead of being reinterpreted as a protocol-relative authority, so injected input can't override `origin`.

### Performance / cleanup
- Drops the per-request `url::Url::parse(BASE)` that ran on every render (the base was reparsed each call); a live request now does a single `url::Url` parse.
- Host-less contexts (static-site generation, route discovery) are parsed base-free — split on the first `?`, with query pairs decoded exactly once via `url::form_urlencoded::parse` (same engine as `url::Url::query_pairs`).
- Removes the now-dead `BASE` constant and the unreachable `LocationProvider::parse` default (`BrowserUrl` already overrides it with the real `window` origin); `parse` becomes a required trait method.

### Compatibility
`path` / `search` / `search_params` are unchanged for normal inputs. `origin()` changes value server-side (fake dummy → real host); like every web framework it reflects the `Host` header.

This may close #4153.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -784,9 +784,7 @@ fn provide_contexts(
     meta_context: &ServerMetaContext,
     res_options: &ResponseOptions,
 ) {
-    let path = leptos_corrected_path(&req);
-
-    provide_context(RequestUrl::new(&path));
+    provide_context(request_url(&req));
     provide_context(meta_context.clone());
     provide_context(res_options.clone());
     provide_context(req);
@@ -794,13 +792,19 @@ fn provide_contexts(
     leptos::nonce::provide_nonce();
 }
 
-fn leptos_corrected_path(req: &HttpRequest) -> String {
+fn request_url(req: &HttpRequest) -> RequestUrl {
+    // Prefix the real request origin (scheme://host) so that `Url::origin()`
+    // is correct server-side and matches the client after hydration.
+    // `connection_info` resolves the scheme and host, honoring reverse-proxy
+    // forwarding headers.
+    let conn = req.connection_info();
+    let origin = format!("{}://{}", conn.scheme(), conn.host());
     let path = req.path();
     let query = req.query_string();
     if query.is_empty() {
-        "http://leptos".to_string() + path
+        RequestUrl::new(&format!("{origin}{path}"))
     } else {
-        "http://leptos".to_string() + path + "?" + query
+        RequestUrl::new(&format!("{origin}{path}?{query}"))
     }
 }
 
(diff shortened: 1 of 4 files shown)


lukexor:
Thanks for fixing this! I've been using a custom use_url wrapper hook for ages and extract::Host from axum_extra (which was recently deprecated) so not having to handle this manually (and the various edge cases) would be great!

gbj:
Thanks!

sabify:
@gbj, can this also close issue #4153?

PR #4761, opened 2026-06-04
perf(leptos_axum): cut per-render allocations in the SSR hot path

sabify:
### 1. Move request `Parts` out instead of cloning
`handle_response_inner` called `generate_request_and_parts(req)` just to get the head, which cloned the full `Parts` (a fresh `HeaderMap`, every header, and `Extensions`) and rebuilt a `Request` the caller immediately discarded. The body is never read during SSR — rendering is driven by the path string and the head we pass to `provide_contexts` — so this used `req.into_parts()` to move the parts out and drop the body.

- **~6 fewer heap allocations per render (~863 bytes), ~192 ns → ~0 ns/render**
- `generate_request_and_parts` is untouched; it's still used where the body is genuinely needed (server fns, `LeptosContext::call`).

### 2. Index route listings by path for O(1) lookup
The `render_route` custom-handler path scanned all route listings linearly (`paths.iter().find(...)`, O(R) string compares per request) and allocated an owned `String` per request from `MatchedPath` just to dodge a borrow. Now the listings are indexed into a `HashMap<String, AxumRouteListing>` once at router-build time and looked up by the borrowed `&str`.

- **Lookup O(R) → O(1); one per-request `String` allocation removed**
- `or_insert` preserves the previous first-match semantics; resolves the existing "should probably be optimized" TODO.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -775,8 +775,22 @@ where
         app_fn.clone(),
     );
 
+    // Index the route listings by path once, at router-build time, so the
+    // per-request handler resolves a listing in O(1) by a borrowed `&str`
+    // instead of scanning `paths` linearly and allocating an owned path
+    // `String` on every request. `or_insert` preserves the first-match
+    // semantics of the previous `paths.iter().find(...)`.
+    use std::collections::HashMap;
+    let by_path: HashMap<String, AxumRouteListing> = {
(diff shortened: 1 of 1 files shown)


gbj:
Looks like there are merge conflicts now that I can't fix, but this looks good to me otherwise.

sabify:
Done.

PR #4763, opened 2026-06-06
Performance optimizations for `leptos`

sabify:
Four focused, independently-reviewable commits targeting per-request/per-view overhead that this crate owns directly. All changes preserve behavior and keep byte-identical output.

### Changes

**1. Memoize `HydrationScripts` hashed asset names**
`HydrationScripts` re-ran `current_exe()` + `stat` + `open`/`read` of the hash file (plus ~5 allocations) on *every* SSR response to recompute process-static filenames. Now resolved once into a `OnceLock<(Arc<str>, Arc<str>)>`, mirroring the existing `SPLIT_MANIFEST` cache. ~12.7µs → ~5ns per render; removes blocking disk I/O from the request path.

**2. `ErrorBoundary`: render children straight into the output buffer**
The sync path rendered the whole subtree into a scratch `String`, then copied it into the real buffer on the (common) no-error path. Now renders directly into `buf` with a length-marker + `truncate` rollback for the error case. −2 allocations and −5.9 KB allocated per render (scales with subtree size); error path is break-even. Adds an SSR regression test.

**3. `TextProp`: non-boxed `Static` variant**
`TextProp` boxed every input into `Arc<dyn Fn>`, even literals and owned strings. Now an enum with a `Static(Oco)` variant; the reactive `Fn` variant is unchanged. Static literals go from 1 allocation to 0; owned strings from 3 to 2. Trade-off: type grows 8→24 bytes. Scoped to `TextProp` only (`ViewFn`/`OptionGetter` aren't free wins — see commit body).

**4. Cache the `LEPTOS_WATCH` flag in `AutoReload`**
`AutoReload` took the global environment lock via `std::env::var` on every page render for a process-static flag. Now cached in a `OnceLock<bool>`. ~89ns → ~0.3ns per render and removes env-lock contention across concurrent workers.

### DIFF
--- a/leptos/src/error_boundary.rs
+++ b/leptos/src/error_boundary.rs
@@ -332,21 +332,24 @@ where
     ) {
         // first, attempt to serialize the children to HTML, then check for errors
         let _hook = throw_error::set_error_hook(self.hook);
-        let mut new_buf = String::with_capacity(Chil::MIN_LENGTH);
+        // Render children straight into `buf`. If a child throws, roll back to
+        // this marker (`String::truncate` keeps the reserved capacity) and emit
+        // the fallback instead. This avoids the scratch `String` allocation plus
+        // a full second pass copying the whole subtree on the common no-error path.
+        let marker = buf.len();
         let mut new_pos = *position;
         self.children.to_html_with_buf(
-            &mut new_buf,
+            buf,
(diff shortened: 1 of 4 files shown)


gbj:
Thanks!

PR #4764, opened 2026-06-06
Performance optimization for `tachys`

sabify:
Four focused commits that remove redundant allocations, synchronization, and linear scans on the SSR and hydration hot paths. All changes are behavior-preserving (byte-identical output / identical semantics) and verified across feature sets.

### Changes

- **`PositionState`: `Arc<RwLock>` → `Rc<Cell>`** — The hydration cursor is a 1-byte `Copy` enum on a `!Send` path, so the locks and atomics were pure overhead. Now `get`/`set` lower to a plain load/store and clones are non-atomic.

- **DOM template caches: `Vec` → `FxHashMap`** — The three client-side template caches were scanned linearly (O(n), and O(n·html_len) for the HTML/SVG string cache). Switched to `FxHashMap` with `entry().or_insert_with()` for O(1) amortized lookups.

- **Out-of-order suspense splice** — `poll_next` formatted two needle strings and ran two full-buffer scans to locate a resolved chunk's placeholder. Now derives the closing marker by flipping one byte of the opening marker and scans only the tail after the opening, dropping one alloc and one re-scan per resolved chunk.

- **Branch markers via `format_args!`** — When `mark_branches` is on (islands, hot-reload diffing), each `AnyView` node and keyed item formatted its `TypeId` into a throwaway `String`. New `open_branch_fmt`/`close_branch_fmt` write directly into the buffer, eliminating a transient heap alloc per node.

### DIFF
--- a/tachys/src/renderer/dom.rs
+++ b/tachys/src/renderer/dom.rs
@@ -8,7 +8,7 @@ use crate::{
     ok_or_debug, or_debug,
     view::{Mountable, ToTemplate},
 };
-use rustc_hash::FxHashSet;
+use rustc_hash::{FxHashMap, FxHashSet};
 use std::{
     any::TypeId,
     borrow::Cow,
@@ -23,7 +23,7 @@ pub struct Dom;
 
 thread_local! {
     pub(crate) static GLOBAL_EVENTS: RefCell<FxHashSet<Cow<'static, str>>> = Default::default();
-    pub static TEMPLATE_CACHE: RefCell<Vec<(Cow<'static, str>, web_sys::Element)>> = Default::default();
+    pub static TEMPLATE_CACHE: RefCell<FxHashMap<Cow<'static, str>, web_sys::Element>> = Default::default();
 }
 
 pub type Node = web_sys::Node;
@@ -512,14 +512,12 @@ impl Dom {
         thread_local! {
             static TEMPLATE_ELEMENT: LazyCell<HtmlTemplateElement> =
                 LazyCell::new(|| document().create_element(Dom::intern("template")).unwrap().unchecked_into());
-            static TEMPLATES: RefCell<Vec<(TypeId, HtmlTemplateElement)>> = Default::default();
+            static TEMPLATES: RefCell<FxHashMap<TypeId, HtmlTemplateElement>> = Default::default();
         }
 
         TEMPLATES.with_borrow_mut(|t| {
-            let id = TypeId::of::<V>();
-            t.iter()
-                .find_map(|entry| (entry.0 == id).then(|| entry.1.clone()))
-                .unwrap_or_else(|| {
+            t.entry(TypeId::of::<V>())
+                .or_insert_with(|| {
                     let tpl = TEMPLATE_ELEMENT.with(|t| {
                         t.clone_node()
                             .unwrap()
@@ -534,9 +532,9 @@ impl Dom {
                         &mut Default::default(),
                     );
                     tpl.set_inner_html(&buf);
-                    t.push((id, tpl.clone()));
                     tpl
                 })
+                .clone()
         })
     }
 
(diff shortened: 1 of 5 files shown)


gbj:
There are two primary reasons the DOM template caches use `Vec<_>` instead of `FxHashMap<_>`:
1. WASM binary size effect. It's been a few years since I have tested this — I recall the difference being a few kb, which was worth not including. I'd be interested to see example numbers with/without the change, on an example that uses `template! {}`
2. For very-small `n` (like the number of `template! {}` invocations in a typical application) the `O(n)` of a `Vec<_>` linear search can be faster than the `O(1)` of a hashmap lookup. This one is simply going to be a tradeoff depending on the number f templates, and they are both going to be *very* fast.

The rest seems fine.

sabify:
You are right on the binary size, but I found an interesting hybrid solution that may be worth considering:

Profiling commit 3677cb8 on the js-framework-benchmark example (trunk release, wasm-opt -Oz) confirmed both of concerns, but showed the two caches sit on opposite sides of the trade-off — so the follow-up commit (`afc63f9f`) keeps `FxHashMap` only for the HTML-string cache and reverts the `TypeId` cache to a `Vec`:

- **Binary size:** hashing both caches cost +3,967 B raw / +1,378 B gzipped, growing ~360 B per `template!` call site (the map machinery re-inlines into every `get_template::<V>` monomorphization); the hybrid costs a flat +1,999 B regardless of template count.
- **Small-n speed:** in wasm, the Vec scan wins below n ≈ 4–8 (5.0 vs 6.9 ns/hit at n = 1 for TypeId), and the TypeId cache holds one entry per `template!` call site — typically ~1 — so the map bought no speed there.
- **String cache is the exception:** its keyspace grows with every distinct static `view!` fragment (`InertElement`), and Vec lookups degrade as O(n·html_len) (579 ns at n = 128, ~10 µs on shared-prefix keys) while the map stays flat at ~130 ns, with crossover already at n ≈ 4.

More profiling result info in the commit message.

gbj:
Am I understanding the last comment here correctly that the trade-off for migrating the HTML string cache (for InertElement) from a vector to a hash map is 2kb of WASM in exchange for an improvement (if `n = 128` for 128 distinct InertElements) from a 10-microsecond lookup to a 579ns lookup?

Keeping it as a vector sounds like a good call to me, in that case.

sabify:
A correction: 2KB is the RAW size and the gzipped version is ~500 bytes which will normally be shipped to browsers. It has *flat* size and *one time cost*.

Using hash shines with shared prefix keys in design systems.

**Key shapes (the cache keys are the fragments' HTML strings):**

| Shape             | What it means                                   | Example                                            | Comparison fails…                                       |
| ----------------- | ----------------------------------------------- | -------------------------------------------------- | ------------------------------------------------------- |
| **Differ-early**  | keys start differently (different tags/classes) | `<button>…` vs `<div>…`                            | after ~1 byte — cheap                                   |
| **Shared-prefix** | keys identical for a long run, differ late      | `<div class="card">…**A**</div>` vs `…**B**</div>` | only after re-reading the whole shared part — expensive |

Real component libraries produce shared-prefix keys: every fragment opens with the same wrapper markup and differs only deep inside.

So with many similar keys the `Vec` is `O(entries × shared-prefix)`, while the hash stays flat.

**Measured cost per lookup (ns, wasm, release+LTO):**

| Entries | Vec, differ-early | Vec, shared-prefix | Hash |
| ------: | ----------------: | -----------------: | ---: |
| 1       | 82                | 71                 | ~90  |
| 128     | 338               | 4,900              | ~100 |
| 1,000   | 1,400             | 17,900             | ~70  |

**When to use which:**

| Distinct entries                   | Typical app                  | Better choice                                  |
| ---------------------------------- | ---------------------------- | ---------------------------------------------- |
| 0–3                                | counter, TodoMVC | `Vec` — faster and ~0.5 KB smaller             |
| 8+ differ-early / 4+ shared-prefix | medium+, design systems      | Hash                                           |
| ~1,000 + full re-render            | large styled lists           | Hash (saves ~18 ms/render) |

**Key points.**

- A single lookup is always sub-millisecond, even at 1,000 entries. The cost only shows in aggregate: a full render does one lookup per fragment, and the `Vec`'s per-lookup cost itself grows with size, making a full render `O(n²)`.
- The hash map's added size is flat ~0.5 KB gzipped — negligible at any scale.

gbj:
I'm convinced. Thanks!

PR #4765, opened 2026-06-06
Performance optimization for `server_fn`

sabify:
Targeted, behavior-preserving performance fixes on the server-function request → dispatch → encode path (per the `server_fn` performance audit). Wire format and public output are unchanged.

### Changes

- **Stream the generic request body as one item, not byte-by-byte** — `stream::iter(Bytes)` was yielding one `u8` per poll and re-chunking into 16-byte `Bytes` (≈N polls + N/16 allocs + a full copy for an already-contiguous buffer). Now handed back as a single zero-copy item. *Biggest win on the generic/wasi backend (streaming, multipart).*

- **Drop the per-request `String` alloc in dispatch lookup** — the registry was keyed by `(String, Method)`, forcing an owned `String` allocation on every server-fn request just to probe the map. Nested the map by `Method` so the path is looked up by `&str` (borrowed) — no allocation per request.

- **Write `ServerFnError` `Display` straight to the formatter** — each match arm built an intermediate `format!` `String` that was then re-printed via `write!`. Now writes directly to the sink: one fewer allocation per error render/log.

- **Reclaim the `Bytes` buffer for `String` decode instead of copying** — replaced `String::from_utf8(bytes.to_vec())` (always copies) with `bytes.into()`, which reclaims a uniquely-owned buffer without copying. Normalizes these sites with the actix/generic paths.

- **Encode websocket frames straight into the tagged buffer** — outbound frames were encoded, then a second buffer was allocated and the whole payload copied into it just to prepend the `[tag]` byte. Added `Encodes::encode_into(value, &mut BytesMut)` (default preserves old behavior) and overrode it on json/cbor/msgpack/postcard/rkyv/serde_lite so the payload serializes directly after the tag — one alloc + one full-payload copy removed per frame per direction. Byte-identical wire format.

### DIFF
--- a/server_fn/src/codec/cbor.rs
+++ b/server_fn/src/codec/cbor.rs
@@ -1,6 +1,6 @@
 use super::{Patch, Post, Put};
 use crate::{ContentType, Decodes, Encodes, Format, FormatType};
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use serde::{Serialize, de::DeserializeOwned};
 
 /// Serializes and deserializes CBOR with [`ciborium`].
@@ -24,6 +24,10 @@ where
         let mut buffer: Vec<u8> = Vec::new();
         ciborium::ser::into_writer(value, &mut buffer)?;
         Ok(Bytes::from(buffer))
+    }
+
+    fn encode_into(value: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        ciborium::ser::into_writer(value, buf.writer())
     }
 }
 
--- a/server_fn/src/codec/json.rs
+++ b/server_fn/src/codec/json.rs
@@ -1,6 +1,6 @@
 use super::{Patch, Post, Put};
 use crate::{ContentType, Decodes, Encodes, Format, FormatType};
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use serde::{Serialize, de::DeserializeOwned};
 
 /// Serializes and deserializes JSON with [`serde_json`].
@@ -22,6 +22,10 @@ where
 
     fn encode(output: &T) -> Result<Bytes, Self::Error> {
         serde_json::to_vec(output).map(Bytes::from)
+    }
+
+    fn encode_into(output: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        serde_json::to_writer(buf.writer(), output)
     }
 }
 
--- a/server_fn/src/codec/msgpack.rs
+++ b/server_fn/src/codec/msgpack.rs
@@ -2,7 +2,7 @@ use crate::{
     ContentType, Decodes, Encodes, Format, FormatType,
     codec::{Patch, Post, Put},
 };
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use serde::{Serialize, de::DeserializeOwned};
 
 /// Serializes and deserializes MessagePack with [`rmp_serde`].
@@ -24,6 +24,10 @@ where
 
     fn encode(value: &T) -> Result<Bytes, Self::Error> {
         rmp_serde::to_vec(value).map(Bytes::from)
+    }
+
+    fn encode_into(value: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        rmp_serde::encode::write(&mut buf.writer(), value)
     }
 }
 
--- a/server_fn/src/codec/postcard.rs
+++ b/server_fn/src/codec/postcard.rs
@@ -2,7 +2,7 @@ use crate::{
     ContentType, Decodes, Encodes, Format, FormatType,
     codec::{Patch, Post, Put},
 };
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use serde::{Serialize, de::DeserializeOwned};
 
 /// A codec for Postcard.
@@ -24,6 +24,14 @@ where
 
     fn encode(value: &T) -> Result<Bytes, Self::Error> {
         postcard::to_allocvec(value).map(Bytes::from)
+    }
+
+    fn encode_into(value: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        // `to_extend` appends into the buffer we already own (which holds any
+        // framing written before it); `mem::take` hands it over by move, so the
+        // existing contents are not copied.
+        *buf = postcard::to_extend(value, std::mem::take(buf))?;
+        Ok(())
     }
 }
 
--- a/server_fn/src/codec/rkyv.rs
+++ b/server_fn/src/codec/rkyv.rs
@@ -2,7 +2,7 @@ use crate::{
     ContentType, Decodes, Encodes, Format, FormatType,
     codec::{Patch, Post, Put},
 };
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use rkyv::{
     Archive, Deserialize, Serialize,
     api::high::{HighDeserializer, HighSerializer, HighValidator},
@@ -39,6 +39,14 @@ where
     fn encode(value: &T) -> Result<Bytes, Self::Error> {
         let encoded = rkyv::to_bytes::<rancor::Error>(value)?;
         Ok(Bytes::copy_from_slice(encoded.as_ref()))
+    }
+
+    fn encode_into(value: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        // Copy the `AlignedVec` straight into the caller's buffer, skipping the
+        // intermediate `Bytes` that `encode` allocates and copies into.
+        let encoded = rkyv::to_bytes::<rancor::Error>(value)?;
+        buf.extend_from_slice(encoded.as_ref());
+        Ok(())
     }
 }
 
--- a/server_fn/src/codec/serde_lite.rs
+++ b/server_fn/src/codec/serde_lite.rs
@@ -3,7 +3,7 @@ use crate::{
     codec::{Patch, Post, Put},
     error::ServerFnErrorErr,
 };
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use serde_lite::{Deserialize, Serialize};
 
 /// Pass arguments and receive responses as JSON in the body of a `POST` request.
@@ -31,6 +31,14 @@ where
         )
         .map_err(|e| ServerFnErrorErr::Serialization(e.to_string()))
         .map(Bytes::from)
+    }
+
+    fn encode_into(value: &T, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        let intermediate = value
+            .serialize()
+            .map_err(|e| ServerFnErrorErr::Serialization(e.to_string()))?;
+        serde_json::to_writer(buf.writer(), &intermediate)
+            .map_err(|e| ServerFnErrorErr::Serialization(e.to_string()))
     }
 }
 
--- a/server_fn/src/error.rs
+++ b/server_fn/src/error.rs
@@ -222,33 +222,35 @@ where
     CustErr: Display,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            match self {
-                ServerFnError::Registration(s) => format!(
-                    "error while trying to register the server function: {s}"
-                ),
-                ServerFnError::Request(s) => format!(
-                    "error reaching server to call server function: {s}"
-                ),
-                ServerFnError::ServerError(s) =>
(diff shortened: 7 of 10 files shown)


lukexor:
I'm always interested in seeing performance improvements but I'm curious if there are any tests or comparisons that show they are improvements?

Specifically in this PR I'm not sure that changing from bytes.to_vec() to bytes.into() actually saves an allocation. Looking at the source, it seems to go through a vtable method and both that I could find do a slice::from_raw_parts of the bytes ptr and then call to_vec. Maybe the compiler is able to reclaim the allocation here but without some test it's hard to guess.

sabify:
No, the compiler cannot optimize out what is structurally different.

**The difference:**

- `bytes.to_vec()` only sees a borrowed slice `&[u8]`. It can't know who owns the buffer, so it **always allocates a new buffer and copies** — then the original `Bytes` is freed on drop.

- `bytes.into()` (`From<Bytes> for Vec<u8>`) owns the `Bytes`, so it can check uniqueness:
  - **uniquely owned** (the common case for a freshly collected request body) → **steals the existing buffer in place: 0 allocations, 0 copy.**
  - **shared** (a clone is alive) → falls back to copy, same as `to_vec()`.

**Net effect:** one fewer allocation + one fewer O(body) memcpy per text/form request and per error decode, with **no regression** on the shared path.

**Why it survives release builds:** it's a code-path difference, not an optimizer trick. One path calls the allocator, the other doesn't — and `-O3` can't remove the allocation because the new buffer escapes into the returned `String`.

**Proven (release):** same pointer in→out (`reclaimed=true`), 0 vs 1 conversion allocs, and 1.35×–1.67× faster from 64 B to 64 KiB.

The profiling code:

<details>

```toml
[package]
name = "bench_40cd37cc"
version = "0.0.0"
edition = "2021"
publish = false

# Match the leptos workspace exactly: bytes 1.11, default-features = false
# (the `From<Bytes> for Vec<u8>` impl is core and identical regardless).
[dependencies]
bytes = { version = "1.11", default-features = false }

[profile.release]
opt-level = 3
lto = true
codegen-units = 1
debug-assertions = false
```

```rust
//! Temporary benchmark/profiler for leptos commit 40cd37cc
//! "perf(server_fn): reclaim Bytes buffer for String decode instead of copying"
//!
//! The commit changed two decode sites from `String::from_utf8(bytes.to_vec())`
//! to `String::from_utf8(bytes.into())`:
//!   - server_fn/src/request/axum.rs : try_into_string (text-codec request path)
//!   - server_fn/src/error.rs        : ServerFnErrorEncoding::decode (error path)
//!
//! Claim under test:
//!   (A) `bytes.to_vec()` ALWAYS heap-allocates + copies the payload.
//!   (B) `bytes.into()` RECLAIMS the existing buffer (0 alloc, 0 copy) when the
//!       `Bytes` is uniquely owned -- the common case for a freshly collected
//!       request body -- and falls back to a copy only when shared.
//!   (C) On shared `Bytes` both paths copy -> no regression.
//!
//! This is built `--release` to answer "is it true even in release mode?":
//! the gap is one allocator call vs zero, which is a property of the two code
//! paths, not a debug artifact.

use bytes::Bytes;
use std::alloc::{GlobalAlloc, Layout, System};
use std::hint::black_box;
use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
use std::time::{Duration, Instant};

// ---------------------------------------------------------------------------
// Counting global allocator: every real heap alloc bumps a counter, so we can
// prove how many allocations each conversion performs, in release mode.
// ---------------------------------------------------------------------------
struct Counting;
static ALLOCS: AtomicUsize = AtomicUsize::new(0);
static ALLOC_BYTES: AtomicUsize = AtomicUsize::new(0);

unsafe impl GlobalAlloc for Counting {
    unsafe fn alloc(&self, l: Layout) -> *mut u8 {
        let p = System.alloc(l);
        if !p.is_null() {
            ALLOCS.fetch_add(1, Relaxed);
            ALLOC_BYTES.fetch_add(l.size(), Relaxed);
        }
        p
    }
    unsafe fn alloc_zeroed(&self, l: Layout) -> *mut u8 {
        let p = System.alloc_zeroed(l);
        if !p.is_null() {
            ALLOCS.fetch_add(1, Relaxed);
            ALLOC_BYTES.fetch_add(l.size(), Relaxed);
        }
        p
    }
    unsafe fn dealloc(&self, p: *mut u8, l: Layout) {
        System.dealloc(p, l);
    }
    unsafe fn realloc(&self, p: *mut u8, l: Layout, n: usize) -> *mut u8 {
        let np = System.realloc(p, l, n);
        if !np.is_null() && n > l.size() {
            ALLOCS.fetch_add(1, Relaxed);
            ALLOC_BYTES.fetch_add(n - l.size(), Relaxed);
        }
        np
    }
}

#[global_allocator]
static GLOBAL: Counting = Counting;

fn allocs() -> usize {
    ALLOCS.load(Relaxed)
}

// ---------------------------------------------------------------------------
// The two conversions, verbatim from the commit (before / after).
// `#[inline(never)]` keeps them as distinct, observable call boundaries so the
// optimizer cannot conflate them; the result escapes via the returned String.
// ---------------------------------------------------------------------------
#[inline(never)]
fn via_to_vec(b: Bytes) -> String {
    // BEFORE 40cd37cc -- always allocates a new Vec and copies.
    String::from_utf8(b.to_vec()).unwrap()
}

#[inline(never)]
fn via_into(b: Bytes) -> String {
    // AFTER 40cd37cc -- reclaims the buffer when uniquely owned.
    String::from_utf8(b.into()).unwrap()
}

// A freshly collected single-chunk body: uniquely-owned, KIND_VEC `Bytes`.
// This mirrors `body.collect().await.to_bytes()` on the common path.
fn fresh_body(len: usize) -> Bytes {
    Bytes::from(vec![b'a'; len])
}

// ---------------------------------------------------------------------------
// (1) PROOF BY POINTER IDENTITY -- the strongest evidence that no copy occurred.
//     If the resulting Vec points at the SAME address as the source Bytes, the
//     buffer was reclaimed in place; a different address means a fresh alloc+copy.
// ---------------------------------------------------------------------------
fn proof_pointer_identity(len: usize) {
    println!("== (1) Pointer-identity proof (len = {len} bytes) ==");

    // unique Bytes -> into(): expect SAME pointer (reclaimed, zero copy)
    let b = fresh_body(len);
    let src = b.as_ptr() as usize;
    let v: Vec<u8> = b.into();
    let dst = v.as_ptr() as usize;
    println!(
        "  into()   on unique Bytes: src={src:#x} dst={dst:#x} -> reclaimed={}",
        src == dst
    );
    black_box(&v);

    // unique Bytes -> to_vec(): expect DIFFERENT pointer (fresh alloc + copy)
    let b = fresh_body(len);
    let src = b.as_ptr() as usize;
    let v = b.to_vec();
    let dst = v.as_ptr() as usize;
    println!(
        "  to_vec() on unique Bytes: src={src:#x} dst={dst:#x} -> reclaimed={}",
        src == dst
    );
    black_box(&v);

    // shared Bytes -> into(): expect DIFFERENT pointer (must copy, can't steal)
    let b = fresh_body(len);
    let keep = b.clone();
    let src = b.as_ptr() as usize;
    let v: Vec<u8> = b.into();
    let dst = v.as_ptr() as usize;
    println!(
        "  into()   on shared Bytes: src={src:#x} dst={dst:#x} -> reclaimed={} (expected false)",
        src == dst
    );
    black_box((&v, &keep));
    println!();
}

// ---------------------------------------------------------------------------
// (2) PROOF BY ALLOCATION COUNT -- measured under the counting allocator.
//     We measure construct-only as a baseline, then construct+convert, and
//     report the per-call allocations attributable to the conversion itself.
// ---------------------------------------------------------------------------
fn per_call_allocs<F: FnMut()>(iters: usize, mut f: F) -> f64 {
    let start = allocs();
    for _ in 0..iters {
        f();
    }
    (allocs() - start) as f64 / iters as f64
}

fn proof_alloc_count(len: usize, iters: usize) {
    println!("== (2) Allocation-count proof (len = {len} bytes, {iters} iters) ==");

    // Baseline: just build + drop the body (1 alloc/iter for the buffer).
    let base_unique = per_call_allocs(iters, || {
        let b = fresh_body(len);
        black_box(&b);
    });

    // into() on unique body: build(1) + reclaim(0) == baseline.
    let mut sink = 0usize;
    let into_unique = per_call_allocs(iters, || {
        let b = fresh_body(len);
        let s = via_into(black_box(b));
        sink ^= s.len();
        black_box(&s);
    });

    // to_vec() on unique body: build(1) + fresh copy(1).
    let tovec_unique = per_call_allocs(iters, || {
        let b = fresh_body(len);
        let s = via_to_vec(black_box(b));
        sink ^= s.len();
        black_box(&s);
    });
    black_box(sink);

    println!("  unique Bytes (common request-body path):");
    println!("    build-only baseline : {base_unique:.3} allocs/call");
    println!(
        "    via_into()  total   : {into_unique:.3} allocs/call  -> conversion = {:.3}",
        into_unique - base_unique
    );
    println!(
        "    via_to_vec() total  : {tovec_unique:.3} allocs/call  -> conversion = {:.3}",
        tovec_unique - base_unique
    );
    println!(
        "    => into() saves {:.3} allocation(s) per call",
        tovec_unique - into_unique
    );

    // ---- shared case: hold a clone alive so neither path can reclaim ----
    // build(1) + clone-promotion(1) baseline; +copy(1) for each conversion.
    let base_shared = per_call_allocs(iters, || {
        let b = fresh_body(len);
        let keep = b.clone();
        black_box((&b, &keep));
    });
    let into_shared = per_call_allocs(iters, || {
        let b = fresh_body(len);
        let keep = b.clone();
        let s = via_into(black_box(b));
        sink ^= s.len();
        black_box((&s, &keep));
    });
    let tovec_shared = per_call_allocs(iters, || {
        let b = fresh_body(len);
        let keep = b.clone();
        let s = via_to_vec(black_box(b));
        sink ^= s.len();
        black_box((&s, &keep));
    });
    black_box(sink);

    println!("  shared Bytes (clone held -> must copy):");
    println!(
        "    via_into()  conversion : {:.3} allocs/call",
        into_shared - base_shared
    );
    println!(
        "    via_to_vec() conversion: {:.3} allocs/call",
        tovec_shared - base_shared
    );
    println!("    => equal: no regression on the shared path");
    println!();
}

// ---------------------------------------------------------------------------
// (3) THROUGHPUT -- wall-clock cost of construct+convert across payload sizes.
//     into() avoids the alloc+memcpy on the unique path, so it wins and the gap
//     grows with payload size.
// ---------------------------------------------------------------------------
fn time_convert(len: usize, iters: usize, into: bool) -> Duration {
    // warmup
    for _ in 0..iters / 10 {
        let b = fresh_body(len);
        let s = if into { via_into(b) } else { via_to_vec(b) };
        black_box(&s);
    }
    let t0 = Instant::now();
    let mut sink = 0usize;
    for _ in 0..iters {
        let b = fresh_body(len);
        let s = if into {
            via_into(black_box(b))
        } else {
            via_to_vec(black_box(b))
        };
        sink ^= s.len();
        black_box(&s);
    }
    black_box(sink);
    t0.elapsed()
}

fn proof_throughput() {
    println!("== (3) Throughput (construct + convert, unique Bytes) ==");
    println!(
        "  {:>10}  {:>14}  {:>14}  {:>9}",
        "len", "to_vec ns/call", "into ns/call", "speedup"
    );
    for &len in &[64usize, 256, 1024, 8 * 1024, 64 * 1024] {
        // More iters for small payloads to keep the signal stable.
        let iters = (64 * 1024 * 1024 / len).clamp(50_000, 5_000_000);
        // interleave + best-of-3 to reduce noise
        let mut to_vec_ns = f64::MAX;
        let mut into_ns = f64::MAX;
        for _ in 0..3 {
            let tv = time_convert(len, iters, false).as_nanos() as f64 / iters as f64;
            let it = time_convert(len, iters, true).as_nanos() as f64 / iters as f64;
            to_vec_ns = to_vec_ns.min(tv);
            into_ns = into_ns.min(it);
        }
        println!(
            "  {:>10}  {:>14.2}  {:>14.2}  {:>8.2}x",
            len,
            to_vec_ns,
            into_ns,
            to_vec_ns / into_ns
        );
    }
    println!();
}

fn main() {
    let mode = if cfg!(debug_assertions) {
        "DEBUG"
    } else {
        "RELEASE"
    };
    println!("bench for commit 40cd37cc  (build = {mode})\n");

    proof_pointer_identity(4096);
    proof_alloc_count(4096, 200_000);
    proof_throughput();

    println!("Summary:");
    println!("  - into() reclaims the buffer in place on the unique path (same pointer, 0 allocs).");
    println!("  - to_vec() always allocates a fresh buffer + copies (different pointer, 1 alloc).");
    println!("  - shared path: both copy -> no regression.");
    println!("  - all of the above holds in RELEASE: it is a code-path difference, not -O level.");
}
```

</details>

Issue #4556, opened 2026-02-02
Inert elements inside `<svg>` will be wrapped in a `<g>` (0.8.3 breaking change)


pheki:
**Describe the bug**

Starting from 0.8.3, the first level of `<svg>` elements get wrapped in a `<g>`. For example `<text x="1" y="50"><tspan>Just an example</tspan></text>`, if a direct child of `<svg>`, gets rendered as `<g><text x="1" y="50"><tspan>Just an example</tspan></text></g>`.

**Leptos Dependencies**

```toml
leptos = { version = "0.8.15", features = ["csr"] }
```

**To Reproduce**

1. Create an example app following the [Getting Started](https://book.leptos.dev/getting_started/index.html) tutorial from the book
2. Replace `src/main.rs` with the following:

```rust
use leptos::prelude::*;

fn main() {
    leptos::mount::mount_to_body(|| view! { <MyOwnSvg /> })
}

#[component]
fn MyOwnSvg() -> impl IntoView {
    view! {
        <svg
            width="200"
            viewBox="0 0 200 100"
            id="lamp"
            xmlns="http://www.w3.org/2000/svg"
            xmlns:xlink="http://www.w3.org/1999/xlink"
        >
            // Inert SVG elements are wrapped in a `<g>` when rendered
            <rect x="12" y="10" width="140" height="32" stroke="black" fill="transparent" />
            <text x="18" y="32"><tspan>Just an example</tspan></text>
            // This is not inert, so it's not wrapped in a `<g>`
            {
                move || view! {
                    <rect x="12" y="110" width="140" height="32" stroke="black" fill="transparent" />
                }
            }
            // This is not inert, so it's not wrapped in a `<g>`
            <rect x=12 y="210" width="140" height="32" stroke="black" fill="transparent" />
        </svg>
    }
}
```

4. Open http://localhost.:8080/ in your browser (tested on both Firefox and Chromium)
5. Check in the HTML inspector that both `<rect>` and `<text>` got wrapped in two different `<g>` elements. The same did not happen for `<tspan>` (nested inside) nor the other `<rect>`s (not inert).

**Next Steps**

- [X] I will make a PR
- [ ] I would like to make a PR, but need help getting started
- [ ] I want someone else to take the time to fix this
- [X] This is a low priority for me and is just shared for your information

**Additional context**

I discovered this while investigating https://github.com/leptos-rs/leptos/issues/4555.

[...]

PR #4767, opened 2026-06-08
fix(tachys): stop wrapping inert SVG children in a spurious <g>

sabify:
Closes: #4556.

Static SVG children of `<svg>` take the inert HTML fast path and are rebuilt on the client by `Dom::create_svg_element_from_html`. That function parsed the fragment inside a temporary `<g>` (to confer the SVG namespace) but returned the `<g>` itself, so every inert SVG child mounted wrapped in an extra `<g>`:

```
mounted:  <g><text x="18" y="32"><tspan>Just an example</tspan></text></g>
expected: <text x="18" y="32"><tspan>Just an example</tspan></text>
```

Dynamic children were unaffected (they skip the inert path), and SSR was already correct (it emits raw markup with no scaffold).

**Fix:** Drop the inner `<g>` and set `inner_html` directly on the SVG-namespaced `<svg>`. The fragment parser keys the namespace off the context element's *namespace*, not its tag, so an `<svg>` context namespaces the children exactly as the `<g>` did. Return `svg.first_element_child()` (with a safe fallback), which is now the authored element itself, mirroring the non-SVG `create_element_from_html`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3946,6 +3946,7 @@ dependencies = [
  "tokio-test",
  "tracing",
  "wasm-bindgen",
+ "wasm-bindgen-test",
  "web-sys",
 ]
 
--- a/tachys/Cargo.toml
+++ b/tachys/Cargo.toml
@@ -169,6 +169,7 @@ tokio = { features = [
   "rt",
   "macros",
 ], workspace = true, default-features = true }
+wasm-bindgen-test = { workspace = true, default-features = true }
 
 [build-dependencies]
 rustc_version = { workspace = true, default-features = true }
--- a/tachys/Makefile.toml
+++ b/tachys/Makefile.toml
@@ -1,1 +1,4 @@
-extend = { path = "../cargo-make/main.toml" }
+extend = [
+  { path = "../cargo-make/main.toml" },
+  { path = "../cargo-make/wasm-test.toml" },
+]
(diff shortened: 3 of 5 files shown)


pheki:
This fixes the issue, but I'm not sure why add the `<g>` to remove it later, as adding it directly to the `svg` is simpler and also seems to work, like the following patch to main:

```diff
diff --git a/tachys/src/renderer/dom.rs b/tachys/src/renderer/dom.rs
index fb7f5a8c4a..7406941f85 100644
--- a/tachys/src/renderer/dom.rs
+++ b/tachys/src/renderer/dom.rs
@@ -586,14 +586,15 @@
                         Self::intern("svg"),
                     )
                     .unwrap();
-                let g = document()
-                    .create_element_ns(
-                        Some(Self::intern("http://www.w3.org/2000/svg")),
-                        Self::intern("g"),
-                    )
-                    .unwrap();
-                g.set_inner_html(&html);
-                svg.append_child(&g).unwrap();
+                svg.set_inner_html(&html);
                 tpl.unchecked_ref::<TemplateElement>()
                     .content()
                     .append_child(&svg)
```

The problem is that we don't know why was that `<g>` added and which case it's intended to fix.

sabify:
You are right. The excessive `<g>` namespacing is redundant. I just changed it with added tests.

PR #4774, opened 2026-06-10
feat: Add the `#[prop(marker)]` option

Baptistemontan:
Add the `#[prop(marker)]` option, this is intended for marker props such as `PhantomData`, it is incompatible with any other prop option:

- skip displaying the prop in the component doc
- set `#[doc(hidden)]` on the builder
- skip the setter method (`#[builder(default, setter(skip))]`)
- set `#[serde(skip)]` 

## Motivations

This code will compile fine:

```rust
trait Foo {
    fn bar() -> impl IntoView;
}

fn Test<T: Foo>() -> impl IntoView {
    T::bar()
}
```

but adding `#[component]` to `Test` will cause a compilation error, because now the underlying builder as an unused type parameter:

```
error[E0392]: type parameter `T` is never used                                                                                                                                                                                                                                 
  --> src\app.rs:89:9
   |
89 | fn Test<T: Foo>() -> impl IntoView {
   |         ^ unused type parameter
   |
   = help: consider removing `T`, referring to it in a field, or using a marker such as `PhantomData`

  --> src\app.rs:89:4
   |
89 | fn Test<T: Foo>() -> impl IntoView {
   |    ^^^^ cannot infer type of the type parameter `T` declared on the function `__component_test`
   |
help: consider specifying the generic argument
   |
89 | fn Test::<_><T: Foo>() -> impl IntoView {
   |        +++++
```


The solution is to have an optional prop:

```rust
#[component]
fn Test<T: Foo>(#[prop(optional)] _marker: PhantomData<T>) -> impl IntoView {
    T::bar()
}
```

Works great, but now `_marker` appears in docs, is serialized in islands, has a setter on the builder, or appears in the builder doc.

This PR add a way to remove those by simply marking the prop with `#[prop(marker)]`.

## Alternatives

One other way that could be done is to have an attribute on the type parameter itself:

```rust
#[component]
fn Test<#[prop(marker)] T: Foo>() -> impl IntoView {
    T::bar()
}
```
> placeholder name for now, any suggestion is welcome

This would aggregate all the marked generics on the builder with a single `__marker: PhantomData<(...T)>` and skip the doc/setter/serde. Or also might allow to skip the generic from the builder entirely and thus allow component with no props to stay without props, and thus stay a `fn() -> impl IntoView`.
This would, in my opinion, be the better option, but would not be as straightforward as the currently proposed solution, but I would be happy to implement it if decided better.
(I actually just thought about it while writing this PR)

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -1000,20 +1000,23 @@ impl ToTokens for UnknownAttrs {
 #[derive(Clone, Debug, FromAttr)]
 #[attribute(ident = prop)]
 struct PropOpt {
-    #[attribute(conflicts = [optional_no_strip, strip_option])]
+    #[attribute(conflicts = [optional_no_strip, strip_option, marker])]
     optional: bool,
-    #[attribute(conflicts = [optional, strip_option])]
+    #[attribute(conflicts = [optional, strip_option, marker])]
     optional_no_strip: bool,
-    #[attribute(conflicts = [optional, optional_no_strip])]
+    #[attribute(conflicts = [optional, optional_no_strip, marker])]
     strip_option: bool,
     #[attribute(example = "5 * 10")]
     default: Option<syn::Expr>,
     into: bool,
     attrs: bool,
     name: Option<String>,
+    #[attribute(conflicts = [optional, optional_no_strip, strip_option, default, into, attrs, name])]
+    marker: bool,
 }
 
 struct TypedBuilderOpts<'a> {
+    marker: bool,
     default: bool,
     default_with_value: Option<syn::Expr>,
     strip_option: bool,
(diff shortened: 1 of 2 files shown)


Baptistemontan:
Currently no doc written, I'm waiting for a feedback before writing it

gbj:
This seems great to me! A really handy little addition. If you want to add the docs for it, perfect.

Baptistemontan:
@gbj Glad to hear the addition is welcome! 
Did you take a look at the alternative I proposed ? Or like this is fine ?

gbj:
I would recommend the current implementation rather than the alternative. `#[prop(marker)]` fits into existing infrastructure and an existing mental model where we annotate props — `<#[prop(marker)] T: Foo>()` would be convenient in one sense but is more complex/unfamiliar, conjures a new prop, etc.

Baptistemontan:
@gbj perfect, I'll write the doc then !

Baptistemontan:
@gbj Docs done, also sent a PR to the book: https://github.com/leptos-rs/book/pull/267

gbj:
Thank you!

PR #4775, opened 2026-06-10
feat(router): allow permanent redirects (301)

Baptistemontan:
Add a `permanent` flag to the integrations redirect functions and  the `permanent` prop to the `<Redirect />` component to allow a permanent redirect (301).

`permanent` default to `false` for the `<Redirect />` component


gbj:
Works for me! Thanks!

PR #4776, opened 2026-06-11
perf(hydration_context): optimize streaming and async SSR resource handling

sabify:
Three independent performance fixes in `hydration_context/src/ssr.rs`, all on the SSR serialization path. Output is byte-identical in every case; each change ships with a regression test.

- **Drive streaming SSR with `FuturesUnordered`** — `AsyncDataStream::poll_next` previously drained and re-polled *every* unresolved resource future on each wake, making total polls O(N²) as resources resolve at distinct times. Routing each future's waker individually via `FuturesUnordered` makes polling O(N) (measured ~65× fewer leaf polls at N=256) and removes a per-poll `Vec` allocation and a second write-lock acquisition. The `write_async` re-entrancy contract (nested Suspense) is preserved and covered by existing tests.

- **Resolve `consume_buffers` resources concurrently** — the non-streaming/custom-hydration path awaited each async resource sequentially, so latency was the *sum* of resource latencies instead of the *max*. Switching to `join_all` polls them concurrently while preserving registration order.

- **Guard the `<`-escape behind a byte scan** — every serialized payload was rewritten with `str::replace('<', "\u003c")`, which allocates and copies even when no `<` is present (the common case). A new `escape_lt` helper returns `Cow::Borrowed` for clean payloads (~1.5–2.6× faster, one fewer allocation per resource) and de-duplicates the three escape sites. XSS escape behavior is unchanged.

### DIFF
--- a/hydration_context/src/ssr.rs
+++ b/hydration_context/src/ssr.rs
@@ -7,6 +7,7 @@ use futures::{
 };
 use or_poisoned::OrPoisoned;
 use std::{
+    borrow::Cow,
     collections::HashSet,
     fmt::{Debug, Write},
     mem,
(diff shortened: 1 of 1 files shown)


gbj:
It's not clear to me that there's a real benefit to d0109aeab4b1939a6f417883de1be62a40604208 — or at least, its comment should be corrected to avoid confusing future readers.

I might be mistaken, but the `Resource` type spawns a Tokio task per `Resource` created, so the actual async work is *already* all happening concurrently. All we're awaiting here is the "am I ready?" `Future`. So there is ~no wall-clock difference between these two approaches.

sabify:
You are right. It should be considered a refactor rather than a performance improvement for the workspace. The only performance benefit will be for custom implementations of the public-facing API `write_async` if not implemented for concurrency (which would be rare).

I will classify it as a refactor with updated docs and a commit message. It's up to you if you want me to revert it.

gbj:
Thanks!

PR #4778, opened 2026-06-11
perf(router): optimize hot-path route matching

sabify:
Optimizes the route matchers in `leptos_router`, which run on every SSR request and client-side navigation.

### Changes

- **Scan bytes instead of decoding chars in segment matchers** — `StaticSegment`, `ParamSegment`, `WildcardSegment`, and `OptionalParamSegment::test` previously compared one `char` at a time, paying for UTF-8 decoding on every step. Since all comparisons are exact-equality and matching only stops at the ASCII `/` or end of input, byte scanning is equivalent. Control flow is unchanged, and behavior was verified identical against the previous implementations over an adversarial corpus (multi-byte UTF-8, trailing/repeated slashes, smooshed segments, empty/root paths).
- **Drop a throwaway `String` in the horizontal tuple matcher** — the matched text of the first segment was copied into a heap `String` whose only use was a single `.len()` call; now the length is added directly, removing one allocation per matched-prefix candidate on every `match_route` call.

### Results (criterion, Apple M2 Pro)

| benchmark                          | before   | after    | change |
| ---------------------------------- | -------- | -------- | ------ |
| `static_segment::test` (hit)       | 42.2 ns  | 15.9 ns  | −62%   |
| `static_segment::test` (miss late) | 33.8 ns  | 10.3 ns  | −69%   |
| `param_segment::test`              | 96.6 ns  | 65.1 ns  | −32%   |
| `wildcard_segment::test`           | 116.0 ns | 64.9 ns  | −44%   |
| `match_route` (hit, last sibling)  | 359.5 ns | 249.1 ns | −31%   |
| `match_route` (miss)               | 142.8 ns | 65.3 ns  | −54%   |

Allocations per matched route drop by one (e.g. 5 → 4 on `/blog/post/42`).

No public API or behavior changes. Adds unit tests covering multi-byte UTF-8 segments and params.

### DIFF
--- a/router/src/matching/horizontal/param_segments.rs
+++ b/router/src/matching/horizontal/param_segments.rs
@@ -40,29 +40,24 @@ impl PossibleRouteMatch for ParamSegment {
     }
 
     fn test<'a>(&self, path: &'a str) -> Option<PartialPathMatch<'a>> {
-        let mut matched_len = 0;
-        let mut param_offset = 0;
-        let mut param_len = 0;
-        let mut test = path.chars();
+        let bytes = path.as_bytes();
+        let has_leading_slash = bytes.first() == Some(&b'/');
+        // the first char of the path is always consumed before scanning, but
+        // only counts toward the match when it is the leading `/`
+        let (param_offset, scan_start) = if has_leading_slash {
+            (1, 1)
+        } else {
+            (0, path.chars().next().map(char::len_utf8).unwrap_or(0))
+        };
+        // the param ends at the next `/`; `/` is ASCII, so scanning bytes is
+        // equivalent to scanning chars without the UTF-8 decoding
+        let param_len = bytes[scan_start..]
+            .iter()
+            .position(|&byte| byte == b'/')
+            .unwrap_or(bytes.len() - scan_start);
+        let matched_len = param_offset + param_len;
 
-        // match an initial /
-        if let Some('/') = test.next() {
-            matched_len += 1;
-            param_offset = 1;
-        }
-        for char in test {
-            // when we get a closing /, stop matching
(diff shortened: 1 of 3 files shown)


gbj:
How much of the performance effect is the `String` allocation vs the byte-scanning-vs-char-decoding? i.e., could you run that same benchmark separately against the two commits?

sabify:
The provided benchmark table is only for the first commit [f5db4d6](https://github.com/leptos-rs/leptos/pull/4778/commits/f5db4d6a97bf2357603390acbf5c542e847b57cc) (`byte-scanning-vs-char-decoding`), excluding the second one. You can also see the bench table in commit message.

```
Results (before -> after):

    static_segment::test (hit)        42.2 ns -> 15.9 ns  (-62%)
    static_segment::test (miss late)  33.8 ns -> 10.3 ns  (-69%)
    param_segment::test               96.6 ns -> 65.1 ns  (-32%)
    wildcard_segment::test           116.0 ns -> 64.9 ns  (-44%)
    match_route /blog/post/42        359.5 ns -> 249.1 ns (-31%)
    match_route miss                 142.8 ns -> 65.3 ns  (-54%)
```

The second commit [92e0b31](https://github.com/leptos-rs/leptos/pull/4778/commits/92e0b31873d2b50c66cb01c32cc268e4f0a70e68) (`String` allocation) has its own benchmark in its commit message:

```
Results (before -> after):
    match_route /blog/post/42   249.1 ns -> 223.6 ns (-10%)
    match_route miss             65.3 ns ->  65.4 ns (no change)
```

Do I miss something in your question?

gbj:
Thanks! Looks good to me.

Issue #4772, opened 2026-06-10
leptos_axum: concurrent regeneration of a `SsrMode::Static` route can serve one render's body with another render's headers


AlexeyMatskevich:
**Describe the bug**

When several requests hit a not-yet-generated `SsrMode::Static` route concurrently, `leptos_axum` can serve a response whose **HTML body comes from one render but whose `ResponseOptions` headers/status come from a different render**.

Every concurrent miss runs its own SSR render and then:

1. writes its HTML to disk and inserts its captured `ResponseOptions` into the process-global `STATIC_HEADERS` map ([`integrations/axum/src/lib.rs#L1707`](https://github.com/leptos-rs/leptos/blob/b4458855f1f23e4b722719b51a7a27b43748fcec/integrations/axum/src/lib.rs#L1707) on `leptos_0.9`),
2. applies the **request-local** `ResponseOptions` captured by *its own* render ([`#L1793`](https://github.com/leptos-rs/leptos/blob/b4458855f1f23e4b722719b51a7a27b43748fcec/integrations/axum/src/lib.rs#L1793)),
3. but serves the body by **re-opening the file from disk** via `ServeFile` ([`#L1822`](https://github.com/leptos-rs/leptos/blob/b4458855f1f23e4b722719b51a7a27b43748fcec/integrations/axum/src/lib.rs#L1822)).

Nothing pairs steps 2 and 3: between request A's disk write and its `ServeFile` open, a concurrent request B can overwrite the file. A then serves **body B under headers A**. The cache-hit path has the same shape (headers read from `STATIC_HEADERS` at [`#L1796`](https://github.com/leptos-rs/leptos/blob/b4458855f1f23e4b722719b51a7a27b43748fcec/integrations/axum/src/lib.rs#L1796), body via a separate `ServeFile` open, no common lock).

This matters whenever a static route's headers are derived from the rendered content — e.g. a `Link`/preload header naming a hashed asset, a CSP nonce, a custom status — the served body then contradicts its own headers. The hardening in #4739 (merged into `leptos_0.9`) does not address this: it bounds the cache with an LRU but the regeneration branch still pairs the request-local snapshot with a separately re-opened file.

`leptos_actix` has the same code shape (`NamedFile::open` + request-local `ResponseOptions` on the regeneration path), so it is most likely affected as well; I only built a runnable reproduction for axum.

**Leptos Dependencies**

```toml
leptos = { version = "0.8.19", features = ["ssr"] }
[...]

PR #4782, opened 2026-06-12
Performance optimizations for the server integrations

sabify:
This PR removes redundant syscalls, allocations, and copies from the request hot paths of `leptos_integration_utils`, `leptos_actix`, and `leptos_axum`, one commit per change with measurements in each commit message.

### Shared (`leptos_integration_utils`) — affects both integrations
- **Streamed chunks are no longer re-copied:** `ready_chunks(32).map(|n| n.join("")))` allocated and memcpy'd every batch even when it held a single chunk — the normal case while streaming. Singleton batches are now moved out directly (saves 1 alloc + a full byte copy per streamed chunk, ~70 ns–1.6 µs depending on chunk size).
- **`is_path_traversal` no longer allocates:** the lowercased copy + four scans per static request is now a single zero-alloc windowed byte scan (2–10× faster, semantics pinned by existing tests).
- **`Accept` header parsing gets a cheap negative gate:** headers that can't possibly accept HTML (e.g. `application/json` from the server-fn client) skip the per-range mime parse (72.6 → 15.8 ns); positives still go through the full parser, so `q=0` and look-alike handling are unchanged.

### `leptos_actix`
- **Static routes:** replaced the `try_exists` pre-check + synchronous `NamedFile::open` with `NamedFile::open_async` keyed off `NotFound` — drops one `stat` per cached request (~1.2 µs warm), closes the check-then-open TOCTOU race, and keeps file opens off the reactor thread.
- **`request_url`:** assembled once in a pre-sized buffer instead of through two intermediate `format!` strings (224 → 56 ns, 2 allocs saved per SSR render).
- Removed a dead re-clone of the user's context closure on every server-fn call.
- Redirect marker header now uses the const `HeaderValue::from_static("")`.

### `leptos_axum`
- **Static routes:** serve-first via `ServeFile`, generating on 404 — same stat removal and TOCTOU fix as the actix twin.
- **Request URL:** origin helper now returns borrowed `(scheme, host)` and the URL is assembled in one pre-sized buffer (175 → 52 ns, 2 allocs saved per SSR render).
- **Server-fn handler:** the path stays borrowed for the service lookup; the owned `String` is only built on the cold not-found branch. Also switches the redirect marker to `HeaderValue::from_static("")`.

### Test fixes
Two pre-existing failures on `leptos_0.9` from #4775 (`redirect` gaining a `permanent` parameter):
- actix: two test call sites still passed one argument and broke the test build.
- axum: `redirect_sets_301/302` sent no `Accept` header, so they exercised the marker-header branch instead of the status branch they assert on.

No behavioral changes otherwise; all existing tests pass and the traversal/Accept-header changes are pinned by the existing unit tests.

Edit:

Closes: #4772.

Added a fix on top of the changes:

Fixes a race in on-demand static-route regeneration: on a cache miss the axum and actix integrations rendered the page, wrote it to disk, then re-read the file to build the response body while taking the headers from the in-memory render — so under concurrent requests for the same path (last writer wins on disk) a response could pair one render's headers with another render's body. `ResolvedStaticPath::build` now returns the rendered HTML to the caller via a new `StaticResponse` enum instead of writing-and-discarding it, and both integrations serve that in-memory HTML together with the headers captured during the same render (the file is still written for later cached hits), guaranteeing body and headers always come from one render. Thanks to @AlexeyMatskevich for detailed bug report and test case.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -376,10 +376,10 @@ pub fn handle_server_fns_with_context(
     additional_context: impl Fn() + 'static + Clone + Send,
 ) -> Route {
     web::to(move |req: HttpRequest, payload: Payload| {
+        // the handler is `Fn`, so it clones the context closure for the
+        // `async move` block below, which owns that clone from then on
         let additional_context = additional_context.clone();
         async move {
-            let additional_context = additional_context.clone();
-
             let path = req.path();
             let method = req.method();
             if let Some(mut service) =
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1441,13 +1441,6 @@ where
                     return HttpResponse::NotFound().finish();
                 };
                 let path = Path::new(&file_path);
-                // Open the file directly instead of probing with `try_exists`
-                // first: a `NotFound` error is the signal that the route still
-                // needs to be generated. This saves a `stat` on every cached
-                // request, closes the TOCTOU window between check and open,
-                // and `open_async` keeps the open/stat syscalls off the
-                // reactor thread (the synchronous `NamedFile::open` would
-                // stall the worker on slow storage).
                 let opened = NamedFile::open_async(path).await;
 
                 match opened {
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -834,14 +834,28 @@ fn request_url(req: &HttpRequest) -> RequestUrl {
     // `connection_info` resolves the scheme and host, honoring reverse-proxy
     // forwarding headers.
     let conn = req.connection_info();
-    let origin = format!("{}://{}", conn.scheme(), conn.host());
+    let scheme = conn.scheme();
+    let host = conn.host();
     let path = req.path();
     let query = req.query_string();
-    if query.is_empty() {
-        RequestUrl::new(&format!("{origin}{path}"))
-    } else {
-        RequestUrl::new(&format!("{origin}{path}?{query}"))
+    // assemble once in a pre-sized buffer; `RequestUrl::new` then makes the
+    // one unavoidable copy into its `Arc<str>`
+    let mut url = String::with_capacity(
+        scheme.len()
+            + "://".len()
+            + host.len()
+            + path.len()
+            + if query.is_empty() { 0 } else { 1 + query.len() },
+    );
+    url.push_str(scheme);
+    url.push_str("://");
+    url.push_str(host);
+    url.push_str(path);
+    if !query.is_empty() {
+        url.push('?');
+        url.push_str(query);
     }
+    RequestUrl::new(&url)
 }
 
 #[allow(clippy::type_complexity)]
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1328,7 +1328,7 @@ const STATIC_HEADERS_DEFAULT_CAPACITY: NonZeroUsize =
 /// A missing, unparseable, or zero value falls back to the default.
 const STATIC_HEADERS_CAPACITY_ENV: &str = "LEPTOS_STATIC_HEADERS_CACHE_SIZE";
 
-static STATIC_HEADERS: LazyLock<RwLock<LruCache<String, ResponseOptions>>> =
+static STATIC_HEADERS: LazyLock<RwLock<LruCache<String, ResponseParts>>> =
     LazyLock::new(|| {
         let capacity = std::env::var(STATIC_HEADERS_CAPACITY_ENV)
             .ok()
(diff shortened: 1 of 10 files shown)


sabify:
Failed workflows are not related to this PR. Can you please re-run them?

AlexeyMatskevich:
First off, thanks for tackling this — the in-memory `StaticResponse` approach is a
  clean fix for the regeneration (cache-miss) path: serving the freshly rendered HTML
  together with the `ResponseOptions` captured in the *same* render guarantees body and
  headers come from one render, and it drops the re-read entirely. 👍

  One thing I wanted to raise (by code reading, not yet reproduced): the fix closes the
  cache-**miss** window, but the cache-**hit** path still reads the two halves of the
  response independently —

  - body:    `ServeFile` / `NamedFile` opening the file
  - headers: `STATIC_HEADERS.get(path)`

  with no lock pairing them. So if a writer overwrites the file *and* the `STATIC_HEADERS`
  entry for the same path in the window between those two reads, a cache hit can still pair
  one render's body with another render's headers — the same class as #4772.

  This only matters while a writer can still touch an *already-existing* file, which I
  think happens in two situations:

  1. **Startup stampede** — concurrent first requests to a not-yet-generated path: an early
     writer creates the file (later requests now hit it) while other writers are still
     overwriting it. This looks like the intentional no-dedup trade-off, so probably out of
     scope — just noting it for completeness.
  2. **`regenerate` (ISR)** — `StaticRoute::regenerate` re-renders and re-writes an existing
     file on every invalidation, so the window reopens each cycle for any request served
     concurrently. It's lightly documented, but it's public API, so anyone who goes deeper
     into it could run into this.

  The in-memory technique can't really cover this path (a cache hit has no in-memory render
  to pair with), so closing it would need either a lock pairing the two reads, or storing
  the headers alongside the body in a single cache entry. (In our ntex port we pair them
  under a striped write lock taken on both the write and the serve side — happy to share if
  it's useful.)

  Mostly I wanted to check whether narrowing the scope to the regeneration path was
  intentional — totally fine if so.

sabify:
You are absolutely right! The cache hit has the same effect and (maybe) bigger. Just applied a fix in [73d19ab](https://github.com/leptos-rs/leptos/pull/4782/commits/73d19aba3f5f7fd4fe039d00e480e1ba5fcf3701).

The static-route header cache (`STATIC_HEADERS`) stored a `ResponseOptions` — an `Arc<RwLock<ResponseParts>>`. The cache-hit path handed out a *shared* clone of that `Arc`, and `extend_response` applies headers with `std::mem::take`, **draining** them. So the first request to serve a cached entry (including the request that generated it) emptied the shared headers, and every cache hit afterward returned the body with **no custom headers or status** at all.

Fix: Cache an immutable `ResponseParts` snapshot instead of the live `Arc<RwLock<…>>` builder, and apply it to each cache hit non-destructively (clone the headers rather than drain them). The request-scoped `extend_response` path keeps its `mem::take` move, so no extra clone lands on streaming or server-fn responses — the copy is confined to the static cache, which already does file I/O.

**What this approach does:**
- Removes the root error (a mutable, lock-guarded, ref-counted value living in a cache) rather than defending against it.
- Leaner per hit — only the unavoidable `HeaderMap` clone; no `Arc`/`RwLock` allocation, no lock-within-a-lock.
- Hot path untouched: the request-scoped `extend_response` keeps its `mem::take` move, so streaming/server-fn responses get no extra clone.

AlexeyMatskevich:
That's exactly how we handle it in our ntex port: we cache an immutable
  `ResponseParts` snapshot instead of the live `Arc<RwLock<…>>`, and apply a clone of
  it on each hit, so the cached entry is never drained and the status/headers replay on
  every hit. Glad we converged on the same approach — looks good.

integrations/actix/src/lib.rs
```diff
@@ -1405,49 +1405,60 @@ where
...
-                                let options = options.clone();
-                                let path = path.to_owned();
-                                let response_options = owner.with(use_context);
-                                async move {
-                                    write_static_route(
-                                        &options,
-                                        response_options,
-                                        path.as_ref(),
-                                        &html,
+                // Open the file directly instead of probing with `try_exists`
```
gbj: Comments should describe what the code does, not what the code used to do and no longer does. Please remove or amend this comment block so it's relevant.

integrations/axum/src/lib.rs
```diff
@@ -1737,9 +1741,15 @@ where
                 return (StatusCode::NOT_FOUND, "Not Found").into_response();
             };
             let path = Path::new(&file_path);
-            let exists = tokio::fs::try_exists(path).await.unwrap_or(false);
-
-            let (response_options, html) = if !exists {
+            // Serve the file directly instead of probing with `try_exists`
```
gbj: Same as my previous comment. There's no need for a comment that describes something the code no longer does.

integrations/axum/src/lib.rs
```diff
@@ -1028,9 +1028,24 @@ where
                 // client after hydration; fall back to a bare path when the
                 // host is unknown. Building the `RequestUrl` here, before the
                 // request is consumed below, keeps the borrow of `req` short.
-                let request_url = match request_origin(&req) {
-                    Some(origin) => RequestUrl::new(&format!("{origin}{path}")),
-                    None => RequestUrl::new(path),
+                let request_url = match request_scheme_and_host(&req) {
```
gbj, line 1037: This is similar to or identical with the Actix change just made in a separate commit in this PR. Is there an opportunity to refactor into a shared helper here?

gbj:
I left a few comments with nits, but have reviewed this otherwise and it seems good.

integrations/axum/src/lib.rs
sabify: Done.

integrations/axum/src/lib.rs
sabify: Done.

integrations/actix/src/lib.rs
sabify: Done.

gbj:
Thanks!

Issue #4293, opened 2025-09-12
router: with set_is_routing, ProtectedRoute does not preload route resources


wrl:
Following up on a conversation with gbj earlier and a discussion in the discord:

When using `Router`'s `set_is_routing` prop, non-protected routes preload their `Resource`s before navigating to the route. With `ProtectedRoute`, they do not – the routing functions as it would were `set_is_routing` not used.

PR #4784, opened 2026-06-13
fix(router): hold is_routing for ProtectedRoute content with set_is_routing

sabify:
Closes: #4293

With `<Router set_is_routing>`, `is_routing` was cleared as soon as the route's
view had been chosen, which misses:

- content behind a `<ProtectedRoute>` (built through `<Transition>` + `Unsuspend`,
  so its resources are only created when the route is *built*),
- nested outlets built from scratch (from the fallback, or deeper into a layout),
- params-only navigations (`/items/1` → `/items/2`), where nothing is chosen or
  built and resources keyed on the params reload,
- superseded navigations, which could clear the flag for a newer navigation to
  the same route.

Two related pre-existing bugs in the nested router lost a route's view: a
navigation cancelled the whole previous preload batch even for outlets it
reused, and a navigation of only fresh outlets dropped its abort handle while
its preloads were pending, so an abandoned preload could later install its view
over the new route.

## Changes

1. **`RouteSettleContext`** (`reactive_graph`): every `<Suspense>`/`<Transition>`
   built into a route registers a task and releases it once it has built its
   children and has no pending tasks. The context belongs to the outlet's view
   (nested) or the route instance (flat), is guarded from the moment the view is
   scheduled until it has been built, and closes when settled or replaced. A
   navigation polls the contexts of the outlets it displays, reused or new. Each
   navigation gets a sequence number; only the newest one completes it.
2. **Params-only navigations**: `AsyncDerived` now registers with the active
   `AsyncTransition` when a source notifies it, and the new
   `AsyncTransition::track` runs a synchronous action with a transition active.
   Both routers wrap the params update in it and hold `is_routing` until the
   reloads and the reused route's view have settled.
3. **Per-outlet preload cancellation** (nested router): each outlet owns the
   abort handle of its pending preload; a navigation cancels exactly the outlets
   it replaces or removes. The initial load only renders its outlets if no
   navigation has started since.

### DIFF
--- a/examples/router/Cargo.toml
+++ b/examples/router/Cargo.toml
@@ -26,3 +26,5 @@ tracing = "0.1.40"
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.50"
+wasm-bindgen = "0.2"
+web-sys = { version = "0.3", features = ["History"] }
--- a/examples/router/tests/web.rs
+++ b/examples/router/tests/web.rs
@@ -1,0 +1,249 @@
+//! Regression test for `<Router set_is_routing>` with `<ProtectedRoute>`.
+//!
+//! When `set_is_routing` is provided, navigation must keep `is_routing` set to
+//! `true` until the destination route's async resources have loaded (so the old
+//! view stays visible / a progress bar can show). A plain `<Route>` does this.
+//!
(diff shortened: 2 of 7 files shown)


sabify:
@gbj @wrl Can you please confirm if this solves the issue?

PR #4785, opened 2026-06-14
Surface failed lazy-route chunk loads to ErrorBoundary instead of panicking

JohnHarrison:
## Problem

A lazy route's view is loaded from a separate WASM chunk (`#[lazy_route]` → `#[lazy]` → `#[wasm_split]`). Today that `#[wasm_split]` is **not** `fallible`, so a single transient chunk-fetch failure — flaky mobile network, a navigation abort, a momentary 5xx — makes the generated wrapper `.expect()` the load and **panic**. On wasm a panic aborts the whole module (`unreachable`), so the app is bricked until a full page reload, and the failure is cached for the session.

(This is the leptos side of WorldSEnder/wasm-split-prototype#35.)

## Change

Make the generated lazy-route view `fallible` and route its error to the nearest `<ErrorBoundary>`, so a failed chunk renders a fallback (and recovers on the next navigation, since the loader no longer caches failures).

- **`leptos_macro` (`lazy.rs`)** — `#[lazy]` accepts an optional `fallible` flag (alongside the optional module-name ident) that forwards to `#[wasm_split(.., fallible)]`. Existing `#[lazy]` / `#[lazy(name)]` uses are byte-identical.
- **`router_macro` (`lazy_route`)** — emit `#[lazy(fallible)]`; give the generated split view fn the type `Result<AnyView, SplitLoaderError>` (`Ok`-wrapping the user's body — their `view` still reads `-> AnyView`); and `into_any()` it so an `Err` renders to the nearest `<ErrorBoundary>` via the existing `impl<T, E> RenderHtml for Result<T, E>`. `preload()` discards the error (best-effort prefetch; the real surfacing happens in `view()` on the uncached retry).
- Bump the `wasm_split_helpers` floor to **0.2.2**, which adds the `fallible` option and re-exports `SplitLoaderError` (`std::error::Error + Send + Sync`, so it already satisfies `Into<throw_error::Error>` — no new error plumbing needed).
- Document the behavior on `#[lazy_route]` and wrap the `lazy_routes` example in an `<ErrorBoundary>` to model the pattern.

## Why it's small and doesn't change any public API

- The **`LazyRoute` trait is untouched** (`view() -> impl Future<Output = AnyView>`, `preload() -> ()`), and **how users write `#[lazy_route]` is unchanged** (`view` still returns `AnyView`).
- Only an internal, `__`-prefixed generated helper's return type changes, and `#[lazy]` gains an optional flag. So this is a non-breaking change.
- One behavioral note (now in the `#[lazy_route]` docs): without an enclosing `<ErrorBoundary>`, a failed chunk renders nothing rather than panicking — still strictly better than aborting, but worth a boundary to show a fallback.

## Validation

Built the official `examples/lazy_routes` against this branch (with the `wasm_split_helpers` 0.2.2 patch) and drove it with Playwright — aborting exactly one fetch of the ViewC view chunk (`split___view_c_view_*.wasm`):

- **before** (stock): nav 1 → `pageerror: unreachable` (panic); route bricked.
- **after** (this PR): nav 1 → the `<ErrorBoundary>` fallback renders, no panic; nav 2 loads the route.

A standalone non-`fallible` `#[lazy]` data fn in the same example (`lazy_data() -> Vec<Album>`) compiles unchanged, confirming backward-compat.

## Status: draft, blocked on a `wasm_split_helpers` release

This depends on the `fallible` option in WorldSEnder/wasm-split-prototype#36, which isn't published yet — so the `wasm_split_helpers = "0.2.2"` floor here won't resolve from crates.io until that lands. Opening as a draft until then; happy to adjust the design to whatever shape that PR settles on.

---

Thank you all for leptos — the lazy-route + `wasm_split` work is what makes mobile-first Leptos apps viable, and building the fix on top of it was a pleasure precisely because the pieces (`RenderHtml for Result`, `ErrorBoundary`, the `#[lazy_route]` codegen) compose so cleanly. Grateful for the care that goes into this project, and happy to revise anything here. cc @gbj (per the ErrorBoundary discussion on the wasm-split PR).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,9 @@ sha2 = { default-features = false, version = "0.11" }
 subsecond = { default-features = false, version = "0.7" }
 dioxus-cli-config = { default-features = false, version = "0.7" }
 dioxus-devtools = { default-features = false, version = "0.7" }
-wasm_split_helpers = { default-features = false, version = "0.2.1" }
+# 0.2.2 adds the `fallible` option that `#[lazy_route]` relies on to surface a
+# failed chunk load to an `<ErrorBoundary>` instead of panicking.
+wasm_split_helpers = { default-features = false, version = "0.2.2" }
 
 [profile.release]
 codegen-units = 1
@@ -187,3 +189,4 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(leptos_debuginfo)',
   'cfg(erase_components)',
 ] }
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -189,4 +189,3 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(leptos_debuginfo)',
   'cfg(erase_components)',
 ] }
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,8 +171,6 @@ sha2 = { default-features = false, version = "0.11" }
 subsecond = { default-features = false, version = "0.7" }
 dioxus-cli-config = { default-features = false, version = "0.7" }
 dioxus-devtools = { default-features = false, version = "0.7" }
-# 0.2.2 adds the `fallible` option that `#[lazy_route]` relies on to surface a
-# failed chunk load to an `<ErrorBoundary>` instead of panicking.
 wasm_split_helpers = { default-features = false, version = "0.2.2" }
 
 [profile.release]
--- a/examples/lazy_routes/e2e/features/chunk_failure.feature
+++ b/examples/lazy_routes/e2e/features/chunk_failure.feature
@@ -1,0 +1,26 @@
+@chunk_failure
+Feature: A failed lazy-route chunk load renders the ErrorBoundary fallback
+
+	Scenario: A chunk-fetch failure shows the fallback instead of crashing the app.
+		Given I see the app
+		Then I see the page is View A
+		When wasm chunk requests fail
+		When I select the link C
+		Then I see the chunk error fallback
+
+	# TODO: enable once a cargo-leptos release ships wasm_split_cli_support >= 0.2.2
+	# (leptos-rs/cargo-leptos#662). Older versions cache a rejected chunk load for
+	# the lifetime of the session, so the re-navigation below cannot succeed.
+	#
+	# Scenario: A transient chunk-fetch failure is not cached, so navigating again recovers.
+	#	Given I see the app
+	#	Then I see the page is View A
+	#	When wasm chunk requests fail
+	#	When I select the link C
+	#	Then I see the chunk error fallback
+	#	When wasm chunk requests succeed again
+	#	When I select the link A
+	#	Then I see the page is View A
+	#	When I select the link C
+	#	When I wait for a second
+	#	Then I see the page is View C
--- a/examples/lazy_routes/e2e/features/chunk_failure.feature
+++ b/examples/lazy_routes/e2e/features/chunk_failure.feature
@@ -8,19 +8,15 @@ Feature: A failed lazy-route chunk load renders the ErrorBoundary fallback
 		When I select the link C
 		Then I see the chunk error fallback
 
-	# TODO: enable once a cargo-leptos release ships wasm_split_cli_support >= 0.2.2
-	# (leptos-rs/cargo-leptos#662). Older versions cache a rejected chunk load for
-	# the lifetime of the session, so the re-navigation below cannot succeed.
-	#
-	# Scenario: A transient chunk-fetch failure is not cached, so navigating again recovers.
-	#	Given I see the app
-	#	Then I see the page is View A
-	#	When wasm chunk requests fail
-	#	When I select the link C
-	#	Then I see the chunk error fallback
-	#	When wasm chunk requests succeed again
-	#	When I select the link A
-	#	Then I see the page is View A
-	#	When I select the link C
-	#	When I wait for a second
-	#	Then I see the page is View C
+	Scenario: A transient chunk-fetch failure is not cached, so navigating again recovers.
+		Given I see the app
+		Then I see the page is View A
+		When wasm chunk requests fail
+		When I select the link C
+		Then I see the chunk error fallback
+		When wasm chunk requests succeed again
+		When I select the link A
+		Then I see the page is View A
+		When I select the link C
+		When I wait for a second
+		Then I see the page is View C
--- a/leptos/src/lazy_view_error.rs
+++ b/leptos/src/lazy_view_error.rs
@@ -7,17 +7,6 @@ use wasm_split_helpers::SplitLoaderError;
 
 /// The error produced when a lazy route's view fails to load its WASM chunk.
 ///
-/// `#[lazy_route]` generates a `fallible` view whose `Err` is rendered by the
-/// nearest [`ErrorBoundary`](crate::error::ErrorBoundary) instead of panicking.
-/// That rendering path requires the error to be `Send + Sync + 'static`, but the
-/// underlying [`SplitLoaderError`] is deliberately allowed to be neither (so it
-/// can, for example, carry a `JsValue` describing the JavaScript failure).
-///
-/// This wrapper bridges the two: it holds the split error in a [`SendWrapper`],
-/// which is `Send + Sync` unconditionally. That is sound here because a lazy
-/// chunk only ever loads — and so this error is only ever constructed and
-/// rendered — on the single WASM thread.
-///
 /// You can match on this type from an `<ErrorBoundary>` fallback (via
 /// `Error::downcast_ref`) to detect a chunk-load failure specifically.
 #[derive(Clone)]
--- a/leptos/src/lazy_view_error.rs
+++ b/leptos/src/lazy_view_error.rs
@@ -1,0 +1,43 @@
+//! Error type bridging a failed lazy-route chunk load into an
+//! [`ErrorBoundary`](crate::error::ErrorBoundary).
+
+use send_wrapper::SendWrapper;
+use std::fmt;
+use wasm_split_helpers::SplitLoaderError;
+
+/// The error produced when a lazy route's view fails to load its WASM chunk.
+///
+/// `#[lazy_route]` generates a `fallible` view whose `Err` is rendered by the
+/// nearest [`ErrorBoundary`](crate::error::ErrorBoundary) instead of panicking.
+/// That rendering path requires the error to be `Send + Sync + 'static`, but the
+/// underlying [`SplitLoaderError`] is deliberately allowed to be neither (so it
+/// can, for example, carry a `JsValue` describing the JavaScript failure).
+///
+/// This wrapper bridges the two: it holds the split error in a [`SendWrapper`],
+/// which is `Send + Sync` unconditionally. That is sound here because a lazy
+/// chunk only ever loads — and so this error is only ever constructed and
+/// rendered — on the single WASM thread.
+///
+/// You can match on this type from an `<ErrorBoundary>` fallback (via
+/// `Error::downcast_ref`) to detect a chunk-load failure specifically.
+pub struct LazyViewError(SendWrapper<SplitLoaderError>);
+
+impl fmt::Debug for LazyViewError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&*self.0, f)
+    }
+}
+
+impl fmt::Display for LazyViewError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&*self.0, f)
+    }
+}
+
+impl std::error::Error for LazyViewError {}
+
+impl From<SplitLoaderError> for LazyViewError {
+    fn from(err: SplitLoaderError) -> Self {
+        Self(SendWrapper::new(err))
+    }
+}
--- a/leptos/src/lazy_view_error.rs
+++ b/leptos/src/lazy_view_error.rs
@@ -20,8 +20,12 @@ use wasm_split_helpers::SplitLoaderError;
 ///
 /// You can match on this type from an `<ErrorBoundary>` fallback (via
 /// `Error::downcast_ref`) to detect a chunk-load failure specifically.
+#[derive(Clone)]
 pub struct LazyViewError(SendWrapper<SplitLoaderError>);
 
+// Delegate `Debug`/`Display` to the inner error rather than deriving them, so
+// diagnostics show the `SplitLoaderError` itself and not the `SendWrapper`
+// implementation detail.
 impl fmt::Debug for LazyViewError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         fmt::Debug::fmt(&*self.0, f)
@@ -34,7 +38,13 @@ impl fmt::Display for LazyViewError {
     }
 }
 
-impl std::error::Error for LazyViewError {}
+impl std::error::Error for LazyViewError {
+    // Preserve the chain to the underlying `SplitLoaderError` so a fallback can
+    // walk `source()` (or downcast it) for the load-failure details.
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&*self.0)
+    }
+}
 
 impl From<SplitLoaderError> for LazyViewError {
     fn from(err: SplitLoaderError) -> Self {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4659,9 +4659,9 @@ dependencies = [
 
 [[package]]
 name = "wasm_split_helpers"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d0cb6d1008be3c4c5abc31a407bfb8c8449ae14efc8561c1db821f79b9614b0a"
+checksum = "69a727977b2d2e2193e8f85f59069d791d9ce091a4217b8c3d6a6f17f1554498"
 dependencies = [
  "async-once-cell",
  "wasm_split_macros",
@@ -4669,9 +4669,9 @@ dependencies = [
 
 [[package]]
 name = "wasm_split_macros"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d0a659ffe5c7f4538aa6357c07e3d73221cc61eba03bd9a081e14bc91ed09b8c"
+checksum = "3e653af7ee4a9ef0fce481a9ec6f43cb78de20d0cdb4f4f5862e1dc6e407e6c8"
 dependencies = [
  "base16",
  "quote",
--- a/examples/lazy_routes/e2e/tests/fixtures/action.rs
+++ b/examples/lazy_routes/e2e/tests/fixtures/action.rs
@@ -21,3 +21,33 @@ pub async fn click_button(client: &Client, id: &str) -> Result<()> {
     btn.click().await?;
     Ok(())
 }
+
+/// Simulates a network failure for lazily loaded WASM chunks by shimming
+/// `window.fetch` to reject requests for `.wasm` files while blocking is
+/// enabled. The main bundle is unaffected: it has already been loaded by the
+/// time this runs, and everything that isn't a `.wasm` request passes through.
+pub async fn set_wasm_chunks_blocked(client: &Client, blocked: bool) -> Result<()> {
+    client
+        .execute(
+            r#"
+            const [blocked] = arguments;
+            if (!window.__realFetch) {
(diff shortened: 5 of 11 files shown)


JohnHarrison:
@gbj I just wanted to ask your advice here on breaking the logjam as this depends on warm_split_helpers 0.2.2 (unreleased) which in turn is waiting for this change to leptos-rs.  Do you have thoughts here?

gbj:
My only thought at first glance is that this is a breaking change so it should be targeted against the `leptos_0.9` branch and can be released as part of 0.9.

I will happily review it in more detail in the near future.

Note that between now and then, there is no logjam: you can use the `[patch.crates-io]` section of your `Cargo.toml` to use whichever forks or patched versions of the relevant dependencies that you want.

JohnHarrison:
Thanks @gbj! Makes sense — I'll retarget this to leptos_0.9 and reshape the codegen against the 0.9 macros so it can land as part of 0.9.  As you know the other work landed in wasp-splitter 0.2.2, so this should tie things together nicely.

JohnHarrison:
Retargeted to `leptos_0.9` per your note, and re-ported the codegen onto the 0.9 macros (the change applied cleanly — the 0.9 `lazy`/`lazy_route` macros differ only cosmetically here). The wasm-split side has landed: WorldSEnder merged the `fallible` mechanism (WorldSEnder/wasm-split-prototype#36), which publishes as `wasm_split_helpers` 0.2.2 (the floor this PR bumps to). CI will stay red on the dep until 0.2.2 is on crates.io; locally I validated via `[patch.crates-io]` as you suggested.

**End-to-end validation on `leptos_0.9` + the merged wasm-split `main`** (built with `cargo leptos --split`, then aborting exactly one ViewC view-chunk fetch via Playwright route interception):

| Variant (only the macro fix differs) | nav 1 — one view-chunk fetch aborted | result |
| --- | --- | --- |
| stock 0.9 macros | `panicked … load callback should succeed: SplitLoaderError(())` → `unreachable`; module aborts, route bricked | ❌ |
| this PR (`fallible` lazy route) | no panic; `<ErrorBoundary>` fallback renders; route recovers on the next navigation (failure isn't cached) | ✅ |

The A/B isolates the fix cleanly — same merged `wasm_split_helpers` 0.2.2, same `<ErrorBoundary>`, only the two macro files change. Notably, #36's not-caching is necessary but **not sufficient** on its own: a non-`fallible` lazy route still `.expect()`s the load and panics, and on wasm a panic aborts the whole module regardless of caching. The macro emitting `fallible` is what turns the failure into a catchable `Err`.

Test case (the exact Playwright script used): https://github.com/JohnHarrison/wasm-split-chunk-repro/blob/main/lazy_route_errorboundary_repro.py

Ready for the detailed review whenever you have a moment — thanks again!

WorldSEnder:
@gbj Does the Error type for the ErrorBoundary have to implement any Send/Sync requirement? As of now, the split error does, but it could capture a `JsValue` in which case it wouldn't any more and I don't want that to be a breaking change.

JohnHarrison:
Not my call to make on the API shape — fully defer to you both on what `SplitLoaderError` should carry. But since this PR is the consumer, here's what the 0.9 `<ErrorBoundary>` path concretely requires today, in case it's useful input:

The error routes to the boundary via `Result<AnyView, E>: RenderHtml`, and the bound chain is `Send + Sync + 'static`:

- `impl RenderHtml for Result<T, E>` requires `E: Into<throw_error::Error> + Send + 'static` (`tachys/src/view/error_boundary.rs:135`)
- `throw_error::Error` is `Arc<dyn std::error::Error + Send + Sync>`, and its only `From` is `T: Into<Box<dyn Error + Send + Sync + 'static>>` (`any_error/src/lib.rs:23,49`) — so `Into<AnyError>` itself implies `Send + Sync + 'static`
- the generated `view()` then calls `into_any()`, and `impl IntoAny for T` requires `T: Send` (`tachys/src/view/any_view.rs:196`)

So your instinct is right: a `SplitLoaderError` that captured a raw `JsValue` (`!Send + !Sync`) would stop satisfying `Into<throw_error::Error>` and break this integration to compile. If you do want to surface richer detail later without that being breaking, keeping the error `Send + Sync` — e.g. stringifying the `JsValue` into an owned `String` at the boundary rather than storing the `JsValue` — would preserve it. The current `SplitLoaderError(())` is of course fine.

Whatever you settle on works for me; I'll adapt the consumer side to match. Thanks for thinking about the forward-compat here — much appreciated.

WorldSEnder:
> and the bound chain is `Send + Sync + 'static`

In that case, I think you could also, on the leptos side, wrap the error value in a `send_wrapper::SendWrapper` and have `SplitLoaderError` itself be exposed as non-`Send`/`Sync`? Let me know if you think that's workable I'd rather keep it open to have a future `JsValue` representing the full javascript error as an option.

If that's not workable, I can also keep it in mind and try to work around it with something along the lines of https://github.com/wasm-bindgen/wasm-bindgen/discussions/3971 in the future.

JohnHarrison:
Done — implemented exactly that, pushed as f912ea8. It works and keeps `SplitLoaderError` free to become non-`Send`/`Sync` later.

`leptos::LazyViewError` is a thin newtype holding `SendWrapper<SplitLoaderError>`:

```rust
pub struct LazyViewError(send_wrapper::SendWrapper<SplitLoaderError>);
// Debug/Display delegate to the inner error; impls std::error::Error;
// From<SplitLoaderError> wraps in SendWrapper.
```

`SendWrapper` is unconditionally `Send + Sync`, which is sound here because a chunk only ever loads on the single WASM thread (leptos already uses `SendWrapper` this way in `#[lazy]`). The generated `#[lazy_route]` view fn now returns `Result<AnyView, LazyViewError>`; the `fallible` `?`/`From` machinery converts `SplitLoaderError` into it for free. One detail worth noting: `SendWrapper` itself doesn't impl `Error`/`Display`, so the newtype provides those by delegating through the wrapper.

I validated that this actually buys the forward-compat you're after, not just the happy path: with `SplitLoaderError` locally changed to `!Send + !Sync` (`PhantomData<*const ()>`),

- routing it **directly** fails to compile — `*const () cannot be sent between threads safely`, `Result<AnyView, SplitLoaderError>: IntoAny` not satisfied (the `E: Into<AnyError> + Send` bound);
- routing it through **`LazyViewError`** compiles cleanly, and the end-to-end repro still passes (no panic → ErrorBoundary fallback → recovery on renavigation).

So you're free to give `SplitLoaderError` a real `JsValue` payload whenever you like; this side won't break. Bonus: users can `downcast_ref::<LazyViewError>()` in a fallback to detect a chunk-load failure specifically. As always, happy to reshape if you or @gbj prefer a different spot for the wrapper. Thanks for the steer!

JohnHarrison:
Marking this ready for review @gbj — `wasm_split_helpers` 0.2.2 is now published (WorldSEnder/wasm-split-prototype#37 merged), so the dependency floor resolves from crates.io and CI should go green on a re-run.

Summary of where it landed since your last look:
- retargeted to `leptos_0.9` and re-ported onto the 0.9 macros
- a failed lazy-route chunk now renders an `Err` to the nearest `<ErrorBoundary>` instead of panicking; the uncached failure retries on the next navigation (validated end-to-end against the published 0.2.2)
- per WorldSEnder's forward-compat point, the load error is wrapped in `leptos::LazyViewError` (`SendWrapper<SplitLoaderError>`) so `SplitLoaderError` is free to become non-`Send`/`Sync` later (e.g. to carry a `JsValue`) without breaking this side

No `LazyRoute` trait or user-facing API changes. Happy to adjust anything — thanks!

gbj:
Hi @JohnHarrison. Thanks for the PR. I think this combination will be a great improvement, especially for things like... navigating around a lazy-loaded site while on the subway.

I'm wondering how you tested the changes to the example, and whether they worked for you?

Here's what I tried:
- open the `lazy_routes` example in the PR branch
- run it with `cargo leptos serve --split`
- open the home page
- kill the server with Ctrl-C, so that any further requests would be network errors, which should trigger the error case and therefore the ErrorBoundary
- navigate to any other route, or click any of the buttons
- it panics, rather than triggering the ErrorBoundary

I think this is because every page uses some other kind of `#[lazy]` function as well, and the PR automatically opts into fallible loading for view functions, but not for other `#[lazy]` functions, right? 

So, I tried annotating one of the functions with `#[lazy(falllible)]`. This was pretty awkward because we don't re-export `SplitLoaderError`, so I had to add `wasm_split_helpers` and import it from there -- that's easy to fix though. I did this:
```rust
            #[lazy(fallible)]
            pub fn second_value() -> Result<String, wasm_split_helpers::SplitLoaderError> {
                Ok("Third value.".to_string())
            }

            spawn_local(async move {
                if let Ok(second_value) = second_value().await {
                    result.set(second_value);
                }
            });
```

Same process:
- start the server
- load the page
- kill the server
- click Third
- failed network request -- error in console, no Rust panic (good)
- restart the server
- if I open a new tab, works as expected
- but back in the original tab, it does *not* send additional requests to load the WASM chunk, it just keeps repeating the same error in the console
- then, clicking another lazy function (like Second) panics with `Uncaught RuntimeError: function signature mismatch`

It's entirely possible that I'm doing something wrong, in which case — let me know. But I can't seem to get this to work as intended.

JohnHarrison:
Thanks for the careful testing, @gbj — confirmed all three, reproduced them locally, and traced each. Two are already fixed upstream; one is a design question for you.

**1. Other `#[lazy]` fns still panic offline.** Right — the PR makes only the route view fallible, so any other `#[lazy]` fn (data loader, button handler) still panics. This is inherent to the signature, not a bug: a `#[lazy] fn -> String` must return a `String`, so a failed load can only panic or hang unless it's fallible (which turns the return type into `Result` and forces callers to handle it — the macro can't do that silently). So the question is whether `#[lazy]` should be **fallible by default** (breaking; every lazy fn returns `Result`) or stay **opt-in** with this PR covering the route-view case. Your call — happy to do either.

**2 & 3 — no retry after reconnect, and the function signature mismatch.** Both reproduced, and both are already fixed in WorldSEnder's `wasm-split-prototype`, not here:

- **No retry:** the cargo-leptos-generated `__wasm_split.js` was caching the failed load (memoizing the promise even on rejection), so the Rust side's retry kept getting the same rejected promise. The crate that emits that JS (`wasm_split_cli_support`) already has the fix — cache successes only. I confirmed by hand: with that one change, a failed route recovers on the next navigation, no reload.
- **Signature mismatch:** looks covered by the recent "canary to imported symbols" change (turns it into a clear `LinkError`).

These just aren't published yet, so it's release plumbing only — WorldSEnder publishes the updated `wasm_split_cli_support`, cargo-leptos bumps that dep. Nothing in this PR changes.

One correction on my part: my earlier "recovers on next navigation" check used a full reload, which hid the no-retry bug. Your client-side test is the right one — good catch.

**TLDR** the PR correctly removes the route-view panic; retry + signature-mismatch are handled upstream pending a release; the one open question for leptos is fallible-by-default vs opt-in. How would you like to handle that last piece?

WorldSEnder:
The "canary" idea is to avoid accidentally loading wasm from one compilation of an executable into a different executable when the conflicting versions of are served by the same web-server at different times. This might be relevant, but without code changes and in the default incremental mode, just killing and restarting the server should actually lead to binary-compatible split modules and not signature mismatches. I haven't looked deeper into the error case gbj alluded to though. A way to avoid triggering a recompile/rerun on the splitting would be to enable "Throttling: Offline" in the Network tab of the browser console (or similar) while trying the network down navigation.

Regarding the "awkward" reexport, the idea was to `Into` a framework defined error such as `LazyViewError` (naming nitpick aside) in lazy functions such as `second_value`. Using the signature

    pub fn second_value() -> Result<String, LazyViewError>

should work with the same implementation and only using leptos API.

JohnHarrison:
Pushed an update (`0874a26`) and did a fair bit of local testing to pin down @gbj's findings. Summary of where things stand, then the decisions I'd like your steer on.

### What I confirmed about the three findings

- **Other `#[lazy]` fns panic offline** — real, and it's inherent to the signature: a `#[lazy] fn -> T` has no `T` to return on a failed load, so it can only panic unless it's `fallible`. The fix is symmetric with the view path: make the loader fallible and let its error flow to the boundary. I converted the example's ViewC data loader to demonstrate it end-to-end (below).
- **No retry after reconnect** — root cause is the cargo-leptos-generated `__wasm_split.js`: its `makeLoad` memoized the load promise *including rejections*, so the Rust side's retry got handed the same rejected promise and never re-fetched. That JS comes from `wasm_split_cli_support` (behind `cargo leptos --split`), and its **current source already has the fix** (memoize successes only). I confirmed by patching the one line locally — client-side retry then recovers with no reload. It just needs a `wasm_split_cli_support` publish + a cargo-leptos dep bump; nothing in this PR.
- **`function signature mismatch`** — I could not reproduce it under clean browser *offline emulation* (no server restart). @WorldSEnder's read looks right: it came from restarting `cargo leptos serve`, which re-split against an already-loaded main module. So my earlier "the canary fixes it" was wrong — apologies; the canary targets a different cross-version case. Under steady-state offline navigation there's no mismatch.

### Demonstrated: an end-to-end offline-resilient lazy route

With both the view (automatic via `#[lazy_route]`) and the data loader made fallible:

```rust
#[lazy(fallible)]
async fn lazy_data() -> Result<Vec<Album>, LazyViewError> { Ok(/* … */) }
// in view: let albums = this.data.await?;   // Err -> <ErrorBoundary>
```

navigating to the route fully offline now renders the `<ErrorBoundary>` fallback with **no panic** (was: panic on the data chunk). This is the `Result<_, LazyViewError>` shape @WorldSEnder suggested, and it works with leptos-only API — no `wasm_split_helpers` import. So the "subway" scenario is achievable today with opt-in; the question is whether it should be the default.

### Surface changes in `0874a26` (kept minimal)

- `LazyViewError` is **no longer in the prelude** — only `leptos::LazyViewError`, so `use leptos::prelude::*` doesn't reserve the name for apps that never touch lazy routes.
- It now derives `Clone` (required to hold it in a `Resource`) and implements `Error::source()` so a fallback can reach the underlying `SplitLoaderError`.
- Example: ViewC demonstrates the fallible data pattern; the `<Routes>` are wrapped in an `<ErrorBoundary>`.

### Decisions I'd like your call on

1. **Default vs opt-in for `#[lazy]` fallibility.** This PR makes the *view* fallible automatically; everything else is opt-in (`#[lazy(fallible)] -> Result<_, LazyViewError>`). Making it the default would cover the offline case transparently but is a breaking, signature-level change (every lazy fn returns `Result`). Opt-in keeps signatures but requires authors to annotate each loader. Happy to implement either.
2. **Where should `LazyViewError` live?** Right now it's `leptos::LazyViewError`, but `lazy_route` is owned by `leptos_router`, and there's already a `leptos::error` module next to `ErrorBoundary`. I can move it to whichever you prefer (`leptos::error`, `leptos_router`, or leave it) — it's the macro's stable codegen target so I'd rather you pick. (And the naming nitpick — open to a better name than `LazyViewError`.)
3. **Heads-up to document:** an existing lazy route with *no* `<ErrorBoundary>` now renders nothing on a failed chunk instead of panicking. Worth a CHANGELOG note; flagging in case you'd rather a different default.

Thanks both — this has been a genuinely fun thread to chase down.

JohnHarrison:
@gbj — gentle nudge on this, no urgency. It's green and ready; the only thing it's waiting on is a thumbs-up from you on one design call. Here's one possible direction:

Ship this as opt-in now, treat fallible-by-default as a separate follow-up.  

This PR makes the route view fallible automatically; other #[lazy] fns opt in via #[lazy(fallible)] -> Result<_, LazyViewError> (demonstrated end-to-end in the example — a fully offline-resilient route). Non-breaking. 

Making all #[lazy] fns fallible-by-default would cover the offline case transparently but is a breaking signature change — better as its own PR with its own discussion than bundled here.
  
Let me know what you prefer and I'll get on it.

thanks,
John

JohnHarrison:
Added e2e coverage for the behavior this PR introduces (`4434386`), since the earlier validation was manual.

**New scenario in `examples/lazy_routes/e2e`:** a `fetch` shim simulates a network failure for `.wasm` chunk requests, then navigates to the lazy route `/c` and asserts the `<ErrorBoundary>` fallback renders instead of the app crashing. Runs in the existing cucumber/fantoccini suite with no new dependencies.

**A second scenario — recovery on re-navigation** (failure isn't cached; going back to `/c` once the network is healthy loads the route) — is included but commented out for now: it also needs the retrying split loader from `wasm_split_cli_support` 0.2.2, which published cargo-leptos releases don't ship yet (opened leptos-rs/cargo-leptos#662 for that). Verified locally against a cargo-leptos built with that bump: both scenarios pass, along with the rest of the suite. Once a cargo-leptos release includes it, the scenario can be uncommented as-is.

WorldSEnder:
I may have found out why the same code leads to incompatible binaries; There are some decisions in the splitter that depend on the iteration order over a hashmap, which will be different across different runs.

I have opened WorldSEnder/wasm-split-prototype#46 to track this upstream. For now, this means that restarting the server can lead to incompatible modules getting loaded even with the additional protection of the canary import. The testing procedure in the PR should not be impacted, as it shims the fetch on client side.

gbj:
It's *always* iteration over a HashMap, isn't it :smile:

leptos/src/lazy_view_error.rs
```diff
@@ -0,0 +1,53 @@
+//! Error type bridging a failed lazy-route chunk load into an
+//! [`ErrorBoundary`](crate::error::ErrorBoundary).
+
+use send_wrapper::SendWrapper;
+use std::fmt;
+use wasm_split_helpers::SplitLoaderError;
+
+/// The error produced when a lazy route's view fails to load its WASM chunk.
+///
+/// `#[lazy_route]` generates a `fallible` view whose `Err` is rendered by the
```
gbj: I think almost all of this doc comment except the first sentence can be removed. It explains the decision-making process behind an internal detail that is not exposed to the user (the `SendWrapper`)

examples/lazy_routes/e2e/features/chunk_failure.feature
```diff
@@ -0,0 +1,26 @@
...
+Feature: A failed lazy-route chunk load renders the ErrorBoundary fallback
+
+	Scenario: A chunk-fetch failure shows the fallback instead of crashing the app.
+		Given I see the app
+		Then I see the page is View A
+		When wasm chunk requests fail
+		When I select the link C
+		Then I see the chunk error fallback
+
+	# TODO: enable once a cargo-leptos release ships wasm_split_cli_support >= 0.2.2
```
gbj: I'm about to release a new version this afternoon, then this can be removed so the test actually runs

Cargo.toml
```diff
@@ -171,7 +171,9 @@ sha2 = { default-features = false, version = "0.11" }
 subsecond = { default-features = false, version = "0.7" }
 dioxus-cli-config = { default-features = false, version = "0.7" }
 dioxus-devtools = { default-features = false, version = "0.7" }
-wasm_split_helpers = { default-features = false, version = "0.2.1" }
+# 0.2.2 adds the `fallible` option that `#[lazy_route]` relies on to surface a
```
gbj: remove this comment about the dependency

gbj:
@JohnHarrison I left some comments above, otherwise I think this is good to go once I release a new `cargo-leptos` version (in the next couple hours, I'm just waiting on CI.)

JohnHarrison:
@gbj gentle nudge now that cargo-leptos v0.3.7 is out (which was the last thing you mentioned waiting on) — is there anything left on my end before this can merge? Happy to rebase if it's drifted.

gbj:
@JohnHarrison I left multiple comments in my review two weeks ago that have not been addressed.

JohnHarrison:
> @JohnHarrison I left multiple comments in my review two weeks ago that have not been addressed.

@gbj Sorry to be dense.  I believe I've responded to your comments in this thread and I don't see any comments on the code itself.  Do you mean you June 19th comment?  I thought that got answered across a series of my responses and changes here?

gbj (commented):
Sorry -- looks like GitHub left this review "pending" rather than submitting it. Here are the comments I'm talking about.

JohnHarrison:
Addressed all three — thanks for re-submitting the review (GitHub's pending-state strikes again):

- **`LazyViewError` docs**: trimmed to the what-it-is sentence plus the `downcast_ref` tip; the `SendWrapper` design narrative is gone from the user-facing docs.
- **Recovery e2e scenario**: enabled now that cargo-leptos 0.3.7 ships the retrying loader. Validated locally against the released tooling — full `cargo make integration-test` run with cargo-leptos 0.3.7 + ChromeDriver/Chrome 150: all scenarios pass, including the newly-enabled one (fail chunk → fallback → unblock → re-navigate → View C renders).
- **Cargo.toml**: dependency comment removed.

Also rebased onto current `leptos_0.9` (was ~86 commits behind; two small conflicts in `lazy.rs`/`lib.rs` from the `preload_name`/prelude refactors, resolved as the union).

gbj:
Thanks for the PR!

PR #4787, opened 2026-06-17
fix(server_fn): return an error instead of panicking on missing multipart boundary

Noethix55555:
## Summary

FromReq for MultipartFormData called .expect() when the request had a missing or invalid Content-Type header. This panics the server instead of returning a proper error to the caller.

The fix replaces .expect() with .ok_or_else()? using ServerFnErrorErr::Deserialization, consistent with how other codecs in this crate handle malformed input.

## Changes

- server_fn/src/codec/multipart.rs: replace .expect() with ok_or_else + ?, import IntoAppError and ServerFnErrorErr


gbj:
Duplicate of #4718

PR #4788, opened 2026-06-17
fix(tachys): don't render keyed-list key as a text node in SSR

Noethix55555:
In `KeyedView::to_html_with_buf`, the SSR sync path iterated `self.ssr_items` (type `Vec<(String, V)>`) and called `to_html_with_buf` on each tuple. Because tuples implement `RenderHtml` for all elements, this rendered both the `String` key and the view `V`, inserting a stray text node for the key on every list item and causing hydration desync.

The async path already destructured correctly with `for (key, item) in self.ssr_items`. This fix mirrors that pattern in the sync path (`for (_, item) in self.ssr_items`) and also fixes `dry_resolve` for consistency.

### DIFF
--- a/tachys/src/view/keyed.rs
+++ b/tachys/src/view/keyed.rs
@@ -256,7 +256,7 @@ where
 
     fn dry_resolve(&mut self) {
         #[cfg(feature = "ssr")]
-        for view in &mut self.ssr_items {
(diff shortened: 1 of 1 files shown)


gbj:
> and causing hydration desync

Do you have a reproducible example of this issue?

PR #4791, opened 2026-06-18
fix(server_fn): use StreamingText content-type in IntoReq and IntoRes

Noethix55555:
`IntoReq<StreamingText>` and `IntoRes<StreamingText>` both passed `Streaming::CONTENT_TYPE` (`"application/octet-stream"`) instead of `StreamingText::CONTENT_TYPE` (`"text/plain"`).

This means any server function using `TextStream` as its output type would send a `Content-Type: application/octet-stream` header instead of `Content-Type: text/plain`, and the client request would also carry the wrong content-type. Middleware or proxies that inspect Content-Type, as well as clients performing content-type validation, would see the wrong type.

### DIFF
--- a/server_fn/src/codec/stream.rs
+++ b/server_fn/src/codec/stream.rs
@@ -217,7 +217,7 @@ where
         Request::try_new_post_streaming(
             path,
             accepts,
-            Streaming::CONTENT_TYPE,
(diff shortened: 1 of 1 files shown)


gbj:
Duplicate of #4721

PR #4793, opened 2026-06-18
fix(leptos_macro): drop by_ref from generated Props struct fields

Noethix55555:
A component prop declared with a `ref` binding pattern (e.g. `fn Foo(ref x: String)`) causes the macro to emit `pub ref x: String` inside the generated `Props` struct. `ref` is valid in function argument position but not in struct field position, so this produces a confusing compile error in macro-generated code.

The fix removes `by_ref` from the destructuring and the `quote!` output in both `prop_builder_fields` and `prop_serializer_fields`.

```rust
// before (emits invalid `pub ref x: String` in Props struct)
let PatIdent { ident, by_ref, .. } = &name;
quote! { #vis #by_ref #ident: #ty, }

// after
let PatIdent { ident, .. } = &name;
quote! { #vis #ident: #ty, }
```

### DIFF
--- a/leptos_macro/src/component.rs
+++ b/leptos_macro/src/component.rs
@@ -1138,7 +1138,7 @@ fn prop_builder_fields(
                     quote!()
                 };
 
-            let PatIdent { ident, by_ref, .. } = &name;
+            let PatIdent { ident, .. } = &name;
 
             quote! {
                 #docs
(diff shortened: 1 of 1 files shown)


gbj:
This now has merge conflicts. If you resolve them I'll be happy to merge it.

PR #4794, opened 2026-06-18
fix(reactive_stores): untrack parent in AtIndex::writer to avoid spurious notifications

Noethix55555:
AtIndex::writer was not calling parent.untrack() before constructing its WriteGuard. This caused the parent collection's notification chain to fire when writing a single element, so effects tracking the whole collection re-ran on every element write. Subfield::writer and KeyedSubfield::writer both call untrack() and use triggers_for_current_path(); AtIndex::writer was inconsistent. Fix: call parent.untrack() and use triggers_for_current_path() so only the element's own trigger chain propagates. Added a regression test.

### DIFF
--- a/reactive_stores/src/iter.rs
+++ b/reactive_stores/src/iter.rs
@@ -71,7 +71,7 @@ where
     type Value = Prev::Output;
     type Reader = MappedMutArc<Inner::Reader, Prev::Output>;
     type Writer =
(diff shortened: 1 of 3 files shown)


gbj:
Thanks! 

Retargeted against `leptos_0.9` branch because this is technically a breaking change (changes an associated type on a trait impl). I don't know why those other commits now appear as part of this PR, but whatever.

PR #4795, opened 2026-06-18
fix(server_fn): don't panic in set_server_url if already set

Noethix55555:
`set_server_url` calls `ROOT_URL.set(url).unwrap()`. `OnceLock::set`
returns `Err` when the lock is already initialised, which causes `.unwrap()`
to panic. In practice this can happen with hot-module reload, in tests that
call `set_server_url` in multiple test cases, or in component libraries that
call it at startup.

The value is idempotent by construction (a `&'static str` set once at
startup), so silently ignoring a duplicate call with `let _ = ...` is safe
and matches how similar one-time initialisers are handled elsewhere.

### DIFF
--- a/server_fn/src/client.rs
+++ b/server_fn/src/client.rs
@@ -8,8 +8,27 @@ static ROOT_URL: OnceLock<&'static str> = OnceLock::new();
 /// Set the root server URL that all server function paths are relative to for the client.
 ///
 /// If this is not set, it defaults to the origin.
+///
+/// # Panics
+///
+/// Panics if the server URL has already been set. Setting it twice is almost
+/// always a bug: requests would otherwise be silently sent to a different URL
+/// than intended. If you need to handle the already-set case gracefully, use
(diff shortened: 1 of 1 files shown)


gbj:
Why is silent failure a good idea here?

You mention three cases:
- hot-module reload
- tests that call set_server_url in multiple test cases
- component libraries that call it at startup

Silent failure sounds like a bad option to me in each of these three cases.
- hot-module reload: if you're changed the value and it fails silently on hot-module reload, now you've got the wrong (old) value
- tests: currently those will fail in a fairly obvious way, and should just call the setup once; if we want to expose a fallible option we can do that
- component libraries that call it at startup: honestly this sounds like malware; why would a component library set the global URL for server functions? That sounds like it would redirect all my calls to someone else's server. If I call `set_server_url()` and a component library has already called it, I *definitely* want it to panic rather than sending all my API requests somewhere else!

PR #4796, opened 2026-06-18
fix(leptos_server): log serialization error instead of panic in resource SSR

Noethix55555:
`resource.rs` and `once_resource.rs` call `Ser::encode(...).unwrap()` to
serialize resource values for hydration. The `Ser` codec is generic and its
`Error` type is only required to implement `Debug`, not `Infallible`, so
`encode` can fail legitimately (e.g. a value that is not serializable by the
chosen codec). Unwrapping panics the SSR task.

A `// TODO handle serialization errors` comment in `resource.rs` already
acknowledged this. `shared.rs` already handles the same call correctly with
a `match` that logs the error. Apply the same pattern to both files.

### DIFF
--- a/leptos_server/src/once_resource.rs
+++ b/leptos_server/src/once_resource.rs
@@ -164,10 +164,13 @@ where
                             .map(|e| e.into_encoded_string())
                         {
                             Ok(s) => s,
-                            #[allow(unused_variables)]
                             Err(e) => {
                                 #[cfg(feature = "tracing")]
                                 tracing::error!(
+                                    "couldn't serialize resource: {e:?}"
+                                );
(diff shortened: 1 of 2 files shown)


gbj:
This is fine, but now fails silently if tracing is not enabled, which is confusing. I'd suggest adding a `#[cfg(not(feature = "tracing"))]` that logs it via `eprintln` or similar.

sabify:
This already has been patched in a2f1664 for leptos_0.9 branch.

gbj:
@sabify Thanks, I'd forgotten about that one -- although the same applies there! (it fails silently if `tracing` is not enabled)

sabify:
I just made a new PR to address it: #4805.

PR #4797, opened 2026-06-18
fix(axum): don't panic in redirect() on invalid header values

Noethix55555:
`redirect()` calls `HeaderValue::from_str(path).expect(...)`. A redirect
target that contains CR, LF, NUL, or other bytes illegal in HTTP headers
causes `from_str` to return an error, which panics the Tokio worker and
takes down the server -- a potential DoS for any app that passes
user-controlled input to `redirect()`.

The identical bug was fixed in the actix integration by #4755; this
applies the same fix to axum: match the result, log a warning, and return
early instead of panicking.

### DIFF
--- a/integrations/axum/src/lib.rs
+++ b/integrations/axum/src/lib.rs
@@ -228,12 +228,22 @@ pub fn redirect(path: &str) {
     if let (Some(req), Some(res)) =
         (use_context::<Parts>(), use_context::<ResponseOptions>())
     {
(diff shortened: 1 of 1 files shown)


gbj:
Duplicate of #4739 (see [2c85b74](https://github.com/leptos-rs/leptos/pull/4739/commits/2c85b74e8a9021f67f3f31f2afe6f25f0cdc609c))

PR #4799, opened 2026-06-18
fix(axum): don't panic on requests without path-and-query

Noethix55555:
`handle_response_inner` (and every `render_app_*` helper built on it) is
documented to be used as an Axum `.fallback()`, which means it receives
every unmatched request regardless of method. For requests with an
asterisk-form URI (`OPTIONS *`) or authority-form URI (`CONNECT
host:port`), `req.uri().path_and_query()` returns `None`; calling
`.unwrap()` then panics the Tokio worker and takes down the server.

Fix: fall back to `/` when no path-and-query component is present.


gbj:
Duplicate of #4739 (b5dcd3c17b01f995955a18324bdf35458d8a1b25)

PR #4800, opened 2026-06-18
fix(router): decode query params exactly once

Noethix55555:
`url::Url::query_pairs()` (SSR) and `web_sys::UrlSearchParams` (browser) both return already percent-decoded key/value pairs. `ParamsMap::insert()` then called `Url::unescape()` on them again, so any value containing a literal `%` was decoded twice and corrupted — e.g. `?x=%2525` yielded `%` instead of `%25`.

**Fix:** add `ParamsMap::insert_raw()` (no decode) and use it in both query-param collection sites (`server.rs` and `history.rs`). Path params continue to use `insert()`, as they are extracted raw from the path string and need exactly one decode.

Two regression tests added in `location/server.rs` (SSR, `#[cfg(feature = "ssr")]`).

### DIFF
--- a/router/src/location/history.rs
+++ b/router/src/location/history.rs
@@ -276,19 +276,18 @@ impl LocationProvider for BrowserUrl {
 fn search_params_from_web_url(
     params: &web_sys::UrlSearchParams,
 ) -> Result<ParamsMap, JsValue> {
-    try_iter(params)?
-        .into_iter()
-        .flatten()
-        .map(|pair| {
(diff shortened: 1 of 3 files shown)


gbj:
Duplicate of #4731 (1cf445322d552cd0b605a54d24c0b11e21a62689)

I don't think you're intentionally back-porting these commits to `main` from `leptos_0.9`, right? You're just accidentally duplicating the same work.

PR #4801, opened 2026-06-18
fix(reactive_graph): don't drop mark_dirty during ArcAsyncDerived notify window

Noethix55555:
## Summary

`notify_subs` temporarily sets `state = Notifying` while iterating downstream subscribers. `mark_dirty` (called when an upstream source changes) skipped both the `Dirty` assignment and the channel `notify()` call whenever it observed `Notifying`. If a source changed during this window, the update was silently lost: `notify_subs` then restored `prev_state` (Clean), and the derived never re-ran.

- `mark_dirty` now always sets `Dirty` and calls `notifier.notify()`, regardless of `Notifying`.
- `notify_subs` now preserves `Dirty` at the end of the window instead of blindly restoring the pre-window state.

The bug is single-threaded: a downstream subscriber whose `mark_dirty` mutates an upstream signal re-enters `mark_dirty` on the derived while `state == Notifying`. The regression test demonstrates this deterministically — it fails on unpatched code (derived stays at 2) and passes after the fix (derived re-runs to 3).

## Test plan

- [ ] `cargo test -p reactive_graph` passes (all 8 `async_derived` integration tests + 64 doc tests)
- [ ] New test `arc_async_derived_notify_window_not_dropped` fails on unpatched code, passes after fix

### DIFF
--- a/reactive_graph/src/computed/async_derived/arc_async_derived.rs
+++ b/reactive_graph/src/computed/async_derived/arc_async_derived.rs
@@ -451,10 +451,14 @@ impl<T: 'static> ArcAsyncDerived<T> {
             waker.wake();
         }
 
-        // if this was marked dirty before notifications began, this means it
-        // had been notified while loading; marking it clean will cause it not to
-        // run on the next tick of the async loop, so here it should be left dirty
-        inner.write().or_poisoned().state = prev_state;
+        // Restore the state that existed before we entered the Notifying window.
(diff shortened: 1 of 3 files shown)


gbj:
I'm not confident that I understand when the supposed bug would arise. Can you provide some *real* example code that demonstrates how it would happen? (i.e., using existing reactive primitives)

PR #4802, opened 2026-06-18
fix(reactive_stores_macro): use keyed-field path in Patch derive path_at_key closure

Noethix55555:
## Summary

- The `Patch` derive for a struct containing a `#[store(key: ...)]` field generated a `path_at_key` closure that passed the **parent struct path** (not the keyed field's path) to `with_field_keys`, and used `|| vec![]` as the initializer.
- Because the path was wrong, `with_field_keys` always returned `None`, so `patch_field_keyed` never invoked the per-item `notify` callback — silently dropping all granular sub-field notifications when `store.patch(new)` was called on a struct with a keyed field.
- Fix: snapshot `new_path` as `field_path` (parent path + field index) before the mutable borrow of the collection, capture `initial_keys` from the collection beforehand, and pass both into the closure. The `fun` callback now also returns the `(idx, segment)` pair so the reverse-index map (`KeyMap.1`) is populated for `get_trigger_unkeyed`.

## Test plan

- [ ] Added `struct_patch_derive_keyed_field_granular_notify` in `reactive_stores/src/keyed.rs`: patches a `Store`-derived struct with a `#[store(key: ...)]` field at the struct level and asserts that the per-key observer of the changed item fires while the unchanged sibling's observer does not.
- [ ] Test fails on the unpatched code and passes after the fix.
- [ ] All 22 existing `reactive_stores` unit tests and 8 doc-tests continue to pass.

### DIFF
--- a/reactive_stores/src/keyed.rs
+++ b/reactive_stores/src/keyed.rs
@@ -1411,6 +1411,67 @@ mod tests {
         }
     }
 
+    // Regression: `Patch` derive for a struct with a `#[store(key: ...)]` field used the parent
+    // struct path (and an empty initializer) in the generated `path_at_key` closure instead of the
+    // keyed-field's own path.  This caused `store.patch(new)` to resolve every per-item path
+    // against the wrong `KeyMap` bucket, making `with_field_keys` return `None` for every key,
+    // so `patch_field_keyed` never called per-item `notify` — silently dropping all granular
+    // sub-field notifications and leaving observers of individual keyed items unnotified.
+    #[tokio::test]
+    async fn struct_patch_derive_keyed_field_granular_notify() {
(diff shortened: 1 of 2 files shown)


gbj:
The regression test that's introduced looks good to me. However, when I test locally the changes seem to cause a failure in an unrelated test:
```
thread 'writing_one_variant_field_does_not_wake_sibling' (17393741) panicked at reactive_stores/tests/enum_subfield.rs:75:5:
assertion `left == right` failed: name effect woke on an age-only write (path collision)
  left: 2
 right: 1
 ```

gbj:
... Never mind that was an unrelated failing local test case I hadn't noticed I'd added 🤦

gbj:
This now has merge conflicts. If you resolve them I'll be happy to merge it.

PR #4803, opened 2026-06-18
fix(leptos_macro): localize view! diagnostics and recover incomplete component tags

sabify:
Improves error reporting and editor support for the `view!`/`template!` macros.

### Error locality
|                                                          | Before                                                                          | After                                                                                |
| -------------------------------------------------------- | ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------ |
| **Unknown event modifier** (`on:click:typo`)             | Reported at `Span::call_site()` → underlines the whole `view!`; aborts the rest | Points at the event attribute; drops the bad modifier and keeps checking the rest    |
| **Missing required prop** (`<Greeting/>` without `name`) | `.build()` had a call-site span → error blamed the whole `view!`                | `.build()` spanned to the component → underlines that specific `<Component/>`        |
| **Multiple errors**                                      | First `abort!` hid the rest; duplicates/cascades unbounded                      | All errors collected, de-duplicated, and capped (one summary line for the remainder) |

### Editor prop completion
**Before:** typing `view! { <Counter ` leaves the tag unclosed, so the parser drops the element and the macro expands to `()` — no props builder, no completions.

**After:** `view::parse_nodes` (debug builds only) retries with a synthetic `/>` and keeps it **only if it strictly reduces parse errors without losing nodes**. A valid view parses with zero errors, so it's never touched; an incomplete one expands into its props builder, letting rust-analyzer complete the component's props — whether the half-typed tag is alone or beside already-written ones.

### Under the hood
- New `view/diagnostics.rs`: errors recorded as spanned values, emitted from a single sink; de-dup + cap; unit-testable.
- Every `emit_error!`/`abort!` site in view lowering migrated to it; `Span::call_site()` now used in exactly one place (the cap summary).
- New tests for locality, output bounding, and recovery; `proc-macro2` `span-locations` enabled for the test build.

### Notes
- Recovery is **debug-only**, so release/wasm output is unchanged.
- Out of scope (follow-up): friendlier "unknown prop" message — currently surfaces as rustc's "no method named `…`" because the `view!` macro can't see a component's prop list at expansion time.

@lpotthast Could you please let me know if this covers #4628 and #4630.

### DIFF
--- a/leptos_macro/Cargo.toml
+++ b/leptos_macro/Cargo.toml
@@ -33,6 +33,11 @@ uuid = { features = ["v4"], workspace = true, default-features = true }
 tracing = { optional = true, workspace = true, default-features = true }
 
 [dev-dependencies]
+# `span-locations` lets the diagnostics locality tests map a recorded span back
+# to its line/column in the source string.
+proc-macro2 = { workspace = true, default-features = true, features = [
+  "span-locations",
+] }
 log = { workspace = true, default-features = true }
 typed-builder = { workspace = true, default-features = true }
 trybuild = { workspace = true, default-features = true }
--- a/leptos_macro/src/lib.rs
+++ b/leptos_macro/src/lib.rs
@@ -324,16 +324,17 @@ fn view_macro_impl(tokens: TokenStream, template: bool) -> TokenStream {
             .chain(tokens)
             .collect()
     };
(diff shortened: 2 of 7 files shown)


sabify:
Failed workflows are not related to this PR.

gbj:
Tried it out. This is quite nice! Thanks very much.

Issue #3786, opened 2025-03-25
Built-in CSRF Protection in server_fn


double-daruma:
**Is your feature request related to a problem? Please describe.**
Currently leptos (or dioxus) applications built with `server_fn` are by default vulnerable to CSRF-Attacks.

**Describe the solution you'd like**
make `server_fn` CSRF protected by default. Since `server_fn` controls both the client and server code this should be doable

**Describe alternatives you've considered**
document how to manually provide CSRF protection in the docs or at least point out that this is a responsibility left to the developer of each leptos app.

ChosunOne:
Vulnerability to CSRF attacks is going to depend on how you mount your leptos routes and your authentication scheme.  If you are using something like `AsyncAuthorizeLayer` in `axum` with a `Bearer` header on each request, then your leptos routes will be fine against CSRF attacks provided you store your access token securely.

double-daruma:
Thank you for your response. Do you think it would be possible to include CSRF protection by default? e.g. when the user does not configure anything include a `X-LEPTOS-CSRF-PROTECTION=1` header and check for that on the server?

ChosunOne:
Leptos doesn't interact with authentication or CSRF protection directly.  I think the right mental model is that you should secure the routes generated by leptos like you would any other framework.  For CSRF in particular, you can't just automatically protect against it because if you deploy your leptos application such that multiple backend servers are responding to `#[server]` functions, you need some backend infrastructure such as a database or key-value store to properly handle CSRF protection since the token is going to be unique per request.  That isn't something a library can do for you, and trying to just store anti-CSRF tokens in memory would probably end up causing more trouble than it is worth.

double-daruma:
Thank you for taking the time to look at this issue!

It makes sense to me that leptos cannot handle all kinds of CSRF protection. But i feel like leptos should still be able to implement basic header/CORS based  defense as described by [owasp](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#employing-custom-request-headers-for-ajaxapi)? This would protect routes in the default case and should not conflict with any other mechanisms since server_fns will not be called by non-leptos clients e.g. forms.

ChosunOne:
[...]

PR #4809, opened 2026-06-23
docs(server_fn): warn that GetUrl/GetCbor can defeat SameSite=Lax CSRF protection

iaohkut-from-NightWolf-Team:
## Summary

Server functions perform no `Origin`/`Referer`/CSRF validation by design — this is a deliberate, already-discussed decision (#3786), since server functions must stay callable from plain HTML `<form>` submissions with no JS/WASM, which rules out a default scheme based on custom headers/tokens.

The default `PostUrl` (and `Json`/`Cbor` POST) encoding is *incidentally* protected from basic cross-site CSRF by modern browsers' `SameSite=Lax` cookie default, since `Lax` blocks cookies on cross-site **POST** navigations. That incidental protection does not extend to the `GetUrl`/`GetCbor` input encodings: `SameSite=Lax` explicitly *allows* cookies on cross-site top-level **GET** navigation (e.g. an attacker page auto-submitting `<form method="GET" action="https://your-app/api/...">`). So a `GetUrl`-encoded function that mutates state can be triggered cross-site purely by a logged-in victim visiting an attacker-controlled page, even though an equivalent `PostUrl` function would not be vulnerable to the same attack.

Nothing currently documents this distinction — `GetUrl`/`GetCbor` are listed in the `#[server]` macro docs as ordinary, unqualified encoding options, with no indication that they behave differently from `PostUrl` with respect to CSRF.

## Changes

- `server_fn/src/codec/url.rs`: add a security note to the `GetUrl` struct doc comment.
- `leptos_macro/src/lib.rs`: note the same risk next to the `"GetUrl"`/`"GetCbor"` legacy `encoding` string options, and add a fuller explanation under the `#[server]` macro's "Important Notes" section, recommending `PostUrl`/`Json`/`Cbor` for anything that mutates state.

No code/behavior changes — this is documentation only, verified with `cargo doc -p server_fn -p leptos_macro --no-deps` (no new warnings).

## Test plan
- [x] `cargo doc -p server_fn -p leptos_macro --no-deps` builds cleanly, new doc comments render correctly (checked generated HTML for `GetUrl`)
- [x] No code changes, so no functional regression risk

### DIFF
--- a/leptos_macro/src/lib.rs
+++ b/leptos_macro/src/lib.rs
@@ -857,9 +857,12 @@ pub fn slot(args: proc_macro::TokenStream, s: TokenStream) -> TokenStream {
 /// - `encoding`: (legacy, may be deprecated in future) specifies the encoding, which may be one
 ///   of the following (not case sensitive)
 ///     - `"Url"`: `POST` request with URL-encoded arguments and JSON response
-///     - `"GetUrl"`: `GET` request with URL-encoded arguments and JSON response
+///     - `"GetUrl"`: `GET` request with URL-encoded arguments and JSON response (see the
+///       CSRF warning under "Important Notes" below before using this for a function
+///       that mutates state)
 ///     - `"Cbor"`: `POST` request with CBOR-encoded arguments and response
-///     - `"GetCbor"`: `GET` request with URL-encoded arguments and CBOR response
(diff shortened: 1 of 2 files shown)


gbj:
I'm fine with including a warning to this effect. I don't think splicing it into the middle of a list of the basics about how they work is the right place for it.

Perhaps a separate sub-heading at the bottom of this section of docs?

Issue #4818, opened 2026-06-28
`<textarea>` hydration errors in release "0.8.19" (version 0.8.20)


adoyle0:
**Describe the bug**
I've noticed new hydration errors with `textarea`s that have child/default content that appears to be [caused by this commit](https://github.com/leptos-rs/leptos/commit/9417ed441e56a0c3216f8c62e0bfce78e274748e).

Cloning main and changing `true` back to `false` on line 406 makes the errors go away

**Leptos Dependencies**

```toml
leptos = { version = "0.8.20" }
leptos_axum = { version = "0.8.10", optional = true }
leptos_meta = { version = "0.8.6" }
leptos_router = { version = "0.8.14" }
```

**To Reproduce**
1. Use [start-axum](https://github.com/leptos-rs/start-axum) template or probably any app that uses hydration and pulls version 0.8.20
2. Put `<textarea>"hey"</textarea>` somewhere
3. Observe hydration errors as well as "<!>" added to the `textarea`'s default value

**Screenshots**

<img width="342" height="145" alt="Image" src="https://github.com/user-attachments/assets/2f5da359-4934-43d0-87ae-7b6962588807" />

<img width="966" height="354" alt="Image" src="https://github.com/user-attachments/assets/db2a9170-7935-4e62-965c-639edcb59271" />

**Next Steps**

- [ ] I will make a PR
- [x] I would like to make a PR, but need help getting started
- [x] I want someone else to take the time to fix this
- [ ] This is a low priority for me and is just shared for your information

**Additional context**
It seems that the hydration markers clash with trying to escape the "child" content of the `textarea`

gbj:
Thanks, see #4819.

PR #4821, opened 2026-06-28
fix(tachys): decouple child escaping from hydration for raw-text elements

sabify:
Follow-up: #4819, #4730

`ElementType::ESCAPE_CHILDREN` conflated two independent concerns: whether child text is HTML-escaped during SSR, and whether children are hydrated on the client. Raw-text elements (`<script>`, `<style>`, `<noscript>`) and escapable-raw-text elements (`<textarea>`, `<title>`) all shared `false`, so escaping a `<textarea>` to close the `</textarea>` breakout XSS necessarily turned its hydration on too — and walking a textarea's raw value as cursor-walkable child nodes desyncs the hydration cursor (hydration markers are swallowed as literal text inside RCDATA/RAWTEXT), causing hydration errors.

Split the axis by HTML content model, uniformly (no per-tag special cases in the engine):

- Add `ElementType::HYDRATES_CHILDREN`, defaulted to `ESCAPE_CHILDREN` so svg/mathml/custom/external impls are unchanged, and gate hydration on it.
- `<textarea>`/`<title>` are escapable raw text -> escaped but not hydrated; `<script>`/`<style>`/`<noscript>` are raw text -> verbatim and not hydrated.
- Raw text cannot be HTML-escaped (entities are not decoded), so add a general, tag-parameterized debug-only guard that catches `</tag` end-tag breakout (plus `<!--`/`<script` for `<script>`). Release builds are unaffected.

This fixes the `<textarea>` XSS without the hydration regression and closes the same latent issue on `<title>`, while giving the raw-text elements the only general, spec-honest defense a serializer can offer.

### DIFF
--- a/tachys/src/html/element/mod.rs
+++ b/tachys/src/html/element/mod.rs
@@ -285,18 +285,6 @@ pub trait ElementType: Send + 'static {
     /// entity escaping.
     const ESCAPE_CHILDREN: bool;
     /// Whether the element's children should be hydrated (walked as DOM nodes on the client).
-    ///
-    /// This is `true` for normal elements, but `false` for raw-text and escapable-raw-text
-    /// elements — `<script>`, `<style>`, `<noscript>`, `<textarea>`, `<title>` — whose textual
-    /// content the HTML parser reconstructs as the element's *value*, not as cursor-walkable child
-    /// nodes. Hydration markers (`<!>` separators, the empty-text space, branch markers) are only
-    /// parsed as nodes in normal "data" context; inside raw-text/escapable-raw-text content they are
-    /// swallowed as literal text, so walking such children desyncs the hydration cursor.
-    ///
-    /// Defaults to [`ESCAPE_CHILDREN`](Self::ESCAPE_CHILDREN). Override it to decouple escaping from
-    /// hydration: e.g. `<textarea>`/`<title>` are escapable raw text, so their children *are*
-    /// HTML-escaped (`ESCAPE_CHILDREN = true`) yet must *not* be hydrated (`HYDRATES_CHILDREN =
-    /// false`).
     const HYDRATES_CHILDREN: bool = Self::ESCAPE_CHILDREN;
     /// The element's namespace, if it is not HTML.
     const NAMESPACE: Option<&'static str>;
@@ -455,10 +443,6 @@ where
             } else if Ch::EXISTS {
                 // children
                 *position = Position::FirstChild;
-                // `ESCAPE_CHILDREN` and `HYDRATES_CHILDREN` are independent: escapable raw-text
-                // elements (`<textarea>`/`<title>`) escape their children but are not hydrated, so
-                // their children must not emit hydration markers (which would surface as literal
-                // text in the raw-text content).
                 self.children.to_html_with_buf(
                     buf,
                     position,
--- a/tachys/src/html/element/mod.rs
+++ b/tachys/src/html/element/mod.rs
@@ -284,6 +284,20 @@ pub trait ElementType: Send + 'static {
     /// like `<style>` and `<script>`, which include other languages that should not use HTML
     /// entity escaping.
     const ESCAPE_CHILDREN: bool;
+    /// Whether the element's children should be hydrated (walked as DOM nodes on the client).
+    ///
+    /// This is `true` for normal elements, but `false` for raw-text and escapable-raw-text
+    /// elements — `<script>`, `<style>`, `<noscript>`, `<textarea>`, `<title>` — whose textual
+    /// content the HTML parser reconstructs as the element's *value*, not as cursor-walkable child
+    /// nodes. Hydration markers (`<!>` separators, the empty-text space, branch markers) are only
+    /// parsed as nodes in normal "data" context; inside raw-text/escapable-raw-text content they are
+    /// swallowed as literal text, so walking such children desyncs the hydration cursor.
+    ///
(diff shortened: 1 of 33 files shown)


gbj:
This solution yields incorrect textarea content, because "is this element being escaped?" is used as a heuristic for "can this element contain HTML children like comment nodes?"

Example, from #4818; no longer panics (because it no longer hydrates) but contains incorrect data.
```rust
#[component]
    view! {
        <textarea>"hey"</textarea>
    }
}
```

Attempting an example with an XSS attempt like this panics at runtime:
```rust
#[component]
    view! {
        <textarea>"hey"</textarea>

        <textarea>"</textarea><script>console.log('log from textarea')"</textarea>
        <style>"</style><script>console.log('log from style')"</style>
        <script>"</script><script>console.log('script')"</script>
    }
}
```

```
thread 'main' (9070092) panicked at leptos/tachys/src/html/element/mod.rs:422:5:
untrusted content appears to break out of `<style>`: its rendered children contain the `</style` end-tag delimiter (or, for `<script>`, a `<!--`/`<script` sequence). Raw-text elements emit children verbatim with no escaping, so dynamic data must be encoded in the embedded language (JS/CSS) before being rendered into them, like `inner_html`.
```

But surely a runtime panic isn't ideal here? Presumably this is intended to be seen by the developer during development — but it will only appear if they're injecting XSS attempts into their own templates to test? If they're doing that they probably are handling it correctly already.

sabify:
I just pushed a commit to address the decoupling correctly by providing an architectural refactor to respect both the escape and hydrate flags. Most of the changes are refactors only to respect the added hydration flag alongside providing a separate struct to aggregate all the flags, instead of adding a new boolean value.

Regarding the debug runtime panic, you are right. It is the developer's responsibility to know where they should escape raw strings. I just removed it.

tachys/src/html/element/mod.rs
```diff
@@ -284,6 +284,20 @@ pub trait ElementType: Send + 'static {
     /// like `<style>` and `<script>`, which include other languages that should not use HTML
     /// entity escaping.
     const ESCAPE_CHILDREN: bool;
+    /// Whether the element's children should be hydrated (walked as DOM nodes on the client).
+    ///
+    /// This is `true` for normal elements, but `false` for raw-text and escapable-raw-text
```
gbj: This comment is very verbose; it probably only needs the first sentence, and maybe *one* more that says which elements it skips.

tachys/src/html/element/mod.rs
```diff
@@ -442,11 +455,16 @@ where
             } else if Ch::EXISTS {
                 // children
                 *position = Position::FirstChild;
+                // `ESCAPE_CHILDREN` and `HYDRATES_CHILDREN` are independent: escapable raw-text
```
gbj: I don't think the comment adds any value here over the code

tachys/src/view/mod.rs
```diff
@@ -120,6 +120,75 @@ impl MarkBranch for StreamBuilder {
     }
 }
 
+/// Flags that control how a view is serialized to HTML.
+///
+/// These are threaded through [`RenderHtml::to_html_with_buf`] and the streaming variants. The
```
gbj, line 125: Most of these last two paragraphs of comments can be removed

tachys/src/view/mod.rs
```diff
@@ -120,6 +120,75 @@ impl MarkBranch for StreamBuilder {
...
+///
+/// This type is `#[non_exhaustive]`: build it with [`RenderFlags::new`] or the presets and adjust
+/// it with the `with_*` methods, and thread it opaquely through child calls. New rendering axes can
+/// then be added as fields without breaking existing [`RenderHtml`] implementors.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct RenderFlags {
+    /// Whether text content should be HTML-escaped.
+    pub escape: bool,
+    /// Whether hydration position/boundary markers should be emitted: the `<!>` separators between
```
gbj: Comment can end after first sentence.

tachys/src/view/mod.rs
```diff
@@ -265,20 +329,25 @@ where
...
-            true,
+            RenderFlags::HYDRATE_BRANCHING,
             vec![],
         );
         builder.finish()
     }
 
     /// Renders a view to HTML, writing it into the given buffer.
+    ///
+    /// `escape` controls whether text content is HTML-escaped. `hydrate` controls whether
```
gbj: Stale comment about `escape` and `hydrate`, neither of which exist any more. Can omit this whole added comment.

gbj (commented):
Thanks! Yes, this looks much better and passes the textarea test case. (And prevents XSS in textareas, although it still allows it in style/script as a result of them not being escaped.)

I left some comments about comments. I have noticed a tendency both to a) verbose comments that repeat the same content over and over, and b) stale comments (because comments aren't tested!)

These two things compound each other, of course.

tachys/src/view/mod.rs
sabify: Done.

tachys/src/view/mod.rs
sabify: Done.

tachys/src/html/element/mod.rs
sabify: Done.

tachys/src/html/element/mod.rs
sabify: Done.

tachys/src/view/mod.rs
sabify: Done.

gbj:
Thanks — this looks great!

PR #4823, opened 2026-06-29
fix: bump wasm-streams for Cloudflare Workers compatibility

Sushi-Mampfer:
This bumps `wasm-streams` to [0.6.0](https://github.com/MattiasBuelens/wasm-streams/releases/tag/v0.6.0).
I do not believe this will cause any changes to users, unless it's the first dependency for some users to require the 2024 edition.

Cloudflare workers depends on 0.6.0 and this does not seem to be compatible with 0.5.0.

To compile the latest template you also need to remove `strip = true` from `Cargo.toml`.

Error when using 0.5.0:
```
PS C:\Users\sushi\privat\rust\editoffice\web-app>  worker-build --release --features ssr
[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
   Compiling web-app v0.1.0 (C:\Users\sushi\privat\rust\editoffice\web-app)
warning: Linking globals named '__wbg_intounderlyingbytesource_free': symbol multiply defined!

error: failed to load bitcode of module "wasm_streams-dc159e6efc4d8fe5.wasm_streams.5fa0d6c7df9a5116-cgu.0.rcgu.o":

warning: `web-app` (lib) generated 1 warning
error: could not compile `web-app` (lib) due to 1 previous error; 1 warning emitted
Error: Compiling your crate to WebAssembly failed

Caused by:
    failed to execute `cargo build`: exited with exit code: 101
      full command: "cargo" "build" "--lib" "--release" "--target" "wasm32-unknown-unknown" "--features" "ssr"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3137,7 +3137,7 @@ dependencies = [
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
- "wasm-streams",
+ "wasm-streams 0.5.0",
  "web-sys",
 ]
 
@@ -3606,7 +3606,7 @@ dependencies = [
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
- "wasm-streams",
+ "wasm-streams 0.6.0",
  "web-sys",
  "xxhash-rust",
 ]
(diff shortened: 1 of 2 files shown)


gbj:
Off the top of my head I think this is a breaking change (bumps major version of a dependency that's used in the public API) but I've been wrong many times before. If so then I'm happy to merge it into the `leptos_0.9` branch -- that has already had an alpha release and will be moving toward some kind of beta and final release over the course of the summer so it shouldn't be too long to wait.

Sushi-Mampfer:
Running `cargo msrv find` on the leptos repo also returns 1.88.0, which is also the currently configured version in Cargo.toml.

Edit: I don't know enough about rust to say with 100% certainty that this isn't a breaking change, but since you couldn't even compile leptos with a lower version before I don't see why. The only thing I see that could be a problem is the edition.

jakubadamw:
It would be great to get this in. leptos still depending on `wasm-streams` 0.5 forces downstream users to pin `worker` to 0.8.3, essentially.

gbj:
For the sake of being conservative about semver while bumping a dependency like this I'm going to release it as part of 0.9 (which is in beta now anyway). I tried to retarget this against the `leptos_0.9` branch and there's a `Cargo.lock` conflict, so for the sake of not having to deal with that I'm just going to bump it in that branch separately. 

Note: `reqwest` still uses `wasm-streams 0.5` so bumping to 0.6 means there are now two versions in the tree. We only use `reqwest` to be able to support non-web clients for server fns, so that likely doesn't cause any issues for the Cloudflare use case described.

PR #4824, opened 2026-06-30
fix(static): pair cache-hit body and headers by file identity

sabify:
Follow-up: https://github.com/leptos-rs/leptos/issues/4772#issuecomment-4827067893
Closes: #4772

This PR also consists of refactoring of shared logic between two integrations into `utils` (diff is larger).

On a static-route cache hit, the file body and its cached headers were two unsynchronized reads — the file was opened up front, the headers fetched separately, with no lock tying them to the same render. During regeneration (ISR, or the tail of a first-render burst), a request could serve one render's body with another render's headers and status.

## Fix

Key cached headers by the identity (`FileId`) of the file they were captured with, and resolve them on a hit using the identity of the exact handle being served — so body and headers always come from the same render, with no lock spanning the file open and the header lookup.

- Atomic writes split into **stage + commit**; the header snapshot is recorded under the new file's `FileId` *before* the file becomes visible.
- The header cache keeps the last few renders per path (current file + the one it replaced). Capacity and retained generations are configurable via `LEPTOS_STATIC_HEADERS_CACHE_SIZE` / `LEPTOS_STATIC_HEADERS_GENERATIONS`. A file older than the retained renders falls back to default headers, same as cache eviction.
- actix takes the identity from `NamedFile` metadata; axum opens and streams the file directly, since tower-http `ServeFile` hides its handle.
- Since axum no longer uses `ServeFile`, conditional requests are re-added on the cache-hit path: `ETag` (mtime + size) and `Last-Modified` on 200s, plus `If-None-Match` (incl. `*`) / `If-Modified-Since` → `304`. actix keeps this via `NamedFile`. Range requests are not reinstated (irrelevant for rendered HTML).


sabify:
The failed workflows are not related to this PR.

PR #4825, opened 2026-07-01
feat(leptos): always expose use_nonce()

Ofenhed:
I'm building a static site, but during testing I'm using `axum`. The static site does not need nonce, but `leptos_axum` enables it by default. This means that my app (which does not care about nonce) _should_ sometimes include nonce and sometimes not. This creates a small amount of friction for me, but I would imagine it to be more friction if I was developing a library. For the sake of this PR, assume that I'm writing a library.

In the best of worlds, I could guard nonce usage with `#[cfg(feature = "leptos/nonce")]`, but this is not supported by Rust. Alternatively I could add a feature `nonce = ["leptos/nonce"]`, but I don't like this method, because it will never detect that `leptos/nonce` is active, just enforce it if the user enables my library's feature. I did find the `OrDefaultNonce` trait, but it's internal to `meta`, and it's not actually detecting whether leptos has nonce support.

This PR makes it so that `use_nonce` and `Nonce` are available regardless of the `nonce` feature flag, but the `Nonce` struct can only be constructed if the `nonce` feature is enabled, and `use_nonce` always returns `None` if it's disabled. The dependencies of the `nonce` feature are still only enabled if the actual feature is enabled.

### DIFF
--- a/leptos/src/nonce.rs
+++ b/leptos/src/nonce.rs
@@ -1,10 +1,3 @@
-use crate::context::{provide_context, use_context};
-use base64::{
-    alphabet,
-    engine::{self, general_purpose},
-    Engine,
-};
-use rand::{rng, RngCore};
 use std::{fmt::Display, ops::Deref, sync::Arc};
 use tachys::html::attribute::AttributeValue;
 
@@ -49,7 +42,12 @@ use tachys::html::attribute::AttributeValue;
 /// }
 /// ```
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg(feature = "nonce")]
 pub struct Nonce(pub(crate) Arc<str>);
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg(not(feature = "nonce"))]
+pub struct Nonce(pub(crate) Arc<str>, std::convert::Infallible);
 
 impl Nonce {
     /// Returns a reference to the inner reference-counted string slice representing the nonce.
@@ -127,6 +125,9 @@ impl AttributeValue for Nonce {
 /// server response. This can be added to inline `<script>` and
 /// `<style>` tags for compatibility with a Content Security Policy.
 ///
+/// This function can be called without the `nonce` feature enabled,
+/// in which case it will always return [`None::<Nonce>`].
+///
 /// ```rust,ignore
 /// #[component]
 /// pub fn App() -> impl IntoView {
@@ -156,21 +157,36 @@ impl AttributeValue for Nonce {
 ///     }
 /// }
 /// ```
+#[inline(always)]
 pub fn use_nonce() -> Option<Nonce> {
-    use_context::<Nonce>()
+    #[cfg(feature = "nonce")]
+    return crate::context::use_context::<Nonce>();
+    #[cfg(not(feature = "nonce"))]
+    return None;
 }
 
 /// Generates a nonce and provides it via context.
+#[cfg(feature = "nonce")]
 pub fn provide_nonce() {
-    provide_context(Nonce::new())
+    crate::context::provide_context(Nonce::new())
 }
 
-const NONCE_ENGINE: engine::GeneralPurpose =
-    engine::GeneralPurpose::new(&alphabet::URL_SAFE, general_purpose::NO_PAD);
-
+#[cfg(feature = "nonce")]
 impl Nonce {
     /// Generates a new nonce from 16 bytes (128 bits) of random data.
     pub fn new() -> Self {
+        use base64::{
+            alphabet,
+            engine::{self, general_purpose},
+            Engine as _,
+        };
+        use rand::{rng, RngCore as _};
+        const NONCE_ENGINE: engine::GeneralPurpose =
+            engine::GeneralPurpose::new(
+                &alphabet::URL_SAFE,
+                general_purpose::NO_PAD,
+            );
+
         let mut rng = rng();
         let mut bytes = [0; 16];
         rng.fill_bytes(&mut bytes);
@@ -188,6 +204,7 @@ impl Nonce {
     }
 }
 
+#[cfg(feature = "nonce")]
 impl Default for Nonce {
     fn default() -> Self {
         Self::new()
--- a/leptos/src/nonce.rs
+++ b/leptos/src/nonce.rs
@@ -47,6 +47,7 @@ pub struct Nonce(pub(crate) Arc<str>);
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg(not(feature = "nonce"))]
+#[expect(missing_docs)]
 pub struct Nonce(pub(crate) Arc<str>, std::convert::Infallible);
 
 impl Nonce {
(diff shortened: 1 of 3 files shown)


leptos/src/nonce.rs
```diff
@@ -1,89 +1,129 @@
...
     type AsyncOutput = Self;
-    type State = <Arc<str> as AttributeValue>::State;
+    cfg_select! {
+        feature = "nonce" => {
+            type State = <Arc<str> as AttributeValue>::State;
+        }
+        _ => {
+            type State = <() as AttributeValue>::State;
+        }
+    }
```
Ofenhed: I'm not entirely sure about whether this is a good idea. The intention was to limit the complexity of any dead code not cleaned up. Might this cause hibernation issues? I've not noticed any issues in my testing.

Ofenhed:
I'm heavily leaning into [`cfg_select!`](https://doc.rust-lang.org/core/macro.cfg_select.html), which I just noticed require a very recent Rust version. Since leptos is specified to run on Rust 1.88, I guess I'll update it with to use regular `#[cfg]` tags, or a `macro_rules!` macro to remove some of the duplicate code.

Ofenhed:
Hold off on this, I've got a modification in mind that should significantly reduce the size of this PR.

leptos/src/nonce.rs
```diff
@@ -156,21 +157,36 @@ impl AttributeValue for Nonce {
 ///     }
 /// }
 /// ```
+#[inline(always)]
 pub fn use_nonce() -> Option<Nonce> {
-    use_context::<Nonce>()
+    #[cfg(feature = "nonce")]
+    return crate::context::use_context::<Nonce>();
+    #[cfg(not(feature = "nonce"))]
+    return None;
```
Ofenhed: This may be superflous after the inclusion of `Infallible` in `Nonce`.

Ofenhed: Never mind, I was mentally in Haskell mode. Rust functions have side effects, so the compiler can't just remove `use_context()`.

gbj:
I don't love the complexity added by two different definitions of the same type based on conditional compilation. What about the possibility of exposing the `Nonce` type always (it's just an `Arc<str>`, that's never a problem), but gating the *constructor* behind `#[cfg(feature = "nonce")]`? Then `use_nonce()` is always available, and can be effectively
```rust
cfg!(feature = "nonce").then(use_context)
```

`provide_nonce` could either be gated, or could have its body gated.

Ofenhed:
> I don't love the complexity added by two different definitions of the same type based on conditional compilation. What about the possibility of exposing the `Nonce` type always (it's just an `Arc<str>`, that's never a problem), but gating the _constructor_ behind `#[cfg(feature = "nonce")]`?

Before I answer your point, I want to point out that this comment sounds like it's referring to my first iteration of this (where I replaced the `Arc<str>` with `()`). Since that version added a lot more complexity than the current version, could you just verify that you're talking about the `Nonce(Arc<str>, Infallible)` type?

The idea of using `Infallible` is that it guarantees (at compile time) that `Nonce` is never created anywhere (to avoid future bugs/complexity) without having to gate any code besides constructors, while still allowing the compiler to cheaply remove all source code where `Nonce` is a non-optional argument. ~~Unfortunately, [it does not seem to realize this](https://play.rust-lang.org/?version=nightly&mode=release&edition=2024&gist=81c27ac4015a400ac961625a43e6127a).~~ I think the current implementation is a descriptive way of doing it, ~~but `rustc` not acting as I would have hoped makes it~~ a hill I really don't care that much about.

> Then `use_nonce()` is always available, and can be effectively
> 
> ```rust
> cfg!(feature = "nonce").then(use_context)
> ```

That is a lot more elegant!

> `provide_nonce` could either be gated, or could have its body gated.

I think that `provide_nonce` should be gated (and not a NoOp), so that users will have to explicitly set the feature flag and can make easier assumptions on what the function does. I can't think of a use case for a library to use `provide_nonce`, but if there actually is such a use case then I would instead create a specific non-gated version like `maybe_provide_nonce`.

Ofenhed:
I did [some more experimentation](https://github.com/Ofenhed/never-test), and it seemingly turns out that the `Infallible` type does make sure that the functions are not included in the final binary (even in debug builds, though type signatures still exist). Surprisingly, though, it turns out that it cannot figure out that `Option<MyContainer(Arc<str>, Infallible)>` is uninhabitable, so it is the same size as `Option<MyContainer(Arc<str>)>` even in release builds.

Still, it's your project, I just wanted to add context for an informed decision.

gbj:
> Before I answer your point, I want to point out that this comment sounds like it's referring to my first iteration of this (where I replaced the `Arc<str>` with `()`). Since that version added a lot more complexity than the current version, could you just verify that you're talking about the `Nonce(Arc<str>, Infallible)` type?

I've only read and am only talking about the current version of the PR, which contains this:
```rust
#[cfg(feature = "nonce")]
pub struct Nonce(pub(crate) Arc<str>);

#[derive(Clone, Debug, PartialEq, Eq, Hash)]
#[cfg(not(feature = "nonce"))]
#[expect(missing_docs)]
pub struct Nonce(pub(crate) Arc<str>, std::convert::Infallible);
```

This is what I mean by "two different definitions of the same type based on conditional compilation."
 
> The idea of using `Infallible` is that it guarantees (at compile time) that `Nonce` is never created anywhere (to avoid future bugs/complexity) without having to gate any code besides constructors

Here's what I meant:
```rust
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub struct Nonce(pub(crate) Arc<str>);

impl Nonce {
    #[cfg(feature = "nonce")]
    pub fn new() -> Self { /*...*/ }
}
```

This *also* guarantees that `Nonce` can't be created anywhere but within this crate, which seems reasonable.

> I think that `provide_nonce` should be gated (and not a NoOp)

Agreed.

Ofenhed:
> Here's what I meant:
> 
> ```rust
> #[derive(Clone, Debug, PartialEq, Eq, Hash)]
> pub struct Nonce(pub(crate) Arc<str>);
> 
> impl Nonce {
>     #[cfg(feature = "nonce")]
>     pub fn new() -> Self { /*...*/ }
> }
> ```
> 
> This _also_ guarantees that `Nonce` can't be created anywhere but within this crate, which seems reasonable.

It does, but the compiler seemingly can't prove that. I've updated [my `Infallible` test](https://github.com/Ofenhed/never-test#release-mode) to include a case where there simply is no constructor. I don't know the real affect of this in a normal project, this almost certainly a micro optimization that's not worth putting energy into. I'm mostly discussing this point because I find it interesting, and because I prefer to do pull requests with the smallest possible negative impact.

Ofenhed:
I did a small modification of the `Nonce` type to make it a bit more elegant:
```rust
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub struct Nonce(pub(crate) Arc<str>, NonceAppendix);


#[cfg(feature = "nonce")]
type NonceAppendix = ();
#[cfg(not(feature = "nonce"))]
type NonceAppendix = std::convert::Infallible;
```
 If you still don't like the idea of using the type system in this way, I'd like to refer you to #4828 instead.

gbj:
Thanks for your work on this. I'll close it in favor of #4828, which I've merged.

PR #4828, opened 2026-07-04
feat(leptos): always expose use_nonce() (simplified version of #4825)

Ofenhed:
This is a simpler implementation of #4825, see that PR for more information.


gbj:
Thanks for the PR! Yes, I think this approach is the right one.

PR #4830, opened 2026-07-05
fix: don't panic in the wasm-split prefetch machinery when its lock is unavailable

### LINKED ISSUE #4829: Server worker panics ('already been disposed') from the wasm-split prefetch machinery when a client aborts an SSR stream
## Summary

Under axum SSR with wasm-split lazy routes, ordinary client behavior — aborting an in-flight page load by navigating away or closing the tab — can panic the server's tokio worker with:

```
thread 'tokio-rt-worker' panicked at reactive_graph-0.2.14/src/traits.rs:707:46:
At leptos_integration_utils-0.8.8/src/lib.rs:65:30, you tried to access a reactive value
which was defined at leptos-0.8.19/src/lib.rs:380:24, but it has already been disposed.
```

and, on the writer side:

```
thread 'tokio-rt-worker' panicked at reactive_graph-0.2.14/src/traits.rs:394:39:
At leptos-0.8.19/src/lib.rs:393:22, you tried to access a reactive value
which was defined at leptos-0.8.19/src/lib.rs:380:24, but it has already been disposed.
```

Both are the `PrefetchLazyFn` machinery (an `ArcStoredValue<HashSet<&'static str>>`):

1. **Reader**: `leptos_integration_utils::ExtendResponse::from_app`'s prefetch epilogue calls `prefetches.0.read_value()` (and `manifest.0.read_value()`) after the stream completes.
2. **Writer**: `leptos::prefetch_lazy_fn_on_server` calls `prefetches.0.write_value().insert(id)` when a lazy component renders.

`ArcStoredValue`'s `read_value()`/`write_value()` panic whenever `try_read`/`try_write` on the inner lock fails — the 'disposed' message is misleading; the lock is simply unavailable (held by the other side racing, or poisoned after an aborted response's unwind). Since the panic runs on the response task, it kills that worker's request mid-flight.

## Impact

Observed in production-like e2e runs: bursts where one server instance fails a series of unrelated requests. In a ~600-test Playwright suite we captured 12 identical panics on one server in a 12-minute run; after patching both call sites the panics went to zero with no behavioral change (the prefetch `<link rel=preload>` hints are best-effort and the affected responses are already dead).

## Fix

Convert the three reads and one write to their `try_` variants and skip the hint emission when the lock is unavailable. PR incoming from JohnHarrison/leptos:prefetch-disposed-read (also have the same change based on the v0.8.19 tag if a 0.8.x patch release is on the table — we're carrying that via `[patch.crates-io]` in the meantime).

Possibly related context: the interaction between lazy-route rendering and client aborts is the same neighborhood as #4785.

JohnHarrison:
Fixes #4829.

The `PrefetchLazyFn` set and `WasmSplitManifest` are `ArcStoredValue`s whose `read_value()`/`write_value()` panic (with a misleading "already been disposed" message) whenever the inner lock's `try_read`/`try_write` fails. Both the prefetch epilogue in `leptos_integration_utils::ExtendResponse::from_app` (reader) and `leptos::prefetch_lazy_fn_on_server` (writer) hit this on the server when a client aborts an in-flight SSR stream — the panic runs on the response task and takes the worker's request down with it. Evidence and repro details in the issue (12 identical panics captured in a 12-minute e2e run; zero after this change).

This converts the three reads and one write to their `try_` variants and skips the best-effort `<link rel=preload>` hint emission when the lock isn't available — the affected response is already dead, so nothing is lost. Read guards are scoped so they drop before the subsequent `.await`.

Happy to also target this at `leptos_0.9` and/or a 0.8.x patch release — the identical change based on the `v0.8.19` tag is on `JohnHarrison:prefetch-disposed-read-0.8` (we're carrying it via `[patch.crates-io]` in production meanwhile).

### DIFF
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -62,38 +62,57 @@ pub trait ExtendResponse: Sized {
                 pending.await;
             }
 
-            if !prefetches.0.read_value().is_empty() {
-                use leptos::prelude::*;
+            // try_read_value, not read_value: the prefetch set's lock can be
+            // unavailable here — a late/OOO-streamed component may hold the
+            // write lock via prefetch_lazy_fn_on_server, and an aborted
+            // response can leave it poisoned. read_value() would panic the
+            // server worker with a spurious "already disposed" message. The
+            // prefetch hints are best-effort <link rel=preload> tags, so
+            // skipping them beats crashing the response task.
+            //
+            // Scoped block: the ReadGuards must drop before the await below.
+            {
+                let prefetch_keys = prefetches.0.try_read_value();
+                if prefetch_keys.as_ref().is_some_and(|p| !p.is_empty()) {
+                    use leptos::prelude::*;
 
-                let nonce =
-                    use_nonce().map(|n| n.to_string()).unwrap_or_default();
-                if let Some(manifest) = use_context::<WasmSplitManifest>() {
-                    let (pkg_path, manifest, wasm_split_file) =
-                        &*manifest.0.read_value();
-                    let prefetches = prefetches.0.read_value();
+                    let prefetches =
+                        prefetch_keys.expect("checked is_some above");
+                    let nonce =
+                        use_nonce().map(|n| n.to_string()).unwrap_or_default();
+                    // The ReadGuard holds its own Arc to the value, so it can
+                    // outlive the context handle it was read from.
+                    if let Some(manifest_read) =
+                        use_context::<WasmSplitManifest>()
+                            .and_then(|m| m.0.try_read_value())
+                    {
+                        let (pkg_path, manifest, wasm_split_file) =
+                            &*manifest_read;
 
-                    let all_prefetches = prefetches.iter().flat_map(|key| {
-                        manifest.get(*key).into_iter().flatten()
-                    });
+                        let all_prefetches =
+                            prefetches.iter().flat_map(|key| {
+                                manifest.get(*key).into_iter().flatten()
+                            });
 
-                    for module in all_prefetches {
-                        // to_html() on leptos_meta components registers them with the meta context,
-                        // rather than returning HTML directly
-                        _ = view! {
-                            <Link
-                                rel="preload"
-                                href=format!("{pkg_path}/{module}.wasm")
-                                as_="fetch"
-                                type_="application/wasm"
-                                crossorigin=nonce.clone()
-                            />
+                        for module in all_prefetches {
+                            // to_html() on leptos_meta components registers them with the meta context,
+                            // rather than returning HTML directly
+                            _ = view! {
+                                <Link
+                                    rel="preload"
+                                    href=format!("{pkg_path}/{module}.wasm")
+                                    as_="fetch"
+                                    type_="application/wasm"
+                                    crossorigin=nonce.clone()
+                                />
+                            }
+                            .to_html();
                         }
-                        .to_html();
-                    }
-                    _ = view! {
+                        _ = view! {
                         <Link rel="modulepreload" href=format!("{pkg_path}/{wasm_split_file}") crossorigin=nonce/>
                     }
                     .to_html();
+                    }
                 }
             }
 
--- a/integrations/utils/src/lib.rs
+++ b/integrations/utils/src/lib.rs
@@ -62,57 +62,40 @@ pub trait ExtendResponse: Sized {
                 pending.await;
             }
 
-            // try_read_value, not read_value: the prefetch set's lock can be
-            // unavailable here — a late/OOO-streamed component may hold the
-            // write lock via prefetch_lazy_fn_on_server, and an aborted
-            // response can leave it poisoned. read_value() would panic the
-            // server worker with a spurious "already disposed" message. The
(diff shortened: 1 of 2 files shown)


integrations/utils/src/lib.rs
```diff
@@ -62,38 +62,57 @@ pub trait ExtendResponse: Sized {
...
-                        // rather than returning HTML directly
-                        _ = view! {
-                            <Link
-                                rel="preload"
-                                href=format!("{pkg_path}/{module}.wasm")
-                                as_="fetch"
-                                type_="application/wasm"
-                                crossorigin=nonce.clone()
-                            />
+            // try_read_value, not read_value: the prefetch set's lock can be
```
gbj: please remove this verbose section of the comment
```rust
           // try_read_value, not read_value: the prefetch set's lock can be
            // unavailable here — a late/OOO-streamed component may hold the
            // write lock via prefetch_lazy_fn_on_server, and an aborted
            // response can leave it poisoned. read_value() would panic the
            // server worker with a spurious "already disposed" message. The
            // prefetch hints are best-effort <link rel=preload> tags, so
            // skipping them beats crashing the response task.
            //
```

integrations/utils/src/lib.rs
```diff
@@ -62,38 +62,57 @@ pub trait ExtendResponse: Sized {
...
+            // prefetch hints are best-effort <link rel=preload> tags, so
+            // skipping them beats crashing the response task.
+            //
+            // Scoped block: the ReadGuards must drop before the await below.
+            {
+                let prefetch_keys = prefetches.0.try_read_value();
+                if prefetch_keys.as_ref().is_some_and(|p| !p.is_empty()) {
+                    use leptos::prelude::*;
+
+                    let prefetches =
```
gbj: why use `if` and then `.expect` rather than `if let`?

leptos/src/lib.rs
```diff
@@ -413,7 +413,14 @@ pub fn prefetch_lazy_fn_on_server(id: &'static str) {
     use reactive_graph::traits::WriteValue;
 
     if let Some(prefetches) = use_context::<PrefetchLazyFn>() {
-        prefetches.0.write_value().insert(id);
+        // try_write_value, not write_value: a lazy component can render after
```
gbj: please remove the comment, it's unnecessary and verbose

gbj (commented):
This looks good overall, I left a a couple notes about the comments and about the `.expect()`

JohnHarrison:
Gentle nudge on this one — it's green and mergeable, and it fixes a production panic (the prefetch machinery panicking when its lock is unavailable, leptos#4829) that we're currently carrying as a git pin. It's a small diff and adjacent to the wasm-split work in #4785 if it's easier to look at them together.

gbj:
@JohnHarrison I left comments in a review last week that you have not replied to or addressed in the PR.

integrations/utils/src/lib.rs
JohnHarrison: Done — removed.

integrations/utils/src/lib.rs
JohnHarrison: No good reason — switched to `if let Some(prefetches) = …try_read_value().filter(|p| !p.is_empty())`, which also made the explicit scoping block unnecessary (the guard now drops at the end of the `if` body, still before the `await` below).

leptos/src/lib.rs
JohnHarrison: Done — removed.

JohnHarrison:
@gbj sorry for the confusion earlier — I'd missed that your July 10 review had inline comments. All three are addressed now: both verbose comments removed, and the `is_some_and` + `.expect()` replaced with a plain `if let` (via `Option::filter`), which also let me drop the explicit scoping block. Rebased onto current main; net diff is now just the `try_read_value`/`try_write_value` guards. Verified locally: clippy clean and the `lazy_routes` example under `--split` still emits all the per-chunk preload / modulepreload hints on lazy routes.

JohnHarrison:
One follow-up: giving the final diff a closer look, two small things surfaced that I'd like your call on before you merge.

1. **A `None` from these locks is silent and never actually means disposal.** `ArcStoredValue::is_disposed` is hardcoded `false`, so `try_read_value()`/`try_write_value()` return `None` only on lock contention or poisoning (the guardian maps `Poisoned` to `None` via `.ok()`). That means a render panic that poisons the prefetch lock would now silently suppress all preload hints on later responses with nothing logged — the panic-avoidance we want, but potentially hard to diagnose. Happy to add a `log::debug!`/`tracing` line on the skip path if you'd like the failure to stay visible, or leave it fully silent since the hints are best-effort.

2. **The write site could reuse the existing helper.** `if let Some(mut set) = prefetches.0.try_write_value() { set.insert(id); }` is exactly `prefetches.0.update_value(|set| { set.insert(id); });` (the blanket `UpdateValue` impl goes through the same `try_write_value` and silently no-ops). Happy to switch to the one-liner or keep the explicit `if let` you already reviewed — your preference.

Say the word on either and I'll push; otherwise the PR is good to go as-is from my side.

gbj:
Thanks!

PR #4832, opened 2026-07-05
Fix/islands router branch marker off by one

### LINKED ISSUE #4831: Off by 1 in Island Router TreeWalker
**Describe the bug**
There is a bug in the island router `replaceBranch` function which miscounts nested branch markers if a route's first node is rendered as a marker comment instead of an element. When navigating away from such a route the new routes content is inserted to the DOM but the old content is never removed due to an off by one error. 

**Leptos Dependencies**

Reproduced on main HEAD at the time of writing

**To Reproduce**
1. Using islands router render a component at position 0 inside of a Route (e.g. <Title>)
2. Navigate to a different route
3. Old content is not removed

**Next Steps**

- [x] I will make a PR

**Additional context**

seanaye:
closes #4831 

This PR adds an e2e test which reproduces the issue described in #4831 then fixes the off by 1 error in `replaceBranch`

We also fix a potential infinite loop which could happen if the  doc walker runs out of nodes


gbj:
The fix looks right to me and definitely works. 

I think that in order to actually run the e2e tests in CI, the example's Makefile.toml needs to be updated to tell CI to run the e2e tests. (I could be wrong, but I don't think I see them running in the CI logs on the PR)

Given that the test is being added *just* as a regression test but otherwise isn't part of the example, I wonder whether it would be a better idea to test something that's semantically meaningful as part of the actual example? for example, adding a `<Title/>` component to the Home page, and then navigating away and making sure that the About page shows.

seanaye:
@gbj I think this should run in CI and provide a more meaningful test

gbj:
Thanks very much!

Issue #4780, opened 2026-06-12
`tachys` `Keyed` SSR rendering injects spurious whitespace text nodes when `islands` feature is not enabled — causes hydration mismatch


omega-pw:
Between `tachys` 0.2.11 and 0.2.15, the internal structure of `Keyed` (the underlying type for `For` components) was refactored to add an `ssr_items: Vec<(String, V)>` field. When the `islands` feature is **not** enabled, the key string is initialized to `String::new()` (empty). During SSR (`to_html_with_buf`), the tuple `(String, V)` is rendered, and the empty string key causes a spurious space character (`' '`) to be emitted **before each child view**. This whitespace-only text node in the SSR output has no corresponding node in the client-side hydration DOM walk, resulting in hydration mismatches.

### Affected versions

`tachys` >= 0.2.12 (the version where `ssr_items` was introduced) up to at least 0.2.15.  
`tachys` 0.2.11 is **not** affected because `Keyed` did not have an `ssr_items` field; it rendered items directly via `view_fn` without a key string prefix.

### Root cause

**In 0.2.11** (`tachys/src/view/keyed.rs`), the `Keyed` struct had no `ssr_items` field:

```rust
pub struct Keyed<T, I, K, KF, VF, VFS, V> {
    items: I,
    key_fn: KF,
    view_fn: VF,
}
```

`to_html_with_buf` rendered each item directly via `view_fn`, producing only the actual view content:

```rust
fn to_html_with_buf(self, buf, position, escape, ...) {
    for (index, item) in self.items.into_iter().enumerate() {
        let (_, view) = (self.view_fn)(index, item);
        view.to_html_with_buf(buf, position, escape, ...);
        *position = Position::NextChild;
    }
    buf.push_str("<!>");
}
```

**In 0.2.15**, the struct gained an `ssr_items` field:

```rust
pub struct Keyed<T, I, K, KF, VF, VFS, V> {
    items: Option<I>,
    #[cfg(feature = "ssr")]
    ssr_items: Vec<(String, V)>,   // ← (key_string, view) tuple
    key_fn: KF,
    view_fn: VF,
}
```

It is initialized in the `keyed()` constructor:

```rust
#[cfg(feature = "ssr")]
ssr_items: items
    .into_iter()
    .enumerate()
    .map(|(i, t)| {
        let key = if cfg!(feature = "islands") {
            let key = (key_fn)(&t);
            key.ser_key()
        } else {
[...]

PR #4833, opened 2026-07-08
fix(tachys): Keyed SSR injects spurious whitespace text nodes, causing hydration panics

8codebubble:
The Bug
Server-rendered pages that use <For> components (or any keyed iteration) have invisible whitespace text nodes injected before every child element. When the WASM hydrates the page, it encounters these unexpected text nodes and panics with:

```
panicked at tachys/src/hydration.rs:227:9:
internal error: entered unreachable code
```
This means any Leptos SSR+hydrate app that uses <For> breaks on page load. The panic is unrecoverable — the page never becomes interactive.

Why It Happens
Keyed::to_html_with_buf (the sync SSR rendering path) iterates self.ssr_items as raw (String, V) tuples:

```
for item in self.ssr_items {
    item.to_html_with_buf(...);
}
```
item here is the full tuple (String, V). When the tuple's RenderHtml::to_html_with_buf runs, it renders the String first, then the view V. The string is the item's key. When the islands feature is not enabled, every key is initialized as String::new() (empty). An empty string rendered in SSR escape mode becomes a space character ' ' — this is the whitespace text node.

The async counterpart to_html_async_with_buf already handles this correctly — it destructures for (key, item) and only renders item. The sync method was missed during the refactor that added the ssr_items field in tachys 0.2.12.

```
The Fix
One-line change — match the async method's destructuring:

// Before
for item in self.ssr_items {

// After
for (_, item) in self.ssr_items {
```
Concrete Symptom
With the bug, SSR renders:

```
<div><!> <span>a</span><!> <span>b</span></div>
```
Note the space ( ) after each <!> marker. Without the bug:

```
<div><!><span>a</span><!><span>b</span></div>
```
Any test that hydrates the page will observe the panic immediately.

Verification
Fix verified in an app I have in development (Playwright E2E tests). Before: 2 hydration panics per page load. After: 0 panics, all tests green.

Related
#4780 (closed "not planned" due to needs reproduction — this PR provides the reproduction)

### DIFF
--- a/tachys/src/view/keyed.rs
+++ b/tachys/src/view/keyed.rs
@@ -301,7 +301,7 @@ where
         }
 
         #[cfg(feature = "ssr")]
-        for item in self.ssr_items {
+        for (_, item) in self.ssr_items {
             if mark_branches && escape {
                 buf.open_branch("item");
             }


gbj:
> this PR provides the reproduction

I don't see a reproducible example. What am I missing?

8codebubble:
> > this PR provides the reproduction
> 
> I don't see a reproducible example. What am I missing?

Sorry, lets see. 

it can be visually inspected ofcourse and I'll follow up with some reproducibility 

1. Open tachys/src/view/keyed.rs at line 304
2. The sync rendering loop: for item in self.ssr_items (line 304)
3. item is (String, V) — compiles but renders the empty key string as space
4. The async rendering loop at line 340: for (key, item) in self.ssr_items — correct
5. Apply the fix: change for item to for (_, item) at line 304


 SSR Output Inspection (quick test)
Create a minimal Leptos 0.8 app with one component:

```
#[component]
fn App() -> impl IntoView {
    let items = vec!["a", "b", "c"];
    view! {
        <div>
            <For each=move || items.clone() key=|item| *item children=|item| view! { <span>{item}</span> }/>
        </div>
    }
}
```
Build with SSR + hydrate (no islands feature). The SSR output for this component:
in frontend/Cargo.toml

[dependencies]
app = { path = "../app", features = ["hydrate"] }
leptos = { version = "0.8", features = ["hydrate"] }  # NO islands
console_error_panic_hook = "0.1"

**OR you can run the attached project.** 

[bug-example.zip](https://github.com/user-attachments/files/29818053/bug-example.zip)
cargo run --release

This is a bug in the library's SSR rendering that breaks any Leptos app that uses `<For>` with SSR+hydrate.

gbj:
The example you provided doesn't cause a hydration error if you add the provided component to a fresh, correctly-configured project (`cargo leptos new --git start-axum`). The .zip file you provided doesn't set up hydration at all.

PR #4836, opened 2026-07-13
fix: include submit input value in FormData when submitting Form

dav-wolff:
This fixes the same issue as https://github.com/leptos-rs/leptos/pull/2268 but for regular `Form`s instead of `ActionForm`s.

This uses the same `form_data_from_event` function as the other PR. Since that function is located in the leptos crate and the `Form` component is in the router crate I needed to make that function public. I think this is a good change as this is a useful utility function that people might want to use outside of the context of `ActionForm`s, just like the `FromFormData` trait.  
Alternatively, it would be possible to reimplement the logic of that function in the router crate and keep it private.

I tested the change working with this change to the router example, seeing the expected `submit_name: "submit_value"` as part of the request in the browser tools:  
https://github.com/dav-wolff/leptos/commit/555602ac7580ce6d588f823e945c4aa5fc4bdbef

### DIFF
--- a/leptos/src/form.rs
+++ b/leptos/src/form.rs
@@ -291,7 +291,8 @@ where
     }
 }
 
-fn form_data_from_event(
+/// Retrieves the `FormData` of a form given the `submit` event
+pub fn form_data_from_event(
     ev: &SubmitEvent,
 ) -> Result<FormData, FromFormDataError> {
     let submitter = ev.submitter();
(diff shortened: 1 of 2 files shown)


dav-wolff:
I fixed the issue of `form` not being unused. Since it's not needed anymore `extract_form_attributes` doesn't need to return it in the first place.

I ran `cargo clippy` but all the remaining warnings didn't seem related to my changes. I tried running `cargo make test` but it already failed before my commits so I couldn't get any information from that.

gbj:
Great, thanks for the PR!

PR #4839, opened 2026-07-19
fix(router): keep `""` and `"/"` static segments as pass-through parents

sabify:
123eec61 made StaticSegment::test return None when the segment is exhausted but the path continues with a non-slash byte, so an overlong path no longer matches a shorter segment ("/foobar" vs "fo"). But the check also rejected the `""` and `"/"` wrapper segments that nested parent routes use, so every route behind such a parent stopped matching and fell through to the fallback:

    // parent wrapper, as produced by nested route trees
    assert!(StaticSegment("").test("/lang").is_some());  // was None
    assert!(StaticSegment("/").test("/about").is_some()); // was None

Restrict the overlong check to other segments; `""` and `"/"` are already treated as trivially matching when initializing has_matched.

Fixes matches_nested_route, chooses_between_nested_routes and arbitrary_nested_routes, which fail on the parent branch.

### DIFF
--- a/router/src/matching/horizontal/static_segment.rs
+++ b/router/src/matching/horizontal/static_segment.rs
@@ -365,7 +365,9 @@ mod tests {
     #[test]
     fn empty_segment_is_passthrough_parent() {
         let def = StaticSegment("");
-        let m = def.test("/lang").expect("empty segment should pass through");
+        let m = def
+            .test("/lang")
+            .expect("empty segment should pass through");
         assert_eq!(m.matched(), "");
         assert_eq!(m.remaining(), "/lang");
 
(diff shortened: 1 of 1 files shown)


gbj:
So, this PR does fix the failing tests; thanks.

I'm alarmed by the fact that a commit that causes 3 tests to fail passed CI. When I look at the CI logs, it appears that `nextest` does in fact run, and reports the test failures — and then that part of the run is still marked green? This suggests to me that the CI script is not propagating the exit codes correctly, and test failures are not actually being reported. I think might be due to the changes in #4740.

sabify:
Interesting! That PR only added a retry capability and it's unclear to me why it couldn't catch it. Let's see if your PR can catch it. Please keep in mind that your patch should be provided to both the main and leptos_0.9 branches in order to be run in both branches. (You may want to merge main into leptos_0.9 afterward.)

lukexor:
So I think this issue also still affects the public release of 0.8.20 (tag 0.8.19) and needs back-porting.

gbj:
Was already patched on the main branch but not released; I've released the patched version just now as `leptos_router 0.8.15, let me know if that doesn't fix it.

PR #4840, opened 2026-07-19
chore: make sure failing tests in CI actually cause failures

gbj:
#4839 notes that #4790 introduced a bug that caused 3 tests to fail. But this PR passed CI. When I look at the CI logs, it appears that `nextest` does in fact run, and reports the test failures — and then that part of the run is still marked green!

For example [in this run](https://github.com/leptos-rs/leptos/actions/runs/29112531543/job/86428108378) at the bottom of the `run` step, we have
```
  ────────────
       Summary [   0.161s] 51 tests run: 48 passed, 3 failed, 0 skipped
          FAIL [   0.075s] (49/51) leptos_router matching::tests::arbitrary_nested_routes
          FAIL [   0.088s] (50/51) leptos_router matching::tests::chooses_between_nested_routes
          FAIL [   0.083s] (51/51) leptos_router matching::tests::matches_nested_route
  error: test run failed
  Error while executing command, exit code: 100
  Command completed after 1 attempt(s).
  ```
But the whole step (therefore the whole PR) is marked green.

I think this must be an issue with changes to our CI setup (maybe #4740?) stopping error exit codes from propagating correctly. I'll try to fix this so we see the real failures, then see how many things were uncaught.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - uses: actions-rust-lang/setup-rust-toolchain@v1
-        with: {toolchain: "nightly-2026-02-24", components: "rustfmt, clippy", target: "wasm32-unknown-unknown", rustflags: ""}
+        with: {toolchain: "nightly-2026-07-20", components: "rustfmt, clippy", target: "wasm32-unknown-unknown", rustflags: ""}
       - name: Install Glib
         run: |
           sudo apt-get update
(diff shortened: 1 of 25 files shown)


gbj:
Not as bad as I thought it might be. A few `clippy` warnings that are new to 1.97, one PR that broke some tests and wasn't caught (which @sabify fixed in #4839; thanks), and one regression test added that actually fails, but which is already fixed on 0.9 anyway (textarea escaping).

Could be worse. I can't say I was worried about false positives in CI before, but I am now 😅

sabify:
@gbj, can you also please run `./scripts/update_nightly.sh`? I think we have diverges in the latest nightly clippy and current one. I cannot catch the reported clippy in https://github.com/leptos-rs/leptos/actions/runs/29706404204/job/88244057487?pr=4784 for my #4784 PR with my latest nightly.

PR #4842, opened 2026-07-22
test(integrations): gate static route tests on the `default` feature

sabify:
Related: https://github.com/leptos-rs/leptos/actions/runs/29874655448/job/88782457685?pr=4522
Related: https://github.com/leptos-rs/leptos/actions/runs/29874655448/job/88782457908?pr=4522

The `leptos_axum` and `leptos_actix` test targets fail when the crates are
built without default features, which the feature-combination CI run does.

- `leptos_axum`: with `default` off, `SsrMode::Static` routes are not
  registered — building the router panics with "Static routes are not
  currently supported on WASM32 server targets.", so every static route test
  fails at setup.
- `leptos_actix`: `default` → `actix-default` → `actix-web/default`, so
  turning it off drops `actix-web/macros`. `#[actix_web::test]` no longer
  resolves and the whole test build fails to compile.

### DIFF
--- a/integrations/actix/src/lib.rs
+++ b/integrations/actix/src/lib.rs
@@ -1838,7 +1838,7 @@ where
     .await
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "default"))]
 mod tests {
     // Targeted imports rather than `use super::*`: the crate root glob-imports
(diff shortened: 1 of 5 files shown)


gbj:
Thanks!

PR #4844, opened 2026-07-25
fix(leptos): nonce incorrectly generated for wasm preload

Ofenhed:
`HydrationScripts` incorrectly adds a nonce to the wasm preload if `leptos_integration_utils` is included as a dependency. This PR removes this nonce (and cleans up some conditional compilation).


Ofenhed:
I also removed the dependency from `leptos_integration_utils` to `leptos/nonce`. It's still explicitly requested from `leptos_actix` and `leptos_axum`.

gbj:
Thanks!

Issue #4851, opened 2026-08-04
Random hanging SSR responses for one resource loaded in two suspenses


ikoziak:
**Describe the bug**

When one `Resource` is read in the children of two `<Suspense>`, a small fraction
of SSR responses stop mid-body and the connection is never closed. One `<Suspense>`
resolves and its chunk is flushed; the other never resolves, so no
`__RESOLVED_RESOURCES` trailer is written. The client sees an incomplete chunked
response. Either one can be the one that hangs.

The process stays healthy throughout — it burns no CPU and keeps serving other
requests in milliseconds while a response is stalled. Not a livelock, not a
whole-process deadlock.

Looks like the same fault as #4673, which was closed for lack of a reproduction.

**Leptos Dependencies**

```toml
leptos = "0.8.20"
leptos_axum = { version = "0.8.10", optional = true }

[features]
ssr = ["dep:axum", "dep:tokio", "dep:leptos_axum", "leptos/ssr"]
```

**To Reproduce**

Repro: attached 

[leptos-hangs.zip](https://github.com/user-attachments/files/30711520/leptos-hangs.zip)

1. `./serve.sh` — builds and starts on 127.0.0.1:3100
2. `./hammer.sh 50 20` — 1000 requests, 20 concurrent

Bodies of stalled responses are kept in `run/`. A complete one ends with
`__RESOLVED_RESOURCES[0] = ...` then `__INCOMPLETE_CHUNKS=[];`. A stalled one ends
after one `<Suspense>`'s swap script, with neither.

**Screenshots**

N/A.

**Next Steps**

- [ ] I will make a PR
- [ ] I would like to make a PR, but need help getting started
- [x] I want someone else to take the time to fix this
- [ ] This is a low priority for me and is just shared for your information

**Additional context**

Two things are required — remove either
and the fault stops:

- the same `Resource` read in the children of more than one `<Suspense>`
- a resource future that actually suspends - remove the sleep so the server fn
  never yields

Each of the following was removed with the fault intact: `leptos_router` and nested routes,
`Resource` keyed on a reactive source, `<Transition>` vs plain `<Suspense>`,
`provide_context` across separate components, response size, the `hydrate` feature
and the wasm bundle.

Raising the sleep raises the rate, weakly (debug, AMD Ryzen 7 5800H, 16 cores,
Linux 7.0.12, rustc 1.99.0-nightly):

| sleep | stalls / 1000 |
|---|---|
| 2ms | 4 |
[...]

PR #4698, opened 2026-05-04
fix(leptos/suspense): always track `tasks` signal to prevent SSR stream hang

PlanBernhard:
## Bug

Under load, an SSR response can hang permanently after the initial
`<Suspense>` chunk is written. The client receives HTTP 200 plus the
first bytes of the body, then the stream stalls forever. The hung future
is `tasks_rx.await` inside `to_html_async_with_buf`. The matching
`tasks_tx.send(())` never fires because the Suspense effect has silently
lost its subscription on the `tasks` `ArcRwSignal<SlotMap>`.

**Root cause** (`leptos/src/suspense_component.rs`): the effect calls
`tasks.track()` only when `double_checking` is `None`, then calls
`tasks.try_read_untracked()`. If a concurrent `tasks.update(...)` (e.g.
from `TaskHandle::drop` when a child `Resource` finishes) holds the
write lock at that moment, `try_read_untracked()` returns `None` and the
effect exits **without having tracked any signal**. The effect is now
unsubscribed: subsequent `tasks.update(|t| t.remove(self.key))` calls no
longer mark it dirty, the effect is never re-scheduled, `tasks_rx` never
receives, and the SSR stream hangs forever. Classic lost-wakeup race.

Reproduces with ≥ 2 parallel `Resource::new_blocking` resolvers under
any `<Suspense>` boundary at roughly 3–5 % per request under load.

## Fix

Move `tasks.track()` unconditionally to the top of every effect run.
`tasks.track()` is idempotent for an already-subscribed effect, so the
previous `if double_checking.is_none()` guard was an unnecessary
micro-optimization that opened the race window. With unconditional
tracking the effect retains a live subscription regardless of which
branch it exits through.

Change is 1 line of behavior in `leptos/src/suspense_component.rs`; no
API surface change, no semver impact, safe to backport to `0.8.x`.

## Reproduction

```rust
#[component]
fn Page() -> impl IntoView {
    let a = Resource::new_blocking(|| (), |_| async { fetch_a().await });
    let b = Resource::new_blocking(|| (), |_| async { fetch_b().await });
    view! {
        <Suspense fallback=|| "loading">
            {move || Suspend::new(async move {
                let (_a, _b) = (a.await, b.await);
                view! { <div>"ok"</div> }
            })}
        </Suspense>
    }
}

Issue #4673, opened 2026-04-02
SSR deadlock: RwLock ordering inversion between EffectInner, ArcAsyncDerivedInner, and SuspenseContext signal


nlflint:
**Describe the bug**
I ran into a true deadlock (0% CPU, all threads blocked in futex wait) during SSR when multiple Suspense boundaries interact with shared Resources. This is distinct from the livelock described in #4578 — threads are fully blocked on `RwLock::write_contended`, not spinning.

The problem manifests as all std out halting from my app, and subsequent reloads/requests in a browser just time out after a minute or two with default browser failures; my leptos app hangs. My app is running SSR mode.

**Leptos Dependencies**
```toml
leptos = { version = "0.8.13", features = ["tracing", "nightly"] }
leptos_axum = { version = "0.8.7", optional = true }
leptos_meta = { version = "0.8.5" }
leptos_router = { version = "0.8.10", features = ["nightly"] }
```

**To Reproduce**
Load my website.  very low and sporadic reproducability. So far I attempted, but have failed, to make a minimally reproducible exampled. 

**Screenshots**
Not needed.

**Next Steps**

- [ ] I will make a PR
- [ ] I would like to make a PR, but need help getting started
- [ X] I want someone else to take the time to fix this
- [ ] This is a low priority for me and is just shared for your information

**Additional context**
I was able to extract a core dump from the hung process, and analyze the back traces for the hung threads to figure this much out. I can show the backtraces if you like, but I won't send the core dump as it will probably contains secrets. 

For now, we're going to switch our app to CSR and see how it goes.

gbj:
Feel free to share a reproduction when you have time. It's impossible to take any action otherwise.

gbj:
This sounds like it was very similar to #4851 and, if so, should be fixed by #4853

PR #4853, opened 2026-08-05
Fix two concurrency issues with Suspense

gbj:
This should address two different kinds of issues with parallelism that are (probably) the causes of the issues described in https://github.com/leptos-rs/leptos/issues/4851 https://github.com/leptos-rs/leptos/issues/4673 and https://github.com/leptos-rs/leptos/pull/4698

1. Fixes a synchronization issue between "loading" and "wakers" on resources
2. Fixes a timing issue between reading from and tracking the sets of active tasks in Suspense

See comments on the commits.


ikoziak:
This resolves #4851, my e2e tests do not randomly hang any more.

PR #4856, opened 2026-08-17
fix(reactive_graph): ignore dropped local-resource listener

adrianncovaci:
We've ran into an unwrap panic under streaming SSR: `LocalResourceNotifier::notify()` unwraps the `oneshot` send, but the receiver isn't guaranteed to still be around.

### DIFF
--- a/reactive_graph/src/computed/async_derived/mod.rs
+++ b/reactive_graph/src/computed/async_derived/mod.rs
@@ -130,7 +130,12 @@ pub mod suspense {
         /// Send the notification. If the inner channel has already been used, this does nothing.
         pub fn notify(&mut self) {
             if let Some(tx) = self.0.lock().or_poisoned().take() {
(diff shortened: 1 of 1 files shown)


gbj:
Could you please provide an example that is failing without this change?

adrianncovaci:
> Could you please provide an example that is failing without this change?

Here's a minimal repro: https://github.com/adrianncovaci/leptos-repro-4856

gbj:
Thanks! The change in this PR is innocuous (it can't hurt anything, and I'm happy to merge it) but in the case of your reproduction it's actually papering over a real issue, which is that resource reads that were used 1) in an attribute value and 2) nested within Signal/Memo specifically, were not being tracked correctly.

#4859 should fix the actual issue, and I'm happy to merge this as well to log rather than panicking.

PR #4860, opened 2026-08-25
fix(router): three route-matching panics / silent match failures

sabify:
Fixes three bugs in `leptos_router`'s matching code, one commit each with a regression test.

- **Base stripping** (`RouteDefs::match_route`): `base="/"`, `"/app/"`, or `"app"` stripped the leading `/` the segment matchers rely on, so every route under such a base silently failed to match. The base is now trimmed of slashes and stripped only as a whole path segment; the remainder keeps its leading `/`.
- **UTF-8 boundary panic** (`ParamSegment`, `WildcardSegment`, `OptionalParamSegment`): a path without a leading `/` was scanned from after its first character but split from byte 0, so a multi-byte first character (`"éx"`, `"日本語"`) panicked in `split_at` and ASCII input lost its last character. Such paths now follow `StaticSegment`'s rule (no match; zero-length match for the optional segment), and a `debug_assert!` pins the char-boundary invariant.
- **Optional-segment fallback** (`NestedRoute::match_nested`): the re-parse after a child fallback called `.unwrap()`; `/:x?/bar` with child `bar` panicked on `/bar`. Now returns `None`.

### DIFF
--- a/router/src/matching/horizontal/param_segments.rs
+++ b/router/src/matching/horizontal/param_segments.rs
@@ -42,29 +42,27 @@ impl PossibleRouteMatch for ParamSegment {
     fn test<'a>(&self, path: &'a str) -> Option<PartialPathMatch<'a>> {
         let bytes = path.as_bytes();
         let has_leading_slash = bytes.first() == Some(&b'/');
-        // the first char of the path is always consumed before scanning, but
-        // only counts toward the match when it is the leading `/`
-        let (param_offset, scan_start) = if has_leading_slash {
-            (1, 1)
-        } else {
-            (0, path.chars().next().map(char::len_utf8).unwrap_or(0))
-        };
+        if !has_leading_slash && !path.is_empty() {
+            return None;
+        }
+        let offset = usize::from(has_leading_slash);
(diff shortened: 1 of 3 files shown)


gbj:
[95b1eb3](https://github.com/leptos-rs/leptos/pull/4860/commits/95b1eb394ddeceecf89495c561ec614d2ae6e62a) looks good, thanks
[1c39cc9](https://github.com/leptos-rs/leptos/pull/4860/commits/1c39cc97b9dda20cdc177e1b4986ef5deac39164) looks fine but it's worth noting that the panic already couldn't be reached from the router in reality because `url.path()` is already URL-encoded ASCII, so routes like `/éx` in params were already fine.
[a56553f](https://github.com/leptos-rs/leptos/pull/4860/commits/a56553f711ed3ee5d0efb24e70db160308f24295) Thanks! For a follow up: if we have a definition like this
```rust
<ParentRoute path=path!("/:foo?/bar") view=/* ... */>
	<Route path=path!("bar") view=/* ... */>
</ParentRoute>
```

It won't match `/bar/bar`, which it should in theory (the optional param + second segment, followed by the child). Not introduced by this PR.

PR #4861, opened 2026-08-26
fix(router): static optional params, `<Form>` submission fidelity, initial route build race

sabify:
Fixes five bugs in `leptos_router`:

- **Static generation panicked on optional params** — `StaticPath::into_paths` had a `todo!()` for `OptionalParam`; it now expands to the path without the parameter plus one path per configured value.
- **`<Form>` ignored submit-button overrides** — reads `formmethod`/`formaction`/`formenctype` (was `method`/`action`/`enctype`, which are not valid on buttons).
- **`<Form>` dropped the clicked button from the data** — `FormData` is now built with the submitter, so `name=value` of the activated button is sent like a native submit.
- **`<Form>` turned cross-origin POSTs into CORS `fetch`es** — cross-origin actions now fall back to the browser's native submission instead of a request that can succeed server-side but fail client-side.
- **Initial route build could clobber a newer navigation** — on a client-side mount, a slow first (lazy) route could overwrite the route the user had already navigated to (`<FlatRoutes>`) or prevent it from rendering at all (`<Routes>`). The initial build now checks it is still current / is cancelled by the first navigation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2271,6 +2271,7 @@ dependencies = [
  "tracing",
  "url",
  "wasm-bindgen",
+ "wasm-bindgen-test",
  "web-sys",
 ]
 
--- a/leptos/src/form.rs
+++ b/leptos/src/form.rs
@@ -18,11 +18,37 @@ use tachys::{
     reactive_graph::node_ref::NodeRef,
 };
 use thiserror::Error;
-use wasm_bindgen::{JsCast, JsValue, UnwrapThrowExt};
-use web_sys::{
-    Event, FormData, HtmlButtonElement, HtmlFormElement, HtmlInputElement,
-    SubmitEvent,
-};
+use wasm_bindgen::{JsCast, JsValue, UnwrapThrowExt, prelude::wasm_bindgen};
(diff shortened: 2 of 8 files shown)


sabify:
The failed CI is not related to this PR.

gbj:
[b01fc2e](https://github.com/leptos-rs/leptos/pull/4861/commits/b01fc2e8e597cd1b4f63b45c91c2d29f5d91d00d) overlaps with an attempt to fix this same issue on `main` in https://github.com/leptos-rs/leptos/pull/4836; I just hadn't merged main into 0.9 recently. That said, I think the approach in `b01fc2e` is actually better and that we should keep it. However, could you move the new two-arg constructor into `leptos/src/form.rs` rather than `router/src/form.rs`, and then have `form_data_from_event` use it instead? Then it fixes both `ActionForm` and `Form` and this PR doesn't need a special helper.

gbj:
(I should add: everything else looks great, and thanks)

sabify:
Changes have been applied.

sabify:
The failed workflows are not related to this PR.

PR #4863, opened 2026-08-28
fix(server_fn): eight panics, leaks and protocol bugs

sabify:
This fixes eight issues in `server_fn`, one commit each, with regression tests.

**Panics on plausible input**
- actix `get_server_fn_service` panicked via `unreachable!()` on any non-standard HTTP method (`PROPFIND /api/foo` dropped the connection, or aborted the process under `panic = "abort"`). It now returns `None`, so the handler answers 400.
- `FromServerFnError::ser` panicked when neither the app error nor its `Serialization` fallback could be encoded (e.g. an internally-tagged newtype enum with `JsonEncoding`). It now degrades to a `text/plain` error body.
- The browser client's `try_into_stream` unwrapped `Response::body()`, aborting the wasm module on a body-less response (204/304/HEAD/opaque). It now returns an error.
- The tower blanket `Service` impl called `call()` without `poll_ready`, so `#[middleware(ConcurrencyLimitLayer::new(1))]` panicked on the first request. Readiness is now awaited (clone-and-replace); `BoxedService` is `Clone` so tower middleware can wrap it.

**Wrong behaviour**
- `StreamingText` decoded each transport chunk with `String::from_utf8`, so a multi-byte character split across chunks produced two `Deserialization` errors. Decoding now carries the incomplete tail over to the next chunk.
- `Http::run_client` returned early on 4xx/5xx before calling the redirect hook, so `redirect("/login"); Err(..)` never navigated. The hook now runs regardless of status.

**Websocket lifecycle**
- The server's output-forwarding task only noticed a disconnected client on a failed send, leaking a task and stream per disconnected client whenever the output was idle. `Req::try_into_websocket` now also returns a future that resolves on close, and the forwarder selects on it.
- The client never told the server its input had ended (a server function that drains input before replying hung forever), and dropping the output stream neither stopped the input pump nor closed the socket. An end-of-input frame is now sent (tag `2` in the existing framing — a Close frame would cut off the reply), and dropping the output stream cancels the pump and closes the socket.

**API notes:** `Req::try_into_websocket` returns a fourth tuple element; `BoxedService.service` is now `Arc<Mutex<Box<dyn Service + Send>>>` and `BoxedService: Clone`; the tower blanket impl requires `Clone` (tower convention — non-`Clone` services such as `RateLimit` no longer satisfy it).

**What changes for users (`BoxedService: Clone`)**

- **Non-breaking for the normal case.** `#[middleware(...)]` with any layer whose service is `Clone` (`ConcurrencyLimitLayer`, `TimeoutLayer`, `LoadShedLayer`, `BufferLayer`, `ServiceBuilder` stacks, most hand-written middleware) keeps compiling — and now actually works instead of panicking.
- **Breaking for non-`Clone` tower services.** A `#[middleware]` whose service is not `Clone` (`tower::limit::RateLimitLayer` is the notable one) is rejected at compile time with an unsatisfied `Clone` bound. Before, it compiled but couldn't limit anything: a fresh `RateLimit` is created per request and its `call` panics once the per-instance quota is exhausted. Rate limiting belongs at the router level anyway.
- **Custom `server_fn::middleware::Service` impls** (not tower-based) are unaffected — the trait itself didn't change.
- **Code that read `BoxedService.service` directly** (the field is `pub`, but nothing in-tree outside `server_fn` touches it) now sees `Arc<Mutex<..>>`; `BoxedService::new` and `run` are unchanged.
- The actix blanket impls are untouched; actix services are often neither `Clone` nor `Send`, and no failure was demonstrated there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4218,6 +4218,7 @@ dependencies = [
  "pin-project-lite",
  "sync_wrapper",
  "tokio",
+ "tokio-util",
  "tower-layer",
  "tower-service",
  "tracing",
--- a/examples/server_fns_axum/src/middleware.rs
+++ b/examples/server_fns_axum/src/middleware.rs
@@ -18,6 +18,7 @@ impl<S> Layer<S> for LoggingLayer {
     }
 }
 
+#[derive(Clone)]
 pub struct LoggingService<T> {
     inner: T,
 }
--- a/server_fn/Cargo.toml
+++ b/server_fn/Cargo.toml
@@ -101,6 +101,18 @@ rustc_version = { workspace = true, default-features = true }
 [dev-dependencies]
 trybuild = { workspace = true, default-features = true }
 
+[target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
+tokio = { workspace = true, features = [
+  "macros",
+  "rt-multi-thread",
+  "net",
+  "time",
+  "sync",
+] }
+axum = { workspace = true, default-features = true, features = ["ws"] }
+tokio-tungstenite = { workspace = true }
+serde_json = { workspace = true }
+
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
 wasm-bindgen-test = { workspace = true, default-features = true }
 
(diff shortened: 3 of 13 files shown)


sabify:
Please re-run the workflow, the failed one is not related to this PR.

gbj:
This looks fine to me overall. Note that the CI failure in `examples/server_fns_axum` is because 42fccb9 really does break the middleware example there. It should be easy to fix.

sabify:
Done.

PR #4867, opened 2026-08-29
feat(reactive_stores): convert a subfield into MaybeProp

svieujot:
`MaybeProp<T>` already accepts `ReadSignal`, `RwSignal`, `Memo` and `Signal`, each in both the `Option<T>` and the plain `T` shape. A store subfield had neither, and `Into` does not chain, so a field feeding a `#[prop(into)] MaybeProp<T>` had to spell the hop out at the call site as `Signal::derive(move || store.field().get())`.


gbj:
Looks good, thanks!

PR #4868, opened 2026-08-31
fix: UnsyncCallback/Callback comments

geovie:
fixes some stale comments found while reading the docs about callback

### DIFF
--- a/reactive_graph/src/callback.rs
+++ b/reactive_graph/src/callback.rs
@@ -9,7 +9,7 @@
 //! - [`Callback`](reactive_graph::callback::Callback)
 //! - [`UnsyncCallback`](reactive_graph::callback::UnsyncCallback)
 //!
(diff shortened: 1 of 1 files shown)


reactive_graph/src/callback.rs
```diff
@@ -47,7 +47,7 @@ pub struct UnsyncCallback<In: 'static, Out: 'static = ()>(
 
 impl<In> fmt::Debug for UnsyncCallback<In> {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        fmt.write_str("Callback")
```
geovie, line 50: Not sure if this is worth changing maybe someone is relying on it 🤷  (same for SyncCallback below)

gbj:
Thanks!

PR #4871, opened 2026-09-11
fix(examples): await the add-1000-counters click

larstobi:
## Why

`CountersPage.addOneThousandCounters()` is `async`, but starts `Locator.click()` without awaiting or returning its promise. The three awaited calls in `add_1k_counters.spec.ts` can therefore overlap instead of waiting for each click to finish.

## Change

Add the missing `await` so the helper waits for the click and propagates any click failure to its caller.

The diff against `main` is one line in one file. Cargo Make tasks, Playwright server configuration, Trunk options, assertions, retries, and dependencies are unchanged. Server-setup changes have been taken out of this PR for separate consideration.

## Validation and remaining issue

Validated the narrowed change in Ubuntu 24.04 Docker with Rust 1.98.1, Trunk 0.21.14, Node 20.20.2, and Playwright 1.63.0.

- A temporary, isolated check with a controlled click promise confirmed that the helper remains pending until the click finishes. This check is not part of the PR diff.
- `trunk build` passed.
- From `examples/counters`: `cargo +nightly-2026-07-20 fmt --check -- --config-path ../..` passed.
- `git diff --check` passed.
- With the existing Trunk server setup, from `examples/counters/e2e`:
  - `pnpm exec playwright test --retries=0 --reporter=list`: 8 passed, 2 failed.
  - `pnpm exec playwright test --retries=0 --trace=on --reporter=list`: 9 passed, 1 failed.

This does **not** fully fix the browser CI failures. The separate Trunk auto-reload issue remains: the traced run shows a navigation during the counter assertion, resetting the count to zero. Reload protection and possible server-lifecycle improvements need a separate follow-up. The earlier passing results for the broader server-setup patch do not apply to this await-only diff.


larstobi:
This should succeed if #4872 is merged.

larstobi:
@gbj it succeeded! :tada:

PR #4872, opened 2026-09-11
fix(examples): prevent Trunk auto-reload during browser tests

larstobi:
## Why

Trunk watches the example directory, including Playwright's generated files under `e2e/test-results`. Writing traces can therefore trigger rebuilds and reload the page in the middle of a test, resetting state or detaching elements.

Reproduced on unmodified `main` (`e238bced`): 4 failures in 30 test executions with tracing enabled. Trunk's debug log confirms that the trace files trigger rebuilds.

## Change

Add `start-client-for-tests`, extending the existing Cargo Make `start-client` task with two test-only Trunk settings:

- `TRUNK_SERVE_NO_AUTORELOAD=true`: prevent automatic page reloads.
- `TRUNK_WATCH_IGNORE=e2e`: keep test artifacts from triggering rebuilds.

The shared integration-test task uses this for `counters` and `error_boundary`. Normal `start-client` and `trunk serve` behavior is unchanged. Server management remains in Cargo Make; there are no Playwright configuration, assertion, retry, Rust, or dependency changes.

This is independent of the missing-`await` fix in #4871. None of that PR's changes are included here.

## Validation

Tested this reload-only diff in Ubuntu 24.04 Docker with Cargo Make 0.37.24, Trunk 0.21.14, Node 20.20.2, and Playwright 1.63.0. Both examples were tested on Rust 1.98.1 and `nightly-2026-07-20`, with and without `RUSTFLAGS="--cfg erase_components"`.

- `cargo make --no-workspace --profile=github-actions integration-test`: all 64 test executions passed across the eight configurations.
- From each `e2e` directory: `pnpm exec playwright test --retries=0 --repeat-each=10 --trace=on --reporter=list`: all 640 test executions passed.
- Confirmed that the running test servers receive both settings and that a separate normal `start-client` invocation does not.
- Nightly formatting checks passed for the workspace and both examples; `git diff --check` passed.

All 704 test executions passed without using retries, and without applying #4871. This validation covers the affected browser integration tests, not the entire workspace CI suite.

**CI selection:** the current workflow filters exclude `examples/cargo-make/**`, so this one-file change does not automatically select the browser jobs on GitHub. The matrix above was run explicitly in Docker; changing those filters is outside this PR's scope.


gbj:
Thanks!

Issue #4873, opened 2026-09-13
Update gloo-utils to 0.3


EvanCarroll:
gloo-utils 0.3 is out, changes include.

```
Version 0.3.0
Fix JsError::try_from panicking on non-string-coercible JsValue inputs (#488) by @ranile
Update MSRV to 1.82, migrate dependencies to workspace versions (#505) by @martinfrances107
```

PR #4875, opened 2026-09-13
chore(deps): update gloo family — gloo-net 0.7, gloo-utils 0.3

### LINKED ISSUE #4874: Update gloo-net to 0.7
Major changes in the gloo-net 0.7 branch include,

```
Version 0.7.0
Remove Clone implementation for EventSource (#417) by @vpochapuis Cloning had a bug where dropping any clone would disconnect all instances.
Add EventSourceBuilder for configuring EventSource with credentials (#530) by @zn3x
Add WebSocket::buffered_amount (#510) by @vi
Add PartialEq, Eq, Hash derives to CloseEvent (#504) by @Alexi24601
Fix RequestBuilder::build producing a URL with an extra & (#497) by @hmacias-avaya
Fix WebSocket binary send panicking when SharedArrayBuffer is enabled (#502) by @Fedeparma74
Update MSRV to 1.82 (#505) by @martinfrances107
Bump pin-project to 1.1 and http to 1.4 (#536) by @Madoshakalaka
```

I'll try to get a patch open soon with #4873

EvanCarroll:
Disclaimer: most of this commit message is written with AI. It's a boring dep bump that I needed because without this the gloo things I need in the gloo ecosystem which ARE NOT exported by leptos caused me to need two copies of gloo-net, and gloo-utils (which is two different versions of the websocket stack) because leptos is behind the gloo ecosystem currently. This brings it up to date.

----

Refreshes every gloo crate in the repo to its current release. gloo-net 0.7.0 and gloo-utils 0.3.0 are the latest on crates.io; the umbrella `gloo` crate only carries them from 0.12, and `gloo-timers`/`gloo-storage` moved to 0.4 in the same MSRV-driven sweep.

Why now: 0.7 fixes two bugs this repo is exposed to.

  * RequestBuilder::build no longer emits a stray `&` when the base URL already has a query string (rustwasm/gloo#497). leptos_router's <Form/> and server_fn's browser client both build URLs this way.
  * WebSocket binary sends no longer panic when SharedArrayBuffer is enabled (rustwasm/gloo#502); Message::Bytes now goes out as a Blob. server_fn's websocket client sends exclusively Message::Bytes.

It also brings EventSourceBuilder (credentialed SSE), WebSocket::buffered_amount, and PartialEq/Eq/Hash on CloseEvent.

Bumped
------
  gloo-net    0.6   -> 0.7   workspace root + 7 excluded packages
  gloo-utils  0.2.0 -> 0.3   examples/axum_js_ssr
  gloo        0.8.0/0.11.0 -> 0.12   benchmarks, todo_app_sqlite,
                                     leptos_dom/examples/{hydration-test,test-bench}
  gloo-timers 0.3   -> 0.4   examples/{counter,fetch,lazy_routes}
  gloo-storage 0.3.0 -> 0.4  projects/login_with_token_csr_only/client

examples/todo_app_sqlite pinned `gloo` to a git ref with no version; it is now a normal crates.io dependency so the build is reproducible.

EventSource lost its Clone impl in 0.7, but the repo's only EventSource (examples/counter_isomorphic) moves ownership into on_cleanup rather than cloning. Message, WebSocketError, EventSourceSubscription and gloo_net::Error are all structurally unchanged. `http` was already at 1.4 and rust-version at 1.88, so the 0.7 requirement bumps of those are no-ops here.

Closes #4874, #4873

Special notes (claude audit of potential breakage)
-------

    POSSIBLE BREAKAGE
    -----------------
    1. wasm-bindgen toolchain. gloo-net 0.7 requires js-sys ^0.3.91, which
       pins wasm-bindgen exactly, so resolution moves:
    
           wasm-bindgen           0.2.108 -> 0.2.128
           js-sys / web-sys       0.3.85  -> 0.3.105
           wasm-bindgen-futures   0.4.58  -> 0.4.78
    
       Anyone building Leptos apps must update wasm-bindgen-cli to match.
       This is the most disruptive part of the change and semver cannot
       express it. CI installs wasm-bindgen-cli unpinned, so CI is unaffected.
    
    2. Public API. gloo-net types leak out of two crates, so downstream code
       that names them must move to gloo-net 0.7 as well:
    
         * leptos_router — the <Form/> `on_error` prop is
           Arc<dyn Fn(&gloo_net::Error)>  (router/src/form.rs)
         * server_fn — `pub use gloo_net::http::Request` and `Response`
           behind the `browser` feature  (src/request/browser.rs,
           src/response/browser.rs); `leptos` re-exports server_fn wholesale
    
       These are patch bumps rather than minor ones because leptos_router
       0.9 would collide with the planned Leptos 0.9 line. That means cargo
       treats the change as compatible and downstream picks it up on a plain
       `cargo update`. Label the PR `breaking` so the semver-check job is
       skipped, and call out the wasm-bindgen floor in the release notes.
    
           leptos        0.8.20 -> 0.8.21
           leptos_router 0.8.15 -> 0.8.16
           server_fn     0.8.13 -> 0.8.14
    
    Verified: cargo check --workspace --all-targets clean; leptos_router and
    server_fn --features browser clean on host and wasm32-unknown-unknown;
    all 11 gloo-using examples clean on both their hydrate/wasm32 and ssr
    sides.


gbj:
Thanks for the PR. The AI text you included gave some good/bad advice here: specifically it notes that `gloo-net` types are reexported, so a breaking release (minor-to-minor version update) in `gloo-net` requires a breaking release in the crates that reexport those types. It then notes that there's already a planned 0.9 release, but rather than folding the change into the 0.9 branch, it says we should disable semver checks and include it in a patch release instead. That doesn't seem like the right move to me; rather, it should be folded into the planned 0.9 release.

So maybe target this against the existing `leptos_0.9` branch instead, and we can fold it in there? 

While making that change, note also that the dependencies intentionally have patch release numbers included in the versions; the PR changes every `0.3.0` to `0.4`, `0.6.0` to `0.7`, etc. but we want to keep the trailing `.0` You can leave the various leptos crate version numbers as they are, and this will be folded into the next release.

EvanCarroll:
> So maybe target this against the existing leptos_0.9 branch instead, and we can fold it in there?

**I'm happy to do that**. But ultimately that doesn't solve my problem with code duplication now and I have to maintain a fork until 0.9.x is released. Moreover, by the time 0.9.x is out it's likely for this to still be a problem with duplication if versioning if not fixed by gloo. I've filed a bug upstream, I'll fix up the patch later this week.

https://github.com/ranile/gloo/issues/584

gbj:
Why would you need to maintain a fork and not just use the 0.9 prerelease versions or use the `leptos_0.9` branch directly? It is in beta, there will be a series of release candidates, etc. You can use the published prerelease versions or use a git dependency on that branch; there is no need to maintain your own fork.

(I'd suggest toning down the hostility in the linked issue; I understand the frustration it's expressing, but it's not expressed in a way that is likely to make the maintainers receptive.)

PR #4877, opened 2026-09-18
fix(ci): use locked dependencies when installing Trunk

larstobi:
When `cargo binstall` falls back to compiling Trunk from source, installing without `--locked` resolves newer transitive dependencies. In [this CI job](https://github.com/leptos-rs/leptos/actions/runs/35147565909/job/104980835799), Trunk 0.21.14 selected `parcel_selectors` 0.28.3, whose `cssparser` 0.37.0 types are incompatible with the `cssparser` 0.33.0 types used by `lightningcss` 1.0.0-alpha.65. Installation failed before the example tests started.

Add `--locked` to the existing Trunk installation command so source builds use the published dependency lockfile. Trunk 0.21.14's published lockfile uses `parcel_selectors` 0.28.1 and `cssparser` 0.33.0 for both dependencies.

Validation: `git diff --check` passed; inspected the lockfile bundled in the published Trunk 0.21.14 crate and confirmed that cargo-binstall forwards `--locked` to `cargo install`. A full source build and CI run have not been performed locally.


gbj:
Thanks!

